PR #89, opened 2018-10-31
Add support for string ids for COCO dataset

ychfan:
The `loadImgs` and `getAnnIds` in `pycocotools` expect a list of ids or single int id. They will treat single string id as list of ids. Always passing a list will work.


hellock:
Thanks for fixing it!

PR #117, opened 2018-11-23
Update nms_wrapper.py

youansheng:
Fix bug for nms_wrapper.

### DIFF
--- a/mmdet/ops/nms/nms_wrapper.py
+++ b/mmdet/ops/nms/nms_wrapper.py
@@ -9,22 +9,21 @@ from .cpu_soft_nms import cpu_soft_nms
 def nms(dets, thresh, device_id=None):
     """Dispatch to either CPU or GPU NMS implementations."""
 
+    is_tensor = False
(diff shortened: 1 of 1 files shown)


mmdet/ops/nms/nms_wrapper.py
```diff
@@ -12,14 +12,17 @@ def nms(dets, thresh, device_id=None):
     if isinstance(dets, torch.Tensor):
         if dets.is_cuda:
             device_id = dets.get_device()
-        dets = dets.detach().cpu().numpy()
-    assert isinstance(dets, np.ndarray)
+        _dets = dets.detach().cpu().numpy()
+    else:
+        _dets = dets.copy()
```
hellock: We don't need to copy the original array since it will not be modified in-place.

hellock:
You may need to fix the code style to pass flake8 check.

mmdet/ops/nms/nms_wrapper.py
youansheng: Thanks!

PR #131, opened 2018-11-29
Update gpu_nms.pyx

youansheng:
Fix bug for gpu_nms.


youansheng:
https://github.com/open-mmlab/mmdetection/blob/64e310d5afae93d39e0a51aaa6a3d7841e1318b3/mmdet/ops/nms/nms_kernel.cu#L54
It will be inconsistent with mask when boxes_dim is not equal to 5.

hellock:
Thanks for the fix! Could you provide some use cases where the dim is not 5?

youansheng:
When I do class-aware NMS, the dets[:, 6] contain the labels of each dets. If I slice the labels, there will make a extra copy. If not, It will encounter the error I have mentioned above.

PR #148, opened 2018-12-07
fix bias checking when init

thangvubk:
Sanity check
```
layer = nn.Conv2d(1, 1, 1, bias=False)
print(hasattr(layer, 'bias'))
>>> True
```


hellock:
Thanks! In fact it has been fixed in [mmcv](https://github.com/open-mmlab/mmcv/blob/master/mmcv/cnn/weight_init.py), you can simply replacing all weight init methods with the ones provided in `mmcv`. Methods in this repo will be deprecated.

hellock:
A PR on GroupNorm is welcome! Also you can replacing the weight init methods in this PR.

PR #160, opened 2018-12-10
add group norm support

thangvubk:
Here is the reported performance on Faster RCNN R50 FPN 1x with group norm. 

| Backbone | Style   | Lr schd | box AP | Download |
|:--------:|:-------:|:-------:|:------:|:--------:|
| R-50-FPN | pytorch | 1x      | 36.4   |  |
|R-50-FPN-GN| pytorch | 1x | 37.0 | [model](https://drive.google.com/file/d/1qkdnqgzfFHRvWI-BTRIlIvd6Hm8EUS9T/view?usp=sharing)|

Config for group norm (simply add ``normalize`` into ``FPN neck`` and ``bbox_head``)

```
# model settings
model = dict(
    type='FasterRCNN',
    pretrained='modelzoo://resnet50',
    backbone=dict(
        type='ResNet',
        depth=50,
        num_stages=4,
        out_indices=(0, 1, 2, 3),
        frozen_stages=1,
        style='pytorch'),
    neck=dict(
        type='FPN',
        in_channels=[256, 512, 1024, 2048],
        out_channels=256,
        num_outs=5,
        normalize=dict(
            type='GN',
            num_groups=32)),
    rpn_head=dict(
        type='RPNHead',
        in_channels=256,
        feat_channels=256,
        anchor_scales=[8],
        anchor_ratios=[0.5, 1.0, 2.0],
        anchor_strides=[4, 8, 16, 32, 64],
        target_means=[.0, .0, .0, .0],
        target_stds=[1.0, 1.0, 1.0, 1.0],
        use_sigmoid_cls=True),
    bbox_roi_extractor=dict(
        type='SingleRoIExtractor',
        roi_layer=dict(type='RoIAlign', out_size=7, sample_num=2),
        out_channels=256,
        featmap_strides=[4, 8, 16, 32]),
    bbox_head=dict(
        type='SharedFCBBoxHead',
        num_fcs=2,
        in_channels=256,
        fc_out_channels=1024,
        roi_feat_size=7,
        num_classes=81,
        target_means=[0., 0., 0., 0.],
        target_stds=[0.1, 0.1, 0.2, 0.2],
        reg_class_agnostic=False,
        normalize=dict(
            type='GN',
            num_groups=32)))
# model training and testing settings
train_cfg = dict(
    rpn=dict(
        assigner=dict(
            pos_iou_thr=0.7,
            neg_iou_thr=0.3,
            min_pos_iou=0.3,
            ignore_iof_thr=-1),
        sampler=dict(
            num=256,
            pos_fraction=0.5,
            neg_pos_ub=-1,
            add_gt_as_proposals=False,
            pos_balance_sampling=False,
            neg_balance_thr=0),
        allowed_border=0,
        pos_weight=-1,
        smoothl1_beta=1 / 9.0,
        debug=False),
    rcnn=dict(
        assigner=dict(
            pos_iou_thr=0.5,
            neg_iou_thr=0.5,
            min_pos_iou=0.5,
            ignore_iof_thr=-1),
        sampler=dict(
            num=512,
            pos_fraction=0.25,
            neg_pos_ub=-1,
            add_gt_as_proposals=True,
            pos_balance_sampling=False,
            neg_balance_thr=0),
        pos_weight=-1,
        debug=False))
test_cfg = dict(
    rpn=dict(
        nms_across_levels=False,
        nms_pre=2000,
        nms_post=2000,
        max_num=2000,
        nms_thr=0.7,
        min_bbox_size=0),
    rcnn=dict(
        score_thr=0.05, nms=dict(type='nms', iou_thr=0.5), max_per_img=100)
    # soft-nms is also supported for rcnn testing
    # e.g., nms=dict(type='soft_nms', iou_thr=0.5, min_score=0.05)
)
# dataset settings
dataset_type = 'CocoDataset'
data_root = 'data/coco/'
img_norm_cfg = dict(
    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
data = dict(
    imgs_per_gpu=2,
    workers_per_gpu=2,
    train=dict(
        type=dataset_type,
        ann_file=data_root + 'annotations/instances_train2017.json',
        img_prefix=data_root + 'train2017/',
        img_scale=(1333, 800),
        img_norm_cfg=img_norm_cfg,
        size_divisor=32,
        flip_ratio=0.5,
        with_mask=False,
        with_crowd=True,
        with_label=True),
    val=dict(
        type=dataset_type,
        ann_file=data_root + 'annotations/instances_val2017.json',
        img_prefix=data_root + 'val2017/',
        img_scale=(1333, 800),
        img_norm_cfg=img_norm_cfg,
        size_divisor=32,
        flip_ratio=0,
        with_mask=False,
        with_crowd=True,
        with_label=True),
    test=dict(
        type=dataset_type,
        ann_file=data_root + 'annotations/instances_val2017.json',
        img_prefix=data_root + 'val2017/',
        img_scale=(1333, 800),
        img_norm_cfg=img_norm_cfg,
        size_divisor=32,
        flip_ratio=0,
        with_mask=False,
        with_label=False,
        test_mode=True))
# optimizer
optimizer = dict(type='SGD', lr=0.02, momentum=0.9, weight_decay=0.0001)
optimizer_config = dict(grad_clip=dict(max_norm=35, norm_type=2))
# learning policy
lr_config = dict(
    policy='step',
    warmup='linear',
    warmup_iters=500,
    warmup_ratio=1.0 / 3,
    step=[8, 11])
checkpoint_config = dict(interval=1)
# yapf:disable
log_config = dict(
    interval=50,
    hooks=[
        dict(type='TextLoggerHook'),
        # dict(type='TensorboardLoggerHook')
    ])
# yapf:enable
# runtime settings
total_epochs = 12
dist_params = dict(backend='nccl')
log_level = 'INFO'
work_dir = './work_dirs/faster_rcnn_r50_fpn_1x'
load_from = None
resume_from = None
workflow = [('train', 1)]
```


hellock:
Thanks for the nice work!

Compared to the official version, there are some modifications needed.
- GN is applied to the backbone as well. (needs pretraining)
- BBoxHead is 4conv1fc instead of 2fc, and GN is added only after conv layers.

If you do not have time or resources to pretrain GN backbones on ImageNet, we can help that. You may try converting [official model](https://github.com/facebookresearch/Detectron/blob/master/projects/GN/README.md#pretrained-models-with-gn) to PyTorch first for further tests.

thangvubk:
Thank for your advice! I will try it and update the progress.

thangvubk:
Here is the check list for adding group norm to backbone. @hellock If you have any comments or suggestions please let me know, thank you!

- [ ] Convert caffe2 resnet-gn to pytorch
- [ ] Check converted model on mmdet
- [ ] Pretrain resnet gn with pytorch
- [ ] Check pretrained model on mmdet

hellock:
The checklist looks great. Thanks for your contribution!

thangvubk:
Group norm (and unit test) for the backbone has been added. However, i encoutered performance issue. 
- Reported top 1: 76%
- Converted from detection: 60%
- Train on imagenet: 75.72

Caffe2 model is taken directly from [official detectron repo](https://s3-us-west-2.amazonaws.com/detectron/ImageNetPretrained/47261647/R-50-GN.pkl).
Training and eval script is borrowed from [pytorch/example/imagenet](https://github.com/pytorch/examples/tree/master/imagenet).

The converted accuracy is so wierd to me, i am not sure wheather i am wrong in subtle things. 
If you are available, could you please look into the code a bit. 
To eval converted caffe2 model: ``python tools/train_imagenet/train_imagenet.py <DATA> --cf_path <caffe_path> --evaluate``

hellock:
I found that GN is trained with pytorch style, putting the stride 2 in the 3x3 conv instead of 1x1 conv, so pytorch style should be used.

thangvubk:
@hellock I believe that the data for caffe2 resnet is processed different from pytorch. Did you try with pytorch style but caffe2 data preprocess?

hellock:
Simply changing the data transforms does not work, maybe further study is needed.

thangvubk:
To make sure that wheather the problem in preprocessing, im training caffe loaded weight with img_norm_config as you provided. Result will be out tomorrow.

thangvubk:
@hellock I suppose that caffe2 input is normalized to zeros mean, has the dynamic range (0, 255) and has inverse channel order compared to pytorch. After doing the above modifications, top1 acc is 75.8%. There are still 0.2% mismatch. Do you have any ideas?

hellock:
I guess there are some possibilities.
- In caffe2 training, opencv is used for preprocessing, which can generate slightly different results from PIL.
- Some operators are configurated with different behaviors, such as the `ceil_mode` of pooling. (not very sure)

thangvubk:
Progress:

|          Model          | Detectron | Detectron converted | Pytorch pretrain |
|:-----------------------:|:---------:|:-------------------:|:----------------:|
| Mask-RCNN R50 FPN GN 2X | 40.3/35.7 |      39.9/36.2      |        -          |

hellock:
An update here. Mask R-CNN R50-GN training from scratch (6x): 40.5/36.7

thangvubk:
Thank you for your suggestion. Running 6x might be faster on distributed training with multiple workers. However i dont have enough resource for that. It would be very nice if you can help to run this experiment.

hellock:
Hi @thangvubk, the above result is reported on this repo, just to show that it is ok.

mmdet/models/utils/norm.py
```diff
@@ -1,17 +1,25 @@
...
     cfg_ = cfg.copy()
-    cfg_.setdefault('eps', 1e-5)
     layer_type = cfg_.pop('type')
 
+    # args name matching
+    if layer_type == 'GN':
+        assert 'num_groups' in cfg
+        cfg_.setdefault('num_channels', num_features)
+    elif layer_type == 'BN':
+        cfg_ = dict()  # rewrite neccessary info for BN from here
```
hellock: This may lead to potential bugs, since `cfg` may contain keys like `momentum`, `affine`, etc.

```python
def build_norm_layer(cfg, num_features):
    assert isinstance(cfg, dict) and 'type' in cfg
    cfg_ = cfg.copy()
    cfg_.setdefault('eps', 1e-5)
    layer_type = cfg_.pop('type')

    if layer_type not in norm_cfg:
        raise KeyError('Unrecognized norm type {}'.format(layer_type))
    elif norm_cfg[layer_type] is None:
        raise NotImplementedError
    else:
        norm_layer = norm_cfg[layer_type]

    if layer_type != 'GN':
        return norm_layer(num_features, **cfg_)
    else:
        assert 'num_groups' in cfg_
        return norm_layer(num_channels=num_features, **cfg_)
```

mmdet/models/backbones/resnet.py
```diff
@@ -235,17 +259,29 @@ def __init__(self,
...
-        self.bn_frozen = bn_frozen
         self.with_cp = with_cp
 
+        assert isinstance(normalize, dict) and 'type' in normalize
+        assert normalize['type'] in ['BN', 'GN']
+        if normalize['type'] == 'GN':
+            assert 'num_groups' in normalize
+        else:
+            assert (set(['type', 'bn_eval', 'bn_frozen'])
+                    == set(normalize))
```
hellock: These assertions can be moved to `build_norm_layer`.

mmdet/models/backbones/resnet.py
```diff
@@ -302,7 +345,7 @@ def forward(self, x):
 
     def train(self, mode=True):
```
hellock: If we freeze BN parameters and set eval mode in `build_norm_layer`, the `train()` method can be significantly simplified.

mmdet/models/backbones/resnet.py
```diff
@@ -312,13 +355,111 @@ def train(self, mode=True):
...
-            self.bn1.bias.requires_grad = False
+
             for i in range(1, self.frozen_stages + 1):
                 mod = getattr(self, 'layer{}'.format(i))
                 mod.eval()
                 for param in mod.parameters():
                     param.requires_grad = False
+
+
+class ResNetClassifier(ResNet):
```
hellock: This class may not be necessary now. We can later add an option to include fc layers in ResNet backbone.

mmdet/models/utils/norm.py
```diff
@@ -1,17 +1,38 @@
...
     cfg_ = cfg.copy()
-    cfg_.setdefault('eps', 1e-5)
     layer_type = cfg_.pop('type')
+    frozen = cfg_.pop('frozen') if 'frozen' in cfg_ else False
+
+    # args name matching
+    if layer_type == 'GN':
+        assert 'num_groups' in cfg
+        cfg_.setdefault('num_channels', num_features)
+    elif layer_type == 'BN':
```
hellock: Other norm layers may also use `num_features` as the first argument.

mmdet/models/utils/norm.py
```diff
@@ -1,17 +1,38 @@
...
         raise KeyError('Unrecognized norm type {}'.format(layer_type))
     elif norm_cfg[layer_type] is None:
         raise NotImplementedError
 
-    return norm_cfg[layer_type](num_features, **cfg_)
+    norm = norm_cfg[layer_type](**cfg_)
+    if frozen:
+        for param in norm.parameters():
+            param.requires_grad = False
+    return norm
```
hellock: A suggestion to include both "frozen" and "eval_mode" in arguments and return the layer and abbreviation.
```python
norm_cfg = {
    'BN': (nn.BatchNorm2d, 'bn'),
    'SyncBN': (None, 'bn'),
    'GN': (nn.GroupNorm, 'gn'),
    # and potentially 'SN'
}


def build_norm_layer(cfg, num_features):
    assert isinstance(cfg, dict) and 'type' in cfg
    cfg_ = cfg.copy()
    layer_type = cfg_.pop('type')
    frozen = cfg_.pop('frozen', False)
    # eval_mode only has effects on BN and its variants
    eval_mode = cfg_.pop('eval_mode', False)

    if layer_type not in norm_cfg:
        raise KeyError('Unrecognized norm type {}'.format(layer_type))
    else:
        norm_layer, abbr = norm_cfg[layer_type]
        if norm_layer is None:
            raise NotImplementedError

    cfg_.setdefault('eps', 1e-5)
    if layer_type != 'GN':
        layer = norm_layer(num_features, **cfg_)
    else:
        assert 'num_groups' in cfg_
        layer = norm_layer(num_channels=num_features, **cfg_)

    if frozen:
        for param in layer.parameters():
            param.requires_grad = False

    if eval_mode:
        param.eval()

    return layer, abbr
```

mmdet/models/backbones/resnet.py
```diff
@@ -302,23 +356,8 @@ def forward(self, x):
 
     def train(self, mode=True):
```
hellock: If we set eval mode in `build_norm_layer`, then this method is not necessary any more.

mmdet/models/backbones/resnet.py
```diff
@@ -279,14 +326,21 @@ def init_weights(self, pretrained=None):
             for m in self.modules():
                 if isinstance(m, nn.Conv2d):
                     kaiming_init(m)
-                elif isinstance(m, nn.BatchNorm2d):
+                elif (isinstance(m, nn.BatchNorm2d)
+                      or isinstance(m, nn.GroupNorm)):
                     constant_init(m, 1)
+
+            # zero init for last norm layer https://arxiv.org/abs/1706.02677
```
hellock: We can add an argument `zero_init_residual` as in [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py).

mmdet/models/backbones/resnet.py
```diff
@@ -279,14 +326,21 @@ def init_weights(self, pretrained=None):
             for m in self.modules():
                 if isinstance(m, nn.Conv2d):
                     kaiming_init(m)
-                elif isinstance(m, nn.BatchNorm2d):
+                elif (isinstance(m, nn.BatchNorm2d)
+                      or isinstance(m, nn.GroupNorm)):
```
hellock: `isinstance(m, (nn.BatchNorm2d, nn.GroupNorm))`

mmdet/models/backbones/resnet.py
```diff
@@ -234,20 +268,17 @@ def __init__(self,
         self.out_indices = out_indices
         assert max(out_indices) < num_stages
         self.style = style
-        self.frozen_stages = frozen_stages
-        self.bn_eval = bn_eval
-        self.bn_frozen = bn_frozen
         self.with_cp = with_cp
-
+        self.is_frozen = [i <= frozen_stages for i in range(num_stages + 1)]
```
hellock: Maybe we can define a method `_freeze_layers()` to handle such things.

mmdet/models/backbones/resnet.py
```diff
@@ -219,8 +251,10 @@ def __init__(self,
                  out_indices=(0, 1, 2, 3),
                  style='pytorch',
                  frozen_stages=-1,
-                 bn_eval=True,
-                 bn_frozen=False,
+                 normalize=dict(
+                     type='BN',
+                     eval=True,
```
hellock: `eval` is a built-in keyword of python, I suggest using `eval_mode`.

mmdet/models/utils/norm.py
thangvubk: If set eval here, it will be overwritten on train() method of backbone.

mmdet/models/backbones/resnet.py
```diff
@@ -29,27 +30,40 @@ def __init__(self,
                  dilation=1,
                  downsample=None,
                  style='pytorch',
-                 with_cp=False):
+                 with_cp=False,
+                 normalize=dict(type='BN'),
+                 frozen=False):
```
hellock: The freezing behavior may be better moved to the `_freeze_layers` method of `ResNet` to make BasicBlock/Bottleneck clean.

mmdet/models/utils/norm.py
hellock: Oh I ignored that, so we still need to set eval mode in `resnet.train()`.

mmdet/models/backbones/resnet.py
thangvubk: I don't see any tradeoff when setting zero_init_residual to False. Since it is recommended, should we use it as defaults?

hellock: I agree that we can set the default value as `True`.

mmdet/models/utils/norm.py
thangvubk: abbr is a nice idea for backbones but might be not helpful at neck and roi_head norms. I think keeping abbr processed in backbones in ok, and build_norm_layer is more straightforward.

hellock: In neck and heads, we usually use ConvModule instead of directly using norm layers. A ConvModule can be more clear if we also adopt "bn", "gn" instead of "norm".

thangvubk: I totally agree. It will be fixed that after the sanity benchmark check.

thangvubk:
@hellock Benchmark on Faster RCNN R50 FPN 1x get 36.4 mAP, similar to the reported result in [model zoo](https://github.com/open-mmlab/mmdetection/blob/master/MODEL_ZOO.md)

mmdet/models/utils/norm.py
thangvubk: If the norm name is changed, says from 'norm' to 'bn',  existing pretrained models won't be loaded directly.

hellock: Existing models in the model zoo have no norm layers in necks and heads, except for this GN model. We can rename keys of the trained GN model without much efforts.

thangvubk: I see. It is updated in [3fdd041](https://github.com/open-mmlab/mmdetection/pull/160/commits/3fdd041cd7f4bd6fc877bac22b28448ae738033d)

mmdet/models/utils/norm.py
```diff
@@ -1,17 +1,48 @@
...
+
+def build_norm_layer(cfg, num_features, postfix=''):
+    """
+    cfg should contain:
+        type (str): identify norm layer type.
+        layer args: args needed to instantiate a norm layer.
+        frozen (bool): [optional] whether stop gradient updates
+            of norm layer, it is helpful to set frozen mode
+            in backbone's norms.
+    """
```
hellock: The docstring needs to be updated.

mmdet/models/backbones/resnet.py
```diff
@@ -279,14 +321,21 @@ def init_weights(self, pretrained=None):
...
                     kaiming_init(m)
-                elif isinstance(m, nn.BatchNorm2d):
+                elif isinstance(m, (nn.BatchNorm, nn.GroupNorm)):
                     constant_init(m, 1)
+
+            # zero init for last norm layer https://arxiv.org/abs/1706.02677
+            if self.zero_init_residual:
+                for m in self.modules():
+                    if isinstance(m, (Bottleneck, BasicBlock)):
+                        last_norm = getattr(m, m.norm_names[-1])
```
hellock: `norm_names` is undefined.

mmdet/models/backbones/resnet.py
```diff
@@ -100,32 +108,39 @@ def __init__(self,
             dilation=dilation,
             bias=False)
 
-        self.bn1 = nn.BatchNorm2d(planes)
-        self.bn2 = nn.BatchNorm2d(planes)
+        # build_norm_layer return: (norm_name, norm_layer)
+        self.norm1, norm1 = build_norm_layer(normalize, planes, postfix=1)
+        self.norm2, norm2 = build_norm_layer(normalize, planes, postfix=2)
+        self.norm3, norm3 = build_norm_layer(normalize, planes*self.expansion,
```
hellock: One space is needed before and after `*`.

mmdet/models/backbones/resnet.py
```diff
@@ -45,11 +52,11 @@ def forward(self, x):
         identity = x
 
         out = self.conv1(x)
-        out = self.bn1(out)
+        out = getattr(self, self.norm1)(out)
```
hellock: I suggest adding some properties so that we can directly use `self.norm1(out)`.
```python
@property
def norm1(self):
    return getattr(self, self.norm1_name)
```

mmdet/models/backbones/resnet.py
```diff
@@ -29,13 +30,19 @@ def __init__(self,
...
                  style='pytorch',
-                 with_cp=False):
+                 with_cp=False,
+                 normalize=dict(type='BN')):
         super(BasicBlock, self).__init__()
         self.conv1 = conv3x3(inplanes, planes, stride, dilation)
-        self.bn1 = nn.BatchNorm2d(planes)
+
+        # build_norm_layer return: (norm_name, norm_layer)
+        self.norm1, norm1 = build_norm_layer(normalize, planes, postfix=1)
```
hellock: The name of `self.norm1` may be misleading, `self.norm1_name` instead?

mmdet/models/backbones/resnet.py
thangvubk: We need to add 3 properties for bottleneck and 2 for basic block, is that ok?

mmdet/models/backbones/resnet.py
thangvubk: How about using self.norm1_ instead of self.norm1_name for short?

mmdet/models/backbones/resnet.py
```diff
@@ -29,27 +30,40 @@ def __init__(self,
...
-                 with_cp=False):
+                 with_cp=False,
+                 normalize=dict(type='BN')):
         super(BasicBlock, self).__init__()
         self.conv1 = conv3x3(inplanes, planes, stride, dilation)
-        self.bn1 = nn.BatchNorm2d(planes)
+
+        self.norm1_name, norm1 = build_norm_layer(normalize, planes, postfix=1)
+        self.norm2_name, norm2 = build_norm_layer(normalize, planes, postfix=2)
+        self.add_module(self.norm1, norm1)
```
hellock: There is a typo.

mmdet/models/backbones/resnet.py
```diff
@@ -235,19 +275,17 @@ def __init__(self,
         assert max(out_indices) < num_stages
         self.style = style
         self.frozen_stages = frozen_stages
-        self.bn_eval = bn_eval
-        self.bn_frozen = bn_frozen
+        assert (isinstance(normalize, dict) and 'eval_mode' in normalize
+                and 'frozen' in normalize)
+        self.norm_eval = normalize.pop('eval_mode')
```
hellock: We can add a default value for `eval_mode` here.

mmdet/models/backbones/resnet.py
```diff
@@ -262,15 +300,42 @@ def __init__(self,
...
+        self.norm1_name, norm1 = build_norm_layer(self.normalize,
+                                                  64, postfix=1)
+        self.add_module(self.norm1_name, norm1)
+        self.relu = nn.ReLU(inplace=True)
+        self.maxpool = nn.MaxPool2d(kernel_size=3, stride=2, padding=1)
+
+        if self.frozen_stages >= 0:
+            for m in [self.conv1, self.norm1]:
+                for param in m.parameters():
+                    param.requires_grad = False
```
hellock: I think L328-331 can also be moved to `_freeze_stages()`.

thangvubk:
@hellock  All existing configs use batch norm as default, so updating config files is not needed.

mmdet/models/backbones/resnet.py
hellock: Yes I know that. I mean we can just left `eval_mode(True)` and `frozen(False)` optional. Besides, the default value of L34 is `normalize=dict(type='BN')`.

hellock:
Oh It is nice to see that no current config file specifies the fields `bn_eval` and `bn_frozen`, so they are compatible to both versions.

mmdet/models/backbones/resnet.py
thangvubk: I intentionally add `normalize=dict(type, eval_mode, frozen)` to let users know how to specify `normalize` dict in config. If we leave `eval_mode(True)` and `frozen(False)` in the body, it will be subtle to add norm mode in to config files.

hellock: I think it is perfect to use `normalize=dict(type, eval_mode, frozen)` as the default value, which can be a good indicator. It may be better to also allow users to modify configs like `normalize=dict(type='BN', eval_mode=False)` instead of writing out all parameters.

thangvubk: `eval_mode` and `frozen` is mandatory for the backbone. But for other modules (neck, head), just leave `normalize=dict(type='BN')` is oke because `frozen` is optional in `build_norm_layer`. Btw, it is noted that `eval_mode` in neck or head is not supported since it will be overridden by parent modules and there will be error informed.

hellock: Okay we can leave the assertion as it is. For necks and heads, there is no pretrained weights so `eval_mode` is usually set to False. Maybe leave a comment for potential errors?

thangvubk: My idea is add following into `build_norm_layer`

```
if eval_mode in normalize:
    raise Exception('eval_mode is supported in backbone only and prohibited in other modules')
```

Do you have any suggestion?

hellock: I think we don't need a strict error here, maybe just a comment. Besides backbones, there are indeed some cases that other modules also have pretrained weights, e.g., res5 head.

thangvubk: Oh. i didnt notice that. However, I think it should be an error because if other modules have pretrained weights, `eval_mode` should be popped out and processed in the `train()` function, not inside `build_norm_layer`. I think the error message here should be changed to `eval_mode for modules without pretrained weights is not supported` to avoid accidentally adding `eval_mode` to non-pretrained modules.

hellock: Looks good to me.

hellock:
Hi @thangvubk , here are the last few comments.
- ResNeXt will be unusable if this PR is merged, so maybe it is better to resolve the issue at this PR.
- Would you allow us to upload the pretrained GN model to AWS and make it avaiable via open-mmlab://resnet50_gn?

thangvubk:
Hi.
1. If resnext has confict with this PR, i will work on it.
2. You are free to use the pretrained GN models.

thangvubk:
Group norm for resnext has been updated. Here are passed unit tests:

- [x] Benchmark on pretrain ``faster_rcnn_x101_32x4d_fpn_1x``
- [x] Norm name in architecture
- [x] Eval mode
- [x] Frozen mode

OceanPang:
Hi @thangvubk , the work 'abbreation' in norm.py should be `abbreviation`. 

Besides, I suggest you change the initialize order in each block to the original `conv1, bn1, conv2, bn2, conv3, bn3`. This modification will make it convenient for you when you want to transform another trained model into mmdetection, due to the ordered keys saved in checkpoints : ).

mmdet/models/backbones/resnet.py
```diff
@@ -235,19 +278,17 @@ def __init__(self,
         assert max(out_indices) < num_stages
         self.style = style
         self.frozen_stages = frozen_stages
-        self.bn_eval = bn_eval
-        self.bn_frozen = bn_frozen
+        assert (isinstance(normalize, dict) and 'eval_mode' in normalize
+                and 'frozen' in normalize)
+        self.norm_eval = normalize.pop('eval_mode')
```
hellock: There is a bug for `normalize.pop('eval_mode')`. We need to first copy the variable `normalize`, otherwise it will raise an error when we call the init function more than once.
```python
model = ResNet(50)
model = ResNet(50)  # will raise an error
```

mmdet/models/backbones/resnet.py
hellock: After writing some config files, I find that putting `eval_mode` in `normalize` is not a good idea. It may be better to follow the original style to use a variable `norm_eval` in the modules where normalization layers may need to be set as eval mode, e.g., backbones.

In this way, we can use a shared variable in the config file.
```python
normalize = xxxx
models = dict(
backbones=(normalize=normalize, xxx),
neck=(normalize=normalize, xxx),
bbox_head=(normalize=normalize, xxx),
mask_head=(normalize=normalize, xxx)
)
```
Also, we do not need to check the `eval_mode` field in `build_norm_layer`.

thangvubk: It is a nice point I think.

mmdet/models/backbones/resnet.py
thangvubk: Fixed in [82e7545](https://github.com/open-mmlab/mmdetection/pull/160/commits/82e75455fda13d3273b8a8d23ebf34d9f10e1ed8)

configs/mask_rcnn_r50_fpn_gn_2x.py
```diff
@@ -167,14 +167,14 @@
...
-        dict(type='TensorboardLoggerHook')
+        # dict(type='TensorboardLoggerHook')
     ])
 # yapf:enable
 # runtime settings
 total_epochs = 24
 dist_params = dict(backend='nccl')
 log_level = 'INFO'
-work_dir = './work_dirs/mask_rcnn_r50_fpn_1x'
+work_dir = './work_dirs/mask_rcnn_r50_fpn_gn_1x'
```
thangvubk: It should be 2x

PR #163, opened 2018-12-10
Add OHEMSampler

yhcao6:
The results are shown in the below table (GTX 1080).

|                       | box AP/mask AP | Train time on 1080Ti (s/iter) |
|-----------------------|----------------|---------------------|
| Faster R-CNN Baseline | 36.4           |  0.591                   |
| Faster R-CNN OHEM     | 36.6           | 0.637               |
| Mask R-CNN Baseline   | 37.3/34.2      | 0.717                    |
| Mask R-CNN OHEM       | 37.6/34.5      | 0.759               |


mmdet/core/bbox/samplers/ohem_sampler.py
```diff
@@ -0,0 +1,80 @@
...
+                cls_score, _ = self.bbox_head(bbox_feats)
+                loss_pos = self.bbox_head.loss(
+                    cls_score=cls_score,
+                    bbox_pred=None,
+                    labels=assign_result.labels[pos_inds],
+                    label_weights=cls_score.new_ones(cls_score.size(0)),
+                    bbox_targets=None,
+                    bbox_weights=None,
+                    reduce=False)['loss_cls']
+                _, topk_loss_pos_inds = loss_pos.topk(num_expected)
```
hellock: You can reuse the above codes by adding a new method.

mmdet/models/detectors/two_stage.py
```diff
@@ -4,7 +4,7 @@
 from .base import BaseDetector
 from .test_mixins import RPNTestMixin, BBoxTestMixin, MaskTestMixin
 from .. import builder
-from mmdet.core import (assign_and_sample, bbox2roi, bbox2result, multi_apply)
+from mmdet.core import (bbox2roi, bbox2result, build_assigner, build_sampler)
```
hellock: The brackets are unnecessary.

mmdet/models/detectors/two_stage.py
```diff
@@ -102,13 +102,26 @@ def forward_train(self,
...
-                proposal_list,
-                gt_bboxes,
-                gt_bboxes_ignore,
-                gt_labels,
-                cfg=self.train_cfg.rcnn)
+            bbox_assigner = build_assigner(self.train_cfg.rcnn.assigner)
+            bbox_sampler = build_sampler(
+                self.train_cfg.rcnn.sampler,
+                dict(bbox_roi_extractor=self.bbox_roi_extractor,
+                     bbox_head=self.bbox_head))
```
hellock: It might be better to modify the interface of `build_sampler` and `build_assigner` to `def build_assigner(cfg, **default_args)`, so that we can just call `build_sampler(self.train_cfg.rcnn.sampler, bbox_roi_extractor=self.bbox_roi_extractor, bbox_head=self.bbox_head)` here.

hellock:
The api of `BaseSampler` needs to be updated in order to fit `OHEMSampler` and other potential samplers.

```python
class BaseSampler(metaclass=ABCMeta):

    def __init__(self, context):
        self.context = context
        self.pos_sampler = self
        self.neg_sampler = self
```
All the child samplers also need to be modified.

In `TwoStageDetector`, we can initialize it with
```python
bbox_sampler = build_sampler(self.train_cfg.rcnn.sampler, context=self)
```

mmdet/models/detectors/two_stage.py
```diff
@@ -102,13 +102,24 @@ def forward_train(self,
...
-                proposal_list,
-                gt_bboxes,
-                gt_bboxes_ignore,
-                gt_labels,
-                cfg=self.train_cfg.rcnn)
+            bbox_assigner = build_assigner(self.train_cfg.rcnn.assigner)
+            bbox_sampler = build_sampler(
+                self.train_cfg.rcnn.sampler, context=self)
+            num_imgs = img.size(0)
+            assign_results = []
```
hellock: `assign_results` is unnecessary since it is not used later.

PR #173, opened 2018-12-12
Support class agnostic in mask_cross_entropy loss

Zehaos:


hellock:
Thanks for the fix!

mmdet/core/loss/losses.py
```diff
@@ -56,6 +56,8 @@ def weighted_sigmoid_focal_loss(pred,
 def mask_cross_entropy(pred, target, label):
     num_rois = pred.size()[0]
     inds = torch.arange(0, num_rois, dtype=torch.long, device=pred.device)
+    if pred.shape[1] == 1:
+        label.zero_()
```
hellock: We could avoid the inplace modification of `label` in case that it will be used afterwards.

hellock:
Hi @Zehaos The fix is fine, just needs to be implemented in some other way. Would you submit a new PR?

Zehaos:
@hellock I made a new pull request #177

PR #177, opened 2018-12-12
Support class agnostic in mask_cross_entropy loss

Zehaos:


Zehaos:
@hellock Thanks, it is fine now.

Issue #79, opened 2018-10-25
Choosing a different backbone network


adityaarun1:
Hi,

How can one change the backbone network, let's say for example VGG or ResNext etc? 

I see a closed issue marked for v0.6 release [here](https://github.com/open-mmlab/mmdetection/issues/3), and I can also see that `mmcv.cnn` already seem to have VGG. If this is a planned feature, when can we expect it to roll out?

adityaarun1:
Hi @hellock, sorry to bug you again, do you have a fresh timeline for this feature?

hellock:
We plan to add VGG for SSD only. For other models (Faster R-CNN, etc.), ResNet runs faster and performs better so we do not think it is necessary, but community contributions are welcome.
ResNeXt is expected to be released in one or two weeks.

PR #192, opened 2018-12-19
add resnext

OceanPang:


hellock:
#79

thangvubk:
@OceanPang Can you share top 1 acc of resnext50 and train script. I tried to train using resnet with group norm in PR [#160](https://github.com/open-mmlab/mmdetection/pull/160) using facebook's train script but get lower than reported top 1 by 0.3%.

hellock:
@thangvubk The pretrained ResNeXt models here are simply converted from the official release, not trained by ourselves.

PR #301, opened 2019-01-30
fix a little bug in coco datasets loader

AresGao:
fix little potential  bug.......


mmdet/datasets/coco.py
```diff
@@ -40,7 +40,7 @@ def get_ann_info(self, idx):
         img_id = self.img_infos[idx]['id']
         ann_ids = self.coco.getAnnIds(imgIds=[img_id])
         ann_info = self.coco.loadAnns(ann_ids)
-        return self._parse_ann_info(ann_info)
+        return self._parse_ann_info(ann_info,self.with_mask)
```
hellock: A space is needed after the comma.

hellock:
Thanks for the fix! There is a minor comment on the code style.

PR #302, opened 2019-01-30
minor fix, print_map_summary

tjsongzw:
hi, this pr would allow pass dataset.CLASSES as it is, like a list of label names


hellock:
Thanks!

PR #316, opened 2019-02-13
Typo

youkaichao:
correct some typoes

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ We provide a training script using the launch utility provided by PyTorch.
 Supported arguments are:
 
 - --validate: perform evaluation every k (default=1) epochs during the training.
-- --work_dir <WORK_DIR>: if not specified, the path in config file will be overwritten.
+- --work_dir <WORK_DIR>: if specified, the path in config file will be replaced.
 
 Expected results in WORK_DIR:
 
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ We provide a training script using the launch utility provided by PyTorch.
 Supported arguments are:
 
 - --validate: perform evaluation every k (default=1) epochs during the training.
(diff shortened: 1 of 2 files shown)


mmdet/models/registry.py
```diff
@@ -22,9 +22,7 @@ def _register_module(self, module_class):
             module (:obj:`nn.Module`): Module to be registered.
         """
         if not issubclass(module_class, nn.Module):
-            raise TypeError(
-                'module must be a child of nn.Module, but got {}'.format(
-                    type(module_class)))
+            raise TypeError(f'module must be a child of nn.Module, but got {module_class}')
```
youkaichao: `module_class ` is already a class and `type(module_class)` always returns `<class 'type'>`

hellock: To be compatible with python 3.4/3.5, we do not use f-strings in this project. When we discard the support for old python versions some day, we will consider introducing some new features.

README.md
```diff
@@ -184,7 +184,7 @@ We provide a training script using the launch utility provided by PyTorch.
 Supported arguments are:
 
 - --validate: perform evaluation every k (default=1) epochs during the training.
-- --work_dir <WORK_DIR>: if specified, the path in config file will be overwritten.
+- --work_dir <WORK_DIR>: if not specified, the path in config file will be overwritten.
```
hellock: I think this is not a typo. If `--work_dir` is not specified, the path in config file will be used, otherwise the specified work_dir will be used.

youkaichao: Oh, I wrongly take the subject of `specified` to be `work_dir`

youkaichao: Oh, the word `overwritten` is confusing. I think it means that the `path` will be overwritten, discarding any content in it. Actually, it means the `value` in `python` will be overwritten with a new `value`.

PR #329, opened 2019-02-18
Update inference.py

AresGao:
fix inference bug, the default keep_ratio is True...But SSD should be false

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -10,7 +10,7 @@ from mmdet.core import get_classes
 def _prepare_data(img, img_transform, cfg, device):
     ori_shape = img.shape
     img, img_shape, pad_shape, scale_factor = img_transform(
-        img, scale=cfg.data.test.img_scale)
+        img, scale=cfg.data.test.img_scale, keep_ratio=cfg.data.test.resize_keep_ratio)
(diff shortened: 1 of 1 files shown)


tjsongzw:
`cfg.data.test.get('resize_keep_ratio', True)` would be better, it will keep default behavior on other config files.

AresGao:
> `cfg.data.test.get('resize_keep_ratio', True)` would be better, it will keep default behavior on other config files.

Yes, u are right....
But why it checks failed..

tjsongzw:
seems it didn't pass flake8 checking

mmdet/apis/inference.py
```diff
@@ -10,7 +10,8 @@
 def _prepare_data(img, img_transform, cfg, device):
     ori_shape = img.shape
     img, img_shape, pad_shape, scale_factor = img_transform(
-        img, scale=cfg.data.test.img_scale)
+        img, scale=cfg.data.test.img_scale,
```
hellock: You have a trailing whitespace here, which is unexpected and fails the flake8 linting.

If you use yapf to auto-format the code, it should look like this.
```python
img, img_shape, pad_shape, scale_factor = img_transform(
    img,
    scale=cfg.data.test.img_scale,
    keep_ratio=cfg.data.test.get('resize_keep_ratio', True))
```

PR #363, opened 2019-03-05
Bug fix for voc_eval

amirassov:


hellock:
I think it is the same as the official evaluation code, not a bug. Could you share some thoughts or evidence?

amirassov:
I use the function `voc_eval` to evaluate a detector on a custom dataset that does not contain `bboxes_ignore`. This causes an error on the line
https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/evaluation/mean_ap.py#L248, because `gt_ignore is not None`.

PR #373, opened 2019-03-08
fix OHEM with cascade rcnn

Morde-kaiser:


hellock:
Thanks! It is also interesting to know whether there is any difference for the final performance when using OHEM, can you share some results?

Morde-kaiser:
> Thanks! It is also interesting to know whether there is any difference for the final performance when using OHEM, can you share some results?

Morde-kaiser:
@hellock Hi, I tested the performance before and after using cascade rcnn ohem.  Results are as follows：
Without OHEM：
Without OHEM

With OHEM：
With OHEM

Performance has improved！

hellock:
Thanks for the report! I have a question that the mAP without OHEM seems much lower than the model zoo (42.1 for R-101). Is your setting different from the official config file?

Morde-kaiser:
> Thanks for the report! I have a question that the mAP without OHEM seems much lower than the model zoo (42.1 for R-101). Is your setting different from the official config file?

I change imgs_per_gpu to 4, and trian on 3 GPUs. I don't know if this will harm the performance.

hellock:
We tested OHEM on Cascade R-CNN with R-50. The results looks reasonable.
- w/o OHEM (baseline): 40.3
- w/ OHEM: 40.6

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -7,8 +7,9 @@
 from .test_mixins import RPNTestMixin
 from .. import builder
 from ..registry import DETECTORS
-from mmdet.core import (assign_and_sample, bbox2roi, bbox2result, multi_apply,
+from mmdet.core import (bbox2roi, bbox2result,
                         merge_aug_masks)
+from mmdet.core import build_assigner, build_sampler
```
hellock: Since all methods are imported from the same module, a single statement is enough.

configs/cascade_rcnn_ohem_r101_fpn_1x.py
```diff
@@ -0,0 +1,210 @@
...
+    ])
+# yapf:enable
+# runtime settings
+total_epochs = 12
+dist_params = dict(backend='nccl')
+log_level = 'INFO'
+work_dir = './work_dirs/cascade_rcnn_r101_fpn_1x'
+load_from = None
+resume_from = None
+workflow = [('train', 1)]
```
hellock: There are already an OHEM example in the config files and using it for Cascade R-CNN is straight forward. I suggest removing this config file, and post the code snippets in this PR instead.

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -109,8 +109,8 @@ def forward_train(self,
                       img,
                       img_meta,
                       gt_bboxes,
+                      gt_bboxes_ignore,
```
hellock: I suggest keeping the original order and `gt_bboxes_ignore` can be an optional argument.

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -121,8 +121,7 @@ def forward_train(self,
             rpn_outs = self.rpn_head(x)
             rpn_loss_inputs = rpn_outs + (gt_bboxes, img_meta,
                                           self.train_cfg.rpn)
-            rpn_losses = self.rpn_head.loss(
-                *rpn_loss_inputs, gt_bboxes_ignore=gt_bboxes_ignore)
+            rpn_losses = self.rpn_head.loss(*rpn_loss_inputs)
```
hellock: Could you show your consideration for this modification?

mmdet/core/bbox/samplers/ohem_sampler.py
```diff
@@ -12,11 +12,16 @@ def __init__(self,
                  context,
                  neg_pos_ub=-1,
                  add_gt_as_proposals=True,
+                 stages=0,
                  **kwargs):
         super(OHEMSampler, self).__init__(num, pos_fraction, neg_pos_ub,
                                           add_gt_as_proposals)
-        self.bbox_roi_extractor = context.bbox_roi_extractor
-        self.bbox_head = context.bbox_head
+        if not hasattr(context, "num_stages"):
```
hellock: Single quote is used by default in this project.

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -135,13 +134,25 @@ def forward_train(self,
...
-                proposal_list,
-                gt_bboxes,
-                gt_bboxes_ignore,
-                gt_labels,
-                cfg=rcnn_train_cfg)
+            sampling_results = []
+            if self.with_bbox or self.with_mask:
+                bbox_assigner = build_assigner(rcnn_train_cfg.assigner)
+                bbox_sampler = build_sampler(
+                    rcnn_train_cfg.sampler, context=self, stages=i)
```
hellock: I suggest that we add a stage indicator like `self.current_stage = i`, thus it can be obtained in the sampler via `context.current_stage` instead of passing an additional argument here.

Morde-kaiser:
> We tested OHEM on Cascade R-CNN with R-50. The results looks reasonable.
> 
> * w/o OHEM (baseline): 40.3
> * w/ OHEM: 40.6
Thanks! The performance of the two is similar, it seems that there is no need to use OHEM. ：)

hellock:
I think supporting OHEM is nice. Applying OHEM to Faster R-CNN brings only a gain of 0.4 to 0.6, so 0.3 for Cascade R-CNN is reasonable. You can reopen the PR with some fixes.

Morde-kaiser:
> I think supporting OHEM is nice. Applying OHEM to Faster R-CNN brings only a gain of 0.4 to 0.6, so 0.3 for Cascade R-CNN is reasonable. You can reopen the PR with some fixes.

:+1: Ok, I will reopen the PR with some fixes.

Morde-kaiser:
Hi, I resubmitted the PR according to your request. Check it, please!

hellock:
It looks good to me.
The permission of `compile.sh` and `dist_train.sh` is changed from 755 to 644, could you recover it?

Morde-kaiser:
Ok, I will fix it.

PR #403, opened 2019-03-13
Added mask visualization part to inference part and add out_file interface.

Luodian:
I find it's the best way to modify `mmdet/apis/inference.py`, to add the existing mask visualization part into it. Also, since I am using remote servers to run my model, I think most of us do in this way, so I set the `show = False` and add an interface `out_file`.
I'm not sure if it's appropriate to directly pull a request, and sorry to post two boring issues to talk about these. Or maybe you can adopt these changes by next time commit.

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -51,7 +51,7 @@ def inference_detector(model, imgs, cfg, device='cuda:0'):
         return _inference_generator(model, imgs, img_transform, cfg, device)
 
 
-def show_result(img, result, dataset='coco', score_thr=0.3, outfile=None):
+def show_result(img, result, dataset='coco', score_thr=0.3, out_file=None):
     img = mmcv.imread(img)
     class_names = get_classes(dataset)
     if isinstance(result, tuple):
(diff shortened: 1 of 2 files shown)


mmdet/apis/inference.py
```diff
@@ -50,18 +51,34 @@ def inference_detector(model, imgs, cfg, device='cuda:0'):
         return _inference_generator(model, imgs, img_transform, cfg, device)
 
 
-def show_result(img, result, dataset='coco', score_thr=0.3):
+def show_result(img, result, dataset='coco', score_thr=0.3, outfile=None):
```
hellock: `out_file` is more preferred.

mmdet/apis/inference.py
```diff
@@ -50,18 +51,34 @@ def inference_detector(model, imgs, cfg, device='cuda:0'):
...
-    bboxes = np.vstack(result)
-    img = mmcv.imread(img)
     mmcv.imshow_det_bboxes(
         img.copy(),
         bboxes,
         labels,
         class_names=class_names,
-        score_thr=score_thr)
+        score_thr=score_thr,
+        show=False,
```
hellock: show=out_file is None

Luodian: Thanks for the wise advice.

PR #490, opened 2019-04-13
add cpp sigmoid focal loss

yhcao6:
The cpp_sigmoid_focal_loss extension is modified from maskrcnn-benchmark and adapt to mmdetection. Compared with original python implementation, speed is faster and memory usage is lower.

| Retinanet R50 FPN     | Mem (GB) | Time (iter/s) |
|-----------------------|----------|---------------|
| Python implementation | 5.0      | 0.374         |
| Cpp implementation    | 3.6      | 0.356         |

### DIFF
--- a/mmdet/ops/sigmoid_focal_loss/functions/sigmoid_focal_loss.py
+++ b/mmdet/ops/sigmoid_focal_loss/functions/sigmoid_focal_loss.py
@@ -1,0 +1,34 @@
+from torch.autograd import Function
+from torch.autograd.function import once_differentiable
+
+from .. import sigmoid_focal_loss
+
+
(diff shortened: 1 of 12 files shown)


mmdet/ops/sigmoid_focal_loss/functions/sigmoid_focal_loss.py
```diff
@@ -0,0 +1,34 @@
+from torch.autograd import Function
+from torch.autograd.function import once_differentiable
+
+from .. import sigmoid_focal_loss
+
+
+class _SigmoidFocalLoss(Function):
```
hellock: We can just follow the naming convention of this repo to name it as `SigmoidFocalLossFunction`.

PR #492, opened 2019-04-14
remove expand loop in bbox head to speed up

yhcao6:
| Model            | Mem (GB) | Time (iter/s) |
|------------------|----------|---------------|
| Mask R50 FPN ori | 4.0      | 0.542         |
| Mask R50 FPN new | 4.0      | 0.468         |

### DIFF
--- a/mmdet/models/bbox_heads/bbox_head.py
+++ b/mmdet/models/bbox_heads/bbox_head.py
@@ -1,9 +1,9 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-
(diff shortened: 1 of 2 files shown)


mmdet/models/bbox_heads/bbox_head.py
```diff
@@ -94,10 +94,16 @@ def loss(self,
                 cls_score, labels, label_weights, reduce=reduce)
             losses['acc'] = accuracy(cls_score, labels)
         if bbox_pred is not None:
+            pos_mask = labels > 0
```
hellock: It may be better to rename it as `pos_inds`.

PR #494, opened 2019-04-15
fix bug in two stage

yhcao6:


mmdet/models/anchor_heads/anchor_head.py
```diff
@@ -127,7 +127,6 @@ def get_anchors(self, featmap_sizes, img_metas):
 
     def loss_single(self, cls_score, bbox_pred, labels, label_weights,
                     bbox_targets, bbox_weights, num_total_samples, cfg):
-        # classification loss
```
hellock: This deletion is unnecessary.

mmdet/core/loss/losses.py
```diff
@@ -23,6 +23,8 @@ def weighted_cross_entropy(pred, label, weight, avg_factor=None, reduce=True):
 
 
 def weighted_binary_cross_entropy(pred, label, weight, avg_factor=None):
+    if pred.size() != label.size():
```
hellock: `pred.dim() != label.dim()` may be more comprehensible.

PR #548, opened 2019-04-26
new model zoo

yhcao6:


thangvubk:
It's amazing. I think it would be nice if you link the PRs for speeding up here.

hellock:
#254
#490 
#492
Using pytorch-nightly instead of pytorch 1.0.1 also boosts the speed.

Issue #571, opened 2019-05-01
Add support for SyncBN (PyTorch 1.1)


hellock:
PyTorch 1.1 ships with an official implementation of SyncBN.

PR #577, opened 2019-05-03
add pytorch 1.1.0 SyncBN support

xvjiarui:
This PR solves https://github.com/open-mmlab/mmdetection/issues/571. 

Actually, this repo could support pytorch 1.1.0 without any changes. The `SyncBN` is shipped with pytorch 1.1.0 release. We adopted the official implementation of `SyncBN` when building models. 

Since `SyncBN` only support `torch.nn.DistributedDataParallel` officially, we set `ddp_gpu_size` on L49:50  to support `MMDistributedDataParallel`.


hellock:
Thanks! There is a potential bug for SyncBN in the backbones. If frozen_stages>=0, the BN layers in the frozen stages will not be set as eval mode.

Calling `_freeze_stages` at the end of `train` may fix the bug.
BTW, `BatchNorm2d ` [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/backbones/resnet.py#L479) can be replaced with `_BatchNorm `.

xvjiarui:
Thanks for pointing out. I deleted the `_freeze_stages` in the `__init__` and add it after `train`.  
I also change the `BatchNorm2d` to protected member `_BatchNorm`.

thangvubk:
If I'm not wrong, ``frozen_stages = frozen + norm_eval``. I think the ``frozen`` (which is used in norm config) should be changed to ``requires_grad`` since ``frozen`` and ``frozen_stages`` have different behaviors and it can be confusing.

Joker316701882:
@thangvubk 
I guess you might mixed up with some definitions here.
In my opinion, frozen is only responsible for weight and bias in **bn** layer, not in **conv** layers, it controls the bn layers across the **whole backbone**, but the running mean and var can still be update. 
frozen_stages are responsible for weight and bias in **both bn and conv** layers **only in specified** stages, not across whole backbone. And in latest commit, it also sets the bn in specified stages into eval mode.
norm_eval brutally set all bn layers in backbone into eval mode.

So, the definition here might be a little complex, but each of them can be extremely useful in some special cases.

Joker316701882:
@hellock 
So with the recent commit in resnet.py

'''

    def _freeze_stages(self):
        if self.frozen_stages >= 0:
            self.norm1.eval()
            for m in [self.conv1, self.norm1]:
                for param in m.parameters():
                    param.requires_grad = False

        for i in range(1, self.frozen_stages + 1):
            m = getattr(self, 'layer{}'.format(i))
            m.eval()
            for param in m.parameters():
                param.requires_grad = False

'''

there is no need to add __freeze_stages() elsewhere, like in train(). And SyncBN will still work fine. Am  I right?

hellock:
@thangvubk I agree with that. `requires_grad` is more accurate for the behavior. You can create another PR for the renaming.

hellock:
@Joker316701882 Most of your understanding is correct.
`frozen` is an argument of the config dict, which is responsible for weight and bias in BN layers. See [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/utils/norm.py#L45) and [here](https://github.com/open-mmlab/mmdetection/blob/master/configs/gn/mask_rcnn_r50_fpn_gn_2x.py#L2).
`frozen_stages()` does two things: 1) set `requires_grad` for all parameters of certain stages in the backbone to False, including conv and BN, and 2) set BN mode to eval.
`norm_eval` controls the mode of all BN layers in the backbone.

In the original implementation, `_freeze_stages()` should also be added at the end of `train()`. Considering the following case that we want to freeze the first stage and use SyncBN at the same time.
`_freeze_stages()` is called in `__init__()`, but when `train()` is called during training, BN layers in the first stage will be set back to train() mode.
Replacing `BatchNorm2d` with `_BatchNorm` at [`train()`](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/backbones/resnet.py#L477) seems to solve the problem.

Joker316701882:
@hellock 
Thanks. I understand the necessity of adding _freeze_stages() in train() now.
Just a mention, if I use the SyncBN from [apex](https://github.com/NVIDIA/apex), there is no need to replace BatchNorm2d with _BatchNorm in train() function. Is it right?

mmdet/models/backbones/resnet.py
```diff
@@ -473,8 +474,12 @@ def forward(self, x):
...
-                    m.eval()
+        if mode:
+            self._freeze_stages()
+            if self.norm_eval:
+                for m in self.modules():
+                    # trick: eval have effect on BatchNorm only
+                    if isinstance(m, _BatchNorm):
+                        m.eval()
+        else:
+            convert_batchnorm(self)
```
hellock: This may lead to a bug since `train()` and `eval()` are called alternately. If `eval()` is called and the SyncBN layers are converted to normal BN, in the next epoch it will not be converted back.

hellock: The necessity of converting SyncBN to BN during testing is caused by current implementation of multi-gpu testing. The advantage that we can run multiple process on each GPU is not significant. Maybe we can discard the implementation and turn to ddp during testing.

chensnathan: Maybe after `convert_batchnorm`, `eval( )` for BN layers is needed. In the `train` function, `eval( )` is called for whole network in `super(ResNet, self).train(mode)`, but after `convert_batchnorm` function, the BN layers are in `train` mode, which may lead to a significance performance drop.

xvjiarui: Thanks very much! That's would be much better than converting! `convert_batchnorm` will be deleted soon.

xvjiarui: That shouldn't be a problem since the `eval` or `train` is called in recursive way. But I would discard the `convert_batchnorm` since the mmddp testing is available now.

PR #607, opened 2019-05-07
 Code for CVPR-2019 Paper Libra R-CNN

OceanPang:


hellock:
This PR is a little mixed. It is better to be split into 3 PRs:
1. Refactoring for losses
2. Add a parent class for FPN
3. Libra R-CNN

PR #653, opened 2019-05-16
update docs

liushuchun:
'mask' option has been removed & replaced with 'segm'


hellock:
Thanks for the fix!

liushuchun:
you are welcome!

PR #668, opened 2019-05-20
Reformat use yapf 0.27.0

OceanPang:


Beginc:
Reformat use yapf 0.27.0

PR #713, opened 2019-05-25
fix bug in iou loss style

thangvubk:


myownskyW7:
Thank you so much for this bug report.

PR #733, opened 2019-05-29
Avoid log(0) in iou_loss, which will lead to Nan.

Joker316701882:


hellock:
Thanks for the fix! It may be better to set `eps=1e-6` as an argument of the method.

PR #734, opened 2019-05-29
Fix access to a wrong key in checkpoint

sovrasov:


hellock:
Thanks for the fix! This is indeed a bug, but it should be `classes` in both L36 and L37.

sovrasov:
Ok, but it's still looks strange. The model has attribute `CLASSES` and I've thought it would be directly written to a checkpoint under the same name.

hellock:
That also sounds good to me. You may modify https://github.com/open-mmlab/mmdetection/blob/master/tools/train.py#L82 as well to unify the names.

PR #750, opened 2019-06-03
Travis CI: Add Python 3.7 to the testing

cclauss:


hellock:
Thanks for the PR! Have you verified that travis has supported the modified setting? I tried that earlier but it was not supported at that time.

cclauss:
https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

PR #673, opened 2019-05-20
Add a lightweight ssd-based face detector

sovrasov:
This PR contains
- Reader for the WIDER face dataset
- Lightweight single head SSD face detector based on the MobileNetV2
- Tools to export SSD into onnx format with ability to deploy the resulting snapshots via [Open VINO](https://software.intel.com/en-us/openvino-toolkit)

cc @snosov1

PR #765, opened 2019-06-06
Add support of the wider face dataset

sovrasov:
This PR contains:
- reader of the WIDER Face in the VOC format
- README with instructions how to prepare directory structure
- sample of a training config

### DIFF
--- a/configs/wider/README.md
+++ b/configs/wider/README.md
@@ -1,0 +1,28 @@
+## WIDER Face Dataset
+
+To use the WIDER Face dataset you need to download it
+and extract to the `data/WIDER` folder. Annotation in the VOC format
+can be found in this [repo](https://github.com/sovrasov/wider-face-pascal-voc-annotations.git).
+You should move the annotation files from `WIDER_train_annotations` and `WIDER_val_annotations` folders
+to the `Annotation` folders inside the corresponding directories `WIDER_train` and `WIDER_val`.
+Also annotation lists `val.txt` and `train.txt` should be copied to `data/WIDER` from `WIDER_train_annotations` and `WIDER_val_annotations`.
+The directory should be like this:
+
+```
+mmdetection
+├── mmdet
+├── tools
+├── configs
+├── data
+│   ├── WIDER
+│   │   ├── WIDER_train
+│   |   │   ├── Annotations
+│   │   ├── WIDER_val
+│   |   │   ├── Annotations
+│   │   ├── val.txt
+│   │   ├── train.txt
+
+```
+
+After that you can train the SSD300 on WIDER by launching training with the `ssd300_wider.py` config or
+create your own config based on the presented one.
--- a/configs/wider/README.md
+++ b/configs/wider/README.md
@@ -16,8 +16,12 @@ mmdetection
 ├── data
 │   ├── WIDER
 │   │   ├── WIDER_train
+│   |   │   ├──0--Parade
+│   |   │   ├── ...
 │   |   │   ├── Annotations
 │   │   ├── WIDER_val
+│   |   │   ├──0--Parade
+│   |   │   ├── ...
 │   |   │   ├── Annotations
 │   │   ├── val.txt
 │   │   ├── train.txt
--- a/configs/wider/ssd300_wider.py
+++ b/configs/wider/ssd300_wider.py
@@ -1,0 +1,135 @@
+# model settings
+input_size = 300
+model = dict(
+    type='SingleStageDetector',
+    pretrained='open-mmlab://vgg16_caffe',
+    backbone=dict(
+        type='SSDVGG',
+        input_size=input_size,
+        depth=16,
+        with_last_pool=False,
+        ceil_mode=True,
+        out_indices=(3, 4),
+        out_feature_indices=(22, 34),
+        l2_norm_scale=20),
+    neck=None,
+    bbox_head=dict(
+        type='SSDHead',
+        input_size=input_size,
+        in_channels=(512, 1024, 512, 256, 256, 256),
+        num_classes=2,
+        anchor_strides=(8, 16, 32, 64, 100, 300),
+        basesize_ratio_range=(0.15, 0.9),
+        anchor_ratios=([2], [2, 3], [2, 3], [2, 3], [2], [2]),
+        target_means=(.0, .0, .0, .0),
+        target_stds=(0.1, 0.1, 0.2, 0.2)))
+cudnn_benchmark = True
+train_cfg = dict(
+    assigner=dict(
+        type='MaxIoUAssigner',
+        pos_iou_thr=0.5,
+        neg_iou_thr=0.5,
+        min_pos_iou=0.,
+        ignore_iof_thr=-1,
+        gt_max_assign_all=False),
+    smoothl1_beta=1.,
+    allowed_border=-1,
+    pos_weight=-1,
+    neg_pos_ratio=3,
+    debug=False)
+test_cfg = dict(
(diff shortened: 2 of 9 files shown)


vadik6666:
@sovrasov how to get val.txt and train.txt ? How they should look like?
By default I this [repo](https://github.com/sovrasov/wider-face-pascal-voc-annotations.git) did not create these annotation files.
```
mmdetection
├── mmdet
├── tools
├── configs
├── data
│   ├── WIDER
│   │   ├── WIDER_train
│   |   │   ├── Annotations
│   │   ├── WIDER_val
│   |   │   ├── Annotations
│   │   ├── val.txt
│   │   ├── train.txt

```

sovrasov:
@vadik6666 thanks for catch! I've forgot to merge them. See PR https://github.com/sovrasov/wider-face-pascal-voc-annotations/pull/2/files

vadik6666:
@sovrasov I made the same annotation files and I get 
> FileNotFoundError: img file does not exist: data/WIDER/WIDER_train/57--Angler/57_Angler_peoplefishing_57_861.jpg

vadik6666:
May be we should have this structure:
```
mmdetection
├── mmdet
├── tools
├── configs
├── data
│   ├── WIDER
│   │   ├── WIDER_train
│   |   │   ├── Annotations
│   |   │   ├── images
│   │   ├── WIDER_val
│   |   │   ├── Annotations
│   |   │   ├── images
│   │   ├── val.txt
│   │   ├── train.txt

```
and in config file change line
img_prefix=[data_root + 'WIDER_train/'],
to
img_prefix=[data_root + 'WIDER_train/images'],
?

vadik6666:
Maybe in [dataset/wider.py](https://github.com/sovrasov/mmdetection/blob/wider/mmdet/datasets/wider.py#L33) we should change
`img_infos.append(
dict(id=img_id, filename=osp.join(folder, filename), width=width, height=height))`
to
`img_infos.append(
dict(id=img_id, filename=root.find('path').text, width=width, height=height))`
?

sovrasov:
Current scripts should work correctly. Do you have all the folders like `0--Parade` right inside the  `WIDER_train` folder?

vadik6666:
Aaa, I see your last modification of readme. Yes it will work now. But I didn't have `0--Parade` inside `WIDER_train`. Thats why I changed `filename=osp.join(folder, filename)` to `filename=root.find('path').text` and it works for me.

BlueBlueFF:
@sovrasov Thank you for your code! And I implemented the experiments about wider-face(backbone:resnet 18, head:fcos), but the result seems not very well. The precision and recall are very low(especially for recall). Do you have implemented some training about wider-face and meet this problem?

sovrasov:
@BlueBlueFF I haven't trained any fcos-based detector yet, so I don't know the optimal training strategy for it. WIDER has huge amount of tiny faces and it can be a problem for the most detectors. 
SSD300 config provided in this PR gives recall 0.434 precision 0.029 and AP 0.347. All measurements are done using tools/test.py and then tools/voc_eval.py

BlueBlueFF:
> @BlueBlueFF I haven't trained any fcos-based detector yet, so I don't know the optimal training strategy for it. WIDER has huge amount of tiny faces and it can be a problem for the most detectors.
> SSD300 config provided in this PR gives recall 0.434 precision 0.029 and AP 0.347. All measurements are done using tools/test.py and then tools/voc_eval.py

Thank you for sharing your result, I will do more expriments on widerface.

sovrasov:
@yhcao6 how is the review process? Do I need to change something?

yhcao6:
> @yhcao6 how is the review process? Do I need to change something?

Sorry to reply to you so late. There are no other suggestions but I am a little wired about following code:
```
 if self.min_size:
     w = bbox[2] - bbox[0]
     h = bbox[3] - bbox[1]
     if w < self.min_size or h < self.min_size:
         ignore = True
```
The logic is that when a box is too small, these boxes will be ignored during training as well as **validation**. Is it a convention in `WIDER` dataset to ignore these small boxes? Otherwise, it may be unusual to ignore boxes by ourselves.

sovrasov:
@yhcao6 these boxes will be ignored in validation only if `min_size` will be directly specified in val config. It can be useful because we don't have implementation of the original WIDER metric, but still can get some additional info about performance in different usecases.
In my SSD300 config small boxes are not ignored in val and test configs. Do you mean the case when somebody performs validation on the train?

sovrasov:
@yhcao6  You can also look at README.md from #673. 
According too measured AP = 0.305 seems like MobilenetV2-based single head detector doesn't work at all, but AP on faces > 30 pix = 0.790, so it works for the target usecase.

yhcao6:
In such case it would be better to add an `assert` to exclude this potential danger.

```
if self.min_size:
    assert not self.test_mode
```

sovrasov:
@yhcao6 done

yhcao6:
There are two other problems:

1. `WIDER` and `WIDER face` are two different datasets. Replacing `wider` with `wider_face` to prevent misunderstanding.

2. The import order in `wider.py` is incorrect, it should be
```
import os.path as osp
import xml.etree.ElementTree as ET

import mmcv

from .xml_style import XMLDataset
```

yhcao6:
Thanks for fixing it! It seems the config file is ignored, could you fix it as well to make it consistent?

hellock:
Thanks! @sovrasov

ElegantGod:
It's unnecessary from widerface annotation to voc annotation!
just create a widerface.py, 
```
import os.path as osp

import mmcv
import numpy as np
import cv2

from .custom import CustomDataset


class WIDERFACEDataset(CustomDataset):
    CLASSES = ('face',)

    def __init__(self, **kwargs):
        self.cat2label = {cat: i + 1 for i, cat in enumerate(self.CLASSES)}
        super(WIDERFACEDataset, self).__init__(**kwargs)

    def load_annotations(self, ann_file):

        img_infos = []
        with open(ann_file, 'r') as file:
            annos = file.readlines()
        self.ann_infos_dict = {}
        count = 0
        while count < len(annos):
            name = str(annos[count]).rstrip()
            filename = name
            filepath = osp.join(self.img_prefix, filename)
            img = cv2.imread(filepath)
            width = img.shape[1]
            height = img.shape[0]
            count += 1
            bboxes = []
            labels = []
            bboxes_ignore = []
            labels_ignore = []
            n_anno = int(annos[count])
            for i in range(n_anno):
                count += 1
                bbox = annos[count].split(' ')[0:4]
                bbox = [int(round(float(x))) for x in bbox]
                x1 = max(0, bbox[0])
                y1 = max(0, bbox[1])
                bbox = [
                    x1,
                    y1,
                    int(min(x1+bbox[2], width)),
                    int(min(y1+bbox[3], height))
                ]
                label = self.cat2label['face']
                bboxes.append(bbox)
                labels.append(label)
            bboxes_ignore = np.zeros((0,4))
            labels_ignore = np.zeros((0, ))
            bboxes = np.array(bboxes, ndmin=2) - 1
            labels = np.array(labels)
            ann = dict(
                bboxes=bboxes.astype(np.float32),
                labels=labels.astype(np.int64),
                bboxes_ignore=bboxes_ignore.astype(np.float32),
                labels_ignore=labels_ignore.astype(np.int64))
            self.ann_infos_dict[name] = ann
            count += 1
            img_infos.append(dict(id=name, filename=filename, width=width, height=height))

        return img_infos

    def get_ann_info(self, idx):
        img_id = self.img_infos[idx]['id']
        return self.ann_infos_dict[img_id]
```

PR #773, opened 2019-06-09
fix download of pre-trained weights

innerlee:
Only download one snapshot.


innerlee:
Close in favor of https://github.com/open-mmlab/mmdetection/pull/781

PR #833, opened 2019-06-19
Add MobileNetV2 SSD-Lite

sovrasov:
This PR adds:
- MobileNetV2 backbone for SSD
- Lite heads for SSD
- Automatic regression/classification losses balancing for SSD (based on [paper](http://openaccess.thecvf.com/content_cvpr_2018/papers/Kendall_Multi-Task_Learning_Using_CVPR_2018_paper.pdf))
- Ability to specify anchors in SSD manually
- Sample configs to demonstrate the above features

### DIFF
--- a/configs/ssd300_mobilenetv2_coco.py
+++ b/configs/ssd300_mobilenetv2_coco.py
@@ -5,7 +5,8 @@ model = dict(
     pretrained='./snapshots/mobilenet_v2.pth.tar',
     backbone=dict(
         type='SSDMobilenetV2',
-        input_size=input_size
+        input_size=input_size,
+        activation_type='relu6'
         ),
     neck=None,
     bbox_head=dict(
@@ -14,7 +15,7 @@ model = dict(
         in_channels=(576, 1280, 512, 256, 256, 128),
         num_classes=81,
         anchor_strides=(16, 30, 60, 100, 150, 300),
-        basesize_ratio_range=(0.15, 0.9),
+        basesize_ratio_range=(0.2, 0.95),
         anchor_ratios=([2], [2, 3], [2, 3], [2, 3], [2], [2]),
         target_means=(.0, .0, .0, .0),
         target_stds=(0.1, 0.1, 0.2, 0.2),
@@ -122,7 +123,7 @@ log_config = dict(
 total_epochs = 24
 dist_params = dict(backend='nccl')
 log_level = 'INFO'
-work_dir = './work_dirs/ssd300_coco'
+work_dir = './work_dirs/ssd300_mobilenet_v2_coco'
 load_from = None
 resume_from = None
 workflow = [('train', 1)]
--- a/configs/ssd300_mobilenetv2_coco.py
+++ b/configs/ssd300_mobilenetv2_coco.py
@@ -1,0 +1,128 @@
+# model settings
+input_size = 300
+model = dict(
+    type='SingleStageDetector',
+    pretrained='./snapshots/mobilenet_v2.pth.tar',
+    backbone=dict(
+        type='SSDMobilenetV2',
+        input_size=input_size
+        ),
+    neck=None,
+    bbox_head=dict(
+        type='SSDHead',
+        input_size=input_size,
+        in_channels=(576, 1280, 512, 256, 256, 128),
+        num_classes=81,
+        anchor_strides=(16, 30, 60, 100, 150, 300),
+        basesize_ratio_range=(0.15, 0.9),
+        anchor_ratios=([2], [2, 3], [2, 3], [2, 3], [2], [2]),
+        target_means=(.0, .0, .0, .0),
+        target_stds=(0.1, 0.1, 0.2, 0.2),
+        depthwise_heads=True)
+        )
+cudnn_benchmark = True
+train_cfg = dict(
+    assigner=dict(
+        type='MaxIoUAssigner',
+        pos_iou_thr=0.5,
+        neg_iou_thr=0.5,
+        min_pos_iou=0.,
+        ignore_iof_thr=-1,
+        gt_max_assign_all=False),
+    smoothl1_beta=1.,
+    allowed_border=-1,
+    pos_weight=-1,
+    neg_pos_ratio=3,
+    debug=False)
+test_cfg = dict(
+    nms=dict(type='nms', iou_thr=0.45),
+    min_bbox_size=0,
+    score_thr=0.02,
+    max_per_img=200)
+# model training and testing settings
+# dataset settings
+dataset_type = 'CocoDataset'
+data_root = 'data/MS_COCO/'
+img_norm_cfg = dict(mean=[0, 0, 0], std=[255, 255, 255], to_rgb=True)
+data = dict(
+    imgs_per_gpu=50,
+    workers_per_gpu=3,
+    train=dict(
+        type='RepeatDataset',
+        times=5,
+        dataset=dict(
+            type=dataset_type,
+            ann_file=data_root + 'annotations/instances_train2017.json',
+            img_prefix=data_root + 'train2017/',
+            img_scale=(300, 300),
+            img_norm_cfg=img_norm_cfg,
+            size_divisor=None,
+            flip_ratio=0.5,
+            with_mask=False,
+            with_crowd=False,
+            with_label=True,
+            test_mode=False,
+            extra_aug=dict(
+                photo_metric_distortion=dict(
+                    brightness_delta=32,
+                    contrast_range=(0.5, 1.5),
+                    saturation_range=(0.5, 1.5),
+                    hue_delta=18),
+                expand=dict(
+                    mean=img_norm_cfg['mean'],
+                    to_rgb=img_norm_cfg['to_rgb'],
+                    ratio_range=(1, 4)),
+                random_crop=dict(
+                    min_ious=(0.1, 0.3, 0.5, 0.7, 0.9), min_crop_size=0.3)),
+            resize_keep_ratio=False)),
+    val=dict(
+        type=dataset_type,
+        ann_file=data_root + 'annotations/instances_val2017.json',
+        img_prefix=data_root + 'val2017/',
+        img_scale=(300, 300),
+        img_norm_cfg=img_norm_cfg,
+        size_divisor=None,
+        flip_ratio=0,
+        with_mask=False,
+        with_label=False,
+        test_mode=True,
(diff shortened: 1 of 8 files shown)


d-li14:
@sovrasov thanks for the contribution. I wonder if you have trained this lightweight detector on the COCO dataset. If the answer is yes, how about its performance.

d-li14:
BTW, there exists a stronger pre-trained backbone model in https://github.com/d-li14/mobilenetv2.pytorch, which achieves slightly higher accuracy than the TensorFlow official model.

sovrasov:
@d-li14 not yet, but I'm going to train it. Current config is adopted form SSD300 and unlikely optimal for this model. 
[offtopic] I'ts an interesting topic: does better imagenet weights for the same backbone improve detection performance? I think that 1-3% difference for the same backbone doesn't matter since imagenet weights are going to be partially destructed by higher (than it was on the final steps of imagenet pretraining) learning rate of coco training. Here warmup is used to prevent completely discarding of old weights, but wieghts are definitely corrupted. At the same moment we can't start with the same learning rate as it was  at the end of pretraining, because it's obviously not enough for good results. @d-li14 what is you intuition about that? Did you perform such experiments with different transfer learning strategies?

d-li14:
Right, training MobileNetV2 SSD-Lite needs a different configuration (such as the prior box sizes, channels of detection heads and opt hyperparameters) from original SSD-VGG16.  I have made some attempts with respect to the training configs around half a year ago but didn't get decent performance.

For [offtopic], I agree that it's not a golden rule that a better pre-trained model renders better detection performance after fine-tuning, though I haven't conducted rigorous ablation experiments about that. I just provide a seemingly better backbone here for reference, maybe you could try with it for comparison.

sovrasov:
I've reproduced the architecture of SSD-Lite (amount of heads, channels, anchors design) close to description in paper, but my main concern is optimization strategy.

mmdet/models/anchor_heads/ssd_head.py
```diff
@@ -21,77 +21,114 @@ def __init__(self,
...
+                anchor_generator.base_anchors = torch.index_select(
+                    anchor_generator.base_anchors, 0,
+                    torch.LongTensor(indices))
+                self.anchor_generators.append(anchor_generator)
 
         self.target_means = target_means
         self.target_stds = target_stds
         self.use_sigmoid_cls = False
         self.cls_focal_loss = False
+        self.loss_balancing = loss_balancing
```
OceanPang: Hi, can you provide a more detail ablation study about the loss balancing method from Kendal, etc? We have also tried the method on Faster R-CNN and other detectors but found unstable results (can't work end-to-end but can indeed learn a useful loss weight that can be used offline and bring improvements vs. baseline). If the method indeed helps in some cases, we hope the method can be implemented in loss evaluator as a general method.

mmdet/datasets/extra_aug.py
```diff
@@ -66,15 +66,17 @@ def __call__(self, img, boxes, labels):
...
             self.mean = mean
         self.min_ratio, self.max_ratio = ratio_range
+        self.prob = prob
 
     def __call__(self, img, boxes, labels):
-        if random.randint(2):
+        if random.uniform(0, 1) > self.prob:
             return img, boxes, labels
 
         h, w, c = img.shape
```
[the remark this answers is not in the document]
sovrasov: It's true, but only of `self.prob == 0.5`. For all other cases we need a real-valued random number here.

mmdet/models/anchor_heads/ssd_head.py
sovrasov: I've tested this method with SSD only. On WIDER dataset it helps to adapt tiny detector for big faces and generally prevents from overfitting on particular object size.
This scheme is not a plug-and-play method and requires tuning of optimization strategy. At least one need to increase learning rate to maintain the same amplitude of gradients.
Also this balancing method does each adjustment of weights based on previous minibatch. May be for huge detectors when the batch size is 2-4 this delay is critical and batches are too different on early stages, so such adjustment makes training unstable.

sovrasov:
@ujsyehao could you give me a link to the official MobilenetV2-SSd config?
I've used [this one](https://github.com/tensorflow/models/blob/master/research/object_detection/samples/configs/ssdlite_mobilenet_v2_coco.config) and it has 5 anchors per SSD head.

mmdet/models/backbones/mobilenetv2.py
```diff
@@ -0,0 +1,253 @@
+import torch
+import torch.nn as nn
+from mmcv.cnn import (constant_init, kaiming_init, normal_init)
+from ..registry import BACKBONES
+
+
+def conv_bn(inp, oup, stride, groups=1, act_fn=nn.ReLU):
```
OceanPang: We have a [ConvModule](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/utils/conv_module.py) in this codebase. Use the designed module to act conv/bn/relu would be better.

mmdet/models/backbones/mobilenetv2.py
```diff
@@ -0,0 +1,253 @@
...
+        if self.use_res_connect:
+            if self.ssd_output:
+                expanded, conv = self.compute_conv(x)
+                return (expanded, x + conv)
+            return x + self.compute_conv(x)
+        else:
+            return self.compute_conv(x)
+
+
+@BACKBONES.register_module
```
OceanPang: Considering mobilenet maybe used in other detectors, can you first write a class for standard mobilenet and later inherit a SSDMobilenetV2 from it?

mmdet/models/backbones/mobilenetv2.py
```diff
@@ -0,0 +1,253 @@
...
+class SSDMobilenetV2(nn.Module):
+    def __init__(self, input_size, width_mult=1.0,
+                 activation_type='relu',
+                 single_scale=False):
```
OceanPang: Similar to SSDVGG, we prefer all hyper-parameters can be defined with one dict. Therefore, it would be great if there are no hard code settings in the __init__().

configs/ssd300_mobilenetv2_coco.py
```diff
@@ -0,0 +1,129 @@
...
+    min_bbox_size=0,
+    score_thr=0.02,
+    max_per_img=200)
+# model training and testing settings
+# dataset settings
+dataset_type = 'CocoDataset'
+data_root = 'data/coco/'
+img_norm_cfg = dict(mean=[0, 0, 0], std=[255, 255, 255], to_rgb=True)
+data = dict(
+    imgs_per_gpu=70,
```
OceanPang: For benchmarking, could you please provide a comparison with ssd300 baseline we provided? It would be great if the two configs share the same hyper-parameters if possible (lr, epoch).

mmdet/models/anchor_heads/ssd_head.py
```diff
@@ -21,78 +21,115 @@ def __init__(self,
...
         reg_convs = []
         cls_convs = []
         for i in range(len(in_channels)):
-            reg_convs.append(
-                nn.Conv2d(
+            if depthwise_heads:
+                reg_conv = nn.Sequential(
+                    nn.Conv2d(in_channels[i], in_channels[i],
+                              kernel_size=3, padding=1, groups=in_channels[i]),
+                    nn.BatchNorm2d(in_channels[i]),
```
OceanPang: BN is commonly not used without multi-gpu synchronization. Are you sure about this setting? Or is there any difference without the bns?

configs/ssd300_mobilenetv2_coco.py
OceanPang: Then number `imgs_per_gpu = 70` seems too large... which gpu are you using..

sovrasov: I use RTX2080Ti with 11GB of memory. MobilenetV2 is much smaller than VGG-16. 70 images is ok, I even have 1GB of free space.

sovrasov: Ok, I'll preserve the number of epochs, but MobileNetV2 requires higher learning rate. It would be the only difference. Currently my best snapshot under these conditions gives 19.6 mAP vs 25.7 (ssd300).

mmdet/models/anchor_heads/ssd_head.py
sovrasov: I didn't tested it without bns. Lite heads are useful in case of small detectors when batch size is sufficiently large. I had no problems with this head in distributed training mode even without synchronization, because BN has enough samples to learn statistics.
In backbone I also have BNs and everything is ok.

configs/ssd300_mobilenetv2_coco.py
BlueBlueFF: > I use RTX2080Ti with 11GB of memory. MobilenetV2 is much smaller than VGG-16. 70 images is ok, I even have 1GB of free space.

Do we need to frozen some stages if we use MobilenetV2 pretrained models?  If so, we should add some codes for frozen functions.

sovrasov: I think we shouldn't. In case of low capacity backbone frozen stages will harm the final performance.

hangg7:
@sovrasov Thanks for your contribution! Do we have any update on this thread? Have you reproed paper’s result?

sovrasov:
I've suspended this activity because unfortunately have no time to make my code compliant with all the requirements.  My current best snapshot (puiblished in this PR) gives 19.6 mAP. As far as I know it's one of the closest attempts to reproduce the result form Tensorflow model zoo in pytorch.

huihuixiaohuihui:
the loss is about 5，how can i low the loss

zhaojw219:
> This PR adds:
> 
> * MobileNetV2 backbone for SSD
> * Lite heads for SSD
> * Automatic regression/classification losses balancing for SSD (based on [paper](http://openaccess.thecvf.com/content_cvpr_2018/papers/Kendall_Multi-Task_Learning_Using_CVPR_2018_paper.pdf))
> * Ability to specify anchors in SSD manually
> * Sample configs to demonstrate the above features

Hello,very thankful for your work!Have you pretrained the mobilenet on coco or other dataset?How can I get the pretrained model for mobilenetv2?

chaoyan1037:
Any follow-up for adding MobileNetV2 SSD-lite?

ZwwWayne:
This PR is closed as it is too old and we will support it in July.

PR #839, opened 2019-06-20
Add reduction_override flag

yhcao6:


mmdet/models/losses/utils.py
```diff
@@ -42,12 +42,14 @@ def weight_reduce_loss(loss, weight=None, reduction='mean', avg_factor=None):
...
-                'avg_factor can only be used with reduction="mean"')
+    # if avg_factor are specified
+    # if reduction is mean, reduce the loss sum by avg factor
+    elif reduction == 'mean':
         loss = loss.sum() / avg_factor
+    elif reduction == 'none':
+        loss = loss
+    else:
+        raise ValueError('avg_factor can not be used with reduction="sum"')
     return loss
```
hellock:
```python
# if avg_factor is not specified, just reduce the loss
if avg_factor is None:
    loss = reduce_loss(loss, reduction)
else:
    # if reduction is mean, then average the loss by avg_factor
    if reduction == 'mean':
        loss = loss.sum() / avg_factor
    # if reduction is 'none', then do nothing, otherwise raise an error
    elif reduction != 'none':
        raise ValueError('avg_factor can not be used with reduction="sum"')

File: mmdet/models/bbox_heads/bbox_head.py [hunk 2]
```diff
@@ -97,12 +97,16 @@ def loss(self,
              label_weights,
              bbox_targets,
              bbox_weights,
-             reduce=True):
+             reduction_override=False):
```
hellock: `reduction_override=None`

PR #881, opened 2019-06-27
Refactor compilation into single setup.py

dseuss:
This pull request removes all `setup.py` files sprinkled throughout the repo and moves all extension building into the main `setup.py` file.

This would make it easier to pip install the repo as it gets rid of the extra `./compile.sh` step. Additionally, I think it's more maintainable.


dseuss:
I rebased the PR off the current master branch in 6b43bd7

setup.py
```diff
@@ -103,10 +129,53 @@ def get_version():
...
         tests_require=['pytest'],
         install_requires=[
             'mmcv>=0.2.6', 'numpy', 'matplotlib', 'six', 'terminaltables',
-            'pycocotools'
+            'pycocotools', 'torch>=1.1'
+        ],
+        ext_modules=[
+            make_cython_ext(
+                name='soft_nms_cpu',
+                modpath=('mmdet', 'ops', 'nms'),
```
hellock: Why not just use `mmdet.ops.nms` so that we do not need `'.'.join(modpath)`.

setup.py
```diff
@@ -103,10 +129,53 @@ def get_version():
...
+            make_cuda_ext(
+                name='sigmoid_focal_loss_cuda',
+                modpath=('mmdet', 'ops', 'sigmoid_focal_loss'),
+                sources=['src/sigmoid_focal_loss.cpp',
+                         'src/sigmoid_focal_loss_cuda.cu']),
+            make_cuda_ext(
+                name='masked_conv2d_cuda',
+                modpath=('mmdet', 'ops', 'masked_conv'),
+                sources=['src/masked_conv2d_cuda.cpp',
+                         'src/masked_conv2d_kernel.cu']),
```
hellock: It seems that the code was not formatted by `yapf` with the specified [configuration](https://github.com/open-mmlab/mmdetection/blob/master/.style.yapf).

setup.py
```diff
@@ -103,10 +129,53 @@ def get_version():
             'Programming Language :: Python :: 3.6',
         ],
         license='Apache License 2.0',
-        setup_requires=['pytest-runner'],
+        setup_requires=['pytest-runner', 'cython', 'torch>=1.1', 'numpy'],
```
hellock: We may assume that users have already installed PyTorch before they install mmdetection. Sometimes the installed pytorch cannot be identified correctly during the installation of mmdetection, so that adding torch in `setup_requires` may lead to unexpected behavior.

setup.py
```diff
@@ -1,7 +1,11 @@
 import os
 import subprocess
 import time
-from setuptools import find_packages, setup
+
+import numpy as np
+from Cython.Build import cythonize
```
hellock: If cython is not installed ahead, an ImportError will be raised.

setup.py
```diff
@@ -1,7 +1,11 @@
...
+from setuptools import find_packages, setup, Extension
```
hellock: System imports should be at the top.

hellock:
Thanks for the PR, it is really helpful!

setup.py
dseuss: Because it's also needed to construct the path to the source files. Alternatively, we can pass in `mmdet.ops.nms` and split on '.'

setup.py
dseuss: Will do

setup.py
dseuss: OK, I wanted to add the version requirement as having 1.0 installed leads to linker errors that are not easy to debug

setup.py
dseuss: That's why I added it to `setup_requires`

hellock: This import statement happens before installing packages specified in `setup_requires`.

dseuss:
OK, I've taken care of all suggestions above. Let me know what you think.

setup.py
```diff
@@ -80,6 +82,29 @@ def get_version():
...
+def make_cuda_ext(name, module, sources):
+
+    return CUDAExtension(
+        name='{}.{}'.format(module, name),
+        sources=[os.path.join(*module.split('.'), p) for p in sources])
+
+
+def make_cython_ext(name, module, sources):
+    from Cython.Build import cythonize
+    import numpy as np
```
hellock: This method will be called before installing dependencies, so that an error will still be raised.
A feasible solution may be installing cython before setting up mmdetection.

dseuss: You're right, it doesn't make a difference putting it there.

The `INSTALL.md` already mentions installing cython. Since we don't list pytorch in the `install_requires` field, we can't simply pip install the package anyway without prior installation of pytorch and cython. So how do you think we should handle those two dependencies?

hellock: Yes it seems nontrivial to avoid this dependency, so we may just assume that users have followed the installation guide to install cython first. Thus I think it's ok to move these import statements outside the method. Thanks!

dseuss: Sounds good! Done in 18c0e5e

dseuss:
BTW, it would be nice put the building processes into CI. It seems like you can at least compile CUDA code on Travis, see e.g. [this post](https://github.com/travis-ci/travis-ci/issues/5911#issuecomment-484837167).

hellock:
Thanks! I will take a look at the travis config.

hellock:
The compiling of cuda extensions takes quite a long time without printing any logs. To be more user friendly, is there any way to print some information when compiling the extensions?

dseuss:
Running `python setup.py install` prints the compiler output on my system. You can achieve the same with pip by running `pip install -v .`. Besides that, we can increase the loglevel for distutils, but I think an unaware user may be overwhelmed by all the additional output.

hellock:
Nice! Maybe we can modify the instructions in `INSTALL.md` from `pip install -e .` to pip install -v -e .`

INSTALL.md
```diff
@@ -41,7 +41,7 @@ d. Install mmdetection (other dependencies will be installed automatically).
 
 ```shell
 python setup.py develop
-# or "pip install -e ."
+# or "pip install -e -v ."
```
hellock: `-v` cannot be between `-e` and `.`, otherwise there will be an error.

dseuss: Sorry, sometimes I think I know what I am doing and that I can just skip testing 😄

OpenMMLab-Assistant-007:
Hi!
@dsuess 
First of all, we want to express our gratitude for your significant PR in the OpenMMLab project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA

If you have WeChat account，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

PR #883, opened 2019-06-27
Fix scale test error

liushuchun:
When setting the config of resize_keep_ratio with False got the scale_factor with the np.array type.
and the type of bboxes is tensor, so need to check the value before the operation.


mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -325,9 +325,14 @@ def simple_test(self, img, img_meta, proposals=None, rescale=False):
...
-                    det_bboxes[:, :4] * scale_factor
-                    if rescale else det_bboxes)
+                if isinstance(scale_factor, float):
+                    _bboxes = (det_bboxes[:, :4] * scale_factor
+                               if rescale else det_bboxes)
+                else:
+                    _bboxes = (det_bboxes[:, :4] * torch.from_numpy(
+                        scale_factor).to(det_bboxes.device)
+                               if rescale else det_bboxes)
+
```
hellock: The formatting seems not to be consistent with the coding style. You may check the [contributing guide](https://github.com/open-mmlab/mmdetection/blob/master/CONTRIBUTING.md#python).

PR #939, opened 2019-07-06
Add batch inference

yhcao6:


chicm:
Does it support GroupSampler?  If not, does it mean we should group the images before inference to avoid that a 400x800 image and a 800x400 image are put in a same batch?

gittigxuy:
@yhcao6 ,so how could  I use this code to multi batch inference?

yhcao6:
> @yhcao6 ,so how could I use this code to multi batch inference?

Which model?

gittigxuy:
faster rcnn

yhcao6:
> faster rcnn

For example if you want to test with 2 images per GPU, then you need to add `imgs_per_gpu_override=2,` after this line. Currently you must obtain the rule that len(test_dataset) % (#GPU x #imgs_per_gpu) == 0
https://github.com/open-mmlab/mmdetection/blob/5b9c5dd45fc5446a1c2f6c1fce571ad504e4d9ce/configs/faster_rcnn_r50_fpn_1x.py#L139

gittigxuy:
Thanks,Furthermore,how about the data augment function,How could I use data-pipeline brench.From now on,it just support ssd,mask rcnn.htc,if I want to support faster rcnn,how could I fix the code?

gittigxuy:
@yhcao6 ,how about multi-batch test in the guide anchor faster rcnn,I fix the code reference by your diff code,but it still error.I think I should fix the guide anchor model file,but how could I fix it?

gittigxuy:
@yhcao6,I think you should Add the code in the models/detectors/test_mixins.py RPNTestMixin class simple_test_rpn function to support multi-batch inference

mosheliv:
will this work for retinanet models?

yhcao6:
Currently this work supports retinanet.

ievbu:
Is this PR compatible with HTC models? If yes, then how could I use it for batch inference?

greathope:
support mask-rcnn?

greathope:
It seems your pr is far behind current branch， can you transfer your code to current branch？

greathope:
I find this pr doesn't accelerate the inference, on the contrary, setting bs to 5, the inference takes more time  than bs 1

zvadaszi:
Is batch inference support for Cascade R-CNN in plans?

bruceyang2012:
Can you add batch inference in the latest version of mmdet？Thank you.

PR #993, opened 2019-07-15
Require an operation(Save or Show the results) when testing

wswday:
It will waste a lot time to inference but do nothing if we didn't assign an operation
I did several times :joy:

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -128,6 +128,10 @@ def parse_args():
 def main():
     args = parse_args()
 
+    assert args.out or args.show,\
+        'Please choice an operation(Save or Show the results)\
+         by argument --out or --show'
+
     if args.out is not None and not args.out.endswith(('.pkl', '.pickle')):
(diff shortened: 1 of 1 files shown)


tools/test.py
```diff
@@ -128,6 +128,10 @@ def parse_args():
 def main():
     args = parse_args()
 
+    assert args.out or args.show,\
+        'Please choice an operation(Save or Show the results)\
+         by argument --out or --show'
```
hellock: The backslash can be avoided by using parentheses.
```python
    assert (args.out or args.show,
            'Please specify at least one operation (save or show the results) '
            'with the argument "--out" or "--show"')
```

hellock:
Thanks! This check is necessary indeed.

tools/test.py
wswday: I make some modify to avoid the assertion of flake8

PR #1010, opened 2019-07-17
Warning when skip image

wswday:
1.Warning when skip image
2.added parameter skip_img_without_anno for dataset. It is useful , if we want to use images that has no annotation to train as negative sample.
3.Fix format

### DIFF
--- a/mmdet/datasets/custom.py
+++ b/mmdet/datasets/custom.py
@@ -56,6 +56,7 @@ class CustomDataset(Dataset):
                  seg_scale_factor=1,
                  extra_aug=None,
                  resize_keep_ratio=True,
+                 skip_img_without_anno=True,
                  test_mode=False):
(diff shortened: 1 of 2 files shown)


mmdet/datasets/custom.py
```diff
@@ -203,7 +206,8 @@ def prepare_train_img(self, idx):
             gt_bboxes_ignore = ann['bboxes_ignore']
 
         # skip the image if there is no valid gt bbox
-        if len(gt_bboxes) == 0:
+        if len(gt_bboxes) == 0 and self.skip_img_without_anno:
+            warnings.warn('Skip the image that has no valid gt bbox')
```
hellock: Maybe also print the filename?

wswday: Yes.
Committed. Please check

PR #1058, opened 2019-07-25
Added Registry use for validation datasets (distributed)

### LINKED ISSUE #1057: Registry datasets bug for distributed training
Hi All,

If a new dataset is used via the Registry, distributed training fails when the --validate flag is used.
This is due to [this line](https://github.com/open-mmlab/mmdetection/blob/3dc0047c29430248047274cf415bbf6c33dbc642/mmdet/apis/train.py#L177). 

Error is:
```bash
 File "/home/precision14/testing/mmdetection/mmdet/apis/train.py", line 177, in _dist_train
    dataset_type = getattr(datasets, val_dataset_cfg.type)
AttributeError: module 'mmdet.datasets' has no attribute 'NewDatasetCoco'
```

The fix should probably be to look for the dataset class type from the registry and not the datasets module directly for adding the evaluation hook.

dmarnerides:
This is to use the registry for preparing the validation datasets instead of the mmdet.datasets module directly. Fixes #1057

### DIFF
--- a/mmdet/core/evaluation/eval_hooks.py
+++ b/mmdet/core/evaluation/eval_hooks.py
@@ -5,7 +5,7 @@ import mmcv
 import numpy as np
 import torch
 import torch.distributed as dist
-from mmcv.runner import Hook, obj_from_dict
+from mmcv.runner import Hook
 from mmcv.parallel import scatter, collate
 from pycocotools.cocoeval import COCOeval
 from torch.utils.data import Dataset
@@ -13,6 +13,7 @@ from torch.utils.data import Dataset
 from .coco_utils import results2json, fast_eval_recall
 from .mean_ap import eval_map
 from mmdet import datasets
+from mmdet.utils import build_from_cfg
 
 
 class DistEvalHook(Hook):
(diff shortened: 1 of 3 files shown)


mmdet/core/evaluation/eval_hooks.py
```diff
@@ -21,8 +22,8 @@ def __init__(self, dataset, interval=1):
         if isinstance(dataset, Dataset):
             self.dataset = dataset
         elif isinstance(dataset, dict):
-            self.dataset = obj_from_dict(dataset, datasets,
-                                         {'test_mode': True})
+            self.dataset = build_from_cfg(dataset, datasets.DATASETS,
```
hellock: We can use [`build_dataset()`](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/builder.py#L30) here.

hellock:
Thanks for the bug fix!

mmdet/core/evaluation/eval_hooks.py
dmarnerides: Yes that's probably best. How can we ensure that test_mode = True when using build_dataset()? It won't accept default_args. Should it be changed too?

dmarnerides:
Glad to help!

mmdet/core/evaluation/eval_hooks.py
hellock: Yes you can add an argument for `build_dataset`.

dmarnerides: Ok done.

Issue #364, opened 2019-03-06
I have a question about the way you wanted to implement.


kye9216789:
Hi.
I have been searched about relation between meaning of validation loss and evaluation result(mAP).
But I failed to find concrete solution.
I found your reply : 
https://github.com/open-mmlab/mmdetection/issues/171#issuecomment-446434197

My question is, Why you choose the second option? My expectancy is,
First : First and second options have exactly same meaning
Second : First option is pointless to enhance training result
Third : Second option is easy to implement and use
Fourth : Ordinarily Object detection methods are using just second option, not first option.

I can not sleep because of my curiosity about this question. 
I need your help.

Thank you.

hellock:
Unlike classification, the loss and performance (mAP) does not mean exactly the same thing for object detection.
You can choose both ways for validation, and we evaluate the mAP by default.
If you don't specify `--validate` and use two phases in the `workflow` field, then it will just calculate the val loss instead of computing mAP. It will be faster.

kye9216789:
You mean, when loss is very low, the performance may not good in object detection?
I cannot understand. 
Can you recommend me any document or paper?
Is it your experience?

I always thank for your brilliant sight and clear answer.

hust-kevin:
> You mean, when loss is very low, the performance may not good in object detection?
> I cannot understand.
> Can you recommend me any document or paper?
> Is it your experience?
> 
> I always thank for your brilliant sight and clear answer.

hi, do you understand the the relation of loss and ap?

Issue #271, opened 2019-01-18
workflow not work


kakaluote:
when i config workflow = [('train', 2),('val',1)], train crash by this
`assert len(data_loaders) == len(workflow)`

because  len(data_loaders) always 1

hellock:
If you specify 2 phases in the workflow, you need 2 dataloaders. In detection tasks we implement mAP evaluation during training to serve as validation, so the val phase is unnecessary. If you want to run val epochs and see the validation loss, you can modify the codes to build 2 dataloaders for the runner.

alexwq100:
hi ,@hellock,
you say " In detection tasks we implement mAP evaluation during training to serve as validation, so the val phase is unnecessary." where is it,in the tools/train.py I haven't found.

tdiekel:
@alexwq100 [Look here](https://github.com/open-mmlab/mmdetection/issues/171#issuecomment-446434197)

Issue #171, opened 2018-12-12
Load train_dataloader and val_dataloader


whikwon:
How can I see the proceeding status (validation) during training?

I added `workflow = [('train, 1), ('val', 1)]` to config file and executed `python tools/train.py configs/faster_rcnn_r50_fpn_1x.py` 

Error below occurred
` assert len(data_loaders) == len(workflow)` 

I think handling only `train_dataset` in the tools/train.py script make this error.

How can I use validation dataset?

thangvubk:
You should you argument --validate instead. Pls refer to readme file for details

hellock:
For validation there are two options. One is to show the loss on validation set, another is to evaluate the mAP on validation set. In mmdetection we adopt the second one and use eval hooks to implement it, see [here](mmdet/core/evaluation/eval_hooks.py) for details.
If you want to adopt the first option and specify two phases in the workflow, two dataloaders are needed.
A minor modification is needed in [L70-L77](tools/train.py) as follows.
```python
train_dataset = get_dataset(cfg.data.train)
val_dataset = get_dataset(cfg.data.val)
train_detector(
    model,
    [train_dataset, val_dataset],
    cfg,
    distributed=distributed,
    validate=args.validate,
    logger=logger)
```

whikwon:
@hellock 
I tried 1st option and got error:
```Traceback (most recent call last):
  File "tools/train.py", line 82, in <module>
    main()
  File "tools/train.py", line 78, in main
    logger=logger)
  File "/home/whikwon/anaconda3/envs/biorobot/lib/python3.6/site-packages/mmdet-0.5.4+c95c637-py3.6.egg/mmdet/apis/train.py", line 59, in train_detector
    _non_dist_train(model, dataset, cfg, validate=validate)
  File "/home/whikwon/anaconda3/envs/biorobot/lib/python3.6/site-packages/mmdet-0.5.4+c95c637-py3.6.egg/mmdet/apis/train.py", line 107, in _non_dist_train
    dist=False)
  File "/home/whikwon/anaconda3/envs/biorobot/lib/python3.6/site-packages/mmdet-0.5.4+c95c637-py3.6.egg/mmdet/datasets/loader/build_loader.py", line 31, in build_dataloader
    sampler = GroupSampler(dataset, imgs_per_gpu)
  File```

hellock:
I updated the code snippets above.

whikwon:
I've followed your instruction and got error

[...]

PR #1093, opened 2019-07-31
Allowing validation dataset for computing validation loss

dmarnerides:
This is to allow the use of more than one dataset in the train api so that validation sets can be used to track validation loss. 

Relevant issues:
#171
#364
#271

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -136,12 +136,14 @@ def build_optimizer(model, optimizer_cfg):
 
 def _dist_train(model, dataset, cfg, validate=False):
     # prepare data loaders
+    dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
     data_loaders = [
         build_dataloader(
-            dataset,
+            ds,
             cfg.data.imgs_per_gpu,
             cfg.data.workers_per_gpu,
             dist=True)
+        for ds in dataset
     ]
     # put model on gpus
     model = MMDistributedDataParallel(model.cuda())
@@ -189,13 +191,15 @@ def _dist_train(model, dataset, cfg, validate=False):
 
 def _non_dist_train(model, dataset, cfg, validate=False):
     # prepare data loaders
+    dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
     data_loaders = [
         build_dataloader(
-            dataset,
+            ds,
             cfg.data.imgs_per_gpu,
             cfg.data.workers_per_gpu,
             cfg.gpus,
             dist=False)
+        for ds in dataset
     ]
     # put model on gpus
     model = MMDataParallel(model, device_ids=range(cfg.gpus)).cuda()
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -139,10 +139,7 @@ def _dist_train(model, dataset, cfg, validate=False):
     dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
     data_loaders = [
         build_dataloader(
-            ds,
-            cfg.data.imgs_per_gpu,
-            cfg.data.workers_per_gpu,
-            dist=True)
+            ds, cfg.data.imgs_per_gpu, cfg.data.workers_per_gpu, dist=True)
         for ds in dataset
     ]
     # put model on gpus
@@ -198,8 +195,7 @@ def _non_dist_train(model, dataset, cfg, validate=False):
             cfg.data.imgs_per_gpu,
             cfg.data.workers_per_gpu,
             cfg.gpus,
-            dist=False)
-        for ds in dataset
+            dist=False) for ds in dataset
     ]
     # put model on gpus
     model = MMDataParallel(model, device_ids=range(cfg.gpus)).cuda()
--- a/tools/train.py
+++ b/tools/train.py
@@ -83,7 +83,9 @@ def main():
     model = build_detector(
         cfg.model, train_cfg=cfg.train_cfg, test_cfg=cfg.test_cfg)
 
-    train_dataset = build_dataset(cfg.data.train)
+    datasets = [build_dataset(cfg.data.train)]
+    if len(cfg.workflow) == 2:
+        datasets.append(build_dataset(cfg.data.val))
     if cfg.checkpoint_config is not None:
         # save mmdet version, config file content and class names in
         # checkpoints as meta data
@@ -95,7 +97,7 @@ def main():
     model.CLASSES = train_dataset.CLASSES
     train_detector(
         model,
-        train_dataset,
+        datasets,
         cfg,
         distributed=distributed,
         validate=args.validate,
--- a/tools/train.py
+++ b/tools/train.py
@@ -92,9 +92,9 @@ def main():
         cfg.checkpoint_config.meta = dict(
             mmdet_version=__version__,
             config=cfg.text,
-            CLASSES=train_dataset.CLASSES)
+            CLASSES=datasets[0].CLASSES)
     # add an attribute for visualization convenience
-    model.CLASSES = train_dataset.CLASSES
+    model.CLASSES = datasets[0].CLASSES
     train_detector(
         model,
         datasets,


mosheliv:
this won't work, as these datasets are being fed through the training, so validation is useless. The change must be much deeper and involve some flag to prevent mmdetection from training on the dataset. As they are now restructuring the whole pipeline thing, I was secretly hoping they will add this...

dmarnerides:
I don't think they are fed to the training. in the mmcv runner function that is called is self.val is you set ('val', 1) in the workflow config and self.train if you set ('train', 1).

I might be missing something though

mosheliv:
i tried this and got 99% accuracy on the val set... believe me, it was not 99% accurate :)
I looked at the code and it looks like you are right, but my experiments showed otherwise. 
I am missing this feature very much as well, currently I am just running inference on all epochs backward to see when results start dropping. not fun.

hellock:
For non-distributed training, there is no validation yet. Eval hooks are only available for distributed training. I noted that this is an expected feature for users with only a single GPU, and will implement an evaluation hook for non-distributed mode.

dmarnerides:
@hellock Thanks for the quick response. This is not for validation using mAP (for either distributed or not). This is for getting the loss on the validation dataset (using the workflow, e.g. [('train', 1), ('val', 1)] which is already implemented in mmcv, but not in the train API of mmdet.

@mosheliv I will look into whether using this leads to the validation dataset being used for training and get back to you (unless @hellock has objections to this being implemented now?).

dmarnerides:
I checked to see if the validation dataset is used, and I don't think it is. I tried to overfit a training set of 5 images and validate with 3 other images with cfg.workflow = [('train', 1), ('val', 1)] and this in the train script:
```python
datasets = [build_dataset(cfg.data.train)]
if len(cfg.workflow) == 2:
    datasets.append(build_dataset(cfg.data.val))
```

This is what i get with the losses, which I think means the validation set is not being used in training:

Screenshot from 2019-08-01 12-20-58

The only problem I find is with the Text Logger. When it's logging the averages for validation it prints Epoch(train) instead of Epoch(val). This is because the logger assumes that the log_buffer will not have 'time' key when validating, but it does, because the log_buffer does not reset the keys. It's an mmcv bug and it doesn't affect the actual process in any other way.

It would be great if you can try it as well just to be sure.

@hellock, I think this works and it's only a minor change that brings a lot of functionality. Should I proceed with getting it ready to merge or do you want it closed?

mosheliv:
It might be that the text logger does not log the accuracy and loss
correctly? as well it might have not reset the counters. did you get
reasonable results in the text log of loss and accuracy, not only
tensorboard?  I will retry it when my GPUs free up.

dmarnerides:
The text logger logs everything accurately. E.g. from the above run:
```bash
2019-08-01 12:21:21,599 - INFO - Epoch [500][1/2]       lr: 0.01000, eta: 3:11:07, time: 2.877, data_time: 1.813, memory: 6186, loss_rpn_cls: 0.1627, loss_rpn_bbox: 0.0493, loss_cls: 0.1157, acc: 95.5078, loss_bbox: 0.1768, loss: 0.5044
2019-08-01 12:21:22,650 - INFO - Epoch [500][2/2]       lr: 0.01000, eta: 3:11:00, time: 1.053, data_time: 0.015, memory: 6186, loss_rpn_cls: 0.1706, loss_rpn_bbox: 0.0284, loss_cls: 0.0953, acc: 96.5332, loss_bbox: 0.1292, loss: 0.4234
2019-08-01 12:21:25,262 - INFO - Epoch(train) [500][1]  loss_rpn_cls: 1.3342, loss_rpn_bbox: 0.1753, loss_cls: 0.4107, acc: 88.1836, loss_bbox: 0.1120, loss: 2.0322
```

The only thing that is wrong is that it says Epoch(train) when it should be saying Epoch(val). The rest is correct. This is a problem from mmcv:

mosheliv:
Something is still not right there. Accuracy of 88%  is really high for
training on just a few images. It doesn't makes sense to me, but I'll look
at it tomorrow.
I understand the over fitting accuracy of 95% on the training, the model
has "memorized" these. But on the val, facing unseen images, it should have
been very low, as the 6 images in the training do not allow for any
generalization. I would expect it to be lower than 50%

dmarnerides:
That accuracy is because the dataset is consisting of only one class and it is objects of relatively simple features. In any case, it doesn't matter. If it was memorizing and also using the validation set for training then it would also memorize the validation set as well (it's only 5 + 3 points). Here it is clearly not doing that. The validation set is not being used for training.

Please let me know if your run confirms it  when you get a chance to run it.

mosheliv:
ok, managed to get the test running. I still think the stats are being
averaged across train and val.
my train had one class (crocodile) and val had another class (rose). here
are the stats:
2019-08-01 23:14:07,006 - INFO - Epoch [1][5/9] lr: 0.00085, eta: 0:01:05,
time: 0.637, data_time: 0.035, memory: 5096, loss_rpn_cls: 0.6951,
loss_rpn_bbox: 0.0594, loss_cls: 3.9696, acc: 50.1562, loss_bbox: 0.0092,
loss: 4.7332
2019-08-01 23:14:17,927 - INFO - Epoch(train) [1][2]    loss_rpn_cls:
0.6187, loss_rpn_bbox: 0.0298, loss_cls: 0.5528, acc: 95.4102, loss_bbox:
0.1035, loss: 1.3048
2019-08-01 23:14:21,079 - INFO - Epoch [2][5/9] lr: 0.00088, eta: 0:00:42,
time: 0.628, data_time: 0.039, memory: 5096, loss_rpn_cls: 0.4337,
loss_rpn_bbox: 0.0470, loss_cls: 0.1992, acc: 98.9844, loss_bbox: 0.0276,
loss: 0.7075
2019-08-01 23:14:31,670 - INFO - Epoch(train) [2][2]    loss_rpn_cls:
0.2151, loss_rpn_bbox: 0.0336, loss_cls: 2.2899, acc: 87.9883, loss_bbox:
0.2751, loss: 2.8137
2019-08-01 23:14:34,813 - INFO - Epoch [3][5/9] lr: 0.00091, eta: 0:00:34,
time: 0.626, data_time: 0.036, memory: 5096, loss_rpn_cls: 0.1516,
loss_rpn_bbox: 0.0186, loss_cls: 0.1513, acc: 97.7734, loss_bbox: 0.0555,
loss: 0.3770
2019-08-01 23:14:45,827 - INFO - Epoch(train) [3][2]    loss_rpn_cls:
0.0889, loss_rpn_bbox: 0.0559, loss_cls: 2.1893, acc: 90.5273, loss_bbox:
0.2365, loss: 2.5705
2019-08-01 23:14:48,977 - INFO - Epoch [4][5/9] lr: 0.00094, eta: 0:00:29,
time: 0.628, data_time: 0.038, memory: 5096, loss_rpn_cls: 0.0635,
loss_rpn_bbox: 0.0381, loss_cls: 0.0664, acc: 97.9688, loss_bbox: 0.0494,
loss: 0.2173
2019-08-01 23:14:59,705 - INFO - Epoch(train) [4][2]    loss_rpn_cls:
0.0680, loss_rpn_bbox: 0.0343, loss_cls: 2.4305, acc: 86.9141, loss_bbox:
0.2787, loss: 2.8115
2019-08-01 23:15:02,798 - INFO - Epoch [5][5/9] lr: 0.00097, eta: 0:00:25,
time: 0.617, data_time: 0.035, memory: 5096, loss_rpn_cls: 0.0402,
loss_rpn_bbox: 0.0454, loss_cls: 0.0344, acc: 97.7344, loss_bbox: 0.0389,
loss: 0.1589

I wouldn't think that is is statistically possible for the detector to
detect rose being trained on crocodiles with 86% accuracy.
It might be the leftovers from the imagenet pre-trained weighs, but still
the numbers are strange and the accuracy is still around 89% even after
many rounds.

On second thought, it might be the pre-trained weights. I assume the
resnext will detect the rose as features and the faster rcnn will use
these? Object detection is half magic.

dmarnerides:
That is indeed weird, but I don't think it has to do with validation data
being used for training. Did you let it run a bit longer to have a look at
tha graphs?

mosheliv:
i did run it for a few more epochs. the accuracy was more or less constant,
90%
I am not sure why this happens - i thought it was a bug that made it run
the val through train, but it is not. another option perhaps is that it
does not reset the statistics between train and val. anyways, something is
fishy there, not sure what.

dmarnerides:
It does look fishy, you are right. On the other hand, a very easy
distinguishing feature is crocodile - green, rose-red, so if it only learns
those two classes, accuracy can be pretty high even with overfitting, given
these easy features, especially if it is not let to overfit for many more
epochs. I have a feeling that performance will drop more the more it's
trained.

mosheliv:
No, there is no rose in training, only in val, and no croc in val, only in
training. So it cant learn about roses from training.

mosheliv:
My GPUS are busy now, but would have been interesting to inference on the
val set and see what it actually find....

hellock:
Accuracy or loss does not necessarily be consistent with mAP, especially for two-stage detectors, where most samples are negative samples and the accuracy can be high. That's why I did not see the benefit of adding validation epochs. Evaluating the mAP after training epochs may be a better alternative, although it can be slower than running a validation epoch.

dmarnerides:
You are right, it's not necessarily consistent with mAP but it's correlated, otherwise we wouldn't be minimizing these losses as a proxy to mAP. Having the validation loss can sometimes be a good indicator for overfitting early on in training, before mAP has a chance to increase significantly (validation or training), e.g. when the dataset is small and you need to add more augmentation etc.. 

Anw, the functionality is already there, I don't see adding these two lines being a big issue, as it's something optional, but of course we can close this request if you don't want it added.

hellock:
Of course I would like to add an option, just to respond to the accuracy issue above.

The api is expected to be a little different. In this PR, if one want to run the validation epochs, the config would look like this and there are duplicated items
```python
data = dict(
    train=[dict(xxx), dict(yyy)],
    val=dict(yyy),
    test=dict(zzz))
```
It would be better if the two dataloaders are built based on `train` and `val` in the original configs. An additional argument may be needed to specify whether to use the val config to build a dataloader and run validation epochs. In most cases, evaluating the mAP after each training epoch is enough, thus users may not want the validation epochs.

mosheliv:
Thank you for your response. I am not sure I follow. In the config example
you gave the validation dataset would be run through training cycle, right?
The whole idea is to have the model run on the validation in "stand alone
mode", like the "test.py" script after each epoch. The model must not train
on the validation and the validation statistics should be separated from
the training statistics.

The point is to detect over fitting, although this is much less of a
problem in object detection.

hellock:
I mean in your current implementation, users have to write the config file like in the above example since you assume `dataset` is a list or tuple, see [here](https://github.com/open-mmlab/mmdetection/pull/1093/files#diff-320359f9a551428df57ff952742a20ceR139) and [here](https://github.com/open-mmlab/mmdetection/blob/master/tools/train.py#L98).
However, since there already exists a validation dataset config, we can just use it for validation instead of specifying a duplicated one.
Also we never use validation dataset for training in such configs.

dmarnerides:
Hi, thanks for the responses!
The way I see it is the following:

Config:
```python
workflow = [('train', 1), ('val', 1)] # This dictates whether to do val loss
data = dict(
    train=dict(xxx),
    val=dict(yyy),
    test=dict(zzz))
```

Train Api (changed in this pull request):
```python
def _dist_train(model, dataset, cfg, validate=False):
    # prepare data loaders
    dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
    data_loaders = [
        build_dataloader(
            ds,
            cfg.data.imgs_per_gpu,
            cfg.data.workers_per_gpu,
            dist=True)
        for ds in dataset
    ]
    ...
```

Train script (did not change it in the pull request, but this is how I use it)
```python
def main():
    ...
    datasets = [build_dataset(cfg.data.train)]
    if len(cfg.workflow) == 2:  # This is not ideal, but works if the workflow is not crazy
        datasets.append(build_dataset(cfg.data.val))
    ...
    train_detector(
        model,
        datasets,
        cfg,
        distributed=distributed,
        validate=args.validate,
        logger=logger)
```

There is probably a much better way to do this, this was just the easiest. I could make the changes to the train script as well, though I don't see it as part of the API, so anyone can use it the way they want.

hellock:
Thanks! You may also include the changes in `tools/train.py` in this PR so that users can specify whether to add validation epochs by simply modifying the config files.

BTW, could you fix the linting errors?

dmarnerides:
Sure, it should be ready to merge now. Thanks for the awesome repo!

tiberium24:
> Accuracy or loss is not necessarily consistent with mAP, especially for two-stage detectors, where most samples are negative samples and the accuracy can be high. That's why I did not see the benefit of adding validation epochs. Evaluating the mAP after training epochs may be a better alternative, although it can be slower than running a validation epoch.

@hellock 
Can you elaborate why loss is not consistent with mAP?
I trained some models and observed that smaller validation loss means lower mAP. 
I was expecting to observe the opposite, meaning smaller loss in validation would give me higher mAP in validation.

PR #1112, opened 2019-08-03
Init instance variable of CustomDataset first

wswday:
So derived-class's functions, which will be call at` __init__` of `CustomDataset`, can access this variable.
such as `load_annotations`

### DIFF
--- a/mmdet/datasets/custom.py
+++ b/mmdet/datasets/custom.py
@@ -63,6 +63,43 @@ class CustomDataset(Dataset):
                  test_mode=False):
         # prefix of images path
         self.img_prefix = img_prefix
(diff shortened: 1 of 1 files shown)


hellock:
Thanks for the fix! We are refactoring the data pre-processing pipeline in #935, and I prefer not to merge this PR but directly switch to the new one.

wswday:
> Thanks for the fix! We are refactoring the data pre-processing pipeline in #935, and I prefer not to merge this PR but directly switch to the new one.

Nice work.
What time do you expect to finish?

hellock:
The code is almost finished and we are verifying the performance, since it is a quite large refactoring.

PR #1114, opened 2019-08-03
manually add type promotion

youkaichao:
``Bool`` type has been separated from ``Byte`` in pytorch. ``valid_flags`` is in ``Byte`` type but arithmetic comparisons return ``Bool``, which causes error under current pytorch version.
To keep compatible across pytorch versions, we have to manually add the type promotion.
(pytorch has always been conservative in automatic type conversion, even if there is an issue [there](https://github.com/pytorch/pytorch/issues/23758), I believe they will not add the type promotion)


hellock:
Thanks! You may also add a `TODO` mark to use bool instead of uint8 after the release of pytorch 1.2.

youkaichao:
But the ``bool`` type is not available for previous versions. I'm sure you don't want to lose compatibility.

PR #1142, opened 2019-08-07
fix variable names in CE Losses

### LINKED ISSUE #1138: Some confusion in CrossEntropyLoss
As the title, there is some confusion in [CrossEntropyLoss](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/cross_entropy_loss.py). For example, the ``weight`` used in ``forward()`` ([L#95](https://github.com/open-mmlab/mmdetection/blob/5c2e68bb4c20573e8144d92b989454efe886fbbb/mmdet/models/losses/cross_entropy_loss.py#L89)) is the ``label`` for ``mask_cross_entropy`` ([#L54](https://github.com/open-mmlab/mmdetection/blob/5c2e68bb4c20573e8144d92b989454efe886fbbb/mmdet/models/losses/cross_entropy_loss.py#L54)). The variable names seems to not reflect their roles. Some suggestions for improvement could be:
1. [#L87](https://github.com/open-mmlab/mmdetection/blob/5c2e68bb4c20573e8144d92b989454efe886fbbb/mmdet/models/losses/cross_entropy_loss.py#L87): change ``cls_score`` to ``pred``.
2. [#L54](https://github.com/open-mmlab/mmdetection/blob/5c2e68bb4c20573e8144d92b989454efe886fbbb/mmdet/models/losses/cross_entropy_loss.py#L54): change ``target`` to ``label``.
3. [#L54](https://github.com/open-mmlab/mmdetection/blob/5c2e68bb4c20573e8144d92b989454efe886fbbb/mmdet/models/losses/cross_entropy_loss.py#L54): change ``label`` to ``mask``, add argument ``weight`` for future use. It would be make ``mask_cross_entropy`` consistent with other CE losses

thangvubk:
Fix #1138

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -141,6 +141,11 @@ def build_optimizer(model, optimizer_cfg):
 
 
 def _dist_train(model, dataset, cfg, validate=False):
+    if validate:
+        raise NotImplementedError('Built-in validation is not implemented '
+                                  'yet in not-distributed training. Use '
+                                  'distributed training or test.py and '
+                                  '*eval.py scripts instead.')
     # prepare data loaders
     dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
     data_loaders = [


hellock:
I ran the new version for twice but it seems that the mAP of Mask R-CNN is 0.1~0.2 lower than the original version. That's weird.

thangvubk:
Is the mAP of original Mask R-CNN also re-trained or from modelzoo?

thangvubk:
@hellock  I just retrain the latest version without this modification and the results are also lower than model zoo 0.1 AP: 
``BboxAP/MaskAP = 37.2/34.1``
It seems that some harmful intermediate commits have made the performance changes.

hellock:
I rerun some baselines and found that the results are not very stable. Sometimes the mAP can be 0.1 or 0.2 lower, and sometimes just the same. Basically the latest version is slightly lower than the results in the model zoo.

PR #1182, opened 2019-08-13
update results of HRNet

wondervictor:


yhcao6:
@wondervictor  I update the url of newly added models. Please have a check to see if there is any problem.

wondervictor:
@yhcao6 Hi, I've checked it. Thanks for your support.

hellock:
The name of backbones can be more precise, e.g., `HRNetV2-W18` or `HRNetV2-W18-FPN`. The 
filename of checkpoints can also be improved.

wondervictor:
@hellock We use HRFPN instead of FPN, will it be confusing when using HRNetV2-W18-FPN. In HRNetV2 paper, we use HRNetV2p-W32 to denote a feature pyramid.

hellock:
Okay I get it, so maybe just use `HRNetV2p-W18` for such a setting.

wondervictor:
I've changed the names of these backbones

hellock:
It seems that the checkpoint filenames of 1x schedule is `xxxxxxxx_hrnetv2_w18_fpn_1x_xxxxxx`, and 2x schedule is `xxxxxxxx_hrnetv2p_w18_2x_xxxxxx`.

wondervictor:
Checkpoints of 1x schedule were uploaded before. They should be changed to `xxxx_hrnetv2p_w18_xxxx`. Sorry about that.

PR #1228, opened 2019-08-21
add coco analyze tool

myownskyW7:


tools/coco_analyze_tool.py
```diff
@@ -0,0 +1,174 @@
...
+        makeplot(recThrs, ps, res_out_dir, 'allclass', iou_type)
+
+
+def main():
+    parser = ArgumentParser(description='COCO Analyze Tool')
+    parser.add_argument('result', help='result file (json format) path')
+    parser.add_argument('out_dir', help='dir to save analyze result images')
+    parser.add_argument(
+        '--ann',
+        default='/data/coco/annotations/instances_val2017.json',
```
hellock: The leading `/` should be removed.

PR #1235, opened 2019-08-21
update CI - pkg build

Borda:
Adding package build with CUDA to Travis and empty testing via `pytest`
Inspiration on how to install CUDA in Travis was taken from https://github.com/jeremad/cuda-travis


hellock:
Thanks for the updates! It is quite useful and makes the CI more complete.
Here are some comments.
- The installation of CUDA is quite long, you may move the linting (flake8, isort, yapf) before the installation, thus if a linting error happens, the CI will exit without waiting for a long time.
- You may leave the demo script along and add a test file, which just import mmdet. We can further add some unit tests.

Borda:
@hellock thx for feedback. In fact, CUDA installation is quite fast since it is cached, but the building this package takes about 5min so I will move the initial check (flake8, isort, yapf) before building

Borda:
it seems that the `nosetest`, as well as `pytest`, are freezing even the package is test-less (except single doctest) so I would temporarily disable the testing for this PR to let it pass...

hellock:
I suggest adding a `test/` directory and just use `pytest test/`. You can write a dummy testing file in the test dir.

Borda:
Sure, so you do not want to do docstring testing? There is already a folder `tests`...

hellock:
There is currently no files in `tests/`, you may create a `test_import.py` and add an empty method in it. We can see if such a testing could pass.

hellock:
The config works now. We may switch back to pytest.

hellock:
@Borda Looks good to me now. Do you have any further comments?

Borda:
@hellock I think it is all related to this PR... :)

Issue #1236, opened 2019-08-21
Precision, Recall, AP from tools/test.py


dhananjaisharma10:
Hi!

I am using a custom dataset and am running a detection model. Currently, in the module `tools/test.py`, I do not see any functionality for calculating the aforementioned statistics on non-COCO datasets. I could find a way to do it by using the function `eval_map` lying in the module `mmdet/core/mean_ap.py`. Please let me know if there is a quicker way to do it.

Thanks!

hellock:
You can use `tools/test.py` to test the dataset and save the results, and use `tools/voc_eval.py` for evaluation.

dhananjaisharma10:
Thanks a lot! Apparently, there seems to be a bug in the `voc_eval.py` module:
```
if not gt_ignore:
        gt_ignore = gt_ignore
```
should be
```
if not gt_ignore:
        gt_ignore = None
```
The former throws an error in the `eval_map` function as that does not correctly handle the bboxes to be ignored. Submitting a pull request for the same.

PR #1242, opened 2019-08-22
Fix for gt_ignore

dhananjaisharma10:
Fix for issue #1236. If list `gt_ignore` was empty, it should become `None` after the following if statement:
```
if not gt_ignore:
        gt_ignore = gt_ignore
```
otherwise, the `eval_map` function throws an `AssertionError`.
Changed it to `gt_ignore = None`.


hellock:
Thanks! It is definitely a typo.

dhananjaisharma10:
> Thanks! It is definitely a typo.

Not sure why all the checks haven't completed yet. Is that an issue?

PR #1249, opened 2019-08-24
Fix expand mutable operation

yhcao6:


adampower48:
This change makes the returned bboxes.dtype == Float64 instead of Float32 (as they come into the function)

`print(boxes.dtype, np.tile((left, top), 2).dtype)` -> `float32 int64`
`print((boxes + np.tile((left, top), 2)).dtype)` -> `float64`

which causes the following error:
 ` File "/data/modules/mmdetection/mmdet/core/bbox/geometry.py", line 48, in bbox_overlaps`
`    lt = torch.max(bboxes1[:, None, :2], bboxes2[:, :2])  # [rows, cols, 2]`

`RuntimeError: Expected object of scalar type Double but got scalar type Float for argument #2 'other'` 

adding `boxes = (boxes + np.tile((left, top), 2)).astype(boxes.dtype)` instead fixes the issue for me

Kaeseknacker:
Can a python pro please explain me this strange behavior? :)

luyao777:
It occurs in ssd300_wider_face.py too.
I trid to use `boxes = (boxes + np.tile((left, top), 2)).astype(boxes.dtype)` but it didn't work.
Is there any suggestion?

Issue #1285, opened 2019-08-27
RuntimeError: train.py ssd300_coco.py


Kaeseknacker:
Hi,

I try to train ssd with standard config file, but I get the following error:

```
  File "/home/user//Code/mmdetection/mmdet/core/bbox/geometry.py", line 53, in bbox_overlaps
    lt = torch.max(bboxes1[:, None, :2], bboxes2[:, :2])  # [rows, cols, 2]
RuntimeError: Expected object of scalar type Double but got scalar type Float for argument #2 'other'

```

Last week (e.g. commit ID: 0b96173...) I had no problems with that. I suspect it has something to do with the new training pipeline.

Kaeseknacker:
I tried different commits:
``python tools/train.py configs/ssd300_coco.py``
works until 0d5233a
Since 381f491 the error occurs.

hellock:
Thanks for the report! We will fix this bug tomorrow.
See also https://github.com/open-mmlab/mmdetection/pull/1249#issuecomment-525320638.

auroua:
The config file  'ssd512_coco.py'  wrongly set `img_scale`  to  (300, 300).

hellock:
Thanks @auroua, you may open a PR to fix that if would like to.

luyao777:
It occurs in ssd300_wider_face.py too.
I trid to use boxes = (boxes + np.tile((left, top), 2)).astype(boxes.dtype) but it didn't work.
Is there any suggestion?

auroua:
@luyao777   
After add op the `boxes` type have changed to `float64`, using the following code.
`
boxes += np.tile((left, top), 2)
`

luyao777:
Oh, sorry! There is something wrong in my branch. It worked after use `boxes = (boxes + np.tile((left, top), 2)).astype(boxes.dtype)` in mmdet/datasets/pipelines/transforms.py, the original `boxes = (boxes + np.tile((left, top), 2))` may have some bugs in ssd. @auroua Thanks.
Hope to fix that!

PR #1288, opened 2019-08-28
Fix ssd bug

yhcao6:


mmdet/datasets/pipelines/transforms.py
```diff
@@ -527,7 +527,7 @@ def __call__(self, results):
         left = int(random.uniform(0, w * ratio - w))
         top = int(random.uniform(0, h * ratio - h))
         expand_img[top:top + h, left:left + w] = img
-        boxes = boxes + np.tile((left, top), 2)
+        boxes = boxes + np.tile((left, top), 2).astype(boxes.dtype)
```
auroua: The old version is 
`boxes += np.tile((left, top), 2)`

hellock:
> The old version is `boxes += np.tile((left, top), 2)`.

Yes the old version introduces a bug, #1285.

PR #1333, opened 2019-09-04
fix bugs for cropping masks in RandomCrop

gfjiangly:
results['gt_masks'] is a python list that cannot be sliced using bool values.


mmdet/datasets/pipelines/transforms.py
```diff
@@ -348,10 +348,11 @@ def __call__(self, results):
...
                 valid_gt_masks = []
-                for i in valid_inds:
-                    gt_mask = results['gt_masks'][i][crop_y1:crop_y2, crop_x1:
-                                                     crop_x2]
-                    valid_gt_masks.append(gt_mask)
+                for i, is_valid in enumerate(valid_inds):
+                    if is_valid:
+                        gt_mask = results['gt_masks'][i]
+                        valid_gt_masks.append(
+                            gt_mask[crop_y1:crop_y2, crop_x1:crop_x2])
```
hellock: This may look simpler.
```python
for i in np.where(valid_inds)[0]:
    gt_mask = results['gt_masks'][i][crop_y1:crop_y2, crop_x1: crop_x2]
    valid_gt_masks.append(gt_mask)
```

hellock:
Thanks for the fix!

mmdet/datasets/pipelines/transforms.py
gfjiangly: Yes, it looks simpler.

PR #1359, opened 2019-09-07
[Modulated DeformableConv] fix a bug for deformable_group>1 

chongruo:
Fix a bug for deformable_group>1.  

The performance of Faster-RCNN R-50-FPN  is increased by 0.2 (40.2 -> 40.4)


chongruo:
btw, how to upload the model ?

hellock:
Thanks for the fix! You could provide a google drive or dropbox link and we can upload it to AWS.

It is better to add another config file instead of modifying the existing configs, since `deformable_groups=1` is the default setting provided by the authors to reproduce DCNv2.

chongruo:
@hellock    Here is the [LINK](https://drive.google.com/drive/folders/1zTxAO0z2TBVu5OQ6T7dSm7TZ48GOlSEF) for the model and log. I also create a new script (configs/dcn/faster_rcnn_mdconv_c3-c5_group4_r50_fpn_1x.py) for this exp. Thanks

configs/dcn/faster_rcnn_mdconv_c3-c5_group4_r50_fpn_1x.py
```diff
@@ -0,0 +1,177 @@
...
+    hooks=[
+        dict(type='TextLoggerHook'),
+        # dict(type='TensorboardLoggerHook')
+    ])
+# yapf:enable
+# runtime settings
+total_epochs = 12
+dist_params = dict(backend='nccl')
+log_level = 'INFO'
+work_dir = './work_dirs/faster_rcnn_mdconv_c3-c5_r50_fpn_1x'
```
yhcao6: `work_dir` should be consistent with the name of config.

chongruo: Fixed. Thanks

PR #1364, opened 2019-09-09
Fix bugs in bbox_flip and aug_test_bboxes.

shenyunhang:
1. Fix bbox_flip, so that it can handle bboxes with augmentation, as bbox_flip in [mmdet/datasets/pipelines/transforms.py#L183](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/transforms.py#L183).

2. Fix scales of proposal in aug_test_bboxes.
The proposals are rescaled before feeding to the model (see [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/test_aug.py#L17)), it should not rescale proposals again in aug_test_bboxes.

### DIFF
--- a/mmdet/core/bbox/transforms.py
+++ b/mmdet/core/bbox/transforms.py
@@ -81,8 +81,8 @@ def bbox_flip(bboxes, img_shape):
     if isinstance(bboxes, torch.Tensor):
         assert bboxes.shape[-1] % 4 == 0
         flipped = bboxes.clone()
-        flipped[:, 0::4] = img_shape[1] - bboxes[:, 2::4] - 1
-        flipped[:, 2::4] = img_shape[1] - bboxes[:, 0::4] - 1
+        flipped[..., 0::4] = img_shape[1] - bboxes[..., 2::4] - 1
+        flipped[..., 2::4] = img_shape[1] - bboxes[..., 0::4] - 1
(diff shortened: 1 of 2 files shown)


yhcao6:
> 1. Fix bbox_flip, so that it can handle bboxes with augmentation, as bbox_flip in [mmdet/datasets/pipelines/transforms.py#L183](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/transforms.py#L183).
> 2. Fix scales of proposal in aug_test_bboxes.
>    The proposals are rescaled before feeding to the model (see [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/test_aug.py#L17)), it should not rescale proposals again in aug_test_bboxes.

I disagree with the second point. Could you explain why the proposals will be rescaled [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/test_aug.py#L17)?

ZwwWayne:
MMDetection has been refactored a lot since last year. The major modification in `mmdet.core.bbox.transforms.py` is already adopted and `test_mixins.py` works fine now. So this PR will be closed. Feel free to create an issue for discussion if you have any further questions.

PR #1367, opened 2019-09-09
Fix bug when using coco_eval evaluate

yhcao6:


mmdet/core/evaluation/coco_utils.py
```diff
@@ -23,7 +23,12 @@ def coco_eval(result_files, result_types, coco, max_dets=(100, 300, 1000)):
         return
 
     for res_type in result_types:
-        result_file = result_files[res_type]
+        if isinstance(result_files, str):
+            result_file = result_files
+        elif isinstance(result_files, dict):
+            result_file = result_files[res_type]
+        else:
+            assert NotImplementedError
```
hellock: May raise a TypeError here.

Issue #1376, opened 2019-09-10
Segmentation network evaluation error (KeyError: 0)


RaymondKirk:
Running [test.py](https://github.com/open-mmlab/mmdetection/blob/8d010d7de9c2643e715aaf6033ff7fd5c60ebdc2/tools/test.py) with `bbox segm` eval options on instance segmentation networks results in `data['score'] = float(mask_score[i]) (KeyError: 0)` in the [coco_utils.py](https://github.com/open-mmlab/mmdetection/blob/8d010d7de9c2643e715aaf6033ff7fd5c60ebdc2/mmdet/core/evaluation/coco_utils.py#L154) file.

Command executed (independant of dataset used)

```bash
python tools/test.py configs/mask_rcnn_r50_fpn_1x.py work_dirs/mask_rcnn_r50_fpn_1x/latest.pth --out results.pkl --eval bbox segm
```

**Environment**
 -  Ubuntu 16.04.6
 - PyTorch version 1.1.0

**Error traceback**
```
Traceback (most recent call last):
  File "/home/user/mmdetection/tools/test.py", line 224, in <module>
    main()
  File "/home/user/mmdetection/tools/test.py", line 201, in main
    result_files = results2json(dataset, outputs, args.out)
  File "/home/user/mmdetection/mmdet/core/evaluation/coco_utils.py", line 170, in results2json
    json_results = segm2json(dataset, results)
  File "/home/user/mmdetection/mmdet/core/evaluation/coco_utils.py", line 154, in segm2json
    data['score'] = float(mask_score[i])
KeyError: 0
```

Possible clone of malformed issue https://github.com/open-mmlab/mmdetection/issues/1140.

yhcao6:
I download the [`mask R-50-FPN`](https://s3.ap-northeast-2.amazonaws.com/open-mmlab/mmdetection/models/mask_rcnn_r50_fpn_1x_20181010-069fa190.pth) model from modelzoo and execute the same test command with you. However, there is no error.

May I ask if you have modified something? If not, could you replace your `lastest.pth` with the [model](https://s3.ap-northeast-2.amazonaws.com/open-mmlab/mmdetection/models/mask_rcnn_r50_fpn_1x_20181010-069fa190.pth) provided in modelzoo and try again?

RaymondKirk:
[...]

PR #1384, opened 2019-09-12
Fixed coco_eval bug where seg length check could be wrong (#1376)

RaymondKirk:
Commit 8d010d7de9c2643e715aaf6033ff7fd5c60ebdc2 added a change where seg would be checked to have a length of 2 (denoting it was a different detector result). When a dataset has two classes this breaks, so this PR simply adds an additional check to ensure in the event the dataset has two classes the script doesn't break. Also ensures the segms counts has the decode method.

### DIFF
--- a/mmdet/core/evaluation/coco_utils.py
+++ b/mmdet/core/evaluation/coco_utils.py
@@ -142,7 +142,7 @@ def segm2json(dataset, results):
 
             # segm results
             # some detectors use different score for det and segm
-            if isinstance(seg, tuple) and label in seg[0]:
+            if isinstance(seg, tuple):
                 segms = seg[0][label]
                 mask_score = seg[1][label]
             else:
--- a/mmdet/core/evaluation/coco_utils.py
+++ b/mmdet/core/evaluation/coco_utils.py
@@ -142,7 +142,7 @@ def segm2json(dataset, results):
 
             # segm results
             # some detectors use different score for det and segm
-            if len(seg) == 2:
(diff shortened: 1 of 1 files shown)


RaymondKirk:
@yhcao6 Please review.

RaymondKirk:
Added condition `label in seg[0]` since it will always be True when detectors use a different score for det and segm and False when standard evaluation should take place.

RaymondKirk:
Ready for review!

yhcao6:
> Ready for review @yhcao6!

I test your pr with this [config](https://github.com/open-mmlab/mmdetection/blob/master/configs/ms_rcnn/ms_rcnn_r50_caffe_fpn_1x.py), but the program still fails. Please ensure there is no bug.

RaymondKirk:
I'm not sure why? `if len(seg) == 2 and label in seg[0]:` should capture this case right?

yhcao6:
> I'm not sure why? `if len(seg) == 2 and label in seg[0]:` should capture this case right?

https://github.com/open-mmlab/mmdetection/issues/1376#issuecomment-531652382

If the detector returns both `segms` and `mask_scores`, then `seg` is a tuple. Otherwise it is a list. This can be used to resolve this bug.

RaymondKirk:
Tested and confirmed working with this [config](https://github.com/open-mmlab/mmdetection/blob/master/configs/ms_rcnn/ms_rcnn_r50_caffe_fpn_1x.py).

RaymondKirk:
Ready to merge.

PR #1390, opened 2019-09-14
Fix dpool

yhcao6:


mmdet/ops/dcn/deform_pool.py
```diff
@@ -21,6 +22,10 @@ def forward(ctx,
                 part_size=None,
                 sample_per_part=4,
                 trans_std=.0):
+        # TODO: support unsquare RoIs
+        assert out_size[0] == out_size[1]
```
hellock: Before the assertion, we need to pair `out_size`, since it can be either a tuple or integer. See [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/ops/roi_align/roi_align.py#L13) for a reference.

PR #1474, opened 2019-09-30
Add two doctest examples for resnet and resnext

Erotemic:
When I'm trying to learn a library it helps to be able to get into the code and start stepping through it.  Doctest examples help with this a lot. 

As a side note: you may want to consider using [xdoctest](https://github.com/Erotemic/xdoctest) instead of Python's builtin doctest module in CI testing. Full disclosure: I wrote xdoctest, so I may be biased, but I do believe it offers a better features and syntax and makes it easier to write and debug doctests in general.

### DIFF
--- a/mmdet/models/backbones/resnet.py
+++ b/mmdet/models/backbones/resnet.py
@@ -352,6 +352,20 @@ class ResNet(nn.Module):
             memory while slowing down the training speed.
         zero_init_residual (bool): whether to use zero init for last norm layer
             in resblocks to let them behave as identity.
+
+    Example:
+        >>> from mmdet.models.backbones.resnet import *  # NOQA
+        >>> import torch
+        >>> self = ResNet(depth=18)
+        >>> self.eval()
+        >>> inputs = torch.rand(1, 3, 32, 32)
+        >>> level_outputs = self.forward(inputs)
+        >>> for level_out in level_outputs:
+        ...     print(tuple(level_out.shape))
+        (1, 64, 8, 8)
+        (1, 128, 4, 4)
+        (1, 256, 2, 2)
+        (1, 512, 1, 1)
     """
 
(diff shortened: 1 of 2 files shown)


mmdet/models/backbones/resnet.py
```diff
@@ -352,6 +352,20 @@ class ResNet(nn.Module):
             memory while slowing down the training speed.
         zero_init_residual (bool): whether to use zero init for last norm layer
             in resblocks to let them behave as identity.
+
+    Example:
+        >>> from mmdet.models.backbones.resnet import *  # NOQA
```
hellock: This may be more straightforward.
```python
from mmdet.models import ResNet
```

hellock:
Thanks for improving the docs and it is much appreciated!

mmdet/models/backbones/resnet.py
Erotemic: Technically, this line isn't even needed because the `import *` is implicit when running a doctest, but I tend to include them so a user can copy-paste the code into an IPython terminal. 

However, I do like importing `ResNet` explicitly for clarity. How about `from mmdet.models.backbones.resnet import ResNet` so the import line references the exact file that it belongs to?

Ultimately, I'm happy with any option so tell me which one you like best: 

1. remove the line and let the import be implicit. 
2. `from mmdet.models import ResNet`
3. `from mmdet.models.backbones.resnet import ResNet`

hellock: Yes an explicit import statement may be more clarified.
Although both 2 and 3 are acceptable to me, personally I prefer `from mmdet.models import ResNet` since it is clear enough and simpler.

Erotemic: I did option 2.

PR #1475, opened 2019-09-30
Add in_channels kwarg to backbones. 

Erotemic:
I added `in_channels` argument to the backbones. Previously they were hard coded to 3, which makes sense for most RGB problems. However, this makes it a pain to use when your input is grayscale or 4 channel RGBA. Adding this as an option makes mmdet a bit more usable in these situations. 

Of course the pretrained weights don't quite fit, but fortunately the existing system handles this well and simply outputs a warning message that a particular layer doesn't fit right (:tada: thanks for implementing that nicely!). 

I've added documentation in the `Args` block for each  case. I also added a docstring example because the HRNet backbone was unclear about how it should be used from the module itself (I think its nice when modules are self-documenting).  

Note, that I was only able to add `in_channels` to resnet, resnext, and hrnet. ssd_vgg depends on the mmcv implementation of VGG, so that parameter would need to be exposed there before it could be exposed in mmdet. I simply added a TODO note in this case.

### DIFF
--- a/mmdet/models/backbones/hrnet.py
+++ b/mmdet/models/backbones/hrnet.py
@@ -201,6 +201,7 @@ class HRNet(nn.Module):
 
     Args:
         extra (dict): detailed configuration for each stage of HRNet.
+        in_channels (int): Number of input image channels. Normally 3.
         conv_cfg (dict): dictionary to construct and config conv layer.
         norm_cfg (dict): dictionary to construct and config norm layer.
         norm_eval (bool): Whether to set norm layers to eval mode, namely,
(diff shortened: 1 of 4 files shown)


mmdet/models/backbones/hrnet.py
```diff
@@ -210,12 +211,52 @@ class HRNet(nn.Module):
             memory while slowing down the training speed.
         zero_init_residual (bool): whether to use zero init for last norm layer
             in resblocks to let them behave as identity.
+
+    Example:
+        >>> from mmdet.models.backbones.hrnet import *  # NOQA
```
hellock: May also adopt the same option.

Erotemic: I rebased on the latest master and fixed this issue.

PR #1476, opened 2019-09-30
Fix robustness data pipeline

michaelisc:
The way it was set up the `main` function in `test_robustness.py` would not use the copy of the config into which the corruption was inserted but the original config without the corruption. This should fix it.


hellock:
Thanks for the fix!

PR #1477, opened 2019-09-30
Expand docstring docs

Erotemic:
I'm attempting to learn how to use the infrastructure of this module independent of its training loop and configuration logic. While I'm sure they are nice for internal research, as an external researcher, I really want to be able to construct raw instances of these classes and use them in my own training / inference / configuration system.  Fortunately, this library has a well thought out architecture that makes this possible, but the docs are a bit lacking. 

As I'm understanding more about what each network expects I'm adding documentation as best as I can. Hopefully this will be beneficial to both the maintainers and other researchers who use this library. 

Note that even though this is mostly a documentation PR, this does contain one small bugfix. In cascade_rcnn.py simple_test, the `img_meta` is expected to be a `List[Dict]`, but the base `forward_test` passes `img_meta` as a `Dict`. I fixed this by checking for this scenario and simply re-wrapping the variable in a list. 

This diff is built on top of #1474 and #1475, so the diff contains a bit more than it should. Merging the other two PRs should reduce the size of this diff.

### DIFF
--- a/mmdet/datasets/pipelines/formating.py
+++ b/mmdet/datasets/pipelines/formating.py
@@ -134,6 +134,28 @@ class DefaultFormatBundle(object):
 
 @PIPELINES.register_module
 class Collect(object):
+    """
+    Populates img_meta, which by default includes:
+
+        - "img_shape": shape of the image input to the network as a tuple (h, w, c).
+            Note that images may be zero padded on the bottom/right if the batch
+            tensor is larger than this shape.
+
+        - "scale_factor": a float indicating the preprocessing scale
+
+        - "flip": a boolean indicating if image flip transform was used
+
+        - "filename": path to the image file
+
+        - "ori_shape": original shape of the image as a tuple (h, w, c)
+
+        - "pad_shape": image shape after padding
+
+        - "img_norm_cfg": a dict of normalization information:
+            - mean - per channel mean subtraction
+            - std - per channel std divisor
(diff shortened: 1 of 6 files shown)


mmdet/datasets/pipelines/formating.py
```diff
@@ -134,6 +134,28 @@ def __repr__(self):
 
 @PIPELINES.register_module
 class Collect(object):
+    """
+    Populates img_meta, which by default includes:
```
hellock: Besides populating img_meta, it also includes other fields such as 'img', 'gt_bboxes', 'gt_labels'.

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -270,7 +298,26 @@ def forward_train(self,
...
+
+        Returns:
+            dict: results
+        """
         x = self.extract_feat(img)
+
+        if isinstance(img_meta, dict):
+            # a single dict was passed in but the rest of this function
+            # expect a list of with one item.
+            img_meta = [img_meta]
```
hellock: `simple_test` is assumed to be called from `forward_test` and `img_meta` is supposed to be a list of dict.

hellock:
I revert it to the original version that assumes `img_meta` is a list of dict, since this modification needs further discussion. I will merge this PR and you may open a new one if there are any other issues.

PR #1479, opened 2019-09-30
Enable doctests in CI with xdoctest

Erotemic:
This PR enables doctests in in pytest. Instead of using the builtin doctest module I used xdoctest, which has pytest integration by default. I also mentioned [xdoctest](https://github.com/Erotemic/xdoctest)  in #1474.

The main benefit of xdoctest over buildin-doctest is you can prefix normal code with `>>> ` and the doctest will just work. In the old-style you would have to prefix certain lines with `>>> ` and other lines with `... `. Hopefully this helps the maintainers write more doctests, which can be both illustrative for newcomers to this library and practical for ensuring the integrity of the library. 

You may have noticed that the doctest in `mmdetection/mmdet/models/losses/utils.py` wasn't working. This is because the builtin doctest module has a very restrictive syntax (which prevented the l1_loss function from being defined) due to fundamental flaws in its parser, however xdoctest (which is a drop-in replacement) does not have this issue. I did have to import torch to make this test work, but if all goes well you should see an increase in the number of test run on the CI with this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ script:
   - yapf -r -d --style .style.yapf mmdet/ tools/ tests/
   - python setup.py check -m -s
   - python setup.py build_ext --inplace
-  - coverage run --source mmdet -m py.test tests -v --doctest-modules
+  - coverage run --source mmdet -m py.test -v --xdoctest-modules tests mmdet
 
 after_success:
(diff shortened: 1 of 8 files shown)


hellock:
I would be glad to accept xdoctest for doctests.

PR #1506, opened 2019-10-07
Fix AnchorHead in_channels

Erotemic:
Based on the conversation in #1491 

While AnchorHead and GuidedAnchorHead are base classes, the meaning of the `in_channels` and `feat_channels` parameters were confusing. This PR makes `in_channels` and `feat_channels` consistent with the rest of the code base. `in_channels` now marks the number of input channels to the nn.module as expected, and `feat_channels` is documented as the number of hidden channels. 

As a check to make sure this change didn't break anything I added a test that loads all configs from the config directory and attempts to call `build_detector`. This test is a bit slow (it takes about 2 minutes), but it guarantees that changes like this don't break the `__init__` for any detector component. It might be a good idea to just pick a subset of configs to test because I'm sure many of them are redundant. Thoughts?

EDIT: I modified this test to explicitly list all of the configs that are tested. The list actually contains all configs I just selectively commented some of them out. This reduces the running time from 2 minutes to 30 seconds while still maintaining high test coverage.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -43,7 +43,7 @@ def test_config_build_detector():
         # 'dcn/faster_rcnn_mdconv_c3-c5_group4_r50_fpn_1x.py',
         # 'dcn/faster_rcnn_mdconv_c3-c5_r50_fpn_1x.py',
         # ---
-        'htc/htc_x101_32x4d_fpn_20e_16gpu.py',
+        # 'htc/htc_x101_32x4d_fpn_20e_16gpu.py',
         'htc/htc_without_semantic_r50_fpn_1x.py',
         # 'htc/htc_dconv_c3-c5_mstrain_400_1400_x101_64x4d_fpn_20e.py',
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -0,0 +1,178 @@
...
+        config_mod.model
+        config_mod.train_cfg
+        config_mod.test_cfg
+        print('Building detector, config_fpath = {!r}'.format(config_fpath))
+
+        # Remove pretrained keys to allow for testing in an offline environment
+        if 'pretrained' in config_mod.model:
+            config_mod.model['pretrained'] = None
+
+        import ubelt as ub
```
hellock: Maybe we can avoid involving this library since it is only used here one for testing.
An alternative of the timer is `mmcv.Timer`.

Erotemic: Of course. This is actually debug code that I didn't mean to commit.

PR #1517, opened 2019-10-09
Pass img_meta as list to simple_test

Erotemic:
As discussed in #1477, there is an issue with the way `forward_test` calls `simple_test` when num_augs is 1. Previously the line was  `self.simple_test(imgs[0], img_metas[0], **kwargs)`, which passes img_metas in as a `dict` and not a `list[dict]`. It results in this error:

```python
~/code/mmdetection/mmdet/models/detectors/base.py in forward_test(self, imgs, img_metas, **kwargs)
     92         if num_augs == 1:
     93             # return self.simple_test(imgs[0], img_metas[0:1], **kwargs)
---> 94             return self.simple_test(imgs[0], img_metas[0], **kwargs)
     95         else:
     96             return self.aug_test(imgs, img_metas, **kwargs)

~/code/mmdetection/mmdet/models/detectors/rpn.py in simple_test(self, img, img_meta, rescale)
     62     def simple_test(self, img, img_meta, rescale=False):
     63         x = self.extract_feat(img)
---> 64         proposal_list = self.simple_test_rpn(x, img_meta, self.test_cfg.rpn)
     65         if rescale:
     66             for proposals, meta in zip(proposal_list, img_meta):

~/code/mmdetection/mmdet/models/detectors/test_mixins.py in simple_test_rpn(self, x, img_meta, rpn_test_cfg)
      8         rpn_outs = self.rpn_head(x)
      9         proposal_inputs = rpn_outs + (img_meta, rpn_test_cfg)
---> 10         proposal_list = self.rpn_head.get_bboxes(*proposal_inputs)
     11         return proposal_list
     12 

~/code/mmdetection/mmdet/core/fp16/decorators.py in new_func(*args, **kwargs)
    125                                 'method of nn.Module')
    126             if not (hasattr(args[0], 'fp16_enabled') and args[0].fp16_enabled):
--> 127                 return old_func(*args, **kwargs)
    128             # get the arg spec of the decorated method
    129             args_info = getfullargspec(old_func)

~/code/mmdetection/mmdet/models/anchor_heads/anchor_head.py in get_bboxes(self, cls_scores, bbox_preds, img_metas, cfg, rescale)
    268                 bbox_preds[i][img_id].detach() for i in range(num_levels)
    269             ]
--> 270             img_shape = img_metas[img_id]['img_shape']
    271             scale_factor = img_metas[img_id]['scale_factor']
    272             proposals = self.get_bboxes_single(cls_score_list, bbox_pred_list,

KeyError: 0
```

Changing this line to `self.simple_test(imgs[0], img_metas[0:1], **kwargs)` fixes the issue. 

I've also added a test, which both demonstrates this errors and tests the fix. These tests build a few select detectors and passes some dummy data to them in a forward pass. I don't check the results in detail, the main purpose it just to make sure the forward pass runs without crashing. 

To make these tests work on the CI I had to fix cuda/cpu issues in ssd_head and ghm_loss. In the first case ssd_head simply failed to pass device to get_anchors. 

In the second case the GHMC/GHMR loss assumed that the edges and acc_sum were always cuda tensors. I fixed this so they default to the CPU, but I registered them as buffers so if the user calls `.cuda(0)` or `.to(0)` then these are appropriately moved to the GPU with the rest of the model parameters and buffers (note a buffer does not compute gradients, so this is safe). In my tests I checked that the new code works on both CPU and GPU (if available).

### DIFF
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -90,7 +90,7 @@ class BaseDetector(nn.Module):
         assert imgs_per_gpu == 1
 
         if num_augs == 1:
-            return self.simple_test(imgs[0], img_metas[0], **kwargs)
+            return self.simple_test(imgs[0], img_metas[0:1], **kwargs)
         else:
             return self.aug_test(imgs, img_metas, **kwargs)
 
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -75,6 +75,15 @@ class BaseDetector(nn.Module):
             logger.info('load model from: {}'.format(pretrained))
 
     def forward_test(self, imgs, img_metas, **kwargs):
+        """
+        Args:
+            imgs (List[Tensor]): the outer list indicates test-time
+                augmentations and inner Tensor should have a shape NxCxHxW,
+                which contains all images in the batch.
+            img_meta (List[List[dict]]): the outer list indicates test-time
+                augs (multiscale, flip, etc.) and the inner list indicates
+                images in a batch
+        """
         for var, name in [(imgs, 'imgs'), (img_metas, 'img_metas')]:
             if not isinstance(var, list):
                 raise TypeError('{} must be a list, but got {}'.format(
@@ -90,12 +99,20 @@ class BaseDetector(nn.Module):
         assert imgs_per_gpu == 1
 
         if num_augs == 1:
-            return self.simple_test(imgs[0], img_metas[0:1], **kwargs)
+            return self.simple_test(imgs[0], img_metas[0], **kwargs)
         else:
             return self.aug_test(imgs, img_metas, **kwargs)
 
     @auto_fp16(apply_to=('img', ))
     def forward(self, img, img_meta, return_loss=True, **kwargs):
+        """
+        Calls either forward_train or forward_test depending on whether
+        return_loss=True. Note this setting will change the expected inputs.
+        When `return_loss=False`, img and img_meta are single-nested (i.e.
+        Tensor and List[dict]), and when `resturn_loss=True`, img and img_meta
+        should be double nested (i.e.  List[Tensor], List[List[dict]]), with
+        the outer list indicating test time augmentations.
+        """
         if return_loss:
             return self.forward_train(img, img_meta, **kwargs)
         else:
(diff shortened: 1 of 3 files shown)


hellock:
Thanks for the nice fix! Could you open a separated PR to fix the ssd_head and ghm_loss?

Erotemic:
Sure, I've pushed those changes up as PR #1578. The tests in this PR are based on having SSD and GHM available on CPU, so I'll have to reorganize this PR a bit.

Erotemic:
This should now be all set to merge.

Erotemic:
@hellock, are there any unresolved issues with this PR? If not, could it get merged? I'm currently manually merging it in my when building my pipeline, so it would be nice if I could remove that step.

hellock:
Thanks a lot!
I am not sure if this modification will break the original testing behavior. The error in #1477 is caused by different usage in the test case. `img_meta` is designed to be a list of list of dict, where the outer list indicates test-time augs (multiscale, flip, etc.) and the inner list indicates images in a batch. Changing 
`img_metas[0]` to `img_metas[0:1]` may leads to errors for the normal testing.

Erotemic:
I see, that makes more sense. So I assume the imgs should be a `List[Tensor]` where the outer list is also corresponds to test-time augmentations and the Tensor should have a shape NxCxHxW, where N is the batch size?

Given this knowledge I'll revert my changes, add a docstring indicating the expected layout of the inputs, and then change the tests accordingly.

hellock:
Yes `imgs` is a `List[Tensor]` during testing. There should have been some docstrings for the testing data preparation.

Erotemic:
I think I see what my stumbling block was: because `forward_test` expects that you are using test-time augmentation the call of `forward` is different when `return_loss=False` vs when `return_loss=True`. That was my main confusion. I think it would be nice if `forward` had the same signature regardless of whether you were in train or test mode, but I'm not sure if anything can be done about it without breaking backwards compatibility.

Either way, I've added docs to make this behavior more clear as well as tests which demonstrate how to use these models in a standalone fashion.  This means there aren't any functional changes anymore, but merging this PR should increase test coverage by quite a bit.

hellock:
I agree that the inconsistent signatures for training and testing may be confusing. I have not figured out an elegant way to support test-time augmentations.

BTW, there is a minor linting error.

PR #1528, opened 2019-10-11
fix a scale bug according to the Non-local papaer

Howal:
According to [the original implementation of non-local](https://github.com/facebookresearch/video-nonlocal-net/blob/c253ed5eb004fa2cae0490ed46f5018a3c3b060f/lib/models/nonlocal_helper.py#L86), the correct formula of "Scaled Dot-Product Attention" should be 
image


hellock:
Thanks for the fix! @OceanPang Please have a double check.

Issue #425, opened 2019-03-21
how to use pure background images,which has no gt bbox?


15605190982:
I get a dataset, some images are pure background, how to use them (no gt bbox)to train a model?

HaoxuZhang:
Same question here.

hellock:
Images without gt bboxes are filtered in the dataset currently. We may consider adding support for that.

SUYEgit:
Have you add support for training with background images? It'd be important for reducing over detection. If not, could you please tell me the place or method to modify your codes to support training with background images?

hbhflw2000:
Could I know if you have already added support for that? As far as I know, the model performance will be better if the number of positive samples vs that of negative ones is 1:1, it will help reduce false positives. (Frankly I'm not sure if this is reasonable.)

Thanks in advance for your help.

pkhigh:
@hellock can you suggest the files which we can update so that images with background only can also be used in training ?

BlakeXiaochu:
I face the same problem~ Could anybody tell me how far has the future plan been achieved, or give some suggestions about modify the code by myself ?

zvadaszi:
I'm also interested in this feature or some guidance on what to change in the code. Thanks!

hellock:
Solved in #1531

PR #1531, opened 2019-10-11
Allow for images to contain zero true detections

Erotemic:
When I went to train a CascadeRCNN on my dataset the loss computation failed when it loaded an image that had no truth boxes on it. I'm a bit surprised that this was an issue. Perhaps I'm using the library incorrectly? Does this library expect that there exists some magic negative bounding box in cases where there are really no objects of interest n an image?

If this is indeed a real issue, I think I fixed it. I also added tests to ensure that these corner cases don't break in the future. The main issue was in `MaxIoUAssigner`, which explicitly disallowed both the number of predicted boxes to be zero and the number of truth boxes to be zero. I modified the code so it instead returns an appropriate empty assignment if either truth or predictions have no boxes. 

 There was also an issue in bbox_head, where it asserted that all images had truth. I simply removed this check, and I believe the rest of the code still functions correctly (but it would be good if someone could double check this). 

Lastly I added some docs to AssignResult to make it clear what the object contains. 

----

EDIT 2019-10-18: I also fixed the ApproxMaxIoUAssigner and PointAssigner and added corresponding tests cases. 

There were issues in CascadeRCNN, where it would crash when trying to RoiAlign the assigned ROIs in the case where the assignment was empty. I simply added some logic to skip that step, which is the correct thing to do.

There was an issue in loss_bbox, where it failed to compute a bbox loss when all boxes are assigned to the background. Again, the fix for this case is a simple check and skipping the computation of that loss term. 

There was also an issue in my previous code where if no truth was made all predicted boxes got a gt_ind of -1, which means "dont care". I fixed this so now they correctly get assigned to 0, which means background. 

Lastly, I added two tests to make sure cascade rcnn could compute losses for batches that had no truth boxes.  I also added a test case for AnchorHead loss to ensure it computes background loss correctly in the case where the batch has no truth. 

---

EDIT 2019-10-21: I found another edge case in convfc_bbox_head, where `x.view(x.shape[0], -1)` raised an error when x.shape[0] was 0. I added a function `_view_flat_trailing_dims` which tests for and handles this case. 

---

EDIT 2019-10-29: I fixed a logic error where I wrote `pos_inds.numel()` instead of `pos_inds.any()`, rebased on master, and added tests for BBoxHead.


Erotemic:
The dashboards seem to be failing for all builds? I don't see any change in the history that might have caused this. Is there an issue with travis?

hellock:
> The dashboards seem to be failing for all builds? I don't see any change in the history that might have caused this. Is there an issue with travis?

Seems to be caused by the prebuilt PyTorch 1.3, which requires CUDA 10.1. It should be fixed in #1534.

hellock:
#425

Erotemic:
Cool, I'll rebase on master once #1534 is merged. Then this PR should fix #425

Erotemic:
@AAnoosheh my main concern in this PR is to increase the robustness of the models regardless of the underlying dataloading / training loop. This PR doesn't fix the mmdet training loop because I think the dataloader explicitly ignores empty images. Perhaps the fix the augmenter / data loader would be best addressed by a separate PR after this one is merged?

AAnoosheh:
@Erotemic 
Sure thing.

Also another question: Do these changes still enforce a background-class loss on images without bounding boxes? Or do they just allow an image to pass through without error, but without computing a background loss?

The codebase is confusing enough that I can't figure out what's actually going on.
Thanks!

Erotemic:
@AAnoosheh, This codebase is certainly complex, but its not insurmountable. Things are pretty well compartmentalized and most function have exactly one job (I give a lot of credit to @hellock et al for this disciplined design), which makes it possible to grok things in small chunks.

To answer your question "do empty-gt batch items generate loss?": Yes. If you look in my `test_anchor_head.py` I test this [explicitly](https://github.com/open-mmlab/mmdetection/blob/6557ba8d7eefcbf63db946cd5d96c74a6abfd40c/tests/test_anchor_head.py#L52). In the case where there is no gt, the class loss is nonzero because we want to encourage the RPN to predict background everywhere. However, note the bbox loss is zero because the shape of the boxes isn't particularly important when everything predicts background. 

If you want to ensure that the builtin mmdet  trainer works with empty-gt, then I think you'll have to look in the [datasets](https://github.com/open-mmlab/mmdetection/tree/master/mmdet/datasets) to ensure that no-gt items aren't skipped.  You'll probably have to test that it works, but I think that's all that needs to be done.

yhcao6:
Thanks for the pr.

To test [faster R-50](https://github.com/Erotemic/mmdetection/blob/dev%2Fallow_empty_gt/configs/faster_rcnn_r50_fpn_1x.py) when there is no gt, I hack the program by inserting
```
gt_bboxes = [gt.new_empty(0, 4) for gt in gt_bboxes]
gt_labels = [gt.new_empty(0, ) for gt in gt_labels]
```
at the begining of [farward_train in two_stage.py](https://github.com/open-mmlab/mmdetection/blob/d9172919d00695f0dddb0ab10c89c3b7a8e26427/mmdet/models/detectors/two_stage.py#L165), but the program fails.

It seems there some are logic errors. For example, even there is no gt, the sampler of bbox head will still sample some proposals for training. As a result,  [cls_score.numel()](https://github.com/Erotemic/mmdetection/blob/4c034b7d4145940069ba74ee89dfd4b3a80de6a4/mmdet/models/bbox_heads/bbox_head.py#L110) is not zero and [pos_inds](https://github.com/Erotemic/mmdetection/blob/4c034b7d4145940069ba74ee89dfd4b3a80de6a4/mmdet/models/bbox_heads/bbox_head.py#L125) is also not zero. 

Could you check and test it again?

LitvinchukAndrey:
Hello! I forked your pull request and tried to train Cascade RCNN + Guided Anchoring. And I still got next error. This error is 100% connected with background pictures, because when I changed flag `skip_img_without_anno` in `Albu` extension to `False` this error occurred. (This means, that before that I skipped all pictures without annotations, using `Albu` augmentations extansion and everything worked fine)
```Traceback (most recent call last):
  File "/home/andrii/mmdetection/tools/train.py", line 108, in <module>
    main()
  File "/home/andrii/mmdetection/tools/train.py", line 104, in main
    logger=logger)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 58, in train_detector
    _dist_train(model, dataset, cfg, validate=validate)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 192, in _dist_train
    runner.run(data_loaders, cfg.workflow, cfg.total_epochs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/runner/runner.py", line 358, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/runner/runner.py", line 264, in train
    self.model, data_batch, train_mode=True, **kwargs)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 38, in batch_processor
    losses = model(**data)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/parallel/distributed.py", line 50, in forward
    return self.module(*inputs[0], **kwargs[0])
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/andrii/mmdetection/mmdet/core/fp16/decorators.py", line 49, in new_func
    return old_func(*args, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/detectors/base.py", line 100, in forward
    return self.forward_train(img, img_meta, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/detectors/cascade_rcnn.py", line 197, in forward_train
    *rpn_loss_inputs, gt_bboxes_ignore=gt_bboxes_ignore)
  File "/home/andrii/mmdetection/mmdet/models/anchor_heads/ga_rpn_head.py", line 53, in loss
    gt_bboxes_ignore=gt_bboxes_ignore)
  File "/home/andrii/mmdetection/mmdet/core/fp16/decorators.py", line 127, in new_func
    return old_func(*args, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/anchor_heads/guided_anchor_head.py", line 420, in loss
    ignore_ratio=cfg.ignore_ratio)
  File "/home/andrii/mmdetection/mmdet/core/anchor/guided_anchor_target.py", line 79, in ga_loc_target
    scale = torch.sqrt((gt_bboxes[:, 2] - gt_bboxes[:, 0] + 1) *
IndexError: too many indices for tensor of dimension 1
```
I guess, that this problem occurred not because of your PR, but because background pictures' problem is much deeper and there is still a lot of work to be done. 
Then I wanted to try Cascade RCNN + Libra RCNN. I got almost the same error(if it is needed, I can reproduce that error).

Erotemic:
@yhcao6  you are right. My faster_rcnn test is broken. Unfortunately it doesn't run on travis-CI because the RoIAlign forward implementation either needs a GPU or `torchvision` with the `_C` backend, so those tests were skipped in the CI runs. 

I did have a logic error in `bbox_head.BBoxHead.loss`.  I was assuming `pos_inds` was an array of indexes, but in fact it was an array of boolean flags, which means I should have used `pos_inds.any()` instead of `numel`. I fixed that error and also added a test in `test_heads.py` which tests the BBoxHead.loss function explicitly. 

Correct me if I'm wrong, but I think the `cls_score.numel()` check is ok. I only added it in case there were no predictions. In the more common case where there are sampled predictions but the truth is empty there should be a cls loss because any prediction should be encouraged to choose background as its class. 

@LitvinchukAndrey Yes, there are probably several other code paths where empty bounding boxes will still cause a problem. Because this problem is pretty big it might make sense to fix it incrementally, which will help prevent the PRs from becoming monolithic. In this PR I'm trying to only focus on issues in models. Furthermore, I'm only fixing the code where I encounter the problem and I can write a unit tests that demonstrates that the problem is fixed. I'm explicitly not fixing the code in apis and datasets. Once we have unit tests demonstrating that the loss functions can handle empty truth, it should be easier to go back and fix those other components of the system. 

However, it does look like the problem you are encountering is in models/guided_anchor_head. Perhaps you can follow my examples in `tests/test_heads.py` and write a unit test that reproduces the issue and then find a fix for it in that particular component?

ternaus:
Very excited for this PR to get merged.

I have strange errors when I am trying to run training of the cascade RCNN on 32 GPUs that may be addressed by this PR.

LitvinchukAndrey:
@Erotemic I tried to remove all features like Guided Anchoring and Libra RCNN and train vanilla Cascade RCNN and Faster RCNN. But with of both them I meet the same problem:
```
  File "/home/andrii/mmdetection/tools/train.py", line 108, in <module>
    main()
  File "/home/andrii/mmdetection/tools/train.py", line 104, in main
    logger=logger)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 58, in train_detector
    _dist_train(model, dataset, cfg, validate=validate)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 192, in _dist_train
    runner.run(data_loaders, cfg.workflow, cfg.total_epochs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/runner/runner.py", line 358, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/runner/runner.py", line 264, in train
    self.model, data_batch, train_mode=True, **kwargs)
  File "/home/andrii/mmdetection/mmdet/apis/train.py", line 38, in batch_processor
    losses = model(**data)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/mmcv/parallel/distributed.py", line 50, in forward
    return self.module(*inputs[0], **kwargs[0])
  File "/home/andrii/anaconda3/envs/dev/lib/python3.7/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/andrii/mmdetection/mmdet/core/fp16/decorators.py", line 49, in new_func
    return old_func(*args, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/detectors/base.py", line 100, in forward
    return self.forward_train(img, img_meta, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/detectors/two_stage.py", line 176, in forward_train
    *rpn_loss_inputs, gt_bboxes_ignore=gt_bboxes_ignore)
  File "/home/andrii/mmdetection/mmdet/models/anchor_heads/rpn_head.py", line 51, in loss
    gt_bboxes_ignore=gt_bboxes_ignore)
  File "/home/andrii/mmdetection/mmdet/core/fp16/decorators.py", line 127, in new_func
    return old_func(*args, **kwargs)
  File "/home/andrii/mmdetection/mmdet/models/anchor_heads/anchor_head.py", line 189, in loss
    sampling=self.sampling)
  File "/home/andrii/mmdetection/mmdet/core/anchor/anchor_target.py", line 63, in anchor_target
    unmap_outputs=unmap_outputs)
  File "/home/andrii/mmdetection/mmdet/core/utils/misc.py", line 24, in multi_apply
    return tuple(map(list, zip(*map_results)))
  File "/home/andrii/mmdetection/mmdet/core/anchor/anchor_target.py", line 116, in anchor_target_single
    anchors, gt_bboxes, gt_bboxes_ignore, None, cfg)
  File "/home/andrii/mmdetection/mmdet/core/bbox/assign_sampling.py", line 32, in assign_and_sample
    gt_labels)
  File "/home/andrii/mmdetection/mmdet/core/bbox/samplers/base_sampler.py", line 78, in sample
    assign_result, gt_flags)
  File "/home/andrii/mmdetection/mmdet/core/bbox/samplers/sampling_result.py", line 17, in __init__
    self.pos_gt_bboxes = gt_bboxes[self.pos_assigned_gt_inds, :]
```
I debugged it and found that the problem is because `self.pos_assigned_gt_inds=tensor([], device='cuda:1', dtype=torch.int64)` and it tries to get empty slice. I wonder how didn't you encounter this problem? May be problem is on my side. But previously everything worked fine(I understand that previously background pictures were filtered out by dataset). Maybe I need to change types of boxes and labels in dataset preparation? Now for background picture I have following assignments:
```
bboxes = np.zeros((0, 4))
labels = np.zeros((0, ))
```
Or we need to fix `SamplingResult` for empty `self.pos_assigned_gt_inds` tensor.

Erotemic:
I've rebased on master, removed test_forward2 and consolidated it with test_forward.

yhcao6:
Really thanks for your hard work. I found another inconsistent [config](https://github.com/open-mmlab/mmdetection/blob/master/configs/faster_rcnn_ohem_r50_fpn_1x.py), please have a check about that.

Erotemic:
@yhcao6 I think I've addressed the issue. 

When I was looking into it I found that I may have been setting assigned result slightly incorrectly. Previously I was setting AssignedResult.max_overlaps as an empty tensor when there were no truth boxes, however I believe it should be a 1D zero tensor with shape equal to the number of predicted boxes to indicate that no pred box had any overlap with the truth. 

While debugging this I had to inspect the contents of the AssignResult class often. To make this easier I added a `__str__` and `__repr__` method that print out nice stats about it. I also added a docstring to AssignResult. 

While inspecting AssignResult, I also noticed that there seemed to be a a bug in `add_gt_`. The `max_overlaps` tensor was extended by a OnesTensor of length `self.num_gts`. This causes `gt_inds`, `labels`, and `max_overlaps` to have different shapes, which I think is not correct. To keep the shapes the same I think it should be extended by `len(gt_labels)` instead. I left the original code in with a commend. Please double check that I've understood what is happening here correctly. 

I also added a doctest to `bbox_overlaps` because I thought there was an issue there, but it turns out that there wasn't. But now there's a nice test for it. 

Finally, I added standalone test for samplers in `test_sampler.py` and I added a test of OHEM with empty truth in `test_forward` to ensure it works end-to-end and in a unit test. 

Please take a look and let me know if there are any other outstanding issues.

mmdet/core/bbox/assigners/assign_result.py
```diff
@@ -13,7 +46,45 @@ def add_gt_(self, gt_labels):
...
+            parts.append('max_overlaps.shape={!r}'.format(
+                tuple(self.max_overlaps.shape)))
+        if self.labels is None:
+            parts.append('labels={!r}'.format(self.labels))
+        else:
+            parts.append('labels.shape={!r}'.format(tuple(self.labels.shape)))
+        return ', '.join(parts)
+
+    def __repr__(self):
+        devnice = self.__nice__()
```
yhcao6: devnice --> device

mmdet/core/bbox/assigners/assign_result.py
```diff
@@ -13,7 +46,45 @@ def add_gt_(self, gt_labels):
...
+        classname = self.__class__.__name__
+        return '<%s(%s) at %s>' % (classname, devnice, hex(id(self)))
```
yhcao6: we prefer `str.format()` instead of `%`.

mmdet/core/bbox/assigners/assign_result.py
```diff
@@ -13,7 +46,45 @@ def add_gt_(self, gt_labels):
...
             1, len(gt_labels) + 1, dtype=torch.long, device=gt_labels.device)
         self.gt_inds = torch.cat([self_inds, self.gt_inds])
+
+        # Was this a bug?
+        # self.max_overlaps = torch.cat(
+        #     [self.max_overlaps.new_ones(self.num_gts), self.max_overlaps])
+        # IIUC, It seems like the correct code should be:
         self.max_overlaps = torch.cat(
-            [self.max_overlaps.new_ones(self.num_gts), self.max_overlaps])
+            [self.max_overlaps.new_ones(len(gt_labels)), self.max_overlaps])
```
yhcao6: why `num_gts` not equal to `len(gt_labels)`?

mmdet/core/bbox/assigners/approx_max_iou_assigner.py
```diff
@@ -81,11 +81,15 @@ def assign(self,
...
         """
-
-        if squares.shape[0] == 0 or gt_bboxes.shape[0] == 0:
-            raise ValueError('No gt or approxs')
         num_squares = squares.size(0)
         num_gts = gt_bboxes.size(0)
+
+        if num_squares == 0 or num_gts == 0:
+            # No predictions and/or truth, return empty assignment
+            overlaps = approxs.new(num_gts, num_squares)
```
yhcao6: `overlaps` initialization is not consistent.
In approx_max_iou_assigner: `overlaps = approxs.new(num_gts, num_squares)`
In max_iou_assigner: `max_overlaps = overlaps.new_zeros((num_bboxes, ))`
In point_assigner: `max_overlaps = None`

mmdet/core/bbox/assigners/max_iou_assigner.py
```diff
@@ -69,9 +69,15 @@ def assign(self, bboxes, gt_bboxes, gt_bboxes_ignore=None, gt_labels=None):
...
+        Example:
+            >>> self = MaxIoUAssigner(0.5, 0.5)
+            >>> bboxes = torch.Tensor([[0, 0, 10, 10], [10, 10, 20, 20]])
+            >>> gt_bboxes = torch.Tensor([[0, 0, 10, 9]])
+            >>> assign_result = self.assign(bboxes, gt_bboxes)
+            >>> expected_gt_inds = torch.LongTensor([1, 0])
+            >>> assert torch.all(assign_result.gt_inds == expected_gt_inds)
         """
-        if bboxes.shape[0] == 0 or gt_bboxes.shape[0] == 0:
-            raise ValueError('No gt or bboxes')
```
yhcao6: why delete these two sentences?

mmdet/models/bbox_heads/convfc_bbox_head.py
```diff
@@ -168,6 +173,36 @@ def forward(self, x):
...
+        >>> import torch
+        >>> x = _view_flat_trailing_dims(torch.empty(3, 5, 7))
+        >>> assert tuple(x.shape) == (3, 35)
+        >>> x = _view_flat_trailing_dims(torch.empty(0, 5, 7))
+        >>> assert tuple(x.shape) == (0, 35)
+        >>> x = _view_flat_trailing_dims(torch.empty(0,))
+        >>> assert tuple(x.shape) == (0, 1)
+    """
+    if x.numel() == 0:
+        num_trailing = reduce(mul, x.shape[1:], 1)
```
yhcao6: `reduce` and `mul` introduce two extrac packages: `functools` and `operator`, there should be a better way to implement the multiply of `x.shape[1:]`.

mmdet/models/detectors/cascade_rcnn.py
```diff
@@ -292,8 +298,10 @@ def forward_train(self,
                 pos_is_gts = [res.pos_is_gt for res in sampling_results]
                 roi_labels = bbox_targets[0]  # bbox_targets is a tuple
                 with torch.no_grad():
-                    proposal_list = bbox_head.refine_bboxes(
-                        rois, roi_labels, bbox_pred, pos_is_gts, img_meta)
+                    import xdev
+                    with xdev.embed_on_exception_context:
+                        proposal_list = bbox_head.refine_bboxes(
+                            rois, roi_labels, bbox_pred, pos_is_gts, img_meta)
```
yhcao6: These statements seem redundant.

Erotemic:
@yhcao6 For whatever reason, github wont let me reply directly to your comments, so I'll reply here. 

* `devnice` was actually spelled as intended. It should be read as "dev"-"nice", but in the context of mmdet that is probably confusing. I simply removed the "dev" prefix and changed the variable to `nice`. 

* I did change the `%s` to `str.format`. Nice catch. 

* If I understand correctly, `num_gts` refers to the number of truth currently known about by the `AssignResult` object. However, `len(gt_labels)` represents the number of truth being added by the `add_gt_` method. The old code makes the length of `self.gt_inds`, `self.max_overlaps`, and `self.labels` go out of sync when `len(gt_labels) != num_gts` (which is a case I ran into when writing tests). I'm fairly confident that my modification to this line is correct, but it never hurts to have someone else verify that. 

* On deleting the two sentences in max_iou_assigner: That's the point of the PR, to be able to return an empty assignment instead of raising an error when there are either no predictions or no truth. 

* `reduce` and `mul` come from Python builtin packages, so they shouldn't cause any dependency issues. Without resorting to numpy, this is the best way to multiply an arbitrary sequence of integers in Python. I suppose torch might have a multiply reduce operation, but I think using pure Python is the cleanest way to write this line of code. 

* The xdev was testing code I forgot to remove. Thanks for catching that. 

------------

* Lastly, on the inconsistent assignment of overlaps. 

> In point_assigner: max_overlaps = None

The PointAssigner doesn't seem to every set the `max_overlaps` property to `AssignResult`, so I followed that pattern. 

> In approx_max_iou_assigner: overlaps = approxs.new(num_gts, num_squares)

This is consistent with `max_iou_assigner` because I'm creating `overlaps` (not `max_overlaps`, and passing it to `MaxIoUAssigner.assign_wrt_overlaps`. I use `new` because either `num_gts` or `num_squares` will be zero, so the entire Tensor will be empty and only its shape matters. The code that creates `max_overlaps` is actually in `max_iou_assigner`, which I talk about next:

> In max_iou_assigner: max_overlaps = overlaps.new_zeros((num_bboxes, ))

This is creating the `max_iou` for each predicted bounding box, which will always be zero because there were either no true or no predicted boxes. So, this is correct. This code is being used by `approx_max_iou_assigner`, so these two functions are consistent. Lastly, `point_assigner` doesn't really have a concept of overlap because they are points not boxes, so its correct that this code is not used there. 

--------

I also rebased on master and resolved the merge conflicts.

tests/test_forward.py
```diff
@@ -168,6 +168,164 @@ def test_retina_ghm_forward():
                 batch_results.append(result)
 
 
+def test_cascade_empty_forward():
```
yhcao6: It may be better to named the func as `test_cascade_forward` since it tests both empty and non-empty gt.

tests/test_sampler.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def _context_for_ohem():
+    try:
+        from test_forward import _get_detector_cfg
+    except ImportError:
+        # Hack: grab testing utils from test_forward to make a context for ohem
+        import sys
+        from os.path import dirname
+        sys.path.insert(0, dirname(__file__))
```
yhcao6: Missing `from test_forward import _get_detector_cfg`?

tests/test_sampler.py
```diff
@@ -0,0 +1,184 @@
...
+        ignore_iof_thr=0.5,
+        ignore_wrt_candidates=False,
+    )
+    bboxes = torch.FloatTensor([
+        [0, 0, 10, 10],
+        [10, 10, 20, 20],
+        [5, 5, 15, 15],
+        [32, 32, 38, 42],
+    ])
+    gt_bboxes = torch.FloatTensor([]).view(0, 4)
```
yhcao6: torch.empty(0, 4)

yhcao6:
> @yhcao6 For whatever reason, github wont let me reply directly to your comments, so I'll reply here.
> 
> * `devnice` was actually spelled as intended. It should be read as "dev"-"nice", but in the context of mmdet that is probably confusing. I simply removed the "dev" prefix and changed the variable to `nice`.
> * I did change the `%s` to `str.format`. Nice catch.
> * If I understand correctly, `num_gts` refers to the number of truth currently known about by the `AssignResult` object. However, `len(gt_labels)` represents the number of truth being added by the `add_gt_` method. The old code makes the length of `self.gt_inds`, `self.max_overlaps`, and `self.labels` go out of sync when `len(gt_labels) != num_gts` (which is a case I ran into when writing tests). I'm fairly confident that my modification to this line is correct, but it never hurts to have someone else verify that.
> * On deleting the two sentences in max_iou_assigner: That's the point of the PR, to be able to return an empty assignment instead of raising an error when there are either no predictions or no truth.
> * `reduce` and `mul` come from Python builtin packages, so they shouldn't cause any dependency issues. Without resorting to numpy, this is the best way to multiply an arbitrary sequence of integers in Python. I suppose torch might have a multiply reduce operation, but I think using pure Python is the cleanest way to write this line of code.
> * The xdev was testing code I forgot to remove. Thanks for catching that.
> 
> * Lastly, on the inconsistent assignment of overlaps.
> 
> > In point_assigner: max_overlaps = None
> 
> The PointAssigner doesn't seem to every set the `max_overlaps` property to `AssignResult`, so I followed that pattern.
> 
> > In approx_max_iou_assigner: overlaps = approxs.new(num_gts, num_squares)
> 
> This is consistent with `max_iou_assigner` because I'm creating `overlaps` (not `max_overlaps`, and passing it to `MaxIoUAssigner.assign_wrt_overlaps`. I use `new` because either `num_gts` or `num_squares` will be zero, so the entire Tensor will be empty and only its shape matters. The code that creates `max_overlaps` is actually in `max_iou_assigner`, which I talk about next:
> 
> > In max_iou_assigner: max_overlaps = overlaps.new_zeros((num_bboxes, ))
> 
> This is creating the `max_iou` for each predicted bounding box, which will always be zero because there were either no true or no predicted boxes. So, this is correct. This code is being used by `approx_max_iou_assigner`, so these two functions are consistent. Lastly, `point_assigner` doesn't really have a concept of overlap because they are points not boxes, so its correct that this code is not used there.
> 
> I also rebased on master and resolved the merge conflicts.

Hi, Erotemic, thanks for your reply, now I know the reasons and logic behind the code. 
I make more comments on the code about testing, feel free to correct me if I make some mistakes.
Could you try to figure out why you cannot reply directly after my comment, so that we can know which are proper resolved, which I misunderstand and which are unresolved.

Erotemic:
@yhcao6 Sorry, github simply isn't giving me the option to reply to your reviews. I was able to click resolve, but there was no way to comment. While I would have preferred to respond directly to each comment, I hope responding here isn't too much of an issue. 

 Looks like you took care of most of those issues. I've taken a look and they all seem correct. 

* On deleting the two statements (`img_ids = rois[:, 0].long().unique(sorted=True)` and `assert img_ids.numel() == len(img_metas)`),  I don't remember removing those statements, and it may have simply been a typo or an artifact of debugging. It looks like you added them back in, so that should be all set. 

* Replacing `_view_flat_trailing_dims` with `torch.flatten` was a nice catch. I wasn't aware of that function. 

* The other refactorings you made with renaming and using `torch.empty` over `FloatTensor.view` all seem fine to me.

With your recent patch, I think this branch is all set to merge (at least it LGTM) :smile:.

Erotemic:
@yhcao6 I fixed the linting errors causing Travis to fail. Any chance this could get merged in the near future?

hellock:
@Erotemic Thanks for your contribution. This is an important and non-trivial improvement so that we have to be cautious. Now we are doing a final test and it is very near to be merged.

Erotemic:
@hellock I agree that it is best to be cautions. I think merging this change is likely to break something, albeit that something should be small considering that test cases do cover a large portion of this functionality. 

I've actually run into one of these issues recently: training MaskRCNN using segmentation masks breaks when there are no truth. The fix is simple, (just don't do mask loss if you have no masks) but I doubt this is the only other issue with enabling this feature.

hellock:
Yes the core logic for supporting empty gts is not likely to break since the coverage of test cases is quite good. We mainly verify if other parts like data pre-processing, logging are working well.

I trained two models for person detection, w/ or w/o using images without person annotations. It is confirmed that those images are definitely used for training. The performance of the two models are similar, and I believe that hyper-parameters need to be tuned when using additional background images, which is out of the scope of this PR.

Overall, this PR looks good to me now. I've also reproduced the errors for empty masks. You may push a quick fix.

Erotemic:
On a software note: I pushed up the fix for empty masks. 

On a research note: I'm not surprised that models trained with / without empty images are comparable. I think on average the negative cases in an image without any objects of interest won't provide SGD with much more information than images with objects of interest. However, it does open the possibility of finding and including truly difficult examples that will benefit SGD. I think that is where adding this feature will really shine in terms of improving PR / ROC curves. Also it makes mmdet more robust to unseen datasets which will often contain images without any annotations. (The main reason I'm interested in getting this merged in a timely fashion is because we want to let the users of our [VIAME](https://github.com/VIAME/VIAME) project to train models on their custom datasets, which will almost certainly contain empty images.)

hellock:
Yes hard negative mining is usually applied to a pool of background images in practice.

I fixed the padding transform for empty masks and you may have a check. I will merge it if there is no further issues.

hellock:
@Erotemic Thanks for your enthusiasm and nice work! It finally got merged.

yangninghua:
@hellock  
Hello hellock, I am using the latest master version. Regarding background training, I use mask_rcnn_r50_fpn_1x. An error occurred, but cascade_rcnn_r50_fpn_1x.py does not appear this kind of error

```
Done (t=0.00s)
creating index...
index created!
2020-02-19 15:51:40,487 - mmdet - INFO - Start running, host: spple@spple, work_dir: /home/spple/mmdetection-master/work_dirs/mask_rcnn_r50_fpn_1x
2020-02-19 15:51:40,487 - mmdet - INFO - workflow: [('train', 1)], max: 12 epochs
  File "/home/spple/mmdetection-master/tools/train.py", line 142, in <module>
    main()
  File "/home/spple/mmdetection-master/tools/train.py", line 138, in main
    meta=meta)
  File "/home/spple/mmdetection-master/mmdet/apis/train.py", line 111, in train_detector
    meta=meta)
  File "/home/spple/mmdetection-master/mmdet/apis/train.py", line 305, in _non_dist_train
    runner.run(data_loaders, cfg.workflow, cfg.total_epochs)
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/mmcv/runner/runner.py", line 371, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/mmcv/runner/runner.py", line 271, in train
    for i, data_batch in enumerate(data_loader):
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 582, in __next__
    return self._process_next_batch(batch)
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 608, in _process_next_batch
    raise batch.exc_type(batch.exc_msg)
ValueError: Traceback (most recent call last):
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/torch/utils/data/_utils/worker.py", line 99, in _worker_loop
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/torch/utils/data/_utils/worker.py", line 99, in <listcomp>
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/home/spple/mmdetection-master/mmdet/datasets/custom.py", line 132, in __getitem__
    data = self.prepare_train_img(idx)
  File "/home/spple/mmdetection-master/mmdet/datasets/custom.py", line 145, in prepare_train_img
    return self.pipeline(results)
  File "/home/spple/mmdetection-master/mmdet/datasets/pipelines/compose.py", line 24, in __call__
    data = t(data)
  File "/home/spple/mmdetection-master/mmdet/datasets/pipelines/transforms.py", line 176, in __call__
    self._resize_masks(results)
  File "/home/spple/mmdetection-master/mmdet/datasets/pipelines/transforms.py", line 159, in _resize_masks
    results[key] = np.stack(masks)
  File "<__array_function__ internals>", line 6, in stack
  File "/home/spple/anaconda3/envs/pytorch1_1_0--py3_6--cuda9_0/lib/python3.6/site-packages/numpy/core/shape_base.py", line 421, in stack
    raise ValueError('need at least one array to stack')
ValueError: need at least one array to stack
```

yangninghua:
The same error will also occur in cascade_mask_rcnn_r50_fpn_1x.py. I can understand that this feature has not been added to the instance segmentation?

yangninghua:
I modified the code to run it, I do n’t know if it is right, please evaluate

mmdetection-master/mmdet/datasets/pipelines/transforms.py

```
class Resize(object):
    ...........
    def __call__(self, results):
        if 'scale' not in results:
            self._random_scale(results)
        self._resize_img(results)
        # ----------------------------------------------------------
        # Added judgment here
        if (len(results['gt_bboxes'])==0 and len(results['gt_labels'])==0 and len(results['gt_masks'])==0):
            return results
        self._resize_bboxes(results)
        self._resize_masks(results)
        self._resize_seg(results)
        return results
    ...........


@PIPELINES.register_module
class RandomFlip(object):
    ...........
    def __call__(self, results):
        if 'flip' not in results:
            flip = True if np.random.rand() < self.flip_ratio else False
            results['flip'] = flip
        if 'flip_direction' not in results:
            results['flip_direction'] = self.direction
        if results['flip']:
            # flip image
            results['img'] = mmcv.imflip(
                results['img'], direction=results['flip_direction'])
            # ----------------------------------------------------------
            # Added judgment here
            if (len(results['gt_bboxes']) == 0 and len(results['gt_labels']) == 0 and len(results['gt_masks']) == 0):
                return results
            # flip bboxes
            for key in results.get('bbox_fields', []):
                results[key] = self.bbox_flip(results[key],
                                              results['img_shape'],
                                              results['flip_direction'])
            # flip masks
            for key in results.get('mask_fields', []):
                results[key] = np.stack([
                    mmcv.imflip(mask, direction=results['flip_direction'])
                    for mask in results[key]
                ])

            # flip segs
                results[key] = mmcv.imflip(
                    results[key], direction=results['flip_direction'])
        return results
    ...........

```

yangninghua:
https://github.com/open-mmlab/mmdetection/tree/v1.0.0



I use version 1.0.0 and added unlabeled background image training. There is no problem when using a single GPU. But when I try to train with multiple GPUs,

when run the script "./tools/dist_train.sh ./configs/mask_rcnn_r50_fpn_1x.py 4", meet the problem
2020-01-14 09: 01: 53,683-INFO-workflow: [('train', 1)], max: 12 epochs

Stuck here for hours, please help

yangninghua:
And the occupancy rate of multiple GPUs is 100%

leemengwei:
@yangninghua
in file : mmdet/models/bbox_heads/bbox_head.py
after 119             pos_inds = labels > 0
add:                    losses['loss_bbox'] = torch.tensor(0.).to(bbox_pred.device)
will solve your problem,  I forget where is the reference for this solve. but it works

joihn:
How should I format a coco-style dataset JSON to take into account these "pure background" images ?

PR #1534, opened 2019-10-12
Travis: build freeze PyTorch 1.2

Borda:
freeze PyTorch version 1.2 since compilation with the 1.3 on Travis CI fails with following error:
```
RuntimeError: cuda runtime error (100) : no CUDA-capable device is detected at /pytorch/aten/src/THC/THCGeneral.cpp:50
```
see https://travis-ci.org/Borda/mmdetection/jobs/596941932

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic  # ubuntu 18.04
 language: python
 
 python:
@@ -6,7 +6,7 @@ python:
   - "3.6"
   - "3.7"
 
-env: CUDA=9.2.148-1 CUDA_SHORT=9.2 UBUNTU_VERSION=ubuntu1604
+env: CUDA=10.1.105-1 CUDA_SHORT=10.1 UBUNTU_VERSION=ubuntu1804
 
 # Ref to CUDA installation in Travis: https://github.com/jeremad/cuda-travis
 before_install:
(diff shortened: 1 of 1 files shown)


hellock:
Thanks!

Borda:
just thinking if it should not be also in the requirements...
```
torch==1.1
```

hellock:
> just thinking if it should not be also in the requirements...
> 
> ```
> torch==1.1
> ```

PyTorch 1.3 should work fine on the local server.

hellock:
I think that just upgrading CUDA from 9.2 to 10.1 may fix the travis issue. Is it necessary to freeze PyTorch to 1.2?

Borda:
I have tried upgrading Cuda to 10.1 but I it was not enough so I freezed pytorch...

hellock:
The extension compiling process of PyTorch 1.3 is a little different from 1.2, and the CI will fail since there is no GPU.

Borda:
@hellock I told you and referred in my commit https://github.com/open-mmlab/mmdetection/pull/1534#issuecomment-541360044 you just did not believe me :]

hellock:
My fault😂 I didn't noticed that you have attached the url to travis log since there is none in the pr. I just wanted to have a test to see the errors.

Borda:
Well, if it does not compile because of some interface differences, it should be frozen in requires too...

hellock:
On local gpu servers it can compile with pytorch 1.3. The issue for travis is that travis servers do not have gpus, and 1.2 can compile with cuda but without any gpus.

PR #1536, opened 2019-10-12
Added pre-commit hook

ternaus:
The [pre-commit hook](https://pre-commit.com) showed a lot of value in the projects that I work on.

One can enable to check for `flake8` errors and enforce formatters like:
1. `yapf`
2. `isort`
3. fix the end of files
4. fix trailing whitespaces

to be enforced automatically on every commit. This saves a lot of files, finding these errors earlier at your computer, rather than waiting for CI tool to find them for you later.

In this PR I added config for the pre-commit hook, and updated contributing guide.

basically one needs to once in the folder with mmdetection:

```
pip install pre-commit
```

```
pre-commit install
```

After this everything will work automatically.


hellock:
Awesome! Thanks for improving the contributing workflow.

PR #1567, opened 2019-10-21
Fix model training on windows

mattdawkins:
Get a bunch of errors without these regardless of what type my labels are


ZwwWayne:
Should already support training on windows. Thus, this PR is closed.

PR #1575, opened 2019-10-23
Fix oom when there are too many gts

yhcao6:


ZwwWayne:
I think the code is OK and could be merged.

mmdet/core/bbox/assigners/max_iou_assigner.py
```diff
@@ -72,6 +77,15 @@ def assign(self, bboxes, gt_bboxes, gt_bboxes_ignore=None, gt_labels=None):
...
         if bboxes.shape[0] == 0 or gt_bboxes.shape[0] == 0:
             raise ValueError('No gt or bboxes')
+        assign_on_cpu = True if (self.gpu_assign_thr > 0) and (
+            gt_bboxes.shape[0] > self.gpu_assign_thr) else False
+        if assign_on_cpu:
+            device = bboxes.device
+            bboxes = bboxes.cpu()
+            gt_bboxes = gt_bboxes.cpu()
+            gt_bboxes_ignore = gt_bboxes_ignore.cpu(
+            ) if gt_bboxes_ignore is not None else None
```
hellock:
```python
if gt_bboxes_ignore is not None:
    gt_bboxes_ignore = gt_bboxes_ignore.cpu()
if gt_labels is not None:
    gt_labels = gt_labels.cpu()
```

PR #1602, opened 2019-10-29
Split error message for Bbox assign

ternaus:


Erotemic:
I believe this is fixed in #1531 which patches the code such that the assignment will actually work if you have no truth or predictions.

ternaus:
Sounds good!

I have some issues debugging the code that breaks on 32 GPUs and the error that I get comes from there lines.

PR #1326, opened 2019-09-03
[DeformConv] Fix zero outputs when not running on cuda:0

Lyken17:
Currently, deform_conv only works for `cuda:0`

```python
from ops.dcn.deform_conv import DeformConv, deform_conv

## Works normally
device = "cuda:0"
out = deform_conv(x, offset, weight, 1, [1, 1], 1, 2, 2)
print(out.mean(), out.var())
# tensor(0.1310, device='cuda:0'), tensor(0.1312, device='cuda:0')

## Fails to work 
device = "cuda:1"
out = deform_conv(x, offset, weight, 1, [1, 1], 1, 2, 2)
print(out.mean(), out.var())
# tensor(0., device='cuda:1'), tensor(0., device='cuda:1')
```

The patch aims to fix the issue.

PR #1603, opened 2019-10-29
Add tests for NMS / Issue with NMS on GPU 1. 

### LINKED ISSUE #1608: NMS fails on any non-default GPU.
I'm creating an issue that corresponds to the problem that came up in #1603. I'm reposting the body of that issue here:

I get an error when I try to run NMS code on any GPU except 0. 

The issue is that I get RuntimeError: cuda runtime error (700) : an illegal memory access was encountered at mmdet/ops/nms/src/nms_kernel.cu:103 when I try to run NMS with a Tensor on any device except CPU or 0. The error happens on this line:

  THCudaCheck(cudaMemcpy(&mask_host[0],
                        mask_dev,
                        sizeof(unsigned long long) * boxes_num * col_blocks,
                        cudaMemcpyDeviceToHost));
But I believe the issue is actually here:

  THCState *state = at::globalContext().lazyInitCUDA(); // TODO replace with getTHCState

  unsigned long long* mask_dev = NULL;
  //THCudaCheck(THCudaMalloc(state, (void**) &mask_dev,
  //                      boxes_num * col_blocks * sizeof(unsigned long long)));

  mask_dev = (unsigned long long*) THCudaMalloc(state, boxes_num * col_blocks * sizeof(unsigned long long));
My guess is that THCudaMalloc is creating the mask_dev array on device 0 and not the device corresponding to the input at::Tensor boxes. It looks like state might encode which device a new cuda array is allocated on, so my intuition would be to try and grab the state from boxes. However, I'm not a CUDA expert, so I'm probably totally off base for how to use THCState objects. I was attempting to look through the pytorch docs / source to see if I could figure something out, but I'm not having any luck.

Any pointers on how this issue might be handled would be appreciated. Note that if you have two GPUs you can reproduce the error by checking out this PR and running: xdoctest -m tests/test_nms.py test_nms_device_and_dtypes_gpu

Erotemic:
I added doctests and unit tests for non-max suppression code. These tests were written to debug an error I'm experiencing. Unfortunately they will not manifest on TravisCI because reproducing the issue requires having at least 2 GPUs. 

The issue is that I get `RuntimeError: cuda runtime error (700) : an illegal memory access was encountered at mmdet/ops/nms/src/nms_kernel.cu:103` when I try to run NMS with a Tensor on any device except CPU or 0. The error happens on this line: 

```c++
  THCudaCheck(cudaMemcpy(&mask_host[0],
                        mask_dev,
                        sizeof(unsigned long long) * boxes_num * col_blocks,
                        cudaMemcpyDeviceToHost));
```

But I believe the issue is actually here: 

```c++
  THCState *state = at::globalContext().lazyInitCUDA(); // TODO replace with getTHCState

  unsigned long long* mask_dev = NULL;
  //THCudaCheck(THCudaMalloc(state, (void**) &mask_dev,
  //                      boxes_num * col_blocks * sizeof(unsigned long long)));

  mask_dev = (unsigned long long*) THCudaMalloc(state, boxes_num * col_blocks * sizeof(unsigned long long));
```

My guess is that `THCudaMalloc` is creating the `mask_dev` array on device 0 and not the device corresponding to the input `at::Tensor boxes`. It looks like `state` might encode which device a new cuda array is allocated on, so my intuition would be to try and grab the state from `boxes`. However, I'm not a CUDA expert, so I'm probably totally off base for how to use `THCState` objects. I was attempting to look through the pytorch docs / source to see if I could figure something out, but I'm not having any luck. 

Any pointers on how this issue might be handled would be appreciated. Note that if you have two GPUs you can reproduce the error by checking out this PR and running: `xdoctest -m tests/test_nms.py test_nms_device_and_dtypes_gpu`. 

------
EDIT:

Fixes #1608 

This problem was solved by using DeviceGaurd as in #1326

### DIFF
--- a/mmdet/ops/nms/nms_wrapper.py
+++ b/mmdet/ops/nms/nms_wrapper.py
@@ -21,6 +21,18 @@ def nms(dets, iou_thr, device_id=None):
     Returns:
         tuple: kept bboxes and indice, which is always the same data type as
             the input.
+
+    Example:
+        >>> dets = np.array([[49.1, 32.4, 51.0, 35.9, 0.9],
+        >>>                  [49.3, 32.9, 51.0, 35.3, 0.9],
+        >>>                  [49.2, 31.8, 51.0, 35.4, 0.5],
+        >>>                  [35.1, 11.5, 39.1, 15.7, 0.5],
+        >>>                  [35.6, 11.8, 39.3, 14.2, 0.5],
+        >>>                  [35.3, 11.5, 39.9, 14.5, 0.4],
+        >>>                  [35.2, 11.7, 39.7, 15.7, 0.3]], dtype=np.float32)
+        >>> iou_thr = 0.7
+        >>> supressed, inds = nms(dets, iou_thr)
+        >>> assert len(inds) == len(supressed) == 3
     """
     # convert dets (tensor or numpy array) to tensor
     if isinstance(dets, torch.Tensor):
@@ -50,6 +62,18 @@ def nms(dets, iou_thr, device_id=None):
 
 
 def soft_nms(dets, iou_thr, method='linear', sigma=0.5, min_score=1e-3):
+    """
+    Example:
+        >>> dets = np.array([[4., 3., 5., 3., 0.9],
+        >>>                  [4., 3., 5., 4., 0.9],
+        >>>                  [3., 1., 3., 1., 0.5],
+        >>>                  [3., 1., 3., 1., 0.5],
+        >>>                  [3., 1., 3., 1., 0.4],
+        >>>                  [3., 1., 3., 1., 0.0]], dtype=np.float32)
+        >>> iou_thr = 0.7
+        >>> supressed, inds = soft_nms(dets, iou_thr, sigma=0.5)
+        >>> assert len(inds) == len(supressed) == 3
+    """
     if isinstance(dets, torch.Tensor):
         is_tensor = True
         dets_np = dets.detach().cpu().numpy()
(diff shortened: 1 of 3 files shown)


hellock:
Not sure but maybe related to https://github.com/open-mmlab/mmdetection/pull/1326?

Erotemic:
@hellock It looks like adding a DeviceGaurd solves the problem on my machine. 

I took a peek at the source and it looks like adding this code effectively saves the current cuda device, calls `cudaSetDevice(device.index())` (from `cuda_runtime_api.h`) on initialization to change the global cuda device to match the input tensor, and then on destruction resets the global cuda device to whatever it was before.

Relevant source files are here: 
https://github.com/pytorch/pytorch/blob/2d485ffb17ebdec54b399df591e1da031e101d46/c10/cuda/impl/CUDAGuardImpl.h

https://github.com/pytorch/pytorch/blob/39a4ec414121cb48ca540dacbc4fcb31f475ab0e/c10/core/DeviceGuard.h  

https://github.com/pytorch/pytorch/blob/39a4ec414121cb48ca540dacbc4fcb31f475ab0e/c10/core/impl/InlineDeviceGuard.h

Also note that there may be potential overhead issues with using DeviceGaurd as noted in: 

https://github.com/pytorch/pytorch/issues/13974

https://github.com/pytorch/pytorch/issues/14234

However, I think these are mainly related to JIT code, so its probably right to use a DeviceGaurd here. Given that the tests pass I think this is ready for merge.

Issue #1644, opened 2019-11-08
test.py shows boxes different that saves to output.pkl


korabelnikov:
mmdetection/tools/test.py 
config.py latest.pth --out out.pkl --show
will shows less boxes than saves to out.pkl

ZwwWayne:
Hi @korabelnikov ,
Please check the discussion in the PR https://github.com/open-mmlab/mmdetection/pull/1645.

PR #1645, opened 2019-11-08
visualization and scale fix for tools/test*.py #1644 #1646

korabelnikov:
pls review change in tools/test_robustness.py . I could't realize what dataset.img_norm_cfg used for

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -21,13 +21,15 @@ def single_gpu_test(model, data_loader, show=False):
     results = []
     dataset = data_loader.dataset
     prog_bar = mmcv.ProgressBar(len(dataset))
+    score_thr = model.module.test_cfg['score_thr']
+
     for i, data in enumerate(data_loader):
         with torch.no_grad():
-            result = model(return_loss=False, rescale=not show, **data)
+            result = model(return_loss=False, rescale=False, **data)
         results.append(result)
 
         if show:
-            model.module.show_result(data, result)
+            model.module.show_result(data, result, score_thr=score_thr)
 
         batch_size = data['img'][0].size(0)
         for _ in range(batch_size):
--- a/tools/test_robustness.py
+++ b/tools/test_robustness.py
@@ -132,8 +132,8 @@ def single_gpu_test(model, data_loader, show=False):
         results.append(result)
 
         if show:
-            model.module.show_result(data, result, dataset.img_norm_cfg,
-                                     score_thr=score_thr)
+            model.module.show_result(
+                data, result, dataset.img_norm_cfg, score_thr=score_thr)
 
         batch_size = data['img'][0].size(0)
         for _ in range(batch_size):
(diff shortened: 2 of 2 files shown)


korabelnikov:
does anybody know what's wrong yapf have found?

ZwwWayne:
Hi @korabelnikov , you could check the details in https://travis-ci.com/open-mmlab/mmdetection/jobs/254451545#L1154

tools/test.py
```diff
@@ -21,13 +21,15 @@ def single_gpu_test(model, data_loader, show=False):
     results = []
     dataset = data_loader.dataset
     prog_bar = mmcv.ProgressBar(len(dataset))
+    score_thr = model.module.test_cfg['score_thr']
+
     for i, data in enumerate(data_loader):
         with torch.no_grad():
-            result = model(return_loss=False, rescale=not show, **data)
+            result = model(return_loss=False, rescale=False, **data)
```
hellock: If `show` is specified to False, which means the results are not used for debugging or demonstration, then we need to rescale the results to the original size.

tools/test.py
```diff
@@ -21,13 +21,15 @@ def single_gpu_test(model, data_loader, show=False):
     results = []
     dataset = data_loader.dataset
     prog_bar = mmcv.ProgressBar(len(dataset))
+    score_thr = model.module.test_cfg['score_thr']
```
hellock: `score_thr` in the test config is usually too small (e.g., 0.05) for visualization, which is usually set to `0.5`.

ZwwWayne:
Hi @korabelnikov ，
We cannot merge this PR because when ` show=True`, the image and the boxes are shown in the scaled size, which might be different from the input images' original sizes. But during test/eval, we need to rescale the `bbox` to the corresponding input images' original sizes to be consistent with the annotation. However, your modification does not follow this logic, so it cannot be merged.

korabelnikov:
hi @ZwwWayne !
Thank you for a quick response.
As mentioned in issues (https://github.com/open-mmlab/mmdetection/issues/1646) now image and bboxes have different scale and it is real bug.

What's the best way to have them with same scale? Should we show both image and bboxes in original scale?

ZwwWayne:
Hi @korabelnikov ，
Different options have different usages. For `show=True`, we suppose you only use the results for visualization and debug, so it might be better to keep the scales that forwarded to the network.

korabelnikov:
@ZwwWayne but how user can realize it?
My behavior as user: I see s script that saves detection to file. ok. I want to visualize these detection, I use a flag '--show', not output file is broken.
We should either keep images and bboxes in same scale or disable saving to file when 'show' is True

ZwwWayne:
Thanks for your kind advice.
Do you think it will be better if we add some instructions in the [guidance](https://github.com/open-mmlab/mmdetection/blob/master/docs/GETTING_STARTED.md) for now?
Refactoring the logic might take a while and we could plan it in the future.

hellock:
@korabelnikov @ZwwWayne  We may raise a warning when both `--show` and `--out` are specified.

korabelnikov:
@ZwwWayne good idea

@hellock it would be better to raise and exception else a user will got broken output which visually OK

PR #1647, opened 2019-11-08
Async inference interface

DXist:
This PR adds async inference interface implemented for two stage detector with RPN head.

Concurrency is supported by two async context managers:

1. `completed` allows to switch execution context during GPU bound computations like convolutions. CUDA stream is restored when the context is switched back.
2. `concurrent` makes it possible to execute block of code concurrently in a dedicated pool of CUDA streams. The pool is constructed and passed via `asyncio.Queue`. It's an alternative to tensor-level batch inference.

I've tested this implementation on my current project that uses single thread for inference. Async variant has ~30% higher throughput than blocking one but the GPU is still underutilized and there is a room for improvement.

Concurrency can be increased by using `BBoxTestMixin.async_test_bboxes`. At the moment it crashes on memory errors. Probably there are some operations that need to be fixed for concurrent execution.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -111,32 +111,32 @@ Async interface allows not to block CPU on GPU bound inference code and enables 
 ```python
 import asyncio
 import torch
-from mmdet.apis import init_detector, inference_detector, show_result
+from mmdet.apis import init_detector, async_inference_detector, show_result
 
 async def main():
-	config_file = 'configs/faster_rcnn_r50_fpn_1x.py'
-	checkpoint_file = 'checkpoints/faster_rcnn_r50_fpn_1x_20181010-3d1b3351.pth'
-	device = 'cuda:0'
-	model = init_detector(config_file, checkpoint=checkpoint_file, device=device)
+    config_file = 'configs/faster_rcnn_r50_fpn_1x.py'
+    checkpoint_file = 'checkpoints/faster_rcnn_r50_fpn_1x_20181010-3d1b3351.pth'
+    device = 'cuda:0'
+    model = init_detector(config_file, checkpoint=checkpoint_file, device=device)
 
-	# queue is used for concurrent inference of multiple images
-	streamqueue = asyncio.Queue()
-	# queue size defines concurrency level
-	streamqueue_size = 3
+    # queue is used for concurrent inference of multiple images
+    streamqueue = asyncio.Queue()
+    # queue size defines concurrency level
+    streamqueue_size = 3
 
-	for _ in range(streamqueue_size):
-		streamqueue.put_nowait(torch.cuda.Stream(device=device))
+    for _ in range(streamqueue_size):
+        streamqueue.put_nowait(torch.cuda.Stream(device=device))
 
-	# test a single image and show the results
-	img = 'test.jpg'  # or img = mmcv.imread(img), which will only load it once
+    # test a single image and show the results
+    img = 'test.jpg'  # or img = mmcv.imread(img), which will only load it once
 
-	async with concurrent(streamqueue):
-		result = await async_inference_detector(model, img)
+    async with concurrent(streamqueue):
+        result = await async_inference_detector(model, img)
 
-	# visualize the results in a new window
-	show_result(img, result, model.CLASSES)
-	# or save the visualization results to image files
-	show_result(img, result, model.CLASSES, out_file='result.jpg')
+    # visualize the results in a new window
+    show_result(img, result, model.CLASSES)
+    # or save the visualization results to image files
+    show_result(img, result, model.CLASSES, out_file='result.jpg')
 
 
 asyncio.run(main())
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -74,6 +74,7 @@ python demo/webcam_demo.py configs/faster_rcnn_r50_fpn_1x.py \
 
 ### High-level APIs for testing images
 
+#### Synchronous interface
 Here is an example of building the model and test given images.
 
 ```python
@@ -102,6 +103,45 @@ for frame in video:
 ```
 
 A notebook demo can be found in [demo/inference_demo.ipynb](../demo/inference_demo.ipynb).
+
+#### Asynchronous interface - supported for Python 3.7+
+
+Async interface allows not to block CPU on GPU bound inference code and enables better CPU/GPU utilization for single threaded application. Inference can be done concurrently either between different input data samples or between different models of some inference pipeline.
+
+```python
+import asyncio
+import torch
+from mmdet.apis import init_detector, inference_detector, show_result
+
+async def main():
+	config_file = 'configs/faster_rcnn_r50_fpn_1x.py'
+	checkpoint_file = 'checkpoints/faster_rcnn_r50_fpn_1x_20181010-3d1b3351.pth'
+	device = 'cuda:0'
+	model = init_detector(config_file, checkpoint=checkpoint_file, device=device)
+
+	# queue is used for concurrent inference of multiple images
+	streamqueue = asyncio.Queue()
+	# queue size defines concurrency level
+	streamqueue_size = 3
+
+	for _ in range(streamqueue_size):
+		streamqueue.put_nowait(torch.cuda.Stream(device=device))
+
+	# test a single image and show the results
+	img = 'test.jpg'  # or img = mmcv.imread(img), which will only load it once
+
+	async with concurrent(streamqueue):
+		result = await async_inference_detector(model, img)
+
+	# visualize the results in a new window
+	show_result(img, result, model.CLASSES)
+	# or save the visualization results to image files
+	show_result(img, result, model.CLASSES, out_file='result.jpg')
+
+
+asyncio.run(main())
+
+```
 
 
 ## Train a model
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -108,10 +108,13 @@ A notebook demo can be found in [demo/inference_demo.ipynb](../demo/inference_de
 
 Async interface allows not to block CPU on GPU bound inference code and enables better CPU/GPU utilization for single threaded application. Inference can be done concurrently either between different input data samples or between different models of some inference pipeline.
 
+See `tests/async_benchmark.py` to compare the speed of synchronous and asynchronous interfaces.
+
 ```python
 import asyncio
 import torch
 from mmdet.apis import init_detector, async_inference_detector, show_result
+from mmdet.utils.contextmanagers import concurrent
 
 async def main():
     config_file = 'configs/faster_rcnn_r50_fpn_1x.py'
--- a/mmdet/models/anchor_heads/anchor_head.py
+++ b/mmdet/models/anchor_heads/anchor_head.py
@@ -1,5 +1,7 @@
 from __future__ import division
 
+from typing import List, Tuple
+
 import numpy as np
 import torch
 import torch.nn as nn
@@ -94,8 +96,18 @@ class AnchorHead(nn.Module):
         bbox_pred = self.conv_reg(x)
         return cls_score, bbox_pred
 
-    def forward(self, feats):
-        return multi_apply(self.forward_single, feats)
+    def forward(
+            self,
+            feats: List[torch.Tensor],
+    ) -> Tuple[List[torch.Tensor], List[torch.Tensor]]:
+        cls_scores = []
+        bbox_preds = []
+        for level, f in enumerate(feats):
+            flat = f.reshape(-1)
+            cls_score, bbox_pred = self.forward_single(f)
+            cls_scores.append(cls_score)
+            bbox_preds.append(bbox_pred)
+        return cls_scores, bbox_preds
 
     def get_anchors(self, featmap_sizes, img_metas, device='cuda'):
         """Get anchors according to feature map sizes.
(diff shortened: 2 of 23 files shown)


hellock:
It would be great to add support for async inference. My main concern is that it would break the backward compatibility for python 3.5&3.6 (although I prefer 3.7 personally). It may be better to allow the normal usage under lower python versions and enable advanced features for python 3.7.

DXist:
@hellock, I prefer Python 3.7 as well. I've added version checks and made async generators available only for Python version >=3.7.

mmdet/models/anchor_heads/anchor_head.py
```diff
@@ -94,8 +95,17 @@ def forward_single(self, x):
         bbox_pred = self.conv_reg(x)
         return cls_score, bbox_pred
 
-    def forward(self, feats):
-        return multi_apply(self.forward_single, feats)
+    def forward(
+            self,
+            feats: List[torch.Tensor],
+    ) -> Tuple[List[torch.Tensor], List[torch.Tensor]]:
```
hellock: We plan to add type hints for the whole project in the future. At this time, it is better to keep the original style.

mmdet/models/detectors/base.py
```diff
@@ -61,6 +61,10 @@ def forward_train(self, imgs, img_metas, **kwargs):
         """
         pass
 
+    @abstractmethod
+    async def async_test(self, img, img_meta, **kwargs):
```
hellock: maybe renamed to `async_simple_test` if we consider supporting aug test later on?

mmdet/utils/contextmanagers.py
```diff
@@ -0,0 +1,129 @@
...
+
+    with torch.cuda.stream(initial_stream):
+        stream = await streamqueue.get()
+        assert isinstance(stream, torch.cuda.Stream)
+
+        try:
+            with torch.cuda.stream(stream):
+                if DEBUG_CONCURRENT:
+                    logger.info("%s %s is starting, stream: %s", trace_name,
+                                name, stream)
```
hellock: `DEBUG_CONCURRENT` may be omitted and we can use `logger.debug()`.

mmdet/utils/contextmanagers.py
```diff
@@ -0,0 +1,129 @@
...
+        torch.cuda.Event(enable_timing=DEBUG_COMPLETED_TIME) for _ in streams
+    ]
+
+    if DEBUG_COMPLETED_TIME:
+        start = torch.cuda.Event(enable_timing=True)
+        stream_before_context_switch.record_event(start)
+
+        cpu_start = time.monotonic()
+    if DEBUG_COMPLETED:
+        logger.info("%s %s starting, streams: %s", trace_name, name, streams)
```
hellock: `DEBUG_COMPLETED` may be omitted and we can use `logger.debug()`.

mmdet/utils/profiling.py
```diff
@@ -0,0 +1,43 @@
+# coding: utf-8
```
hellock: Since we only supports python 3, this may be useless.

hellock:
Thanks! Could you also update the [docs](https://github.com/open-mmlab/mmdetection/blob/master/docs/GETTING_STARTED.md#high-level-apis-for-testing-images) and add an example for async tests?

mmdet/models/anchor_heads/anchor_head.py
DXist: Ok.

This change is leaked from JIT support PR. Hints are required to customize type inference in TorchScript.

DXist:
@hellock I've added an async test https://github.com/DXist/mmdetection/blob/729b562235b55d22360a38d2ee5214d71f96eeb3/tests/test_async.py#L65 and enabled it only for CUDA environments.

DXist:
I tried to debug memory issues I see during concurrent run. 

`cuda-memcheck --tool initcheck` for my test program returned a lot of global memory unitialized access errors like
```
========= Host API memory access error at host access to 0x132a42f000 of size 716800 bytes                                                                                                                                       
=========     Uninitialized access at 0x132a47d1c0 on access by cudaMemcopy source.                                                                                                                                              
=========     Saved host backtrace up to driver entry point at error                                                                                                                                                             
=========     Host Frame:/usr/lib/x86_64-linux-gnu/libcuda.so.1 (cuMemcpyDtoDAsync_v2 + 0x1b9) [0x266a69]                                                                                                                        
=========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/../../../../libcudart.so.10.1 [0x3897e]                                                                                                                
=========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/../../../../libcudart.so.10.1 [0x14b13]                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/../../../../libcudart.so.10.1 (cudaMemcpyAsync + 0x176) [0x54086]                                                                                      
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so (THCStorage_resize + 0xfb) [0x3f28f3b]                                                                                                     
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x568880a]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so (_ZN2at6native12resize_cuda_ERNS_6TensorEN3c108ArrayRefIlEE + 0x42) [0x5688012]                                                            
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x157d6f4]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x59b2700]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so (_ZN2at6native23upsample_nearest2d_cudaERKNS_6TensorEN3c108ArrayRefIlEE + 0xfe) [0x59b620e]                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x3e67532]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x1a64943]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x157db62]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x358fc9d]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch.so [0x1a64943]                                                                                                                                
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_python.so [0x1cf769]                                                                                                                          
 =========     Host Frame:/opt/conda/lib/python3.7/site-packages/torch/lib/libtorch_python.so [0x3b2ab8]                                                                                                                          
 =========     Host Frame:/opt/conda/bin/python (_PyMethodDef_RawFastCallKeywords + 0x254) [0x164744]                                                                                                                             
 =========     Host Frame:/opt/conda/bin/python (_PyCFunction_FastCallKeywords + 0x21) [0x164861]                                                                                                                                 
 =========     Host Frame:/opt/conda/bin/python (_PyEval_EvalFrameDefault + 0x4ecd) [0x1d02bd]                                                                                                                                    
 =========     Host Frame:/opt/conda/bin/python (_PyEval_EvalCodeWithName + 0xac9) [0x114d09]                                                                                                                                     
 =========     Host Frame:/opt/conda/bin/python (_PyFunction_FastCallKeywords + 0x387) [0x163f57]                                                                                                                                 
 =========     Host Frame:/opt/conda/bin/python (_PyEval_EvalFrameDefault + 0x14dc) [0x1cc8cc]                                                                                                                                    
 =========     Host Frame:/opt/conda/bin/python (_PyEval_EvalCodeWithName + 0xac9) [0x114d09]                                                                                                                                     
 =========     Host Frame:/opt/conda/bin/python (_PyFunction_FastCallDict + 0x1d5) [0x115635] 
... frames omitted ...
```    

These messages could be false positives.

DXist:
I configured my detector to use nondefault CUDA stream. In this setup I get wrong and nondeterministic detection results.

I run Nvidia Visual Profiler. I looks like that `RoIAlignForward` operation runs in default CUDA stream instead of configured. @hellock I guess we need CUDAStreamGuard in the op implementation?

Screenshot from 2019-12-13 18-44-49

hellock:
That maybe the truth. I find that in the [FlowNet2 repo](https://github.com/NVIDIA/flownet2-pytorch/blob/master/networks/correlation_package/correlation_cuda.cc#L76) by nvidia, they explicitly specify the current stream in kernels for the correlation op.

DXist:
@hellock , Yes I found this in CUDA Toolkit Programming guide:

> Kernel launches and host <-> device memory copies that do not specify any stream parameter, or equivalently that set the stream parameter to zero, are issued to the default stream.

hellock:
You may update it in this PR, or it's ok if you raise another PR to fix it.

DXist:
I've set current cuda stream for all kernel launches in custom operators. Also I fixed synhronisation issue with memory copy in nms operator. When nms is run in nondefault stream `cudaMemcpy` is run in the default stream and the copy was unsynchronized:
Screenshot from 2019-12-17 10-39-41

I've changed copy function to `cudaMemcpyAsync` and passed current cuda stream.

After that I haven't seen any synchronization issues and was able successfully use `async_test_bboxes` method in `TwoStageDetector.async_simple_test`.

I've tested an async implementation of my object detector and found that maximum inference throughput is achieved when 3 images are processed concurrently (`streamqueue_size=3`).

It's about ~30% faster than initial synchronous variant.

docs/GETTING_STARTED.md
```diff
@@ -103,6 +104,45 @@ for frame in video:
...
 A notebook demo can be found in [demo/inference_demo.ipynb](../demo/inference_demo.ipynb).
 
+#### Asynchronous interface - supported for Python 3.7+
+
+Async interface allows not to block CPU on GPU bound inference code and enables better CPU/GPU utilization for single threaded application. Inference can be done concurrently either between different input data samples or between different models of some inference pipeline.
+
+```python
+import asyncio
+import torch
+from mmdet.apis import init_detector, inference_detector, show_result
```
hellock: A typo here: `inference_detector` -> `async_inference_detector`

docs/GETTING_STARTED.md
```diff
@@ -103,6 +104,45 @@ for frame in video:
...
+
+async def main():
+	config_file = 'configs/faster_rcnn_r50_fpn_1x.py'
```
hellock: Tabs should be replaced with 4 spaces.

docs/GETTING_STARTED.md
```diff
@@ -103,6 +104,45 @@ for frame in video:
...
+	# visualize the results in a new window
+	show_result(img, result, model.CLASSES)
+	# or save the visualization results to image files
+	show_result(img, result, model.CLASSES, out_file='result.jpg')
+
+
+asyncio.run(main())
+
+```
+
```
hellock: There must be users asking if it is faster than synchronized inference. This example may not show the advantages and the time cost should be similar. Maybe we can demonstrate how to test a list of images and/or the speed benchmark?

mmdet/models/detectors/test_mixins.py
```diff
@@ -1,11 +1,33 @@
...
+if sys.version_info >= (3, 7):
+    from mmdet.utils.contextmanagers import completed
+
 
 class RPNTestMixin(object):
 
+    if sys.version_info >= (3, 7):
+
+        async def async_test_rpn(self, x, img_meta, rpn_test_cfg):
+            sleep_interval = rpn_test_cfg.pop("async_sleep_interval", 0.025)
```
hellock: How are the values of `sleep_interval` set? I noticed that 0.025, 0.017 and 0.035 are used.

mmdet/ops/dcn/src/deform_conv_cuda_kernel.cu
```diff
@@ -261,7 +262,7 @@ void deformable_im2col(
         const scalar_t *data_offset_ = data_offset.data<scalar_t>();
         scalar_t *data_col_ = data_col.data<scalar_t>();
 
-        deformable_im2col_gpu_kernel<<<GET_BLOCKS(num_kernels), CUDA_NUM_THREADS>>>(
+        deformable_im2col_gpu_kernel<<<GET_BLOCKS(num_kernels), CUDA_NUM_THREADS, 0, at::cuda::getCurrentCUDAStream()>>>(
```
hellock: There is an inconsistency. In some kernels `at::cuda::getCurrentCUDAStream()` is specified in the execution configuration, while in others `cudaStream_t stream` is added as an argument in the function.

hellock:
@DXist Thanks for all your efforts! I have some comments above and believe that it is very close to be merged.

mmdet/models/detectors/test_mixins.py
DXist: I've measured GPU time required for operations using temporary `profile_time` decorator and set `sleep_interval` a bit higher. I used K80 GPU and single image for measurements.

mmdet/ops/dcn/src/deform_conv_cuda_kernel.cu
DXist: Fixed. Now I use the first style - in the execution configuration.

DXist:
I've rebased and added a benchmark to compare async and sync interfaces. For a MaskRCNN model async interface takes 0,83-0.85 of the inference time required by sync interface.

Extra optimizations that are out of scope of this PR but handy for production use:
- don't use RLE encoding for masks - might be a good option to disable it in the API
- decode jpeg images by TurboJPEG library (in my case about 3x faster than opencv)

Issue #1672, opened 2019-11-15
[HTC] Mismatch in AP of object scale between model zoo and htc paper


thangvubk:
I found that the mask AP between model zoo and HTC paper is comparable but the AP for different object scales is not consistent. 

|        Model       |  AP  | AP50 | AP75 | APS  | APM  | APL  |
|:------------------:|:----:|:----:|------|------|------|------|
| HTC from model zoo | 38.1 | 59.4 | 41.0 | 17.9 | 40.7 | 56.6 |
| HTC from paper     | 38.0 | 59.4 | 40.7 | 20.3 | 40.9 | 52.3 |

The ``HTC from model zoo`` is tested using latest code (@[4d8416](https://github.com/open-mmlab/mmdetection/commit/4d84161f142b7500089b0db001962bbc07aa869d)). I guess the evaluation code has been changed somehow. I report it here and hope we can figure out what is the problem.

Updated: I also test with Mask R-CNN and it also has the same problem.

hellock:
There can also be other reasons.
1. They are not the same checkpoint, we re-trained HTC models before the code release.
2. Code refactoring happens during that period.

hellock:
It is interesting that the AP on small and large objects are very different. There maybe some refactoring which leads to that.

hellock:
Thanks for the fix!

PR #1679, opened 2019-11-18
fix mAP w.r.t. object scales

thangvubk:
This PR related to #1672 that the mAP w.r.t. object scales has been changed after some code refactoring. I think it is important, especially for publications since reviewers may ask about abnormal performance changes.

|      Model     | AP   | AP50 | AP75 | APS  | APM  | APL  |
|:--------------:|------|------|------|------|------|------|
| HTC from paper | 38.0 | 59.4 | 40.7 | 20.3 | 40.9 | 52.3 |
| Before fix     | 38.1 | 59.4 | 41.0 | 17.9 | 40.7 | 56.6 |
| After fix      | 38.1 | 59.4 | 41.0 | 20.3 | 41.1 | 52.8 |

The bug stems from commit [#976629](https://github.com/open-mmlab/mmdetection/commit/976629d4e604f07280eb621b8fc7e6023fc94a90) which separates the results into independent files for ``bbox`` and ``segm``. However, the segmentation results do not have ``bbox`` field which makes the cocotools create default ``bbox`` by the segmentation results (Check [here](https://github.com/cocodataset/cocoapi/blob/636becdc73d54283b3aac6d4ec363cffbb6f9b20/PythonAPI/pycocotools/coco.py#L339)). 

I would suggest keeping the behavior as before that uses a single file for the results. If you agree i am willing to do that.

### DIFF
--- a/mmdet/core/evaluation/coco_utils.py
+++ b/mmdet/core/evaluation/coco_utils.py
@@ -188,6 +188,7 @@ def segm2json(dataset, results):
             for i in range(bboxes.shape[0]):
                 data = dict()
                 data['image_id'] = img_id
+                data['bbox'] = xyxy2xywh(bboxes[i])
                 data['score'] = float(mask_score[i])
                 data['category_id'] = dataset.cat_ids[label]
                 if isinstance(segms[i]['counts'], bytes):


hellock:
Thanks for the fix! It is indeed very important for reporting performance in technical reports.
The reason for using two files for `bbox` and `segm` results is that some methods use different scores for bbox and masks, e.g., Mask Scoring R-CNN. In such case a single file is not applicable for coco evaluation.

thangvubk:
I see. How about generating the separate files only if it is nescessary?

hellock:
We considered only generating separate files when necessary, but it will make the interface too complicated and not consistent. Maybe your current fix is a good tradeoff.

Issue #1687, opened 2019-11-20
Mask-RCNN: Mask target resizing fails with invalid resize


dsuess:
**Describe the bug**

While training a Mask-RCNN on a custom dataset, I get the following error:

```
cv2.error: OpenCV(4.1.1) /io/opencv/modules/imgproc/src/resize.cpp:3720: error: (-215:Assertion failed) !ssize.empty() in function 'resize'
```

This is raised from [mask_target_single](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/mask/mask_target.py#L31) and is due to one of the bounding boxes in `pos_proposals` having a coordinate `-1`.

A similar error at a different position was raised #370 for bounding-box datasets, but since the bounding boxes for instance segmentation datasets are computed on-the-fly, it's not fixed by it.

**Reproduction**
1. What command or script did you run?
Standard training script.
2. Did you make any modifications on the code or config? Did you understand what you have modified?
No. 
3. What dataset did you use?
Proprietary, but I think any dataset with instance masks close enough to the boundary + some rotation augmentation should replicate this.

**Error traceback**
If applicable, paste the error trackback here.
```
...
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/mmcv/runner/runner.py", line 358, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/mmcv/runner/runner.py", line 264, in train
    self.model, data_batch, train_mode=True, **kwargs)
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/mmdet/apis/train.py", line 38, in batch_processor
    losses = model(**data)
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 152, in forward
    outputs = self.parallel_apply(replicas, inputs, kwargs)
  File "/home/users/daniel/.local/conda/envs/mmpond/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 162, in parallel_apply
    return parallel_apply(replicas, inputs, kwargs, self.device_ids[:len(replicas)])
[...]

PR #1688, opened 2019-11-20
Clip proposal boxes to valid range in single_mask_target

dsuess:
This addresses issue #1687


yhcao6:
Thanks for the fixing.
Max clip value should be `max_h - 1` or `max_w - 1`, you may refer to [here](https://github.com/open-mmlab/mmdetection/blob/2b89634fcf2f7b93d6627cb0628e9fb585671fe6/mmdet/models/bbox_heads/bbox_head.py#L151).

dsuess:
Right! It doesn't really matter because the image-slicing below takes that into account, but it's better to be consistent.

PR #1708, opened 2019-11-24
Unified img_metas api

stevehjc:
img_metas (List(dict))
img_meta  (dict)


Erotemic:
FWIW, I'm in support of this change. In the past I've found it annoying that sometimes the variable is called img_meta and sometimes it is img_metas. 

However, note that in  mmdet/models/detectors/base.py : simple_test, the type of img_meta really is supposed to be a List[List[dict]] for test-time augmentation. So, some review is needed to ensure that all changes have been correctly made. I only took a quick look at the PR, but that was the only major error I saw. 

This also should be rebased on master to remove conflicting files.

Issue #1673, opened 2019-11-15
BBox with shape [n, num_classes*4] resize error when keep_ratio=False 


RangiLyu:
**Describe the bug**
I trained libra_faster_rcnn with setting `(type='Resize', img_scale=(512, 512), keep_ratio=False)`. Training is okay but when I run test and inference, it reports RuntimeError. The error is:
```Traceback (most recent call last):
  File "/home/rangilyu/Projects/mmdetection/mmdet/apis/inference.py", line 86, in inference_detector
    result = model(return_loss=False, rescale=True, **data)
  File "/home/rangilyu/anaconda3/envs/mmd/lib/python3.7/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/rangilyu/Projects/mmdetection/mmdet/core/fp16/decorators.py", line 49, in new_func
    return old_func(*args, **kwargs)
  File "/home/rangilyu/Projects/mmdetection/mmdet/models/detectors/base.py", line 119, in forward
    return self.forward_test(img, img_meta, **kwargs)
  File "/home/rangilyu/Projects/mmdetection/mmdet/models/detectors/base.py", line 102, in forward_test
    return self.simple_test(imgs[0], img_metas[0], **kwargs)
  File "/home/rangilyu/Projects/mmdetection/mmdet/models/detectors/two_stage.py", line 273, in simple_test
    x, img_meta, proposal_list, self.test_cfg.rcnn, rescale=rescale)
  File "/home/rangilyu/Projects/mmdetection/mmdet/models/detectors/test_mixins.py", line 60, in simple_test_bboxes
    cfg=rcnn_test_cfg)
  File "/home/rangilyu/Projects/mmdetection/mmdet/core/fp16/decorators.py", line 127, in new_func
    return old_func(*args, **kwargs)
  File "/home/rangilyu/Projects/mmdetection/mmdet/models/bbox_heads/bbox_head.py", line 158, in get_det_bboxes
    bboxes /= torch.from_numpy(scale_factor).to(bboxes.device)
RuntimeError: The size of tensor a (324) must match the size of tensor b (4) at non-singleton dimension 1
```

**Environment**
 - OS: [Ubuntu 18.04]
 - GCC [5.5.0]
 - PyTorch version [1.3.1]
- How you installed PyTorch [conda]
- GPU model [1080Ti]
- CUDA version 10.1

**Bug fix**
[...]

PR #1730, opened 2019-11-28
fix scale factor bug

ZwwWayne:
Fix https://github.com/open-mmlab/mmdetection/issues/1673,
Using the mask r50 from [model zoo](https://github.com/open-mmlab/mmdetection/blob/master/docs/MODEL_ZOO.md) and setting `keep_ratio=False` could obtain 34.3 mask mAP.

### DIFF
--- a/mmdet/models/bbox_heads/bbox_head.py
+++ b/mmdet/models/bbox_heads/bbox_head.py
@@ -155,7 +155,9 @@ class BBoxHead(nn.Module):
             if isinstance(scale_factor, float):
                 bboxes /= scale_factor
             else:
-                bboxes /= torch.from_numpy(scale_factor).to(bboxes.device)
+                scale_factor = torch.from_numpy(scale_factor).to(bboxes.device)
+                bboxes = (bboxes.view(bboxes.size()[0], -1, 4) /
+                          scale_factor).view(bboxes.size()[0], -1)
 
         if cfg is None:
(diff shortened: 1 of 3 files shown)


mmdet/models/mask_heads/fcn_mask_head.py
```diff
@@ -159,6 +159,9 @@ def get_seg_masks(self, mask_pred, det_bboxes, det_labels, rcnn_test_cfg,
             scale_factor = 1.0
 
         for i in range(bboxes.shape[0]):
+            if not isinstance(scale_factor, float) and not isinstance(
+                    scale_factor, np.ndarray):
```
hellock: `if not isinstance(scale_factor, (float, np.ndarray))`

mmdet/models/bbox_heads/bbox_head.py
```diff
@@ -155,7 +155,9 @@ def get_det_bboxes(self,
             if isinstance(scale_factor, float):
                 bboxes /= scale_factor
             else:
-                bboxes /= torch.from_numpy(scale_factor).to(bboxes.device)
+                scale_factor = torch.from_numpy(scale_factor).to(bboxes.device)
+                bboxes = (bboxes.view(bboxes.size()[0], -1, 4) /
```
hellock: `bboxes.size(0)`

PR #1781, opened 2019-12-09
Adopt WIDER-SSD300 config to run on 8 GPUs

### LINKED ISSUE #1740: Cannot reproduce the results of SSD-300 on WIDER_FACE
Hi @sovrasov,
Recently, I'm reproducing the results of SSD-300 on WIDER_FACE.
I use the provided config based on mmdetection and get Recall 36.9, Precision 0.023 and AP 27.3.
The results are worse than Recall 43.4, Precision 0.029 and AP 34.7 (shown in https://github.com/open-mmlab/mmdetection/pull/765#issuecomment-502579220).

Generally, the models provided by mmdetection are trained on 8 gpus. So, is the SSD-300 on WIDER_FACE also trained on 8 gpus? If so, maybe I will change the learning rate because I trained the model on 4 gpus.

Moreover, any other advice?

Thanks a lot.

sovrasov:
Resolves #1740
Learning rate is changes according to @sunke123's experiments.

### DIFF
--- a/configs/wider_face/ssd300_wider_face.py
+++ b/configs/wider_face/ssd300_wider_face.py
@@ -107,7 +107,7 @@ data = dict(
         img_prefix=data_root + 'WIDER_val/',
         pipeline=test_pipeline))
 # optimizer
-optimizer = dict(type='SGD', lr=1e-3, momentum=0.9, weight_decay=5e-4)
+optimizer = dict(type='SGD', lr=0.012, momentum=0.9, weight_decay=5e-4)
 optimizer_config = dict()
 # learning policy
 lr_config = dict(


configs/wider_face/ssd300_wider_face.py
```diff
@@ -107,7 +107,7 @@
         img_prefix=data_root + 'WIDER_val/',
         pipeline=test_pipeline))
 # optimizer
-optimizer = dict(type='SGD', lr=1e-3, momentum=0.9, weight_decay=5e-4)
+optimizer = dict(type='SGD', lr=8*1.5e-3, momentum=0.9, weight_decay=5e-4)
```
hellock: To be consistent with other configs, maybe just `0.02` instead of `8*1.5e-3`?

sovrasov: Ok, but 8*1.5e-3 = 0.012

Issue #1752, opened 2019-12-04
Config of LibraRCNN


littletomatodonkey:
Hi , the following code shows the config of Libra-sampling, the negative sampling thres is set as -1, it seems that may 0 be more reasonable? (There not exists porposals whose overlaps with gt box are less than 0, I think.) 
```
    rcnn=dict(
        assigner=dict(
            type='MaxIoUAssigner',
            pos_iou_thr=0.5,
            neg_iou_thr=0.5,
            min_pos_iou=0.5,
            ignore_iof_thr=-1),
        sampler=dict(
            type='CombinedSampler',
            num=512,
            pos_fraction=0.25,
            add_gt_as_proposals=True,
            pos_sampler=dict(type='InstanceBalancedPosSampler'),
            neg_sampler=dict(
                type='IoUBalancedNegSampler',
                floor_thr=-1,
                floor_fraction=0,
                num_bins=3)),
        pos_weight=-1,
        debug=False)
```

ZwwWayne:
Hi @littletomatodonkey ,
You can check the docstring in the [iou_balanced_neg_sampler.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/bbox/samplers/iou_balanced_neg_sampler.py#L21)
Set -1 means all using IoU balanced sampling.

littletomatodonkey:
Thanks for your reply. @ZwwWayne 
The code is as follows.
```
```

ZwwWayne:
Yes, so at this time **all using IoU balanced sampling**.

littletomatodonkey:
But personally I think, may it be more reasonable for the range starts from 0 and ends with 0.5 ( Since the bg_thres is set as 0.5 )?
It seems that from the figure in thesis, The interevals are acqured from 0 to 0.5 rather than from -1 to 1. 
Did I miss any important notes? @ZwwWayne 

![image](https://user-images.githubusercontent.com/14270174/70214633-c4968500-1776-11ea-8f2d-572db6189346.png)

ZwwWayne:
Yes, thanks for your advice. For now, it seems that the modification does not change the performance much,  so we might keep it or change the config & model zoos in the future.

littletomatodonkey:
OK, thanks for you quick reply again!

PR #1784, opened 2019-12-10
Fix IoU sampling

OceanPang:
Fix one bug in IoU balanced sampling.


hellock:
Fixes #1752

mmdet/core/bbox/samplers/iou_balanced_neg_sampler.py
```diff
@@ -98,6 +98,8 @@ def _sample_neg(self, assign_result, num_expected, **kwargs):
                 floor_set = set()
                 iou_sampling_set = set(
                     np.where(max_overlaps > self.floor_thr)[0])
+                # for sampling interval calculation
+                self.floor_thr == 0
```
qianyizhang: I think you meant, 
self.floor_thr = 0

PR #1806, opened 2019-12-14
Remove keep all stage code in HTC and Cascade RCNN

yhcao6:


hellock:
You may also remove the `keep_all_stages` config.

yafz:
May I ask why? What is wrong with keep_all_stage?

hellock:
@yafz There are no errors but the `keep_all_stages` option makes the inference code complex and also makes the maintenance more painful. However it can be easily implemented by users for debugging purpose.

PR #1819, opened 2019-12-16
Fix loader for images with no proposals

goolhanrry:
Since proposals of an image should be a [k, 4] or [k, 5] tensor, the dummy proposals ought to have the same number of dimensions as well.


hellock:
Thanks for the fix.

PR #1834, opened 2019-12-19
Per class colors in Mask results viz

karolmajek:
Small change to have consistent colors for all of the objects in the class.
I needed this to create a video for YT.


hellock:
Thanks for your contribution. Could you fix the linting error?

PR #1852, opened 2019-12-22
Split SegResizeFlipPadRescale into different existing transforms

yhcao6:


hellock:
Cropping transforms may also need to deal with semantic segmentation masks.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -401,24 +415,10 @@ def __init__(self, scale_factor=1):
...
-                interpolation='nearest')
-        if results['flip']:
-            gt_seg = mmcv.imflip(gt_seg)
-        if gt_seg.shape != results['pad_shape']:
-            gt_seg = mmcv.impad(gt_seg, results['pad_shape'][:2])
-        if self.scale_factor != 1:
-            gt_seg = mmcv.imrescale(
-                gt_seg, self.scale_factor, interpolation='nearest')
-        results['gt_semantic_seg'] = gt_seg
+        for key in results.get('seg_fields'):
```
hellock: `for key in results.get('seg_fields', [])`

PR #1894, opened 2019-12-30
Encapsulate DCN into a ConvModule & Conv_layers

ZwwWayne:
To make the residual module more clean in the [resnet.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/backbones/resnet.py) and [resnext.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/backbones/resnext.py) and enable more flexible usage of Deformable convolution, the DCN is encapsulated into the DCNModule in the [conv_module.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/utils/conv_module.py).

For backward compatibility, the DCNModule also modifies the `_load_from_state_dict` method; thus, it can test the models in the [DCN Model Zoo](https://github.com/open-mmlab/mmdetection/tree/master/configs/dcn) to get the same mAP as before. 

The training reimplementation is ongoing to ensure the initialization and other details are also correct.

**Notice:** A related PR https://github.com/open-mmlab/mmcv/pull/164 for [mmcv](https://github.com/open-mmlab/mmcv) is necessary for this PR to enable the `_load_from_state_dict`. The backward compatibility is tested with these two PRs together.


mmdet/models/backbones/resnet.py
```diff
@@ -158,30 +156,20 @@ def __init__(self,
...
-                self.deformable_groups * offset_channels,
-                kernel_size=3,
-                stride=self.conv2_stride,
-                padding=dilation,
-                dilation=dilation)
-            self.conv2 = conv_op(
+            assert self.conv_cfg is None, 'conv_cfg cannot be None for DCN'
+            # use ConvModule for backward/forward compatibility for now
+
+            # TODO: change to use build_conv_layer when new benchmark
```
hellock: The above two comments are outdated.

mmdet/models/backbones/resnet.py
```diff
@@ -459,6 +437,7 @@ def __init__(self,
 
         self.feat_dim = self.block.expansion * 64 * 2**(
             len(self.stage_blocks) - 1)
+        print(self)
```
hellock: This line should be removed.

mmdet/ops/dcn/deform_conv.py
```diff
@@ -260,6 +279,33 @@ def forward(self, x):
...
+                    prefix[:-1] + '_offset.weight')
+            if (prefix + 'conv_offset.bias' not in state_dict
+                    and prefix[:-1] + '_offset.bias' in state_dict):
+                state_dict[prefix +
+                           'conv_offset.bias'] = state_dict.pop(prefix[:-1] +
+                                                                '_offset.bias')
+
+        if version is not None and version > 1:
+            from mmdet.apis import get_root_logger
+            logger = get_root_logger
```
hellock: `get_root_logger()`

mmdet/ops/dcn/deform_conv.py
```diff
@@ -324,14 +389,43 @@ def __init__(self, *args, **kwargs):
...
```
hellock: `get_root_logger()`

PR #1900, opened 2019-12-31
Code for Cascade RPN - NeurIPS 2019

thangvubk:


hellock:
Thanks for the nice work!
I suggest wrapping all stages of the cascade rpn head into a single module, which can serve as a direct substitute to RPN without inserting a for-loop in detectors. It will be cleaner and easier to integrate it into other detectors. (Honestly the current implementation of Cascade R-CNN may not be the best choice either.)

thangvubk:
Thanks for your suggestion. I will try on that.

yhcao6:
Hi, @thangvubk,
Recently, MMDetection V2.0 is released. Would you like to migrate the current code to the V2.0 version? We plan to merge this PR in the next update.

thangvubk:
Hi @yhcao6,
Sorry for the long delay. Currently, I am busy with another project. I will solve the conflict of the current PR with the master branch ASAP.

yhcao6:
never mind, good luck

thangvubk:
Hi @yhcao6,
I would like to wrap the for loop of CascadeRPN in ``two_stage.py`` into ``CascadeRPNHead.py``. However, it requires passing ``img_meta`` in the ``forward_train()`` of the detector in ``two_stage.py``. I think it is not clean behavior. 

I suggest that in ``dense_head`` module, we can create a ``BaseDenseHead.py`` which defines the ``forward_train()`` and ``forward_test()`` methods. Other dense head modules inherit ``BaseDenseHead``. By this way, we can apply more flexible logic for ``forward_train()`` and ``forward_test()``.  The code is more consistent in ``two_stage.py`` between RPN and RoI parts. The code for ``one_stage.py`` can also be shortened.

yhcao6:
Thanks for the suggestion!

If I understand correctly, the problem is that the `forward` method of `anchor_head.py` has no parameter `img_metas`, but `CascadeRPNHead` needs `img_metas` right?

You can write a detector `CascadeRPN` which inherits `two_stage.py` and override its `forward_train` method, then passing `img_metas` when calling `self.rpn_head()`.

thangvubk:
Yes. You are right. But it seems a problem when i want to integrate CascadeRPN into a two-stage detector.

yhcao6:
Does `Intergrade CascadeRPN into a two-stage detector` means we don't write detector `CascadeRPN` but only replace `RPNHead` with `CascadeRPNHead`?

I also can't think a good way to integrate CascadeRPN into a two-stage detector. It seems the only way is to add `img_metas` to the `forward` method of every dense_head.py, but that is not a good way. So I think it is better to write a detector `CascadeRPN`.

thangvubk:
Yes. I think the solution could be my suggestion above, with a ``BaseDenseHead``. Some pseudo code can be as follow:
```
class BaseDenseHead(nn.Module):

    def forward_train(x, img_metas, gt_bboxes, **kwargs):
        rpn_outs = self(x)
        rpn_loss_inputs = rpn_outs + (gt_bboxes, img_metas)
        rpn_losses = self.loss(
            *rpn_loss_inputs, gt_bboxes_ignore=gt_bboxes_ignore)

        proposal_cfg = ...
        proposal_list = self.get_bboxes(
            *rpn_outs, img_metas, cfg=proposal_cfg)
        return rpn_losses, propsal_list
```
Then other dense head (e.g., AnchorHead, RPNHead, GARPNHead...) just need to inherit the ``BaseDenseHead`` and I can override forward_train in ``CascadeRPNHead``.

hellock:
I vote for the solution to avoid writing a new detector for each DenseHead.

yhcao6:
I also think `BaseDenseHead` is a better way.

thangvubk:
Hi. If you agree with the solution, may I open a PR for BaseDenseHead?

yhcao6:
Sure.

hachreak:
Hi @thangvubk 
any news about this PR? :smile:

thangvubk:
I have completed the code or Cascade RPN and make it compatible with mmdet2 APIs. However, when i check Fast R-CNN with precomputed proposals, the program is hanging with 100% GPUs utils (Trainings with RPN and Faster R-CNN are normal). I am not sure what is the problems. Do you have any suggestions?

thangvubk:
ping @hellock, @yhcao6

yhcao6:
To be more clear, you train [fast rcnn](https://github.com/thangvubk/mmdetection/blob/cascade_rpn_pr/configs/fast_rcnn/fast_rcnn_r50_fpn_1x_coco.py) with the proposals of [rpn](https://github.com/thangvubk/mmdetection/blob/cascade_rpn_pr/configs/rpn/rpn_r50_fpn_1x_coco.py) right?

thangvubk:
yes. you are right. It is similar to guided anchoring.

yhcao6:
I try to train fast_rcnn but met no problems, here are my steps:
1. Get proposals of RPN with the command: 
`./tools/dist_test.sh configs/rpn/rpn_r50_fpn_1x_coco.py models/rpn_r50_fpn_1x_coco_20200218-5525fa2e.pth 8 --eval proposal_fast --out data/rpn_val.pkl`
2. Modify [fast_rcnn](https://github.com/open-mmlab/mmdetection/blob/master/configs/fast_rcnn/fast_rcnn_r50_fpn_1x_coco.py) to load the val proposals in training stage
```
data = dict(
    samples_per_gpu=2,
    workers_per_gpu=2,
    train=dict(
        ann_file=data_root + 'annotations/instances_val2017.json',
        img_prefix=data_root + 'val2017/',
        proposal_file='data/rpn_val.pkl',
        pipeline=train_pipeline),
    val=dict(
        proposal_file='data/rpn_val.pkl',
        pipeline=test_pipeline),
    test=dict(
        proposal_file='data/rpn_val.pkl',
        pipeline=test_pipeline))
```
3. training fast rcnn with command:
`./tools/dist_train.sh configs/fast_rcnn/fast_rcnn_r50_fpn_1x_coco.py 8`

thangvubk:
@yhcao6,
I also can train with your steps, however, when changing number of max proposals to 300 (``num_max_proposals=300``) the training get stuck after few iterations. Can you verify it?

yhcao6:
@thangvubk,
I have fixed that. You can have a try #3670

thangvubk:
Hi @yhcao6,
I have resolved some conflicts and updated the pretrained models. Could you give this PR a review? Thank you!

yhcao6:
> Hi @yhcao6,
> I have resolved some conflicts and updated the pretrained models. Could you give this PR a review? Thank you!

Sure, I will review it asap

mmdet/models/dense_heads/cascade_rpn_head.py
```diff
@@ -0,0 +1,580 @@
...
+        return cls_reg_targets
+
+    def anchor_offset(self, anchor_list, anchor_strides, featmap_sizes):
+        """ Get offest for deformable conv based on anchor shape
+        NOTE: currently support deformable kernel_size=3 and dilation=1
+
+        Args:
+            anchor_list (list[list[tensor])): [NI, NLVL, NA, 4] list of
+                multi-level anchors
+                anchor_strides (list): anchor stride of each level
```
yhcao6, line 298: add a blank line

yhcao6:
Dear @thangvubk, Thanks for your hard work.
Could you add a docstring to every method, you can just detailed the docstring of some important methods, but add a simple description to other methods or add a reference mark, for example see [here](https://github.com/open-mmlab/mmdetection/blob/ced1c57dd212a8542f5a6fbc504eb22338a49e65/mmdet/core/mask/structures.py#L302).
Besides, can you add unit test to make it more rigorous?

hellock:
Task linked: [CU-4qn8k3 [Review] Cascade RPN](https://app.clickup.com/t/4qn8k3)

mmdet/models/detectors/rpn.py
```diff
@@ -83,7 +83,8 @@ def forward_train(self,
         Returns:
             dict[str, Tensor]: A dictionary of loss components.
         """
-        if self.train_cfg.rpn.get('debug', False):
+        if (isinstance(self.train_cfg.rpn, dict)
+                and self.train_cfg.rpn.get('debug', False)):
```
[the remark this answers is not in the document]
thangvubk: In Cascade RPN the ``train_cfg.rpn`` is a ``list[dist]`` not a ``dict``. So I think this change is needed.

thangvubk:
Hi @yhcao6,
Thank you for reviewing my PR. I have fixed the docstrings and method descriptions as required. Look forward to hearing from your feedback.

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+    indicating the ground truth index.
+
+    - -1: don't care
+    - 0: negative sample, no assigned gt
+    - positive integer: positive sample, index (1-based) of assigned gt
+
+    Args:
+        center_ratio: ratio of the region in the center of the bbox to
+            define positive sample.
+        ignore_ratio: ratui of the region to define ignore samples.
```
yhcao6, line 67: ratui --> ratio

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+
+    # TODO update docs
```
yhcao6, line 70: TODO can be deleted

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+    """Get the flag indicate whether anchors are outside img."""
+    img_h, img_w = img_shape[:2]
+    if allowed_border >= 0:
+        inside_flags = valid_flags & \
+            (flat_anchors[:, 0] >= -allowed_border) & \
+            (flat_anchors[:, 1] >= -allowed_border) & \
+            (flat_anchors[:, 2] < img_w + allowed_border) & \
+            (flat_anchors[:, 3] < img_h + allowed_border)
+    else:
+        inside_flags = valid_flags
```
yhcao6, line 48: The above code seems duplicate with [`anchor_inside_flag`](https://github.com/thangvubk/mmdetection/blob/40d5ef19f14d7668a408c3f3aba6bf0ba4fa64e6/mmdet/core/anchor/utils.py#L20-L46)

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+            gt_bbox = gt_bboxes[gt_id, :4]
+
+            # Compute regions
+            ignore_region = calc_region(gt_bbox, r2, stride, featmap_size)
+            ctr_region = calc_region(gt_bbox, r1, stride, featmap_size)
+
+            # 2. Assign -1 to ignore flags
+            ignore_flags = anchor_ctr_inside_region_flags(
+                anchors, stride, ignore_region)
+            mlvl_assigned_gt_inds[lvl][ignore_flags > 0] = -1
```
yhcao6, line 167: It seems the statement can be simplified as `mlvl_assigned_gt_inds[lvl][ignore_flags] = -1`

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+
+            # 3. Assign gt_bboxes to pos flags
+            pos_flags = anchor_ctr_inside_region_flags(anchors, stride,
+                                                       ctr_region)
+            mlvl_assigned_gt_inds[lvl][pos_flags > 0] = gt_id + 1
```
yhcao6, line 172: `mlvl_assigned_gt_inds[lvl][pos_flags] = gt_id + 1`

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+                u_lvl = lvl + 1
+                u_anchors = mlvl_anchors[u_lvl]
+                u_featmap_size = featmap_sizes[u_lvl]
+                u_stride = anchor_strides[u_lvl]
+                u_ignore_region = calc_region(gt_bbox, r2, u_stride,
+                                              u_featmap_size)
+                ignore_flags = anchor_ctr_inside_region_flags(
+                    u_anchors, u_stride, u_ignore_region)
+                mlvl_ignore_flags[u_lvl][ignore_flags > 0] = 1
+
```
yhcao6, line 195: Can we merge code of `lvl > 0` and `lvl < num_lvls - 1`?

mmdet/core/bbox/assigners/region_assigner.py
```diff
@@ -0,0 +1,221 @@
...
+        # 4. (cont.) Assign -1 to ignore adjacent lvl
+        for lvl in range(num_lvls):
+            ignore_flags = mlvl_ignore_flags[lvl]
+            mlvl_assigned_gt_inds[lvl][ignore_flags > 0] = -1
```
yhcao6, line 199: `mlvl_assigned_gt_inds[lvl][ignore_flags] = -1`

mmdet/models/losses/iou_loss.py
```diff
@@ -9,7 +9,7 @@
 
 
 @weighted_loss
-def iou_loss(pred, target, eps=1e-6):
+def iou_loss(pred, target, linear=False, eps=1e-6):
```
yhcao6, line 12: Arguments in docstring should match the parameters of the function.

mmdet/models/dense_heads/cascade_rpn_head.py
```diff
@@ -0,0 +1,623 @@
...
+        in_channels (int): number of input channels.
+        out_channels (int): number of output channels.
+        adapt_cfg (dict): adaptation type and arguments.
+            dilation (uniform anchor) or offset (arbitrary anchors).
+    """
+
+    def __init__(self,
+                 in_channels,
+                 out_channels,
+                 adapt_cfg=dict(type='dilation', dilation=3)):
```
yhcao6, line 28: Can we align the interface with other Convs?

mmdet/models/dense_heads/cascade_rpn_head.py
```diff
@@ -0,0 +1,623 @@
...
+        Returns:
+            offset_list (list[tensor]): [NLVL, NA, 2, 18]: offset of DeformConv
+                kernel.
+        """
+
+        def _shape_offset(anchors, stride):
+            # currently support kernel_size=3 and dilation=1
+            ks = 3
+            dilation = 1
+            pad = (ks - 1) // 2
```
yhcao6, line 311: It may be better to make these parameters configurable and add an assertion to indicate only support kernels = 3

mmdet/core/bbox/assigners/region_assigner.py
thangvubk: I think we should separate the code for clearer interpretation.

mmdet/models/dense_heads/cascade_rpn_head.py
thangvubk: Do you have any suggestion to align with the interface of other Convs

yhcao6: Sorry for the late reply. For example, you can follow the init method of Deform2d: https://github.com/open-mmlab/mmcv/blob/987cb58341878538a86c584757a49539bb1caab6/mmcv/ops/deform_conv.py#L181-L195.
The default value of the init method should be also the same to avoid ambiguity.
Since `AdaptiveConv` is only tested with `kernel_size=3`, we can add an assertion `assert kernel_size == 3, 'Adaptive conv only supports kernels 3'`.

yhcao6:
Please add CascadeRPN to the supported models https://github.com/open-mmlab/mmdetection/blob/master/README.md#benchmark-and-model-zoo

mmdet/models/dense_heads/cascade_rpn_head.py
thangvubk: I see. Also thank you for your new commit.

yhcao6: You are welcome. I am training with the latest codes, should be able to merge if the performance matches.

yhcao6 (approved):

configs/cascade_rpn/crpn_fast_rcnn_r50_caffe_fpn_1x_coco.py
```diff
@@ -0,0 +1,74 @@
...
+            loss_cls=dict(
+                type='CrossEntropyLoss', use_sigmoid=False, loss_weight=1.5),
+            loss_bbox=dict(type='SmoothL1Loss', beta=1.0, loss_weight=1.0))))
+# model training and testing settings
+train_cfg = dict(
+    rcnn=dict(
+        assigner=dict(pos_iou_thr=0.65, neg_iou_thr=0.65, min_pos_iou=0.65),
+        sampler=dict(num=256)))
+test_cfg = dict(rcnn=dict(score_thr=1e-3))
+dataset_type = 'CocoDataset'
```
ZwwWayne, line 25: unnecessary.

mmdet/models/dense_heads/cascade_rpn_head.py
```diff
@@ -0,0 +1,646 @@
...
+        stride (int or tuple, optional): Stride of the convolution. Default: 1
+        padding (int or tuple, optional): Zero-padding added to both sides of
+            the input. Default: 1
+        dilation (int or tuple, optional): Spacing between kernel elements.
+            Default: 3
+        groups (int, optional): Number of blocked connections from input
+            channels to output channels. Default: 1
+        bias (bool, optional): If set True, adds a learnable bias to the
+            output. Default: False.
+        type (str): Type of adaptive conv, can be either 'offset'
```
ZwwWayne, line 31: （str, optional)

mmdet/models/dense_heads/cascade_rpn_head.py
```diff
@@ -0,0 +1,646 @@
...
+class StageCascadeRPNHead(RPNHead):
+    """Stage of CascadeRPNHead.
+
+    Args:
+        in_channels (int): Number of channels in the input feature map.
+        anchor_generator (dict): anchor generator config.
+        adapt_cfg (dict): adaptation config.
+        bridged_feature (bool): wheater update rpn feature.
+        with_cls (bool): wheather use classification branch.
+        sampling (bool): wheather use sampling.
```
ZwwWayne, line 101: (bool, optional)

mmdet/models/losses/iou_loss.py
```diff
@@ -19,13 +19,18 @@ def iou_loss(pred, target, eps=1e-6):
         pred (torch.Tensor): Predicted bboxes of format (x1, y1, x2, y2),
             shape (n, 4).
         target (torch.Tensor): Corresponding gt bboxes, shape (n, 4).
+        linear (bool): If True, use linear scale of loss instead of log scale.
```
ZwwWayne, line 22: (bool, optional)

mmdet/models/losses/iou_loss.py
```diff
@@ -19,13 +19,18 @@ def iou_loss(pred, target, eps=1e-6):
...
+            Default: False.
         eps (float): Eps to avoid log(0).
```
ZwwWayne, line 24: float, optional

configs/cascade_rpn/crpn_fast_rcnn_r50_caffe_fpn_1x_coco.py
thangvubk: Hi. What is unnecessary here?

PR #1907, opened 2020-01-02
Code for paper "Is Sampling Heuristics Necessary in Training Deep Object Detectors"

hust-kevin:
we re-implement the `sample-free`, more detail please refer to [offical repo](https://github.com/ChenJoya/sampling-free)

**What is sample-free**

- for detector which use focal loss, eg retinanet, sample-free means use bce loss rather than focal loss
- for detector which use down-sample, eg faster r-cnn, sample-free means not apply down-sample in both rpn and roi stage

<br>

**How to get comparable result when sample-free**

- `Bias Initialization, Guided Loss, Class-Adaptive Threshold`

<br>

**Re-implement Note**

- to keep the origin mmdetection code structure, we compute prior and set score_thr before training, so for a specific detector and dataset, please run `configs/sample_free/sample_free.py` before training
- we notice the `offical repo` is different with the `paper` in some details especially in `Class-Adaptive Threshold`, so we just follow the offical repo rather then the paper
- the computed prior here is a little different with the offical repo, but the final result is comparable(we do experiment on voc, but we don't have much gpu do experiment on coco), which means the value of prior is robust
- due to the num of the train sample is changed, so the avg factor of bbox loss is changed, we must rescale the bbox loss for rpn and roi stage, otherwise the the rpn loss will larger than roi-stage loss, and the final result is worse
- the bbox ap for coco is 37.8 (36.4 --> 37.8), which is lower than the office repo 38.4 (36.8 --> 38.4)
- due to the hardware limit, we train all models use 2 GPUs


ZwwWayne:
Thanks for your great work. The code does not pass the CI, can you reformat the code?
More detailed instruction for contribution is [here](https://github.com/open-mmlab/mmdetection/blob/master/.github/CONTRIBUTING.md)

hust-kevin:
@ZwwWayne This is my first pr, I have reformated the code, but there still some error, and it seems due to the version of torchvision, while this files have error is original file in mmdetection, how to fix it ?

ZwwWayne:
Hi @hust-kevin ,
The pillow has upgraded to the newest version 7.0 and removed its `PILLOW_VERSION` as you can find in this issue https://github.com/python-pillow/Pillow/issues/4130. From torchvision's issue https://github.com/pytorch/vision/issues/1714, it seems they will fix this in the early next week, so you can try again in the next week.

hugovk:
[torchvision v0.5.0](https://github.com/pytorch/vision/releases/tag/v0.5.0) has been released with the fix:

1. Require `torchvision>=0.5.0`
2. If Pillow was temporarily pinned, remove the pin

configs/sample_free/init_prior.json
```diff
@@ -0,0 +1,2 @@
+{"FasterRCNN_ResNet50_FPN_COCO": 0.0001584489073138684,
```
ZwwWayne, line 1: It seems that we do not need a json file to store these values, making it a hyperparameter to the head should work.

ZwwWayne:
Hi @hust-kevin ,
Recently, MMDetection V2.0 is released. Would you like to migrate the current code to the V2.0 version? We plan to merge this PR in the next update.

hust-kevin:
@ZwwWayne  this pr is not finished yet, in my original plan,  it should done in February, but due to the covid-19,  I stay home from February, and I can't connect the server in school, on the other hand, I am doing an internship，so I don't have time and gpu to finish this pr

PR #1969, opened 2020-01-10
Reorganize requirements, make albumentations optional

Erotemic:
The albumentations requires an old version of the `imgaug` library, which is incompatible with a lot of other software stacks (namely the one I'm using) which uses the latest `imgaug` library. 

It looks like `albumentations` is not tightly coupled with the system and mmdet can work just fine if it isn't installed. Therefore I'd like to propose to make it an optional requirement instead of a forced requirement. 

This PR does that and also reorganizes the requirements into a format that should be easier to use. Namely, requirements are broken down by what they are required for (e.g. runtime, tests, build, and options) and they all live in a `requirements` folder now. The base requirements.txt file uses the `-r` syntax to refer to each of those files, so `pip install -r requirements.txt` will install all requirements. 

I added an enhanced version of the `parse_requirements` function being used in the `setup.py` script, which handles the `-r` syntax. I also used `extras_require` to register each type of requirements separately. So by default when you `pip install mmdet` you will only get the runtime requirements, but if you `pip install mmdet[all]` you will get everything. Similarly you can `pip install mmdet[optional]` to only grab runtime and optional requirements.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -42,8 +42,7 @@ cd mmdetection
 d. Install mmdetection (other dependencies will be installed automatically).
 
 ```shell
-pip install mmcv
-python setup.py develop  # or "pip install -v -e ."
+pip install -v -e .[all]
 ```
 
 Note:
@@ -55,6 +54,8 @@ It is recommended that you run step d each time you pull some updates from githu
 
 3. If you would like to use `opencv-python-headless` instead of `opencv-python`,
 you can install it before installing MMCV.
+
+4. Some dependencies are optional simply running `pip install -v -e .` will only install the minimum runtime requirements. To use optional dependencies like `albumentations` and `imagecorruptions` either install them manually with `pip install -r requirements/optional.txt` or tell pip that optional dependencies are desired with `pip install -v -e .[optional]`.
 
 ### Another option: Docker Image
 
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -42,7 +42,7 @@ cd mmdetection
 d. Install mmdetection (other dependencies will be installed automatically).
 
 ```shell
-pip install -v -e .[all]
+pip install -v -e .
 ```
 
 Note:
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -131,11 +133,83 @@ def make_cython_ext(name, module, sources):
     return extension
 
 
-def get_requirements(filename='requirements.txt'):
-    here = os.path.dirname(os.path.realpath(__file__))
-    with open(os.path.join(here, filename), 'r') as f:
-        requires = [line.replace('\n', '') for line in f.readlines()]
-    return requires
+def parse_requirements(fname='requirements.txt', with_version=False):
```
hellock: Better to set the default value of `with_version` to True?

requirements/runtime.txt
```diff
@@ -0,0 +1,10 @@
+imagecorruptions
```
hellock: Nice work! `imagecorruptions` may also be made optional.

hellock:
Awesome! The installation guide may also be updated accordingly.

requirements/runtime.txt
Erotemic: done

setup.py
Erotemic: I've generally found that its best to not specify a version in the `install_requires`. This is mostly due to pip's limited ability to resolve dependencies. I've often found that it causes more problems than it solves. However, doing a `pip install -r requirements.txt` will respect the version requirements. 

That being said, only a few requirements actually have version dependencies in the txt files. Also it looked like the old parsing script did include the versions, so perhaps its ok to include them in this instance. I've enabled it to ensure that the existing behavior is preserved and we can always disable it if there are any problems.

Erotemic:
I modified `INSTALL.md` and made `imagecorruptions` optional.

docs/INSTALL.md
```diff
@@ -42,8 +42,7 @@ cd mmdetection
 d. Install mmdetection (other dependencies will be installed automatically).
 
 ```shell
-pip install mmcv
-python setup.py develop  # or "pip install -v -e ."
+pip install -v -e .[all]
```
hellock: For most users, minimal installation should be enough. Maybe we can just set the default command to `pip install -v -e .`?

Erotemic: Sounds reasonable. Changed.

Erotemic:
I rebased on the latest master.

docs/INSTALL.md
```diff
@@ -42,8 +42,7 @@ cd mmdetection
 d. Install mmdetection (other dependencies will be installed automatically).
 
 ```shell
-pip install mmcv
-python setup.py develop  # or "pip install -v -e ."
+pip install -v -e .
```
hellock: I test the installation in a new environment and find that `pip install cython` is necessary before installing mmdet, otherwise a `ModuleNotFoundError` error will be raised.

Erotemic: I added a step where the user runs `pip install -r requirements/build.txt`. This will ensure that they have all build requirements before installing mmdetection. That should resolve the issue.

PR #1976, opened 2020-01-14
Non color images

melikovk:
Hi, 
Thank you for great project. This pull request adds ability to load images that are not 3 channel color images. Since opencv treats 'grayscale' images as 2D matrices an empty third dimension have to be added before converting them to tensors.

### DIFF
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -10,8 +10,9 @@ from ..registry import PIPELINES
 @PIPELINES.register_module
 class LoadImageFromFile(object):
 
-    def __init__(self, to_float32=False):
+    def __init__(self, to_float32=False, color_type='color'):
         self.to_float32 = to_float32
+        self.color_type = color_type
 
     def __call__(self, results):
         if results['img_prefix'] is not None:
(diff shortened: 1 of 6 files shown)


mmdet/datasets/pipelines/loading.py
```diff
@@ -10,16 +10,17 @@
...
+        self.color_type = color_type
 
     def __call__(self, results):
         if results['img_prefix'] is not None:
             filename = osp.join(results['img_prefix'],
                                 results['img_info']['filename'])
         else:
             filename = results['img_info']['filename']
-        img = mmcv.imread(filename)
+        img = mmcv.imread(filename, self.color_type)
```
hellock: I suggest expanding dims here to simplify the formatting.

melikovk: Unfortunately this will not work, because opencv will drop a dummy dimension. For example this code 
img = np.random.rand(320,320,1)
cv2.resize(img, (300,300)).shape
will output (300,300)
Now I realize that more changes are necessary to accommodate grayscale (single-channel) images. In particular changes should be introduced into at least some of the transforms (I added proposed changes to RandomCrop). In addition the way I handle it, 'image_shape' and 'pad_shape' will be 2-tuples and corresponding changes might be necessary (I made some for 'pad_shape'). Clearly one can convert grayscale images to 3 channel images by simply replicating the image, but I think it would be useful to support single channel images directly. I will be happy to do it if there is an interest.

hellock: Nice! Did you tried training any models with grayscale images?

melikovk: Yes. I trained small RetinaNet (retinanet_r50_fpn_1x). It works fine, but I used only Resize, RandomCrop, RandomFlip, Pad and Normalize transforms. I will need to check how other transforms behave with grayscale images.

PR #1988, opened 2020-01-15
fix gt_masks type in maskiou_head._get_area_ratio

tyomj:
Current version throws an exception bellow:

`File "/home/user/mmdetection/mmdet/models/mask_heads/maskiou_head.py", line 160, in _get_area_ratio
gt_instance_mask_area = gt_masks.sum((-1, -2)) 
AttributeError: 'list' object has no attribute 'sum'`

`gt_masks` is a list of masks, so the simplest fix is to convert them into np.array.

(I'm actually surprised that no one faced with this bug before ¯\_(ツ)_/¯)

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -833,9 +833,8 @@ class Albu(object):
                     results[label] = np.array(
                         [results[label][i] for i in results['idx_mapper']])
                 if 'masks' in results:
-                    results['masks'] = [
-                        results['masks'][i] for i in results['idx_mapper']
-                    ]
+                    results['masks'] = np.array(
+                        [results['masks'][i] for i in results['idx_mapper']])
 
                 if (not len(results['idx_mapper'])
                         and self.skip_img_without_anno):


yhcao6:
Currently the `Pad` transform will convert `gt_mask` into `ndarray`, you can look up this line https://github.com/open-mmlab/mmdetection/blob/e032ebb757474fbd83ddf93a01977408b72c5193/mmdet/datasets/pipelines/transforms.py#L295
https://github.com/open-mmlab/mmdetection/blob/e032ebb757474fbd83ddf93a01977408b72c5193/configs/ms_rcnn/ms_rcnn_r50_caffe_fpn_1x.py#L136

I tried to run this config`https://github.com/open-mmlab/mmdetection/blob/master/configs/ms_rcnn/ms_rcnn_r50_caffe_fpn_1x.py` with the latest mmdetection, but the program reported no error.

Have you modified something?

tyomj:
Yes, you are right. Thanks for your help.
I was just confused by the docstring:

https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/mask_heads/maskiou_head.py#L115-L116

The problem I faced comes when applying `albu` transformation after `Pad` like this:
https://github.com/open-mmlab/mmdetection/blob/e032ebb757474fbd83ddf93a01977408b72c5193/configs/albu_example/mask_rcnn_r50_fpn_1x.py#L163-L165

So I caught my own bug in this transform which has to be fixed.
I have checked the same pipeline for both mask-rcnn and ms-rcnn. Everything is ok now.

But I see failure during CI building stage, which is not likely to be connected with my small fix. Could you please help with that?

hellock:
#1818

tyomj:
Thanks! Waiting #1818 to be merged.

PR #1995, opened 2020-01-17
Enhance AssignResult and SamplingResult

Erotemic:
#### Added `random` classmethods to SamplingResult and AssignResult. 

These allow for the simple creation of "demo data" that can be use for testing, debugging, and potentially more (although I don't think these are as useful as random boxes might be). 

#### Fix issue in `SamplingResult.__init__`

I've been hitting issues where sometimes `gt_bboxes` looks like its had its dimensions squashed. This causes an error when doing the index selection. Adding two checks fixes it. 

#### Add rng as attribute of RandomSampler

To ensure that the new random methods could be seeded to consistently reproduce an issue if needed, I had to ensure that the `RandomSampler` class could be seeded. I used `ensure_rng`, which I added in a previous PR to accomplish this simplify. 

#### Misc 

I also added a `to` function to `SamplingResult`, where it moves the data onto a different device. I wrote this to test something that turned out not to be an issue, but I figure its a nice API to have so I left it in.

I also added `info` properties to SamplingResult and RandomSampler as well as implementing the `__nice__` debugging string for SamplingResult. 

I did end up using my [ubelt](https://github.com/Erotemic/ubelt) library to accomplish some of these tasks. The core fix and the random clasmethods don't rely on ubelt, so if there is any resistance to adding a dependency I can remove the features that depend on ubelt and still have a functionally useful PR. However, I've been maintaining and keeping ubelt stable for ~2 years, it has 100% test coverage, it itself is fairly small, and its own dependencies are minimal.

### DIFF
--- a/mmdet/core/bbox/assigners/assign_result.py
+++ b/mmdet/core/bbox/assigners/assign_result.py
@@ -1,8 +1,8 @@
 import torch
-import ubelt as ub
+from mmdet.utils import util_mixins
 
 
-class AssignResult(ub.NiceRepr):
+class AssignResult(util_mixins.NiceRepr):
     """
     Stores assignments between predicted and truth boxes.
 
@@ -108,7 +108,7 @@ class AssignResult(ub.NiceRepr):
         Example:
             >>> from mmdet.core.bbox.assigners.assign_result import *  # NOQA
             >>> self = AssignResult.random()
-            >>> print('{}'.format(ub.repr2(self.info, nl=1)))
+            >>> print(self.info)
         """
         from mmdet.core.bbox import demodata
         rng = demodata.ensure_rng(kwargs.get('rng', None))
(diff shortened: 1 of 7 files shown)


hellock:
Thanks for the PR as well as previous ones! ubelt is indeed an excellent util library, but it may be too early to introduce it since we only need two methods/classes of it. We can redefine `__repr__` and `__str__` to achieve the same goal of `NiceRepr`.

Erotemic:
Sounds good. I removed ubelt as a dependency. 

Instead of redefining `__repr__` and `__str__` (which imo would unnecessarily bloat the code) I simply copied the definition of NiceRepr into mmdet.utils. This removes the dependency on ubelt, but keeps the succinct and easy-to-modify `__nice__` definition. It also prevents code redundancy in both `SamplingResult` and `AssignResult`. 

I was also using `ubelt.repr2` in this PR, which unfortunately is a bit too complex to simply copy-paste into a code base, so I simply wrote out some logic that produces a similar effect. Its less pretty than having a simple function that nicely handles nested representations, but it does prevent the extra dependency.

PR #1999, opened 2020-01-18
Encapsulate the second stage into RoI heads

ZwwWayne:
Hi all,
After a discussion with @hellock and @yhcao6 , we decide to encapsulate box and mask heads into RoI head to reduce the codebase's complexity. This refactoring has several advantages as follows.

- The refactoring reduces most of the duplicate code in the detectors. For example, when adding new methods like Double head R-CNN and Cascade R-CNN, the previous implementation always need to create a new file in the detectors directory and copy the same code for RPN. Since most of these modifications are about the RCNN process, it seems to be better to encapsulate the second stage (including box head, mask head, mask scoring head, and potentially key point head, etc.) into the RoI heads to reduce the duplicate codes and logics.

- The refactoring only creates box assigners and samplers once in RoI heads. The previous implementation creates assigner and sampler during each iteration in the training process, this is unnecessary since they are the same during the training process. Now, the assigner and sampler are created once as a sub-module of RoI heads to reduce the unnecessary overhead.

- The refactoring will ease the processes when code refactoring and adding new methods as it reduces duplicate code and logic and seems to be more conceptually simple than before.

**Notice**: The structure of the detectors are changed, as all configs of two-stage detectors are changed. Thus, it might have compatibility issues with the previous model zoo. We will re-benchmark all the models in the configs when all the refactorings are finished, which will be released in early February.


configs/_base_/models/cascade_mask_rcnn_r50_fpn.py
```diff
@@ -28,62 +27,73 @@
...
+                loss_cls=dict(
+                    type='CrossEntropyLoss',
+                    use_sigmoid=False,
+                    loss_weight=1.0),
+                loss_bbox=dict(type='SmoothL1Loss', beta=1.0,
+                               loss_weight=1.0)),
+            dict(
+                type='Shared2FCBBoxHead',
+                in_channels=256,
+                conv_out_channels=256,
```
xvjiarui: `Shared2FCBBoxHead` doesn't have conv.

ZwwWayne:
In commit 3486716, the configs are ensured to have the same key-values as the previous commit 77d073a. The new configs systems also fix some bugs in old commits by reducing duplicated codes.

mmdet/models/roi_heads/mask_scoring_roi_head.py
```diff
@@ -0,0 +1,86 @@
...
+    """Mask Scoring RoIHead for Mask Scoring RCNN.
+
+    https://arxiv.org/abs/1903.00241
+    """
+
+    def __init__(self, mask_iou_head, **kwargs):
+        super(MaskScoringRoIHead, self).__init__(**kwargs)
+        assert mask_iou_head is not None
+        self.mask_iou_head = builder.build_head(mask_iou_head)
+        self.mask_iou_head.init_weights()
```
xvjiarui: Is there any way to move this into the overridden `init_weights`?

mmdet/models/detectors/two_stage.py
```diff
@@ -106,25 +88,10 @@ def forward_dummy(self, img):
...
-        # mask head
-        if self.with_mask:
-            mask_rois = rois[:100]
-            mask_feats = self.mask_roi_extractor(
-                x[:self.mask_roi_extractor.num_inputs], mask_rois)
-            if self.with_shared_head:
-                mask_feats = self.shared_head(mask_feats)
-            mask_pred = self.mask_head(mask_feats)
-            outs = outs + (mask_pred, )
+        proposals = torch.randn(1000, 4).cuda()
```
xvjiarui: It is better to be on the same device as `img`.

ZwwWayne: Sure

mmdet/models/roi_heads/base_roi_head.py
```diff
@@ -0,0 +1,72 @@
+from abc import ABCMeta, abstractmethod
+
+import torch.nn as nn
+
+
+class BaseRoIHead(nn.Module, metaclass=ABCMeta):
+    """Base class for RoIHeads"""
+
+    def __init__(self):
+        super(BaseRoIHead, self).__init__()
```
hellock: It can have a default implementation.

mmdet/models/roi_heads/base_roi_head.py
```diff
@@ -0,0 +1,72 @@
...
+    @property
+    def with_shared_head(self):
+        return hasattr(self, 'shared_head') and self.shared_head is not None
+
+    @abstractmethod
+    def init_weights(self, pretrained):
+        pass
+
+    @abstractmethod
+    def init_bboxhead(self):
```
hellock: `init_bbox_head`

mmdet/models/roi_heads/standard_roi_head.py
```diff
@@ -0,0 +1,290 @@
...
+            self.shared_head = builder.build_shared_head(shared_head)
+
+        if bbox_head is not None:
+            self.init_bboxhead(bbox_roi_extractor, bbox_head)
+
+        if mask_head is not None:
+            self.init_maskhead(mask_roi_extractor, mask_head)
+
+        self.train_cfg = train_cfg
+        self.test_cfg = test_cfg
```
hellock: The above two lines can be moved to the top (after `super().__init__()`)

mmdet/models/roi_heads/standard_roi_head.py
```diff
@@ -0,0 +1,290 @@
...
+        return losses
+
+    def _bbox_forward(self, x, rois):
+        # TODO: a more flexible way to decide which feature maps to use
+        bbox_feats = self.bbox_roi_extractor(
+            x[:self.bbox_roi_extractor.num_inputs], rois)
+        if self.with_shared_head:
+            bbox_feats = self.shared_head(bbox_feats)
+        cls_score, bbox_pred = self.bbox_head(bbox_feats)
+        return cls_score, bbox_pred, bbox_feats
```
hellock: I suggest returning a dict.

mmdet/models/roi_heads/standard_roi_head.py
```diff
@@ -0,0 +1,290 @@
...
+                                                     gt_masks, self.train_cfg)
+            pos_labels = torch.cat(
+                [res.pos_gt_labels for res in sampling_results])
+            loss_mask = self.mask_head.loss(mask_pred, mask_targets,
+                                            pos_labels)
+            return loss_mask
+        else:
+            return None
+
+    def extract_mask_feats(self, x, sampling_results, bbox_feats):
```
hellock: Use unified apis. There are `_bbox_forward_train` and `_bbox_forward`, but `_mask_forward_train` and `extract_mask_feats`.

PR #2004, opened 2020-01-20
WIP: Build manylinux 2014 wheels

Erotemic:
This PR adds a script to build 2014 manylinux wheels. 

It builds on top of #1969 which reorganizes the requirements.txt file to make this script simpler to implement. 

The basic idea is that you execute `./run_manylinux_build.sh`, which then starts a docker container based on the `pytorch/manylinux-cuda101` image. In the docker container we mount the mmdetection source directory. Then we activate a virtual environment for the requested python version, install dependencies, and run `python setup.py bdist_wheel`. After the wheel is build it lives in the `dist` directory, but its not quite ready yet. 

Before we can upload that wheel to pypi we need to ensure that it conforms to the manylinux standard. This is done via auditwheel. The last step runs auditwheel on the first wheel in the `dist` directory which then outputs the manylinux2014 complient wheel in the `wheelhouse` directory. From there you could upload that wheel to pypi via twine, but but it outside the scope of this PR. 

There is one loose end that I haven't looked into / determined how to resolve. The base image I'm starting from is using cuda 10.1. However, there are also images for cuda 9.2 and cuda 10.0. I'm not sure how different cuda dependencies are reflected in wheels (although I believe pytorch does do special handling of these). 

EDIT: There seems to be issues with the wheels that I built. I'm changing the PR tile to WIP while I'm testing and debugging them.


hellock:
Looking forward to it!

Interestingly, I find that there is a package named mmdet uploaded by someone else in pypi. Once this PR finishes, I will try contacting the maintainer to discuss about it.

Erotemic:
To update on the progress of this: 

I'm getting closer. I'm currently working on a different branch: https://github.com/Erotemic/mmdetection/tree/dev/scikit-build 

In addition to adding wheel support that branch also ports the setup.py file use scikit-build instead of the distutils c-extension classes. Scikit-build is just a thin wrapper around CMake, which means that building cross platform libraries will be much easier. It also allows for building the shared objects outside of setup.py, which makes it a bit easier to debug. 

Unfortunately, it seems like I'm not able to use auditwheel to repair the wheels and make them manylinux2014 compatible. However I was able to follow the pytorch build process and perform similar manual patchelf operations. Using this I was able to build a binary on a manylinux machine that also works on my ubuntu 18 machine. 

I also determined how pytorch makes different wheels available for different cuda versions. It seems like only the 10.1 wheels are hosted on pypi, but the 9.2 and cpu-only wheels are hosted on a custom wheel repository: https://download.pytorch.org/whl/torch_stable.html I'm not sure how mmdet would want to handle that case. Perhaps the best option is to simply host the 10.1 wheels (or whatever the current torch version is at) and then provide instructions on how to build wheels for other versions of cuda if the user wants that?

dsuess:
Any progress on this? We've had working wheels for versions 1.x.x, which weren't too hard to create since we only had to patch the `libcudart.so` dependency. But now with the changes to the libtorch in pytorch 1.5, it's become much more difficult to maintain a build-script out-of-source.

As @Erotemic mentioned, we'll probably have to figure out which CUDA versions we want to support and if we want to limit wheel support to the latest pytorch version. 

We've also fiddled with building conda packages (https://anaconda.org/silverpond/mmdetection), which is a whole other can-of-worms.

Erotemic:
I was able to make working wheels for ubuntu systems, but I wasn't able to get manylinux working properly. The ubuntu wheels were good enough for my purposes, so I haven't continued work on it since then. I'm going to work on updating my code to mmdet 2.x soon, and I may revisit this problem then. @dsuess, can you point me to the code that builds working wheels for 1.x.x?

dsuess:
In the `pytorch/manylinux-cuda101` docker image and with mmdetection 1.1, this worked:
```
export TORCH_CUDA_ARCH_LIST="3.7;6.0;6.1;6.2;7.0;7.5"
export FORCE_CUDA=1
export PYTHON="/opt/python/cp36-cp36m/bin/python"
$PYTHON -m pip install --cache-dir /cache/pip t"orch==1.3" cython numpy wheel twine
$PYTHON setup.py bdist_wheel
auditwheel repair --plat manylinux2014_x86_64 dist/mmdet-*.whl
# FIXING WHEEL :(
cd wheelhouse/
$PYTHON -m wheel unpack *.whl
rm  *.whl
cd mmdet*/mmdet
cp /usr/local/cuda/lib64/libcudart.so.10.1 ../mmdet.libs
for SOFILE in `find . -type f | grep .so`; do patchelf --replace-needed libcudart-1b201d85.so.10.1.243 libcudart.so.10.1 $SOFILE; done
cd ../*.dist-info
export SHA256=($(sha256sum ../mmdet.libs/libcudart.so.10.1))
echo "mmdet.libs/libcudart.so.10.1,sha256=$SHA256,`stat --printf="%s" ../mmdet.libs/libcudart.so.10.1`" >> RECORD
cd ../..
$PYTHON -m wheel pack *
```
For 2.x.x the situation is a bit more challenging if you don't want to include all the pytorch so files and make the package 700MB large. I have something working, but I haven't tested it outside of Ubuntu.

ZwwWayne:
For now, op has been moved into MMCV. And this PR will be closed as it is too old.

PR #2039, opened 2020-02-01
Scale Factor Fix

ZwwWayne:
1. This PR uses separate scale factors for width and height because the scale factors of width and height always have minor differences due to quantization.
2. This PR fixes legacy issues about scale factors in many detectors and heads. Some heads did not adopt the change when the scale factors are NumPy array; thus could cause errors in the test.
3. This PR marginally affects performance in both training and testing. So this PR is compatible with the old benchmarked models or those in the future.


Mxbonn:
@hellock, @ZwwWayne, any plans to merge this? (Seems like it was already approved but simply not merged)

hellock:
@Mxbonn Since this is a BC-breaking change, we plan to merge it in the next release.

PR #2056, opened 2020-02-06
Rewrite Soft NMS with pytorch extension

xvjiarui:
1. Replaced Cython soft NMS with PyTorch extension
2. Remove Cython dependency.


setup.py
```diff
@@ -261,6 +235,10 @@ def gen_packages_items():
                 name='nms_cuda',
                 module='mmdet.ops.nms',
                 sources=['src/nms_cuda.cpp', 'src/nms_kernel.cu']),
+            make_cuda_ext(
+                name='soft_nms_cpu',
+                module='mmdet.ops.nms',
+                sources=['src/soft_nms_cpu.cpp']),
```
hellock: I suggest to use only one extension for NMS and SoftNMS.

xvjiarui: I will merge them into one cpp file.

mmdet/ops/nms/src/nms_cpu.cpp
```diff
@@ -66,6 +66,149 @@ at::Tensor nms(const at::Tensor& dets, const float threshold) {
   return result;
 }
 
+template <typename scalar_t>
```
hellock: We can add a line "Modified from https://xxxx, Soft-NMS is added." at the top of this file.

mmdet/ops/nms/nms_wrapper.py
```diff
@@ -76,10 +75,10 @@ def soft_nms(dets, iou_thr, method='linear', sigma=0.5, min_score=1e-3):
     """
```
hellock: We may add a standard docstring for this method like in `nms()`.

```python
def soft_nms(arg1, arg2=0.5):
    """A brief description.

    Detailed description.

    Args:
        <arg1> (<type>): <Description>.
        <arg2> (<arg2>): <Description>. <Default: 0.5>.

    Returns:
        <type>: <Description>.
    """
    pass
```

tests/test_soft_nms.py
```diff
@@ -0,0 +1,42 @@
...
+    assert len(inds) == len(new_dets) == 4
+
+    dets = torch.FloatTensor(base_dets)
+    new_dets, inds = soft_nms(dets, iou_thr)
+    assert dets.dtype == new_dets.dtype
+    assert len(inds) == len(new_dets) == 4
+
+    dets = base_dets.astype(np.float64)
+    new_dets, inds = soft_nms(dets, iou_thr)
+    assert dets.dtype == new_dets.dtype, "{} vs {}".format(
```
hellock: Single quote is preferred in this codebase.

PR #2066, opened 2020-02-08
fix worker_init_fn in data_loader if set seed

yhcao6:


xvjiarui:
Should torch seed also be set here? Like [here](https://github.com/facebookresearch/detectron2/blob/989f52d67d05445ccd030d8f13d6cc53e297fb91/detectron2/utils/env.py#L31)

PR #2068, opened 2020-02-08
Support Better-Aligned ROIAlign Layer

ZwwWayne:
The implementation is modified from [Detectron2](https://github.com/facebookresearch/detectron2) to better align the image features or mask targets for the second stage detectors.

To use the better-aligned version, the users can set `aligned=True` in the configs. 

By default, the current ROIAlign layer remains to use the legacy implementation in MMDetection, which is set as`aligned=False`. Thus, this new feature will not affect the current models in the model zoo, but we will use the new ROIAlign as default in the future after all the models are re-benchmarked with new ROIAlign


mmdet/ops/roi_align/roi_align.py
```diff
@@ -46,13 +46,54 @@ def backward(ctx, grad_output):
...
+            rois,
+            spatial_scale,
+            output_size[0],
+            output_size[1],
+            bs,
+            ch,
+            h,
+            w,
+            sampling_ratio,
+            ctx.aligned,
```
yhcao6: remove last comma.

yhcao6: remove last comma.

mmdet/ops/roi_align/src/roi_align_cuda.cpp
```diff
@@ -19,17 +20,30 @@ int ROIAlignBackwardLaucher(const at::Tensor top_grad, const at::Tensor rois,
                             const int pooled_height, const int pooled_width,
                             at::Tensor bottom_grad);
 
+at::Tensor ROIAlign_forward_cuda_v2(const at::Tensor& input,
```
yhcao6: It may be better to rearrange code similar to the original ROIAlign that contains following two functions:
`ROIAlign_forwardV2` and `ROIAlignForwardV2Laucher`.

mmdet/ops/roi_align/roi_align.py
```diff
@@ -46,13 +46,45 @@ def backward(ctx, grad_output):
...
-                                    out_w, spatial_scale, sample_num,
-                                    grad_input)
+            roi_align_cuda.backward_v1(grad_output.contiguous(), rois, out_h,
+                                       out_w, spatial_scale, sample_num,
+                                       grad_input)
 
         return grad_input, grad_rois, None, None, None
 
 
+class RoIAlignFunctionV2(Function):
```
hellock: It may be merged with `RoIAlignFunction`.

PR #2094, opened 2020-02-14
Speed up sampler

yhcao6:
This pr reimplements `random_choice` with `torch.randperm`. 
It speeds up [libra faster r50](https://github.com/open-mmlab/mmdetection/blob/master/configs/libra_rcnn/libra_faster_rcnn_r50_fpn_1x.py) by around 0.02s\iter on TITAN XP.

### DIFF
--- a/mmdet/core/bbox/samplers/random_sampler.py
+++ b/mmdet/core/bbox/samplers/random_sampler.py
@@ -1,4 +1,3 @@
-import numpy as np
 import torch
 
 from .base_sampler import BaseSampler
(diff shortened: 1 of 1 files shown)


mmdet/core/bbox/samplers/random_sampler.py
```diff
@@ -18,20 +17,18 @@ def __init__(self,
         self.rng = demodata.ensure_rng(kwargs.get('rng', None))
 
     def random_choice(self, gallery, num):
-        """Random select some elements from the gallery.
-
-        It seems that Pytorch's implementation is slower than numpy so we use
-        numpy to randperm the indices.
-        """
+        """Random select some elements from the gallery. """
```
hellock: Add a standard docstring.

Issue #2072, opened 2020-02-10
rois.shape[0] == 0 in deformable pool


txytju:
In deformable pooling operation, if `rois.shape[0]==0`, there will be error when you view x(a tensor of shape (0, C, pooled_h, pooled_w)) to shape of (0, -1) in [this line](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/ops/dcn/deform_pool.py#L165) and the error is as followes. What should I do ?

> RuntimeError: cannot reshape tensor of 0 elements into shape [0, -1] because the unspecified dimension size -1 can be any value and is ambiguous

ZwwWayne:
Maybe you should check why the `rois.size(0) =0` first.
You can also add an `if` check at someplace, e.g., `if tensor.size(0) == 0`, you can manually skip the deformable pooling.

hellock:
Thanks for the report. We will fix it to handle the exception.

PR #2099, opened 2020-02-15
fix (dpool): directly return empty if rois's length is 0

ZwwWayne:
Fix https://github.com/open-mmlab/mmdetection/issues/2072


tests/test_dpool.py
```diff
@@ -0,0 +1,120 @@
...
+CommandLine:
+    pytest tests/test_dpool.py
+"""
+import torch
+
+from mmdet.ops.dcn.deform_pool import (DeformRoIPoolingPack,
+                                       ModulatedDeformRoIPoolingPack)
+
+
+def test_dpool_empty_cpu():
```
hellock: CPU inference for DCN is currently not supported, thus this may be unnecessary.

tests/test_dpool.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    CommandLine:
+        xdoctest -m tests/test_dpool.py test_dpool_empty_cpu
```
hellock: No need to use xdoctest to perform unit tests. It is mainly used to test the cases in docstrings.

tests/test_dpool.py
```diff
@@ -0,0 +1,120 @@
...
+    assert data.shape[0] == rois.shape[0]
+
+    data = dpool_trans_layer(feats, rois)
+    assert data.shape[0] == rois.shape[0]
+
+    data = mdpool_trans_layer(feats, rois)
+    assert data.shape[0] == rois.shape[0]
+
+
+def test_dpool_empty_gpu():
```
hellock: Non-empty inputs may also be tested, which is unavailable on travis but available locally.

PR #2101, opened 2020-02-16
Add sampler wrapper to customize sampling process

xvjiarui:


xvjiarui:
The current implementation is not elegant enough. 
We may also consider just making another dataset wrapper like [here](https://github.com/xvjiarui/mmdetection/blob/pre-release-v1.1-lvis/mmdet/datasets/dataset_wrappers.py).

Issue #2017, opened 2020-01-25
UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead


Alibhji:
Hello,
I have installed:
Pytorch 1.4.0
cuda: 10.1

after running:
python3 tools/train.py   ./configs/ssd300_coco.py

Training will be started but with this warning, Is there any Idea to remove it?

2020-01-24 18:31:56,041 - INFO - Start running, host: ali@ali-ubt, work_dir: /home/ali/00_2020/mmdetection/work_dirs/ssd300_coco
2020-01-24 18:31:56,041 - INFO - workflow: [('train', 1)], max: 24 epochs
**/pytorch/aten/src/ATen/native/IndexingUtils.h:20: UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.
/pytorch/aten/src/ATen/native/IndexingUtils.h:20: UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.
/pytorch/aten/src/ATen/native/IndexingUtils.h:20: UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.
/pytorch/aten/src/ATen/native/IndexingUtils.h:20: UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.**

ZwwWayne:
A first guess is [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/anchor/anchor_target.py#L162), for now, you can try to remove the `.type(torch.uint8)` and also set related variables to be bool somewhere.
Thanks for your kind notification, we will also fix this warning in the future.

CMobley7:
I can confirm this issue. I've tried PyTorch versions from 1.4 to 1.2 and still receive the error.

CMobley7:
When this change is implemented, can it be back-ported into 1.0?

ZwwWayne:
This change should be independent, so you can fix it in your environment.

prateek-77:
You can add this in train.py while training (to ignore the warning):
```
import warnings
warnings.filterwarnings("ignore", category=UserWarning)
```

EvenStrangest:
The below task also produces this error.

```
root@1bac3eaaf785:/mmdetection# python tools/train.py configs/htc/htc_x101_64x4d_fpn_20e_16gpu.py
[...]

PR #2105, opened 2020-02-16
Uint8 fix

ZwwWayne:
Fix https://github.com/open-mmlab/mmdetection/issues/2017
In PyTorch 1.4, using tensor of type `torch.uint8` will cause a massive warning.
To avoid that, the `inside_flags` is converted to `bool` type. However, ***`Tensor.bool()` is not supported in the PyTorch 1.1, the `and` operation is also invalid for bool tensor, as well as `.any()`***.
Therefore, we use `tensor.type(torch.bool)` here.
The modification has been tested with two environments: PyTorch 1.1 & PyTorch 1.4.

### DIFF
--- a/mmdet/core/anchor/anchor_target.py
+++ b/mmdet/core/anchor/anchor_target.py
@@ -165,13 +165,13 @@ def anchor_inside_flags(flat_anchors,
                         allowed_border=0):
     img_h, img_w = img_shape[:2]
     if allowed_border >= 0:
-        inside_flags = valid_flags & \
-            (flat_anchors[:, 0] >= -allowed_border).type(torch.uint8) & \
-            (flat_anchors[:, 1] >= -allowed_border).type(torch.uint8) & \
-            (flat_anchors[:, 2] < img_w + allowed_border).type(torch.uint8) & \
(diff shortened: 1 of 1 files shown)


wduo:
In PYTORCH="1.3" ARG CUDA="10.1" ARG CUDNN="7", there are still massive warnings here.
How can I fix it?

wduo:
@ZwwWayne

ZwwWayne:
In PYTORCH="1.4" there is no warning, I will also try your environments to see what happened.

iamRyanChia:
In pytorch 1.4, training with config "cascade_rcnn_r101_fpn_1x.py" still exist this warnings.

EvenStrangest:
The below task also produces this error.

```
root@1bac3eaaf785:/mmdetection# python tools/train.py configs/htc/htc_x101_64x4d_fpn_20e_16gpu.py
2020-03-10 14:49:23,916 - mmdet - INFO - Environment info:
------------------------------------------------------------
TorchVision: 0.4.1a0+d94043a
OpenCV: 4.2.0
MMCV: 0.3.2
MMDetection: 1.1.0+e08a1ee
MMDetection Compiler: GCC 5.4
MMDetection CUDA Compiler: 10.1
```

PR #2128, opened 2020-02-21
Fix issue with list of metrics in CustomDataset.evaluate

melikovk:
Changes to be committed:
modified:   mmdet/datasets/custom.py

### DIFF
--- a/mmdet/datasets/custom.py
+++ b/mmdet/datasets/custom.py
@@ -178,12 +178,14 @@ class CustomDataset(Dataset):
             scale_ranges (list[tuple] | None): Scale ranges for evaluating mAP.
                 Default: None.
         """
+        metrics = metric if isinstance(metric, list) else [metric]
         allowed_metrics = ['mAP', 'recall']
-        if metric not in allowed_metrics:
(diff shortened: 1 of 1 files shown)


mmdet/datasets/custom.py
```diff
@@ -178,12 +178,14 @@ def evaluate(self,
             scale_ranges (list[tuple] | None): Scale ranges for evaluating mAP.
                 Default: None.
         """
+        metrics = metric if isinstance(metric, list) else [metric]
```
hellock: Thanks for the fix! Since "mAP" and "recall" cannot be evaluated at the same time, we may simply cast metric to a str like in [voc dataset](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/voc.py#L30-L32).
```python
if not isinstance(metric, str):
    assert len(metric) == 1
    metric = metric[0]
```

melikovk: Thank you! I did not immediately realized that recall and mAP accept different parameters. I will make changes and resubmit.

PR #2134, opened 2020-02-21
Add SSDLite

Mxbonn:
Added an option to use depthwise separable convolutions in the SSDHEAD, which results in what is called **SSDLite** [MobileNetV2: Inverted Residuals and Linear Bottlenecks](https://arxiv.org/abs/1801.04381).

Currently running the added config, but so far (10/24 epochs in) the loss is nearly identical to the baseline.

I plan on implementing the full mobilenetv2-ssdlite, and maybe do some refactoring of the SSDHEAD module to easily support more backbones (e.g. get rid of the hard coded coco and voc 300/512 parameters), but that's for a future pull-request.

### DIFF
--- a/mmdet/models/utils/depthwise_separable_conv.py
+++ b/mmdet/models/utils/depthwise_separable_conv.py
@@ -1,0 +1,40 @@
+import torch.nn as nn
+
+
+class DepthwiseSeparableConv2d(nn.Module):
+
+    def __init__(self,
+                 in_channels,
+                 out_channels,
+                 kernel_size,
+                 stride=1,
+                 padding=0,
+                 dilation=1,
+                 bias=True,
+                 padding_mode='zeros'):
+        super(DepthwiseSeparableConv2d, self).__init__()
+        self.depthwise_conv = nn.Conv2d(
+            in_channels,
+            in_channels,
+            kernel_size,
+            stride=stride,
+            padding=padding,
+            dilation=dilation,
+            groups=in_channels,
+            bias=bias,
+            padding_mode=padding_mode)
+        self.pointwise_conv = nn.Conv2d(
+            in_channels,
+            out_channels,
+            1,
+            stride=stride,
+            padding=0,
+            dilation=1,
+            groups=1,
+            bias=bias,
+            padding_mode=padding_mode)
+
+    def forward(self, x):
+        x = self.depthwise_conv(x)
+        x = self.pointwise_conv(x)
+        return x
--- a/configs/ssdlite_vgg16_300_coco.py
+++ b/configs/ssdlite_vgg16_300_coco.py
@@ -23,7 +23,7 @@ model = dict(
         anchor_ratios=([2], [2, 3], [2, 3], [2, 3], [2], [2]),
         target_means=(.0, .0, .0, .0),
         target_stds=(0.1, 0.1, 0.2, 0.2),
-        use_depthwise=False))
+        use_depthwise=True))
 cudnn_benchmark = True
 train_cfg = dict(
     assigner=dict(
@@ -130,7 +130,7 @@ log_config = dict(
 total_epochs = 24
 dist_params = dict(backend='nccl')
 log_level = 'INFO'
-work_dir = './work_dirs/ssd300_coco'
+work_dir = './work_dirs/ssdlite_vgg16_300_coco'
 load_from = None
 resume_from = None
 workflow = [('train', 1)]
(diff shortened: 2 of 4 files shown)


Mxbonn:
Training log: [20200221_153314.log](https://github.com/open-mmlab/mmdetection/files/4242359/20200221_153314.log).


---
Final evalution is : 
```
python tools/test.py configs/ssdlite_vgg16_300_coco.py work_dirs/ssdlite_vgg16_300_coco/latest.pth --eval bbox
Done (t=0.62s)
creating index...
index created!
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 5000/5000, 27.8 task/s, elapsed: 180s, ETA:     0s
Evaluating bbox...
Loading and preparing results...
DONE (t=6.26s)
creating index...
index created!
Running per image evaluation...
Evaluate annotation type *bbox*
DONE (t=74.84s).
Accumulating evaluation results...
DONE (t=10.76s).
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.247
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.416
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.254
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.062
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.267
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.402
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.233
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.342
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.363
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.109
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.404
```
---

xvjiarui:
Hi @Mxbonn 
Thanks for your PR and I apologize for my late review. 
We had a refactor about `models` and `ops` recently. Could you please merge the mater to resolve conflicts? 
You may also add 1 row in the [MODEL_ZOO](docs/MODEL_ZOO.md) on the performance of SSDLite under SSD section.

configs/ssd300_coco.py
```diff
@@ -22,7 +22,8 @@
         basesize_ratio_range=(0.15, 0.9),
         anchor_ratios=([2], [2, 3], [2, 3], [2, 3], [2], [2]),
         target_means=(.0, .0, .0, .0),
-        target_stds=(0.1, 0.1, 0.2, 0.2)))
+        target_stds=(0.1, 0.1, 0.2, 0.2),
+        use_depthwise=False))
```
xvjiarui: We may keep this config unchanged since `use_depthwise=True` by default.

mmdet/models/utils/depthwise_separable_conv.py
```diff
@@ -0,0 +1,40 @@
+import torch.nn as nn
+
+
+class DepthwiseSeparableConv2d(nn.Module):
```
xvjiarui: It could be added into `conv_cfg` in [conv.py](https://open-mmlab/mmdetection/mmdet/ops/conv.py). So we SSDLite head could be built by `conv_cfg`.

xvjiarui: We may add it into `conv_cfg` in [conv.py](https://github.com/open-mmlab/mmdetection/tree/master/mmdet/ops/conv.py). So we could use build_conv_layer to construct model.

Mxbonn:
@xvjiarui I have since this pull request implemented MobileNetV2 SSDLite using mmdetection.
However this included a major rewrite of the SSDHead.
Also I have really been struggling with getting it to train properly.
Currently I have achieved an mAP of 18.5% which is similar to https://github.com/ShuangXieIrene/ssds.pytorch, which is as far as I know the reported opensource PyTorch implementation of MobileNetV2 SSDLite.
However it is still training and could improve.
The thing is that I am using a custom optimizer (I use RMSProp but the epsilon parameter is defined differently in PyTorch than in Tensorflow and Im trying to stay as close as possible to the original TF implementation so I'm using a port of the RMSProp optimizer with the same behavior as in TF).

How do you want me to proceed to get MobileNetV2SSD(Lite) in mmdetection?

xvjiarui:
@Mxbonn Thanks for getting back to us. 
I suggest you improve the performance first. MobileNetV2 would be supported in the near future. Your PR is very welcomed by then.

PR #2135, opened 2020-02-21
Fix IOU assigners when ignore_of_thr > 0 and no pred boxes

Erotemic:
When ignore_of_thr is enabled and there are no predicted bboxes, the `.max` function fails because the `ignore_overlaps` tensor is empty. The error is as follows:

```
~/code/mmdetection/mmdet/core/bbox/assigners/max_iou_assigner.py in assign(self, bboxes, gt_bboxes, gt_bboxes_ignore, gt_labels)
    104                 ignore_overlaps = bbox_overlaps(
    105                     bboxes, gt_bboxes_ignore, mode='iof')
--> 106                 ignore_max_overlaps, _ = ignore_overlaps.max(dim=1)
    107             else:
    108                 ignore_overlaps = bbox_overlaps(

RuntimeError: cannot perform reduction function max on tensor with no elements because the operation does not have an identity
```

This can be fixed by simply checking that bboxes isn't empty before running the code. I've added this fix for both approx and max iou assigners. I've also added a test, which currently passes, but will fail if the patch is removed.

### DIFF
--- a/mmdet/core/bbox/assigners/approx_max_iou_assigner.py
+++ b/mmdet/core/bbox/assigners/approx_max_iou_assigner.py
@@ -118,8 +118,8 @@ class ApproxMaxIoUAssigner(MaxIoUAssigner):
 
         bboxes = squares[:, :4]
 
-        if (self.ignore_iof_thr > 0) and (gt_bboxes_ignore is not None) and (
-                gt_bboxes_ignore.numel() > 0 and bboxes.numel() > 0):
+        if (self.ignore_iof_thr > 0 and gt_bboxes_ignore is not None
+                and gt_bboxes_ignore.numel() > 0 and bboxes.numel() > 0):
             if self.ignore_wrt_candidates:
                 ignore_overlaps = bbox_overlaps(
                     bboxes, gt_bboxes_ignore, mode='iof')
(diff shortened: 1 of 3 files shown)


mmdet/core/bbox/assigners/approx_max_iou_assigner.py
```diff
@@ -119,7 +119,7 @@ def assign(self,
         bboxes = squares[:, :4]
 
         if (self.ignore_iof_thr > 0) and (gt_bboxes_ignore is not None) and (
-                gt_bboxes_ignore.numel() > 0):
+                gt_bboxes_ignore.numel() > 0 and bboxes.numel() > 0):
```
hellock: Thanks for the fix! Since there are lots of conditions, we can just use one pair of brackets to hold them.
```python
if (self.ignore_iof_thr > 0 and gt_bboxes_ignore is not None
    and gt_bboxes_ignore.numel() > 0 and bboxes.numel() > 0):
    pass
```

hellock:
BTW, it seems that the latest ubelt (0.8.9) is broken. An import error will be raised for `import ubelt`.

Erotemic:
Yes, ubelt 0.8.9 was broken. I had incorrectly used find_packages in the setup.py file. I just release 0.9.0, which (I believe) fixes the issue. I've also removed the extra parens.

Issue #2041, opened 2020-02-01
How to use the custom optimizer?


AlanJie:
I want to customize the optimizer(RAdam etc.), but I find that the optimizer in mmdetection is imported from `torch.optim`(only a few of optimizer are implemented),  can you provide an official example to do this? Thank you

hellock:
We will introduce an optimizer registry like dataset or model registry to support customizing the optimizer. As a walkaround now, you may just modify [this line](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/train.py#L190) and get your own optimizer class.

AlanJie:
OK, I'll try, thanks

hiyyg:
It seems the custom optimizer documentation is out of date. How to use custom optimizer after the cleaning of optimizer code?

hiyyg:
I figured out using this way:
```
from mmcv.runner.optimizer import build_optimizer, OPTIMIZERS
from my_optimizer import MyOptimizer
OPTIMIZERS.register_module()(MyOptimizer)

optim_cfg = dict(type="MyOptimizer", lr=0.001)
optimizer = build_optimizer(model, optim_cfg)
```

PR #2139, opened 2020-02-22
add optimizer registry

xvjiarui:
This PR added support for both PyTorch native optimizers and customize optimizers. 
Fix https://github.com/open-mmlab/mmdetection/issues/2041


mmdet/optimizer/registry.py
```diff
@@ -0,0 +1,3 @@
+from mmdet.utils import Registry
+
+OPTIMIZERS = Registry('optimizer')
```
hellock: We may register all built-in optimizers of PyTorch here to simplify the builder.

mmdet/optimizer/mmsgd.py
```diff
@@ -0,0 +1,84 @@
+import torch
+from torch.optim.optimizer import Optimizer, required
+
+from .registry import OPTIMIZERS
+
+
+@OPTIMIZERS.register_module
+class MMSGD(Optimizer):
```
hellock: It can be just named `CopyOfSGD` and derived from `torch.optim.SGD`. We may also document how to use custom optimizers.

mmdet/core/optimizer/registry.py
```diff
@@ -0,0 +1,22 @@
...
+OPTIMIZERS = Registry('optimizer')
+
+
+def register_torch_optimizers():
+    torch_optimizers = []
+    for module_name in dir(torch.optim):
+        if module_name.startswith('__'):
+            continue
+        _optim = getattr(torch.optim, module_name)
+        if inspect.isclass(_optim):
```
hellock: May add a condition `issubclass(_optim, torch.optim.Optimizer)`

mmdet/core/optimizer/__init__.py
```diff
@@ -0,0 +1,6 @@
+from torch.optim import *  # noqa: F401, F403
+
+from .copy_of_sgd import CopyOfSGD
+from .registry import OPTIMIZERS, TORCH_OPTIMIZERS
```
hellock: It may not be necessary to expose `TORCH_OPTIMIZERS`.

PR #2149, opened 2020-02-24
Fix a documentation error in GETTING_STARTED.md

seanremy:
Need to specify --format-only rather than --out in order to actually get .json output


hellock:
Thanks for the fix!

Issue #2131, opened 2020-02-21
speedup fp16


rydenisbak:
You do not need this line.
https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/fp16/hooks.py#L100

because if input is half tensor, weight is float tensor, output anyway is half tesnor.
But if I comment this line I have speed up near 20% on 1060 and near 50% on 2080.

Check it out please.

rydenisbak:
I checked it on single stage retina-like detector. All worked right.

yhcao6:
Thanks for pointing out that! 
In early versions of pytorch, such as 1.0, 1.1, it is necessary to explicitly convert the half input to float. Otherwise, it will report an error. The error disappears when upgrading pytorch to 1.3.

Would you like to create a pr? We need to compare the version of pytorch to decide if we need to comment this line. I can help you benchmark.

clw5180:
> Thanks for pointing out that!
> In early versions of pytorch, such as 1.0, 1.1, it is necessary to explicitly convert the half input to float. Otherwise, it will report an error. The error disappears when upgrading pytorch to 1.3.
> 
> Would you like to create a pr? We need to compare the version of pytorch to decide if we need to comment this line. I can help you benchmark.

I try to comment '`module.float()`' in Pytorch 1.3, but also got error '**Expected tensor for argument #1 'input' to have the same type as tensor for argument #2 'weight'; but type torch.cuda.FloatTensor does not equal torch.cuda.HalfTensor (while checking arguments for cudnn_batch_norm)**'  how to deal with that ? @yhcao6

rydenisbak:
> Thanks for pointing out that!
> In early versions of pytorch, such as 1.0, 1.1, it is necessary to explicitly convert the half input to float. Otherwise, it will report an error. The error disappears when upgrading pytorch to 1.3.
> 
> Would you like to create a pr? We need to compare the version of pytorch to decide if we need to comment this line. I can help you benchmark.

strange, it seemed to me that I tried this on pytorch 1.0 and 1.1 and 1.2 in other my project. I can check it today in several hours

clw5180:
> Thanks for pointing out that!
> In early versions of pytorch, such as 1.0, 1.1, it is necessary to explicitly convert the half input to float. Otherwise, it will report an error. The error disappears when upgrading pytorch to 1.3.
> 
> Would you like to create a pr? We need to compare the version of pytorch to decide if we need to comment this line. I can help you benchmark.

[...]

PR #2151, opened 2020-02-25
fp16_speedup

rydenisbak:
PR from this issue
https://github.com/open-mmlab/mmdetection/issues/2131

### DIFF
--- a/mmdet/core/fp16/hooks.py
+++ b/mmdet/core/fp16/hooks.py
@@ -97,7 +97,7 @@ def wrap_fp16_model(model):
 def patch_norm_fp32(module):
     if isinstance(module, (nn.modules.batchnorm._BatchNorm, nn.GroupNorm)):
         module.float()
-        torch_version = [int(v) for v in torch.__version__.split('.')]        
+        torch_version = [int(v) for v in torch.__version__.split('.')]
         if isinstance(module, nn.GroupNorm) or torch_version < [1, 3]:
             module.forward = patch_forward_method(module.forward, torch.half,
                                                   torch.float)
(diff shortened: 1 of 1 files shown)


mmdet/core/fp16/hooks.py
```diff
@@ -97,8 +97,12 @@ def wrap_fp16_model(model):
 def patch_norm_fp32(module):
     if isinstance(module, (nn.modules.batchnorm._BatchNorm, nn.GroupNorm)):
         module.float()
-        module.forward = patch_forward_method(module.forward, torch.half,
-                                              torch.float)
+        path_forward_condition = isinstance(module, nn.GroupNorm)
+        major, minor = [int(v) for v in torch.__version__.split('.')][:2]
+        path_forward_condition |= major < 1 or major == 1 and minor < 3
+        if path_forward_condition:
```
hellock: The condition can be rewritten in just one line:
```python
if torch.__version__ < (1, 3) or isinstance(module, nn.GroupNorm):
    module.forward = patch_forward_method(module.forward, torch.half, torch.float)
```

rydenisbak: image
but I can write
image
It looks more pretty, thanks

hellock:
Thanks! Could you fix the linting error?

rydenisbak:
@hellock I found some bug in the code.
My torch version is '1.3.1+cu100' and it cause problem in line
https://github.com/rydenisbak/mmdetection/blob/fp16_speedup/mmdet/core/fp16/hooks.py#L100
```python
torch_version = [int(v) for v in torch.__version__.split('.')]
```
We need to change it, for example like this
```python
torch_version = [int(v) for v in torch.__version__.split('.') if v.isdigit()]
```
or like this
```python
torch_version = [int(v) if v.isdigit() else int(v.split('+')[0]) for v in torch.__version__.split('.') if v.isdigit()]
```
or like something else

hellock:
Oh the pytorch versions installed by pip and conda are different. Maybe simply
```python
torch_version = [int(v) for v in torch.__version__.split('.')[:2]]
```

rydenisbak:
ok, good

Issue #2143, opened 2020-02-24
GIoULoss cannot be used directly


iamRyanChia:
I found GIoULoss has been added to the iou_loss.py, but detectors actually output the offsets of the bboxes, so it cannot be used directly. 

Any plans to make GIoULoss available? Thank you!

xvjiarui:
Fixed by #2156

Issue #1848, opened 2019-12-21
giou


ZYN-guy:
I add some codes in iou_loss.py
### ### ### ### ### ### ### ### ### ### ### 
import torch
import torch.nn as nn

from mmdet.core import bbox_overlaps
from ..registry import LOSSES
from .utils import weighted_loss

@weighted_loss
def iou_loss(pred, target, eps=1e-6):
    """IoU loss.

    Computing the IoU loss between a set of predicted bboxes and target bboxes.
    The loss is calculated as negative log of IoU.

    Args:
        pred (Tensor): Predicted bboxes of format (x1, y1, x2, y2),
            shape (n, 4).
        target (Tensor): Corresponding gt bboxes, shape (n, 4).
        eps (float): Eps to avoid log(0).

    Return:
        Tensor: Loss tensor.
    """
    ious = bbox_overlaps(pred, target, is_aligned=True).clamp(min=eps)
    loss = -ious.log()
    return loss

@weighted_loss
def bounded_iou_loss(pred, target, beta=0.2, eps=1e-3):
    """Improving Object Localization with Fitness NMS and Bounded IoU Loss,
    https://arxiv.org/abs/1711.00164.

    Args:
        pred (tensor): Predicted bboxes.
        target (tensor): Target bboxes.
        beta (float): beta parameter in smoothl1.
        eps (float): eps to avoid NaN.
    """
    pred_ctrx = (pred[:, 0] + pred[:, 2]) * 0.5
    pred_ctry = (pred[:, 1] + pred[:, 3]) * 0.5
    pred_w = pred[:, 2] - pred[:, 0] + 1
    pred_h = pred[:, 3] - pred[:, 1] + 1
    with torch.no_grad():
        target_ctrx = (target[:, 0] + target[:, 2]) * 0.5
        target_ctry = (target[:, 1] + target[:, 3]) * 0.5
        target_w = target[:, 2] - target[:, 0] + 1
        target_h = target[:, 3] - target[:, 1] + 1

    dx = target_ctrx - pred_ctrx
    dy = target_ctry - pred_ctry

    loss_dx = 1 - torch.max(
        (target_w - 2 * dx.abs()) /
        (target_w + 2 * dx.abs() + eps), torch.zeros_like(dx))
    loss_dy = 1 - torch.max(
        (target_h - 2 * dy.abs()) /
        (target_h + 2 * dy.abs() + eps), torch.zeros_like(dy))
[...]

PR #2156, opened 2020-02-25
Unify regression loss

xvjiarui:
This PR unified the regression loss in `bbox_head`. 
Attribute `reg_delta_target = True` indicates the regression target is the delta between prediction and target, otherwise regression target is the ground truth bounding box.
Fixed https://github.com/open-mmlab/mmdetection/issues/2143 so that `IoULoss` could be applied directly in `bbox_head`.

### DIFF
--- a/mmdet/models/anchor_heads/anchor_head.py
+++ b/mmdet/models/anchor_heads/anchor_head.py
@@ -309,7 +309,7 @@ class AnchorHead(nn.Module):
 
         # anchor number of multi levels
         num_level_anchors = [anchors.size(0) for anchors in anchor_list[0]]
-        # concat all level anchors and flags to a single tensor
+        # concat all level anchors to a single tensor
         concat_anchor_list = []
         concat_valid_flag_list = []
         for i in range(num_imgs):
--- a/mmdet/models/anchor_heads/anchor_head.py
+++ b/mmdet/models/anchor_heads/anchor_head.py
@@ -48,6 +48,7 @@ class AnchorHead(nn.Module):
                      type='DeltaXYWHBBoxCoder',
                      target_means=(.0, .0, .0, .0),
                      target_stds=(1.0, 1.0, 1.0, 1.0)),
+                 reg_decoded_bbox=False,
                  background_label=None,
                  loss_cls=dict(
                      type='CrossEntropyLoss',
@@ -77,6 +78,7 @@ class AnchorHead(nn.Module):
 
         if self.cls_out_channels <= 0:
             raise ValueError('num_classes={} is too small'.format(num_classes))
+        self.reg_decoded_bbox = reg_decoded_bbox
 
         self.background_label = (
             num_classes if background_label is None else background_label)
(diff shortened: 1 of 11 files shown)


yhcao6:
Currently only supports two-stage model, may also need to support single-stage models.

mmdet/models/bbox_heads/bbox_head.py
```diff
@@ -95,6 +97,25 @@ def get_target(self, sampling_results, gt_bboxes, gt_labels,
             target_stds=self.target_stds)
         return cls_reg_targets
 
+    def get_reg(self, rois, bbox_pred, bbox_targets, img_metas):
+        labels, label_weights, bbox_targets, bbox_weights = bbox_targets
+        if not self.reg_delta_target:
```
hellock: `roi2bbox` may help.

hellock:
#1848

xvjiarui:
Support vanilla IoU type loss in `AnchorHead` is not that elegant.....

melikovk:
Hi @xvjiarui 

It looks great. If I understand correctly, the idea is to set reg_decoded_bbox flag for IOU losses and then to use inverse transform of predictions to calculate loss. One potential drawback of this approach vs my suggestion in #2417 is that you will be back-propagating the inverse transform.

xvjiarui:
> Hi @xvjiarui
> 
> It looks great. If I understand correctly, the idea is to set reg_decoded_bbox flag for IOU losses and then to use inverse transform of predictions to calculate loss. One potential drawback of this approach vs my suggestion in #2417 is that you will be back-propagating the inverse transform.

Yes. It would be BP the inverse transform. For faster R-CNN style method, the outputs of bbox head are always regarded as offset w.r.t to rois. If you would like to regress bbox coordinate directly, a new head should be implemented.

mmdet/models/anchor_heads/anchor_head.py
```diff
@@ -48,6 +48,7 @@ def __init__(self,
                      type='DeltaXYWHBBoxCoder',
                      target_means=(.0, .0, .0, .0),
                      target_stds=(1.0, 1.0, 1.0, 1.0)),
+                 reg_decoded_bbox=False,
```
hellock, line 51: The docstring is outdated.

melikovk:
@xvjiarui 
I am probably not clearly expressing myself. I am not suggesting to regress to bbox coordinates directly. I am suggesting that IOU is scale invariant and can be used directly with R-CNN outputs w.r.t. to rois with one caveat, that you should not do log transform of w and h, i.e. in bbox2delta instead of doing
```python 
dw = torch.log(gw / pw)
dh = torch.log(gh / ph)
```
you should have an option to do
```python
dw = gw / pw
dh = gh / ph
```
The log transform is unnecessary for IOU type losses (it is meant to control relative contribution from large objects when using L2 or L1 losses). If you have this as an option you will not need to BP inverse transform. It is possible though that the time and memory impact is minor. Thank you for the great work you are doing.

xvjiarui:
Hi @melikovk 
My bad. It is a good idea, which could be easily supported by adding a new `BBoxCoder`. 
This PR is just merged. You are very welcomed to try it out.

melikovk:
Hi @xvjiarui 
This is great. I will test it on my dataset as I am curious to see if IOU or GIOU loss will improve performance. I will give a try to my approach and will let you know if it has any impact. Again, thank you for your great work.

PR #2180, opened 2020-02-29
add affine_grid

xvjiarui:
This PR adds affine_grid as an extension operator.
Specifically, provides affine_grid(align_corners=False) when PyTorch version < 1.3.


mmdet/ops/__init__.py
```diff
@@ -1,8 +1,10 @@
+from .affine_grid_generator import affine_grid
```
hellock: `affine_grid` and `grid_sample` are currently unused. We may remove it from `ops/__init__.py` to speedup the loading of mmdet.

PR #2196, opened 2020-03-04
Fix bug of simple_test with proposals

WangChuanYuan:
fix the bug of simple_test with proposals
remove the redundant codes in fast_rcnn


mmdet/models/detectors/fast_rcnn.py
```diff
@@ -27,35 +27,3 @@ def __init__(self,
             mask_roi_extractor=mask_roi_extractor,
             mask_head=mask_head,
             pretrained=pretrained)
-
```
yhcao6: No need to delete these.

mmdet/models/detectors/base.py
```diff
@@ -119,6 +119,12 @@ def forward_test(self, imgs, img_metas, **kwargs):
         assert imgs_per_gpu == 1
 
         if num_augs == 1:
+            """
+            proposals (List[List[Tensor | None]]): predefiend proposals for
```
yhcao6: The comment seems wrong. Type of `proposals` should be `List[Tensor]`.

mmdet/models/detectors/base.py
```diff
@@ -119,6 +119,12 @@ def forward_test(self, imgs, img_metas, **kwargs):
...
+                each test-time augmentation and each item.
+            """
+            if 'proposals' in kwargs:
+                kwargs['proposals'] = kwargs['proposals'][0]
```
yhcao6: When using RPN to produce proposals, the type of `proposals` is `List[Tensor]`.
When using predifined proposals, `proposals[0]` is of type `List[Tensor]`, so `proposals[0]` becomes `Tensor`.
Though the program works fine in this case, it may be better to keep the type of `proposals` consistent in two cases.

WangChuanYuan: Thanks for your reply. In my opinion, the type of proposals produced by RPN is List[Tensor], but when passed to the pipeline MultiScaleFlipAug, it is wrapped as List[List[Tensor]]. The outer list is for the TTA, and the inner is for the batch size.

yhcao6: I print the type of `kwargs['proposals']` using the following code:
```
print(type(kwargs['proposals']))
print(type(kwargs['proposals'][0]))
```
Whether or not use TTA, the outputs are the same as follow:
```
<class 'list'>
<class 'torch.Tensor'>
```
Could you reconfirm that?

PR #2213, opened 2020-03-07
Fixed wrong attribute name

whikwon:


mmdet/datasets/pipelines/transforms.py
```diff
@@ -872,5 +872,5 @@ def __call__(self, results):
 
     def __repr__(self):
         repr_str = self.__class__.__name__
-        repr_str += '(transformations={})'.format(self.transformations)
+        repr_str += '(transformations={})'.format(self.transforms)
```
hellock: Thanks! You may also change the key here from `transformations` to `transforms`.

whikwon: Thanks for the comment. Updated PR.

PR #2216, opened 2020-03-07
add new config sample and check scripts

xvjiarui:
This PR introduces new config system wiht bases. 
Check script will be deleted before merge. 
To test configs in this PR, please refer use mmcv from here https://github.com/open-mmlab/mmcv/pull/194

During refactor, the following principles are applied. 
* The base model (e.g. r50) of each method is composed of components. 
* The variations of base models are inherited from the closest base model.  
* For the inherited configs, only the updated key-value pairs are included. 
* `_delete_` indicates the original dict will be **deleted** instead of **updated**, which is useful when two dicts have non-overlapped keys. 
 
Notes:
`faster_rcnn_r50_caffe_fpn_1x.py`, `guided_anchoring/ga_retinanet_r50_fpn_1x.py`, `guided_anchoring/ga_faster_r50_fpn_1x.py`, `retinanet_r50_caffe_fpn_1x.py`, `mask_rcnn_r50_caffe_fpn_1x.py`, `ms_rcnn/ms_rcnn_r50_fpn_1x.py`, `rpn_r50_caffe_fpn_1x.py`, `guided_anchoring/ga_rpn_r50_fpn_1x.py` are not in the current configs. They are added for the convenience of inheritance.


tools/config_check.py
```diff
@@ -0,0 +1,80 @@
...
+    def __enter__(self):
+        self.before_modules = sys.modules.copy()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.after_modules = sys.modules.copy()
+
+        for m in list(
+                set(self.after_modules.keys()) -
+                set(self.before_modules.keys())):
+            del sys.modules[m]
```
hellock: Maybe we should delete the imported module in `mmcv.Config`?

xvjiarui:
Please refer to the updated PR first comment.

PR #2217, opened 2020-03-08
fix bug of no attribute `contrast_range` in __repr__ of PhotoMetricDistortion

zlmzju:


mmdet/datasets/pipelines/transforms.py
```diff
@@ -526,7 +526,7 @@ def __repr__(self):
         repr_str = self.__class__.__name__
         repr_str += ('(brightness_delta={}, contrast_range={}, '
                      'saturation_range={}, hue_delta={})').format(
-                         self.brightness_delta, self.contrast_range,
+                         self.brightness_delta, (self.contrast_lower, self.contrast_upper),
```
yhcao6: Thanks for fixing the bug. Could you fix the flake8 error.
Here is the error info:
```
./mmdet/datasets/pipelines/transforms.py:529:80: E501 line too long (91 > 79 characters)
The command "flake8 ." failed and exited with 1 during .
```

zlmzju: Done!

Issue #2153, opened 2020-02-25
RuntimeError:Expected to have finished reduction in the prior iteration before starting a new one


vincentwei0919:
Thanks for your error report and we appreciate it a lot.

**Checklist**
1. I have searched related issues but cannot get the expected help.
2. The bug has not been fixed in the latest version.

**Describe the bug**

**Reproduction**
1. What command or script did you run?
I have change the config name from faster_rcnn_r50_fpn_1x.py to element.py
```
CUDA_VISIBLE_DEVICES=1,2,3 ./tools/dist_train.sh configs/element.py 3 --autoscale-lr
```
2. Did you make any modifications on the code or config? Did you understand what you have modified?
only num_classes and work_dir in config

3. What dataset did you use?
my own dataset which is made the same as VOC format
**Environment**
![image](https://user-images.githubusercontent.com/24283259/75251117-eb63d500-5814-11ea-9bbc-361989af994e.png)

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment infomation and paste it here.
2. You may add addition that may be helpful for locating the problem, such as
    - How you installed PyTorch [e.g., pip, conda, source]
    - Other environment variables that may be related (such as `$PATH`, `$LD_LIBRARY_PATH`, `$PYTHONPATH`, etc.)

**Error traceback**

```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing its output (the return value of `forward`). You can enable unused parameter detection by passing the keyword argument `find_unused_parameters=True` to `torch.nn.parallel.DistributedDataParallel`. If you already have this argument set, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's `forward` function. Please include the structure of the return value of `forward` of your module when reporting this issue (e.g. list, dict, iterable). (prepare_for_backward at /pytorch/torch/csrc/distributed/c10d/reducer.cpp:408)
frame #0: std::function<std::string ()>::operator()() const + 0x11 (0x7f92f4501441 in /usr/local/lib/python3.6/dist-packages/torch/lib/libc10.so)
frame #1: c10::Error::Error(c10::SourceLocation, std::string const&) + 0x2a (0x7f92f4500d7a in /usr/local/lib/python3.6/dist-packages/torch/lib/libc10.so)
[...]

PR #2225, opened 2020-03-10
Allow find_unused_parameters to be set in config

Mxbonn:
Allow `find_unused_parameters` to be set in the configuration file. If not present it defaults to the pytorch default of `False`.

This comes in handy when the underlying reason for the unused parameters is hard to find or not resolved well by pytorch. Current way to overcome this is to change `apis/train.py` in a user maintained fork which is not ideal.

This would also be useful for users affected by https://github.com/open-mmlab/mmdetection/issues/2153

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -129,7 +129,7 @@ def _dist_train(model,
             seed=cfg.seed) for ds in dataset
     ]
     # put model on gpus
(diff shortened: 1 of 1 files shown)


hellock:
We are planning a solution to solve unused parameters. As a walk around, we will merge it first.

Mxbonn:
Thanks! Feel free to tag me in any future issues or WIP pull requests related to a better solution (or if you want help).

PR #2226, opened 2020-03-10
Update default work_dir

mengfu188:
Update default work_dir config to `__file__.replace('configs', 'work_dirs').rstrip('.py')`.


configs/albu_example/mask_rcnn_r50_fpn_1x.py
```diff
@@ -242,7 +242,7 @@
 total_epochs = 12
 dist_params = dict(backend='nccl')
 log_level = 'INFO'
-work_dir = './work_dirs/mask_rcnn_r50_fpn_1x'
+work_dir = __file__.replace('configs', 'work_dirs').rstrip('.py')
```
hellock: `rstrip('.py')` is risky. If the config name ends with `p` or `y` the results are unexpected. `.split('.')[0]` would be more safe.

BTW, we also support json and yaml format as config files.

mengfu188: Thanks for pointing out.

mengfu188: But it's not safe when my config path is `C:/Users/mengf/code/mmdetection.cmf/configs/albu_example/mask_rcnn_r50_fpn_1x.py`.
It will take `C:/Users/mengf/code/mmdetection`


I think it's better `'.'.join(__file__.replace('configs', 'work_dirs').split('.')[:-1])`

hellock: Nice.

PR #2232, opened 2020-03-11
Update default work_dir config

mengfu188:
Update default work_dir config to `'.'.join(__file__.replace('configs', 'work_dirs').split('.')[:-1])`


xvjiarui:
Hi @mengfu188 , thanks for the PR.
It would be better to delete the field of `work_dir` in all config files, which will reduce the redundancy.  And set your implementation as the default value for `cfg.work_dir`. If no `work_dir` is passed in from argument parser, the default value should be used.

mengfu188:
Nice

PR #2233, opened 2020-03-11
Config fix

xvjiarui:
This PR added `norm_cfg` for ResNet PyTorch style models. 
It also deleted redundant fields like `device_id` in FreeAnchor, `auto_resume` in RepPoints,  `evaluation` in CARAFE, `pretrained` in DoubleHeads.

### DIFF
--- a/configs/albu_example/mask_rcnn_r50_fpn_1x.py
+++ b/configs/albu_example/mask_rcnn_r50_fpn_1x.py
@@ -8,6 +8,7 @@ model = dict(
         num_stages=4,
         out_indices=(0, 1, 2, 3),
         frozen_stages=1,
+        norm_cfg=dict(type='BN', requires_grad=True),
         style='pytorch'),
     neck=dict(
         type='FPN',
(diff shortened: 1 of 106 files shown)


hachreak:
Hi @xvjiarui,
can I ask why `norm_cfg` is added in all configuration?
I'm using an old version of mmdetection and I'd like to know if it's introduced to maintains the same performance of before (because you call it `Fix`), or now, with BN, there'll be a boost?
Just to understand what I should expect with the update of my master to the last version.
Thanks :smile:

xvjiarui:
> Hi @xvjiarui,
> can I ask why `norm_cfg` is added in all configuration?
> I'm using an old version of mmdetection and I'd like to know if it's introduced to maintains the same performance of before (because you call it `Fix`), or now, with BN, there'll be a boost?
> Just to understand what I should expect with the update of my master to the last version.
> Thanks 😄

Hi @hachreak ,
It's just to keep all config files consistent in fields.  
Relax, it won't affect the performance.

hachreak:
Hi @xvjiarui,
thanks for the super quick reply! :smile: 
It means that in the past the normalization was applied by default?

xvjiarui:
@hachreak 
Yes, we just explicitly write out the default setting for easy understanding.

PR #2239, opened 2020-03-11
add act_cfg

xvjiarui:
This added config for activation so that user could use activation like  `relu` and `leaky_relu`.


mmdet/models/utils/activation.py
```diff
@@ -0,0 +1,33 @@
+import torch.nn as nn
+
+activation_cfg = {
+    # format: layer_type: (abbreviation, module)
+    'relu': nn.ReLU,
+    'leaky_relu': nn.LeakyReLU,
```
hellock: May also add PReLU, RReLU, ReLU6, SELU, CELU, GELU.

mmdet/models/utils/activation.py
```diff
@@ -0,0 +1,33 @@
...
+}
+
+
+def build_activation_layer(cfg):
+    """ Build activation layer
+
+    Args:
+        cfg (dict): cfg should contain:
+            type (str): Identify activation layer type.
```
hellock: Additional indents.

mmdet/models/utils/conv_module.py
```diff
@@ -151,7 +150,7 @@ def norm(self):
         return getattr(self, self.norm_name)
 
     def init_weights(self):
-        nonlinearity = 'relu' if self.activation is None else self.activation
+        nonlinearity = self.act_cfg['type'] if self.with_activation else 'relu'
```
hellock: nonlinearity can only be `relu` or `leaky_relu`.

mmdet/models/utils/activation.py
xvjiarui: `kaiming_init` and `inplace` option only support `relu` and `leaky_relu`.

hellock: Yes, so we need to handle such cases in `init_weights`. Here we list all possibilities for potential usage (not necessarily in mmdet).

mmdet/models/utils/activation.py
xvjiarui: The indents are for the key-value in `cfg`.

mmdet/models/utils/activation.py
xvjiarui: All added except for GELU, which is not supported in pytorch 1.1

mmdet/models/bbox_heads/double_bbox_head.py
```diff
@@ -41,7 +41,7 @@ def __init__(self,
             out_channels,
             kernel_size=1,
             bias=False,
-            activation=None,
+            act_cfg=None,
             conv_cfg=conv_cfg,
             norm_cfg=norm_cfg)
```
hellock: We may follow the argument order: conv_cfg, norm_cfg, act_cfg.

mmdet/models/necks/fpn_carafe.py
```diff
@@ -52,7 +52,6 @@ def __init__(self,
         self.out_channels = out_channels
         self.num_ins = len(in_channels)
         self.num_outs = num_outs
-        self.activation = activation
```
hellock: We may also save act_cfg as a class attribute.

mmdet/models/utils/conv_module.py
```diff
@@ -60,7 +61,7 @@ class ConvModule(nn.Module):
             False.
         conv_cfg (dict): Config dict for convolution layer.
         norm_cfg (dict): Config dict for normalization layer.
-        activation (str or None): Activation type, "ReLU" by default.
+        act_cfg (str or None): Activation type, "ReLU" by default.
```
hellock: It is no longer a str.

mmdet/models/utils/activation.py
```diff
@@ -0,0 +1,38 @@
+import torch.nn as nn
+
+activation_cfg = {
+    # format: layer_type: (abbreviation, module)
```
hellock: This comment is incorrect.

PR #2246, opened 2020-03-12
Add dataset browser

mengfu188:
This PR add a simple dataset browser, which can check some data augmentation.


tools/browse_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+def main():
+    args = parse_args()
+
+    cfg = clear_config(args.config, args.skip_type)
+
+    dataset = build_dataset(cfg.data.train)
+
+    for item in tqdm(dataset):
+        # print(item['img_meta'].data['filename'], item['gt_bboxes'], item['gt_labels'])
+        mmcv.imshow_det_bboxes(item['img'], item['gt_bboxes'], item['gt_labels'], thickness=2)
```
yhcao6: May also pass `dataset.CLASSES` to show class names instead category id.

tools/browse_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+        .replace("dict(type='Normalize', **img_norm_cfg),",
+                 """dict(type='Normalize', **img_norm_cfg),
+                    dict(type='ToUint8'),""")
+    for s in skip_type:
+        c = c.replace(s, 'Pass')
+    print(c)
+    ft = tempfile.NamedTemporaryFile('w+t', suffix='.py', delete=False)
+    print(ft.name)
+    ft.write(c)
+    ft.close()
```
yhcao6: The logic and code of this function seem not so good. 
Renamed this func as `retrieve_data_cfg`.
The logic can be  as follow:
1. Read config by `cfg = Config.fromfile(args.cfg)`
2. Retrieve cfg of `train.data.train` as `train_data_cfg`, which contains `pipeline` field.
3. Delete unused pipeline like `Normalize, DefaultFormatBundle, Collect`.

mengfu188:
Hi @yhcao6

Thanks for your advice, I'll revise it as soon as possible

tools/browse_dataset.py
```diff
@@ -0,0 +1,84 @@
...
+
+    plt.imshow(img)
+    if args.output_dir is None:
+        plt.show(block=False)
+        plt.pause(1)
+    else:
+        mmcv.mkdir_or_exist(args.output_dir)
+        out = os.path.join(args.output_dir, filename)
+        plt.savefig(out)
+        plt.cla()
```
yhcao6: There is a similar function called `imshow_det_bboxes` in mmcv.

tools/browse_dataset.py
```diff
@@ -0,0 +1,84 @@
...
+    cfg = retrieve_data_cfg(args.config, args.skip_type)
+
+    dataset = build_dataset(cfg.data.train)
+
+    class_name = {k+1: v for k, v in zip(range(len(dataset.CLASSES)), dataset.CLASSES)}
+
+    progress_bar = mmcv.ProgressBar(len(dataset))
+    for item in dataset:
+        filename = Path(item['filename']).name
+        plot_annotation(args, filename, item['img'], item['gt_bboxes'], item['gt_labels'], class_name)
```
yhcao6: You can specify whether to show and the output_dir using `imshow_det_bboxes`. For example:
```
            filename = Path(item['filename']).name
            imshow_det_bboxes(
                item['img'],
                item['gt_bboxes'],
                item['gt_labels'] - 1,
                class_names=dataset.CLASSES,
                show=False,
                out_file=os.path.join(args.output_dir, filename))
```

hellock:
FYI, https://github.com/open-mmlab/mmdetection/blob/master/.github/CONTRIBUTING.md#code-style

PR #2259, opened 2020-03-14
refactor util

xvjiarui:


mmdet/models/plugins/non_local.py
```diff
@@ -2,7 +2,7 @@
 import torch.nn as nn
 from mmcv.cnn import constant_init, normal_init
 
-from ..utils import ConvModule
+from ..common import ConvModule
```
hellock: `plugins` may also be merged into `common`.

Issue #2262, opened 2020-03-15
Error building roi_align_v2 on windows


c1aris:
The error log is as follows. I am happy to provide more info if needed. Thank you!

```
mmdet/ops/roi_align/src/roi_align_kernel_v2.cu(290): error: no instance of function template "at::cuda::ATenCeilDiv" matches the argument list
                argument types are: (long long, long)

mmdet/ops/roi_align/src/roi_align_kernel_v2.cu(290): error: no instance of overloaded function "std::min" matches the argument list
                argument types are: (<error-type>, long)

mmdet/ops/roi_align/src/roi_align_kernel_v2.cu(330): error: no instance of function template "at::cuda::ATenCeilDiv" matches the argument list
                argument types are: (int64_t, long)

mmdet/ops/roi_align/src/roi_align_kernel_v2.cu(330): error: no instance of overloaded function "std::min" matches the argument list
                argument types are: (<error-type>, long)

4 errors detected in the compilation of "C:/Users/li/AppData/Local/Temp/tmpxft_00002ba8_00000000-10_roi_align_kernel_v2.cpp1.ii".
    roi_align_kernel_v2.cu

C:\Users\li\.conda\envs\mmdet\lib\site-packages\torch\utils\cpp_extension.py:209: UserWarning: Error checking compiler version for cl: 'utf-8' codec can't decode byte 0xd3 in position 0: invalid continuation byte
      warnings.warn('Error checking compiler version for {}: {}'.format(compiler, error))
error: command 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.1\\bin\\nvcc.exe' failed with exit status 1
```

c1aris:
I've found that modifying two lines in `mmdet/ops/roi_align/src/roi_align_kernel_v2.cu` works on Windows, but I'm not sure whether it would be the same on Linux or macOS.

Just change
```
dim3 grid(std::min(at::cuda::ATenCeilDiv(output_size, 512L), 4096L));
```
in line 290 to
```
dim3 grid(std::min(at::cuda::ATenCeilDiv(static_cast<int64_t>(output_size), static_cast<int64_t>(512)), static_cast<int64_t>(4096)));
```
and change
```
dim3 grid(std::min(at::cuda::ATenCeilDiv(static_cast<int64_t>(output_size), static_cast<int64_t>(512)), static_cast<int64_t>(4096)));
```
 in line 330 to
```
[...]

PR #2263, opened 2020-03-15
Fix type errors in roi_align_v2

c1aris:
This PR fixes the issue mentioned in #2262.


ZwwWayne:
Hi @c1aris ,
I have tested this PR in the CentOS system and it also works well. So the RoIAlign should work with both Windows and Linux after merging this PR.

c1aris:
@ZwwWayne That's great! May I ask that do you have any plans to re-benchmark the models using RoIAlignV2?

ZwwWayne:
Yes, the re-benchmarked models with RoIAlignV2 and all the BC-breaking changes will be released in about two weeks.

c1aris:
Glad to hear that. Thank you so much!

PR #2281, opened 2020-03-17
Set initial values for default meta keys when loading images

melikovk:
The aim of this change is to simplify use of pipelines that do not require image
scaling, normalization, flipping or padding

On branch set_default_metakeys
Changes to be committed:

### DIFF
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -27,6 +27,12 @@ class LoadImageFromFile(object):
         results['img'] = img
         results['img_shape'] = img.shape
         results['ori_shape'] = img.shape
+        # Set initial values for default meta_keys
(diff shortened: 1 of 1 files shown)


mmdet/datasets/pipelines/loading.py
```diff
@@ -27,6 +27,12 @@ def __call__(self, results):
         results['img'] = img
         results['img_shape'] = img.shape
         results['ori_shape'] = img.shape
+        # Set initial values for default meta_keys
+        results['pad_shape'] = img.shape
+        results['flip'] = False
+        results['scale_factor'] = 1.0
+        cnum = 1 if len(img.shape) < 3 else img.shape[2]
```
hellock: Use a more specific variable name like `num_channels`.

melikovk: Done

PR #2283, opened 2020-03-17
Add class to load multi-channel images from a list of files

melikovk:
This class will be usefull for loading a multi-channel image from a list of separate images. This
is often necessary for medical and scientific images, since OpenCV or PIL do not handle correctly
many specialized medical or scientific file formats.

On branch load_multichannel
Changes to be committed:

### DIFF
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -35,6 +35,39 @@ class LoadImageFromFile(object):
 
 
 @PIPELINES.register_module
+class LoadMultiChannelImageFromFiles(object):
+    """ Load multi channel images from a list of separate channel files.
+    Expects results['filename'] to be a list of filenames
+    """
+
+    def __init__(self, to_float32=True, color_type='unchanged'):
+        self.to_float32 = to_float32
+        self.color_type = color_type
+
+    def __call__(self, results):
+        if results['img_prefix'] is not None:
+            filename = [
+                osp.join(results['img_prefix'], fname)
+                for fname in results['img_info']['filename']
+            ]
+        else:
+            filename = results['img_info']['filename']
+        img = np.stack(
+            [mmcv.imread(name, self.color_type) for name in filename], axis=-1)
+        if self.to_float32:
+            img = img.astype(np.float32)
+        results['filename'] = filename
+        results['img'] = img
+        results['img_shape'] = img.shape
+        results['ori_shape'] = img.shape
+        return results
+
+    def __repr__(self):
+        return '{} (to_float32={}, color_type={})'.format(
+            self.__class__.__name__, self.to_float32, self.color_type)
+
+
+@PIPELINES.register_module
 class LoadAnnotations(object):
 
     def __init__(self,


hellock:
Thanks for adding the new feature. Could you show a usage example, e.g., a medical dataset?

melikovk:
Here are two examples of datasets using this approach:
1. Broad Bioimage Benchmark Collection: compound-profiling Cell Painting experiment
https://data.broadinstitute.org/bbbc/BBBC022/
Here each sample contains 5 images
2.  Human Protein Atlas Image Classification dataset
https://www.kaggle.com/c/human-protein-atlas-image-classification/data
Here image sample requires loading of 4 different images

To use this new class one would populate filename field with the list [fname1, fname2...]

melikovk:
Multiple images are combined into a single 3d array. The only difference with the usual color images is the size of the third dimension. For example after loading images from first example we will have array with dimensions (H, W, 5). Many opencv functions work as expected with such arrays and thus many transforms in the mmdet work. In particular all geometric transforms work (Resize, Pad, Crop).  I expect that many albumentation transforms will work as well although I have not used them. I also have written a number of transforms that I use in my work and will be happy to share them. They differ from transforms in albumentation package in that they allow per-channel application.

bnumaomei:
@melikovk Hi,  Thank you very much for your commit. But I still have some questions about the use of LoadMultiChannelImageFromFiles(). How to use this funtion ? What is format of input images and annotations ? I want to use concat a jpg and png, finally the dimension(H,W,4), I have no idea of the input annotations. Thank you ~

melikovk:
Hi. You will use it in your pipeline instead of LoadImageFromFile. LoadImageFromFile expects Dataset to provide a record with 'img_info' field that has 'filename' field with a single filename that is used to load image. In case of LoadMultiChannelImageFromFiles 'filename' field should be a list of filenames. In your case it should be ['path_to_jpg', 'path_to_png']. Keep in mind that images should have the same dtype before stacking. The easiest way to ensure this is to set to_float32 argument to True. To get a better idea about input annotations and data loading in mmdet take a look on 
[GETTING_STARTED/use-my-own-datasets](https://mmdetection.readthedocs.io/en/latest/GETTING_STARTED.html#use-my-own-datasets)
and
[TECHNICAL_DETAILS/data-pipeline](https://mmdetection.readthedocs.io/en/latest/TECHNICAL_DETAILS.html#data-pipeline)

bnumaomei:
@melikovk  Thanks~
I have changed my annotations alter the filenames as a list['*.jpg','.png'] and use the LoadImageFromFile. But I encounter the KeyError with the 'img_norm_cfg'.  In config file it is img_norm_cfg = dict(
    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True) What can I do to make it for my w*H*4 img?

melikovk:
If you are using LoadImageFromFile and put filenames into the list, you should be getting an error from LoadImageFromFile. It would help if you would provide your pipeline, example record from your annotation file, what Dataset you are using and error message.

LucasLee-ff:
@melikovk Hi,
I still have some confusion on how to use `LoadMultiChannelImageFromFiles` . Currently I have two folders containing 3-channel images and 1-channel images respectively. 
```
├── data
│   ├── my_dataset
│   │   ├── img_1_channel
│   │   │   ├── train
│   │   │   ├── val
│   │   │   ├── test
│   │   ├── img_3_channel
│   │   │   ├── train
│   │   │   ├── val
│   │   │   ├── test
```
If I want to load them together as a 4-channel input, what exactly do I need to change in my dataset config file except from simply replacing `LoadImageFromFile` with `LoadMultiChannelImageFromFiles`?
Thank you for your time!

bnumaomei:
这是来自徐杨的自动回复邮件。

melikovk:
Hi Lucas,

Unfortunately the current version of the filter will not work for your
case. It needs all files to be single channel images. The filter is
easy to modify to fit your purpose or you can split your 3ch image into
three 1ch images. Also your dataset should provide a list of filenames
in "filename" field rather than a single name.

Hope this helps and best regards.

LucasLee-ff:
Hi melikovk,
I've already figured out a way to solve this case. Thank you for your reply!

PR #2285, opened 2020-03-17
Add class to wrap result fields into lists for evaluation

melikovk:
This class is useful to create an evaluation pipeline that does not rely on 
MultiScaleFlipAugmentation 

On branch wrap_into_lists
Changes to be committed:
modified:   mmdet/datasets/pipelines/formating.py

### DIFF
--- a/mmdet/datasets/pipelines/formating.py
+++ b/mmdet/datasets/pipelines/formating.py
@@ -194,6 +194,21 @@ class Collect(object):
 
 @PIPELINES.register_module
 class WrapFieldsToLists(object):
+    """
+    Wrap fields of the data dictionary into lists for evaluation.
+
+    This class can be used as a last step of a test or validation
+    pipeline for single image evaluation or inference.
+    Example:
+        >>> test_pipeline = [
+        >>>    dict(type='LoadImageFromFile'),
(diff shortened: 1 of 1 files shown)


yhcao6:
Thanks for the pr. Could you show a usage example?

melikovk:
Here is an example test pipeline usage
`test_pipeline = [
    dict(type='LoadImageFromFile', color_type='unchanged', to_float32=True),
    dict(type='Normalize', **img_norm_cfg),
    dict(type='Pad', size_divisor=32),
    dict(type='ImageToTensor', keys=['img']),
    dict(type='Collect', keys=['img']),
    dict(type='WrapIntoLists')
]`

mmdet/datasets/pipelines/formating.py
```diff
@@ -190,3 +190,16 @@ def __call__(self, results):
     def __repr__(self):
         return self.__class__.__name__ + '(keys={}, meta_keys={})'.format(
             self.keys, self.meta_keys)
+
+
+@PIPELINES.register_module
+class WrapFieldsToLists(object):
```
hellock: Docstring is missing.

melikovk: I have added a docstring

PR #2291, opened 2020-03-18
Add onnx convert script

xvjiarui:
This PR is based on https://github.com/open-mmlab/mmdetection/pull/1082. 
The experimental ONNX conversion is added. The customized operators are not supported for now.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -360,7 +360,7 @@ Please refer to [ROBUSTNESS_BENCHMARKING.md](ROBUSTNESS_BENCHMARKING.md).
 
 ### Convert to ONNX (experimental)
 
-We provide a script convert model to [ONNX](https://github.com/onnx/onnx) format. The converted model could be visualized by tools like [Netron](https://github.com/lutzroeder/netron).
+We provide a script to convert model to [ONNX](https://github.com/onnx/onnx) format. The converted model could be visualized by tools like [Netron](https://github.com/lutzroeder/netron).
 
 ```shell
 python tools/pytorch2onnx.py ${CONFIG_FILE} ${CHECKPOINT_FILE} --out ${ONNX_FILE} [--shape ${INPUT_SHAPE}]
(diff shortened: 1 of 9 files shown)


hellock:
> image
> It may be better to create a new directory under configs named `onnx` since the original config uses a different setting by default.

We may also update the configs online.

tools/pytorch2onnx.py
```diff
@@ -0,0 +1,115 @@
...
+    all_passes = optimizer.get_available_passes()
+    passes = ['fuse_bn_into_conv']
+    assert all(p in all_passes for p in passes)
+    onnx_model = optimizer.optimize(onnx_model, passes)
+    return onnx_model
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='MMDet test (and eval) a model')
```
hellock: Update the description.

tools/pytorch2onnx.py
```diff
@@ -0,0 +1,115 @@
...
+    model.cpu().eval()
+    # TODO: a better way to override forward function
+    if hasattr(model, 'forward_dummy'):
+        model.forward = model.forward_dummy
+    else:
+        raise NotImplementedError(
+            'ONNX conversion is currently not currently supported with '
+            '{}'.format(model.__class__.__name__))
+
+    input_data = torch.ones(()).new_empty(
```
hellock: Why not just `torch.empty()`

tools/pytorch2onnx.py
```diff
@@ -0,0 +1,115 @@
+# Modified from https://github.com/facebookresearch/detectron2/blob/master/detectron2/export/api.py # noqa
```
hellock: Refer to https://github.com/open-mmlab/mmdetection/pull/1082

docs/GETTING_STARTED.md
```diff
@@ -358,6 +358,15 @@ The final output filename will be `faster_rcnn_r50_fpn_1x_20190801-{hash id}.pth
...
 
+### Convert to ONNX (experimental)
+
+We provide a script convert model to [ONNX](https://github.com/onnx/onnx) format. The converted model could be visualized by tools like [Netron](https://github.com/lutzroeder/netron).
+
+```shell
+python tools/pytorch2onnx.py ${CONFIG_FILE} ${CHECKPOINT_FILE} --out ${ONNX_FILE} [--shape ${INPUT_SHAPE}]
+```
+
+**Note**: This tool is still experimental. Only CPU operators are supported for now. For example, you may need to set `use_torchvision=True` for `RoIAlign`.
```
hellock: We can do for users this on-the-fly.

PR #2319, opened 2020-03-25
Add norm eval in configs

xvjiarui:
This PR add `norm_eval= True` in all configs files.

### DIFF
--- a/configs/_base_/models/cascade_mask_rcnn_r50_fpn.py
+++ b/configs/_base_/models/cascade_mask_rcnn_r50_fpn.py
@@ -10,6 +10,7 @@ model = dict(
         out_indices=(0, 1, 2, 3),
         frozen_stages=1,
         norm_cfg=dict(type='BN', requires_grad=True),
+        norm_eval=True,
         style='pytorch'),
     neck=dict(
         type='FPN',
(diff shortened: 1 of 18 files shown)


sudo-rm-covid19:
What is the reason for fixing the norms of backbones if pretrained weights are used during training?

xvjiarui:
Hi @sudo-rm-covid19 
In resnet based model, batch size is usually small, so batch statistics is not accurate enough.

sudo-rm-covid19:
> Hi @sudo-rm-covid19
> In resnet based model, batch size is usually small, so batch statistics is not accurate enough.
> So fixed BN is a common practice to address that.
Hi jiarui @xvjiarui ,
Thanks for your prompt reply. From my oberservation, if norm_eval sets True, the loss decreases at a lower rate. But if I set it False, the training error could diverge suddenly at some point (it might happen several epochs after the start of the training and every thing seems good in prior to that.)
Does it mean it will provide more stable training process but the results might be sub-optimal (in comparison to the scenario when we have larger batch size)?

xvjiarui:
Hi @sudo-rm-covid19 
Yes. So we could better result by using SyncBN with `norm_eval=False`. But we need to use `4Conv1FcHead` instead of `2FCHead` for that.

sudo-rm-covid19:
> Hi @sudo-rm-covid19
> Yes. So we could better result by using SyncBN with `norm_eval=False`. But we need to use `4Conv1FcHead` instead of `2FCHead` for that.

Hi @xvjiarui 
Thanks for your advice!

PR #2324, opened 2020-03-25
Add support for evaluation in non-dist training

c1aris:
This PR adds support for doing evaluation during single-gpu training, by adding an `EvalHook` and slightly modifying `DistEvalHook` for reusing the code.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -188,7 +188,7 @@ According to the [Linear Scaling Rule](https://arxiv.org/abs/1706.02677), you ne
 ### Train with a single GPU
 
 ```shell
-python tools/train.py ${CONFIG_FILE}
+python tools/train.py ${CONFIG_FILE} [optional arguments]
 ```
 
 If you want to specify the working directory in the command, you can add an argument `--work_dir ${YOUR_WORK_DIR}`.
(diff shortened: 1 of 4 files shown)


hyz-xmaster:
> This PR adds support for doing evaluation during single-gpu training, by adding an `EvalHook` and slightly modifying `DistEvalHook` for reusing the code.

Hi c1aris, thanks for this work. I was also working on this feature on my own project. 

I have tested the feature and found that this feature does not work with multiple gpus training, say 2 gpus in one machine, which is exactly what you have said "support for doing evaluation during single-gpu training". I was wondering it is possible to add such a support for multiple-gpu training.

c1aris:
Hi @hyz-xmaster ! Thanks for your attention.

May I ask that are you using `MMDataParallel` or `MMDistributedDataParallel` for '2 gpus in one machine'? As far as I know, if you have more than one gpu (no matter in single or multiple machines), you may want to use `MMDDP` for training, which is the default setting in mmdet as well. Also, `MMDP` works with `EvalHook` while `MMDDP` works with `DistEvalHook `, so if you are using 2 gpus in one machine, you can simply use `MMDDP` together with `DistEvalHook`.

hyz-xmaster:
Hi @c1aris, thanks for your quick reply. 
I was in fact always using _non_dist_train() (--gpus 2) to train the network with 2 gpus on one machine for the debugging convenience, resulting in the failure of using either EvalHook or DistEvalHook. I think I should use dist_train() instead to be compatible with DistEvalHook and my trial has verified this.

PR #2330, opened 2020-03-26
Set default values for meta keys in LoadMultiChannelImageFromFiles

melikovk:
Fix LoadMultiChannelImageFromFiles to set default values for meta keys similar to LoadImageFromFile.
Refactor Image loading classes to reduce code duplication

Changes to be committed:


mmdet/datasets/pipelines/loading.py
```diff
@@ -32,8 +36,10 @@ def __call__(self, results):
         results['flip'] = False
```
ZwwWayne, line 36: The random flip only does flip when `results` do not have the key `flip` at the beginning. This line should be removed, otherwise, the random flip will never be executed.

melikovk: If we remove it here, then a naive pipeline without flip transform will fail, because 'flip' is default key. One will have to either insert dummy flip transform or supply custom list of default key. 

I suggest to change the behavior of flip to skip on False flip key, rather than its absence. WDYT?

ZwwWayne: A simple way is to keep the flip transform but set the flip prob as 0.

ZwwWayne: But thanks for your suggestion, we will think about it.

melikovk: Thank you for review. I have made a change you requested. I understand that one can just set p=0, that is what I meant by dummy flip transform. I think it is not a great solution from the point of the user of the project, since it is not clear why one **should** have a flip transform in the pipeline. Thank you.

ZwwWayne: Thanks for your kind suggestion and contribution. We will take it into consideration in our future release.

melikovk:
I am not quite sure why file_client initialization is not in __init__

ZwwWayne:
Since we still do not have an elegant to fully reduce the code. This PR is closed.

Issue #2288, opened 2020-03-18
train only person class with COCO2017 and get abnormal result


alyssa121:
**Describe the bug**
I extract person class from COCO2017 train and validation dataset, and get 64115 pics for training and 2693 pics for testing. However, the testing result seems really abnormal:
**************************************************************************************
```
loading annotations into memory...
Done (t=0.06s)
creating index...
index created!
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 2693/2693, 12.4 task/s, elapsed: 217s, ETA:     0s
Evaluating bbox...
Loading and preparing results...
DONE (t=1.90s)
creating index...
index created!
Running per image evaluation...
Evaluate annotation type *bbox*
DONE (t=41.17s).
Accumulating evaluation results...
DONE (t=4.33s).
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.002
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.009
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.000
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.000
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.000
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.004
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.004
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.026
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.059
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.000
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.000
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.161
```
**************************************************************************************

**Reproduction**
1. What command or script did you run?
training command is:
 `python tools/train.py configs/aaaa/faster_rcnn_r50_fpn_1x_coco.py`

testing command is: 
[...]

Issue #2166, opened 2020-02-26
Cannot train retinanet with a single class


manycoding:
**Describe the bug**
I cannot get any results from training retinanet with 1 class.
Log doesn't contain anything after `2020-02-26 14:50:59,227 - mmdet - INFO - workflow: [('train', 1)], max: 12 epochs`, however gpu seems loaded.

**Reproduction**
1. What command or script did you run?
[colab notebook](https://colab.research.google.com/drive/1joj2udGzFRyrHSF3tnXEHHyvN_2D3CSo)
[config on gist](https://gist.github.com/manycoding/a38ff0a0f659dbcbc1c4387c55b63769)
```
!python tools/train.py "{home}/retinanet_fp16.py"
!python tools/test.py "{home}/retinanet_fp16.py" {checkpoint_file} --eval mAP
```
2. Did you make any modifications on the code or config? Did you understand what you have modified?
I set `num_classes=2,`, yes I understand.
3. What dataset did you use?
I am using my custom dataset with about 70 images and bboxes.

I pickled annotations, here's the first line:
```
{'ann': {'bboxes': array([[ 25.373676, 293.4756  , 190.30257 , 336.04767 ]], dtype=float32),
  'labels': array([0])},
 'filename': '000001-2020-02-14T16:50:17.554Z.jpg',
 'height': 399,
 'width': 599}
```

**Environment**
Google Colab with GPU

1.[`python mmdet/utils/collect_env.py`](https://gist.github.com/manycoding/e36495b7ae41aee1300f89fef2303e46)

liyiwei979621500:
Have you modify the CustomDataset dataset?
The CustomDataset does not has the CLASSES attribute with it,
which might leads to your issus.

take the CocoDataset as an example:
```
@DATASETS.register_module
class CocoDataset(CustomDataset):

    CLASSES = ('person', 'bicycle', 'car', 'motorcycle', 'airplane', 'bus',
               'train', 'truck', 'boat', 'traffic_light', 'fire_hydrant',
               'stop_sign', 'parking_meter', 'bench', 'bird', 'cat', 'dog',
               'horse', 'sheep', 'cow', 'elephant', 'bear', 'zebra', 'giraffe',
               'backpack', 'umbrella', 'handbag', 'tie', 'suitcase', 'frisbee',
               'skis', 'snowboard', 'sports_ball', 'kite', 'baseball_bat',
[...]

Issue #424, opened 2019-03-21
'person' single class training and testing by using resnet50-based retinanet


WilliamYi96:
I wonder if I just need to use 'person' label to train and test by setting resnet50-based retinanet as backbone, what do I need to modify?

I've attempted to modifiy the following:

- retinanet_r50_fpn_1x.py: convert num_classes=81 =>> number_classes=2; 
- mmdet/datasets/coco.py: self.cat_ids = self.coco.getCatIds() =>> self.cat_ids = self.coco.getCatIds(catNms=['person']) and self.img_ids = self.coco.getImgIds() =>> self.img_ids = self.coco.getImgIds(CatIds=self.CatIds).

However, it still can not work.
Has anyone faced the same problem and solved it?

wizholy:
when install the mmdetection, you shold use 
python setup.py develop

WilliamYi96:
@wizholy Thanks for your reminder. However, I'm using `python setup.py develop` indeed. Currently, I've solved the problem by just kicking out irrelevant categories and annotations in COCO17 dataset.

PR #2340, opened 2020-03-28
Support to train subsets of coco dataset

ZwwWayne:
There are some needs to train models on the subset of the COCO dataset, which only contain several categories, e.g., people, bicycles and cars. These models are taken as pre-trained models and further finetuned on small datasets such as KITTI. 

However, some bugs are reported when the users train models by themselves on subset of COCO dataset (e.g., in PR https://github.com/open-mmlab/mmdetection/issues/2288, https://github.com/open-mmlab/mmdetection/issues/2166, and https://github.com/open-mmlab/mmdetection/issues/424)
Therefore, this PR supports to set class names to be used when training on the subset of the COCO dataset, which could release the burden to subtract subsets before training. If the `class_names` is None, the data loader will use all the categories specified by the `self.CLASSES` by default, which uses all the data. If the `class_names` is a non-empty list, the dataset will use the `class_names` to find those images contains the specified categories and use them to train the model.

**Notice**: Currently, this behavior is only supported on the COCO dataset.


mmdet/datasets/coco.py
```diff
@@ -32,13 +32,25 @@ class CocoDataset(CustomDataset):
                'vase', 'scissors', 'teddy_bear', 'hair_drier', 'toothbrush')
 
     def load_annotations(self, ann_file):
+        if not self.class_names:
```
xvjiarui, line 35: Why strip `_`?

mmdet/datasets/coco.py
```diff
@@ -32,13 +32,25 @@ class CocoDataset(CustomDataset):
...
+            self.class_names = [res.replace('_', ' ') for res in self.CLASSES]
```
xvjiarui, line 36: Add comment about this.

igonro:
Hi @ZwwWayne,

we have created a PR (#2408) that partially overlaps yours. We believe that the changes we want to add can also solve the use case you are dealing with in this Pull Request and it may be a good idea to generalize this solution to include also other use cases.

In order to fully match your use case, you would need to apply the following changes (the modified lines are marked with an asterisk, *):

In **datasets/coco.py:**

In order to filter the desired `self.cat_ids` (just like you do), we would have to modify the following in `load_annotations()` function:
```python
def load_annotations(self, ann_file):
    self.coco = COCO(ann_file)
*   self.cat_ids = self.coco.getCatIds(catNms=self.CLASSES)
    self.cat2label = {
        cat_id: i + 1
        for i, cat_id in enumerate(self.cat_ids)
    }
    self.img_ids = self.coco.getImgIds()
    img_infos = []
    for i in self.img_ids:
        info = self.coco.loadImgs([i])[0]
        info['filename'] = info['file_name']
        img_infos.append(info)
    return img_infos
```
Then, to filter the images that do not include labels of the desired categories, we could do the following in the `_filter_imgs()` function:
```python
def _filter_imgs(self, min_size=32):
    """Filter images too small or without ground truths."""
    valid_inds = []
    ids_with_ann = set(_['image_id'] for _ in self.coco.anns.values())
*   ids_with_valid_ann = set(_['image_id']
*                               for _ in self.coco.anns.values()
*                               if _['category_id'] in self.cat_ids)
    for i, img_info in enumerate(self.img_infos):
        if (self.filter_empty_gt and self.img_ids[i] not in ids_with_ann
*           ) or (self.custom_classes
*                   and self.img_ids[i] not in ids_with_valid_ann):
            continue
        if min(img_info['width'], img_info['height']) >= min_size:
            valid_inds.append(i)
    return valid_inds
```

In **datasets/custom.py:**

Also, to support the changes introduced in `filter_imgs()`, it would be necessary to modify the `__init__()` of the `CustomDataset` class as follows:
```python
def __init__(..., classes=None):
    [...]
    self.custom_classes = False

    if classes is not None:
        self.CLASSES = classes
        self.custom_classes = True
    [...]
```

I don't know if you think that this would cover your use case or there is something missing, but it would be nice if we can find an intermediate solution. Thank you!

configs/faster_rcnn_r50_fpn_1x_coco-subset.py
```diff
@@ -0,0 +1,180 @@
...
+        roi_layer=dict(type='RoIAlign', out_size=7, sample_num=2),
+        out_channels=256,
+        featmap_strides=[4, 8, 16, 32]),
+    bbox_head=dict(
+        type='SharedFCBBoxHead',
+        num_fcs=2,
+        in_channels=256,
+        fc_out_channels=1024,
+        roi_feat_size=7,
+        num_classes=81,
```
Dragonsson, line 41: The num of classes you used still is 81.

ZwwWayne:
> Hi @ZwwWayne,
> 
> we have created a PR (#2408) that partially overlaps yours. We believe that the changes we want to add can also solve the use case you are dealing with in this Pull Request and it may be a good idea to generalize this solution to include also other use cases.
> 
> In order to fully match your use case, you would need to apply the following changes (the modified lines are marked with an asterisk, *):
> 
> In **datasets/coco.py:**
> 
> In order to filter the desired `self.cat_ids` (just like you do), we would have to modify the following in `load_annotations()` function:
> 
> ```python
> def load_annotations(self, ann_file):
>     self.coco = COCO(ann_file)
> *   self.cat_ids = self.coco.getCatIds(catNms=self.CLASSES)
>     self.cat2label = {
>         cat_id: i + 1
>         for i, cat_id in enumerate(self.cat_ids)
>     }
>     self.img_ids = self.coco.getImgIds()
>     img_infos = []
>     for i in self.img_ids:
>         info = self.coco.loadImgs([i])[0]
>         info['filename'] = info['file_name']
>         img_infos.append(info)
>     return img_infos
> ```
> 
> Then, to filter the images that do not include labels of the desired categories, we could do the following in the `_filter_imgs()` function:
> 
> ```python
> def _filter_imgs(self, min_size=32):
>     """Filter images too small or without ground truths."""
>     valid_inds = []
>     ids_with_ann = set(_['image_id'] for _ in self.coco.anns.values())
> *   ids_with_valid_ann = set(_['image_id']
> *                               for _ in self.coco.anns.values()
> *                               if _['category_id'] in self.cat_ids)
>     for i, img_info in enumerate(self.img_infos):
>         if (self.filter_empty_gt and self.img_ids[i] not in ids_with_ann
> *           ) or (self.custom_classes
> *                   and self.img_ids[i] not in ids_with_valid_ann):
>             continue
>         if min(img_info['width'], img_info['height']) >= min_size:
>             valid_inds.append(i)
>     return valid_inds
> ```
> 
> To filter the annotations, you have to add the following code (exactly like you do) in `_parse_ann_info()` function (inside the for loop):
> 
> ```python
> if ann['category_id'] not in self.cat_ids:
>     continue
> ```
> 
> In **datasets/custom.py:**
> 
> Also, to support the changes introduced in `filter_imgs()`, it would be necessary to modify the `__init__()` of the `CustomDataset` class as follows:
> 
> ```python
> def __init__(..., classes=None):
>     [...]
>     self.custom_classes = False
> 
>     if classes is not None:
>         self.CLASSES = classes
>         self.custom_classes = True
>     [...]
> ```
> 
> I don't know if you think that this would cover your use case or there is something missing, but it would be nice if we can find an intermediate solution. Thank you!
> 
> **EDIT**: I forgot to include the code to filter the annotations.
> **EDIT2**: Hidden comment because after testing it, we realized that with minor fixes it also solves our use case.

Thanks for your kind suggestion. We plan to merge this PR after merging https://github.com/open-mmlab/mmdetection/pull/2408

mmdet/datasets/coco.py
```diff
@@ -60,6 +69,26 @@ def _filter_imgs(self, min_size=32):
                 valid_inds.append(i)
         return valid_inds
 
+    def get_imgs_by_cat(self, imgIds=[], catIds=[]):
```
hellock, line 72: - `get_subset_by_classes()` looks better.
- `img_ids`, `class_ids`
- What is `img_ids` for? It seems to be unused.
- Use tuple or None as default values.
- Use the standard docsting style.

mmdet/datasets/custom.py
```diff
@@ -156,6 +154,27 @@ def prepare_test_img(self, idx):
         self.pre_pipeline(results)
         return self.pipeline(results)
 
+    def get_classes(self, classes):
```
hellock, line 157: Make it a class method and return a list of class names.

mmdet/datasets/custom.py
```diff
@@ -156,6 +154,27 @@ def prepare_test_img(self, idx):
...
+        """Get classes of current dataset
+
+        Args:
+            classes (None | str | tuple | list): If classes is None, use
```
hellock, line 161: `Sequence[str] | str | None`

mmdet/datasets/coco.py
```diff
@@ -17,25 +17,34 @@
...
+        self.cat_ids = self.coco.getCatIds(catNms=self.CLASSES)
+
         self.cat2label = {cat_id: i for i, cat_id in enumerate(self.cat_ids)}
-        self.img_ids = self.coco.getImgIds()
+        # send cat ids to the get img id
+        # in case we only need to train on several classes
+        if self.custom_classes:
+            self.img_ids = self.get_imgs_by_cat(catIds=self.cat_ids)
+        else:
+            self.img_ids = self.coco.getImgIds()
```
hellock, line 46: Is the condition necessary?

mmdet/datasets/coco.py
```diff
@@ -60,6 +66,26 @@ def _filter_imgs(self, min_size=32):
                 valid_inds.append(i)
         return valid_inds
 
+    def get_subset_by_classes(self, class_ids=None):
+        '''Get img ids that contain any category in class_ids.
+
+        Different from the coco.getImgIds, this function returns the id if
+        the img contains one of the cat rather than all.
```
hellock, line 73: Different from the `coco.getImgIds()`, this function returns the id if the img contains one of the categories rather than all.

configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco-ped-bicycle-car.py
```diff
@@ -0,0 +1,7 @@
+_base_ = './faster_rcnn_r50_fpn_1x_coco.py'
+classes = ['person', 'bicycle', 'car']
```
hellock, line 2: Use tuple instead of list.

configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco-person.py
```diff
@@ -0,0 +1,7 @@
+_base_ = './faster_rcnn_r50_fpn_1x_coco.py'
+classes = ('person')
+data = dict(
+    train=dict(classes=classes),
+    val=dict(classes=classes),
+    test=dict(classes=classes))
+load_from = 'https://s3.ap-northeast-2.amazonaws.com/open-mmlab/mmdetection/models/faster_rcnn_r50_fpn_1x_20181010-3d1b3351.pth'  # noqa
```
hellock, line 7: Remove `load_from`.

tests/test_dataset.py
```diff
@@ -0,0 +1,65 @@
...
+        classes=None,
+        test_mode=True,
+        img_prefix='VOC2007' if dataset == 'VOCDataset' else '')
+
+    assert custom_dataset.CLASSES == original_classes
+
+    # Test sending file path
+    import tempfile
+    import os.path as osp
+    tmp_file = tempfile.TemporaryDirectory()
```
hellock, line 51: Use a temp file instead of a temp dir.

mmdet/datasets/custom.py
```diff
@@ -156,6 +154,31 @@ def prepare_test_img(self, idx):
         self.pre_pipeline(results)
         return self.pipeline(results)
 
+    @classmethod
+    def get_classes(cls, classes):
```
hellock, line 158: Provide a default value `classes=None`.

tests/test_wrappers.py
```diff
@@ -11,6 +11,11 @@
 
 
 def test_conv_2d():
```
hellock, line 13: test_conv2d

tests/test_wrappers.py
```diff
@@ -26,7 +31,7 @@ def test_conv_2d():
         wrapper_out = wrapper(x_empty)
 
         # torch op with 3-dim input as shape reference
-        x_normal = torch.randn(3, in_cha, in_h, in_w)
+        x_normal = torch.randn(3, in_cha, in_h, in_w).requires_grad_(True)
```
hellock, line 34: `torch.randn(3, in_cha, in_h, in_w, requires_grad=True)`

mmdet/datasets/xml_style.py
```diff
@@ -30,6 +30,26 @@ def load_annotations(self, ann_file):
             height = int(size.find('height').text)
             data_infos.append(
                 dict(id=img_id, filename=filename, width=width, height=height))
+
+        data_infos = self.get_subset_by_classes(data_infos)
```
hellock, line 34: Parsing the annotation files twice may be slow.

mmdet/datasets/custom.py
```diff
@@ -156,6 +154,31 @@ def prepare_test_img(self, idx):
...
+
+        """
+        if classes is None:
+            return cls.CLASSES
+
+        if isinstance(classes, str):
+            # take it as a file path
+            class_names = mmcv.list_from_file(classes)
+        elif isinstance(classes, (tuple, list)):
+            class_names = classes
```
xvjiarui, line 176: We may raise an exception there.

mmdet/datasets/xml_style.py
```diff
@@ -44,6 +75,8 @@ def get_ann_info(self, idx):
         labels_ignore = []
         for obj in root.findall('object'):
             name = obj.find('name').text
+            if name not in self.CLASSES:
+                continue
```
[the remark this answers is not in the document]
ZwwWayne: Yes. Because sometimes the annotation of an image contains instances of not only customized classes but also other classes. Those unused classes should be marked as background.

mmdet/datasets/custom.py
```diff
@@ -156,6 +158,38 @@ def prepare_test_img(self, idx):
...
+        else:
+            raise ValueError('Unsupported type {} of classes.'.format(
+                type(classes)))
+
+        assert set(class_names).issubset(set(cls.CLASSES))
+        return class_names
+
+    def get_subset_by_classes(self):
+        pass
```
hellock, line 191: return self.data_infos

mmdet/datasets/custom.py
```diff
@@ -156,6 +158,38 @@ def prepare_test_img(self, idx):
...
+        Args:
+            classes (Sequence[str] | str | None): If classes is None, use
+                default CLASSES defined by builtin dataset. If classes is a
+                string, take it as a file name. The file contains the name of
+                classes where each line contains one class name. If classes is
+                a tuple or list, override the CLASSES defined by the dataset.
+
+        """
+        if classes is None:
+            cls.custom_classes = True
```
hellock, line 174: True/False is wrong?

mmdet/datasets/xml_style.py
```diff
@@ -30,8 +30,27 @@ def load_annotations(self, ann_file):
             height = int(size.find('height').text)
             data_infos.append(
                 dict(id=img_id, filename=filename, width=width, height=height))
+
         return data_infos
 
+    def get_subset_by_classes(self):
+        """Filter imgs by user-defined categories
+        """
+        clean_data_infos = []
```
hellock, line 39: subset_data_infos

PR #2345, opened 2020-03-29
Refactor gcnet/emperical into plugin

xvjiarui:
This PR refactored GCNet and empirical attention into plugin.

### DIFF
--- a/configs/empirical_attention/faster_rcnn_r50_fpn_attention_0010_1x_coco.py
+++ b/configs/empirical_attention/faster_rcnn_r50_fpn_attention_0010_1x_coco.py
@@ -1,13 +1,13 @@
 _base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
 model = dict(
-    backbone=dict(
-        plugins=dict(conv2=[
-            dict(
-                cfg=dict(
-                    type='GeneralizedAttention',
-                    spatial_range=-1,
-                    num_heads=8,
-                    attention_type='0010',
-                    kv_stride=2),
-                stages=(False, True, True, True)),
-        ])))
+    backbone=dict(plugins=[
+        dict(
+            cfg=dict(
+                type='GeneralizedAttention',
+                spatial_range=-1,
+                num_heads=8,
+                attention_type='0010',
+                kv_stride=2),
+            stages=(False, True, True, True),
+            position='after_conv2')
+    ]))
--- a/configs/empirical_attention/faster_rcnn_r50_fpn_attention_0010_1x_coco.py
+++ b/configs/empirical_attention/faster_rcnn_r50_fpn_attention_0010_1x_coco.py
@@ -1,7 +1,12 @@
 _base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
 model = dict(
     backbone=dict(
-        gen_attention=dict(
-            spatial_range=-1, num_heads=8, attention_type='0010', kv_stride=2),
-        stage_with_gen_attention=[[], [], [0, 1, 2, 3, 4, 5], [0, 1, 2]],
-    ))
+        plugin=dict(conv2=[
+            dict(
+                type='GeneralizedAttention',
+                spatial_range=-1,
+                num_heads=8,
+                attention_type='0010',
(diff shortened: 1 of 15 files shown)


mmdet/models/backbones/resnet.py
```diff
@@ -180,14 +179,25 @@ def __init__(self,
...
-        # gen_attention
-        if self.with_gen_attention:
-            self.gen_attention_block = GeneralizedAttention(
-                planes, **gen_attention)
+        self.conv1_plugin_names = self.make_plugin(planes, self.conv1_plugin)
+        self.conv2_plugin_names = self.make_plugin(planes, self.conv2_plugin)
+        self.conv3_plugin_names = self.make_plugin(planes * self.expansion,
+                                                   self.conv3_plugin)
+
+    def make_plugin(self, in_channels, plugin):
```
hellock: Add a docstring.

mmdet/models/backbones/resnet.py
```diff
@@ -180,14 +179,25 @@ def __init__(self,
...
+        assert isinstance(plugin, list)
+        plugin_names = []
+        for plugin in plugin:
```
hellock: Avoid using the same variable name.

mmdet/models/backbones/resnet.py
```diff
@@ -96,17 +93,17 @@ def __init__(self,
                  conv_cfg=None,
                  norm_cfg=dict(type='BN'),
                  dcn=None,
-                 gcb=None,
-                 gen_attention=None):
+                 plugin=None):
```
hellock: plugins

hellock: Add a docstring for this class.

mmdet/models/backbones/resnet.py
```diff
@@ -308,10 +318,8 @@ def __init__(self,
                  norm_eval=True,
                  dcn=None,
                  stage_with_dcn=(False, False, False, False),
-                 gcb=None,
-                 stage_with_gcb=(False, False, False, False),
-                 gen_attention=None,
-                 stage_with_gen_attention=((), (), (), ()),
+                 plugin=None,
+                 stage_with_plugin=(False, False, False, False),
```
hellock: In this way, all plugins share the same setting, which is not flexible if we have multiple plugins at the same time.

It is more flexible (but also complicated) with this type:
```python
plugin = dict(
    conv1=[
        dict(
            cfg=dict(type='xxx', arg1='xxx'), stages=(False, True, True, True))
    ],
    conv2=[
        dict(cfg=dict(type='yyy')),
        dict(cfg=dict(type='yy2'), stages=(False, False, True, True))
    ])
```

configs/empirical_attention/faster_rcnn_r50_fpn_attention_0010_1x_coco.py
```diff
@@ -1,7 +1,13 @@
 _base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
 model = dict(
     backbone=dict(
-        gen_attention=dict(
-            spatial_range=-1, num_heads=8, attention_type='0010', kv_stride=2),
-        stage_with_gen_attention=[[], [], [0, 1, 2, 3, 4, 5], [0, 1, 2]],
-    ))
+        plugins=dict(conv2=[
```
ZwwWayne: I just come up with another idea of plugins. It might not be better than the current implementation but I think it is worth consideration. It can be something like this:
```python
plugins=[
    dict(cfg=dict(
        type='GeneralizedAttention',
        spatial_range=-1,
        num_heads=8,
        attention_type='0010',
        kv_stride=2),
        stages=(False, True, True, True),
        position='after_conv2'),
    dict(cfg=dict(
        type='XXXXY',
        spatial_range=-1,
        num_heads=8,
        attention_type='0010',
        kv_stride=2),
        stages=(False, False, True, True),
        position='after_conv3'),
]
```
Then we can use something like `conv2_modules`, `conv3_modules` to add them in.
This seems to reduce the hierarchy.

hellock: Sounds good to me.

PR #2364, opened 2020-03-31
Modify the code casting digit string as int to convert to float and t…

marload:
Occasionally, certain labeltool types can be floating points. If the coordinates are floating points, an error will occur. Therefore, it was corrected to prevent the error from occurring.

Thx !


hellock:
Thanks! You may leave a comment in the codes to explain the intention.

marload:
@hellock 
Added comment. Thank you very much

hellock:
The CI failed.

marload:
@hellock 
I fixed the problem!

PR #2376, opened 2020-04-01
L1 loss

ZwwWayne:
This PR support to use L1 loss when the beta is set to be very small.


mmdet/models/losses/smooth_l1_loss.py
```diff
@@ -43,3 +50,31 @@ def forward(self,
...
+        self.reduction = reduction
+        self.loss_weight = loss_weight
+
+    def forward(self,
+                pred,
+                target,
+                weight=None,
+                avg_factor=None,
+                reduction_override=None,
+                **kwargs):
```
hellock, line 69: There is no kwargs for L1Loss.

PR #2199, opened 2020-03-04
add CPU only mode which can be activated during install

yossibiton:
This is very useful feature, as it make the repository usable to much more environments. 
For example, i was able to run the webcam demo on my mac using different pretrained models.

PR #2385, opened 2020-04-03
Support CPU mode for inference

xvjiarui:
Based on @yossibiton 's PR #2199, support CPU mode in `webcam_demo.py`


yossibiton:
Did you check my latest code version? I'm not sure what is the added value
of this PR.

xvjiarui:
Hi @yossibiton ,
You could checkout `mmdet/ops`, in v2.0 branch, we refactor all customized operator to support compile in both CPU and CUDA mode. So we don't need to check `CPU_ONLY` when import ops.

yossibiton:
Thank you very much for Thai update, I'm sure many users will find this feature usefull.
By the way, the CPU mode is not relevant to webcam since you can even run small training demo if you don't use any of the GPU only operations.

fabmeyer:
@hellock Is it possible now to do inference with mask r-cnn in cpu mode? What argument do you have to pass in `device=`?
I've got an error when building my model: `model = init_detector(config_file, checkpoint_file, device='cpu')`

output:
`Traceback (most recent call last):
  File "program/mask_r-cnn.py", line 18, in <module>
    model = init_detector(config_file, checkpoint_file, device='cpu')
  File "/Users/fabianmeyer/.local/lib/python3.7/site-packages/mmdet-1.1.0+7ed8d51-py3.7-macosx-10.9-x86_64.egg/mmdet/apis/inference.py", line 35, in init_detector
    model = build_detector(config.model, test_cfg=config.test_cfg)
  File "/Users/fabianmeyer/.local/lib/python3.7/site-packages/mmdet-1.1.0+7ed8d51-py3.7-macosx-10.9-x86_64.egg/mmdet/models/builder.py", line 43, in build_detector
    return build(cfg, DETECTORS, dict(train_cfg=train_cfg, test_cfg=test_cfg))
  File "/Users/fabianmeyer/.local/lib/python3.7/site-packages/mmdet-1.1.0+7ed8d51-py3.7-macosx-10.9-x86_64.egg/mmdet/models/builder.py", line 15, in build
    return build_from_cfg(cfg, registry, default_args)
  File "/Users/fabianmeyer/.local/lib/python3.7/site-packages/mmdet-1.1.0+7ed8d51-py3.7-macosx-10.9-x86_64.egg/mmdet/utils/registry.py", line 79, in build_from_cfg
    return obj_cls(**args)
TypeError: __init__() got an unexpected keyword argument 'bbox_roi_extractor'`

As far as I have read there isn't a flag to set cpu mode during installation?

BerryRB:
@fabmeyer  Do you solva this problem? I meet the same error

fabmeyer:
@BerryRB MMDetection 2.0.0 supports inference in CPU mode. If you're working with 1.x, you can use something like [smd](https://github.com/akarazniewicz/smd)

SebastienLinker:
Hi, Does that work for every model? It seems we're having some problems here: https://github.com/open-mmlab/mmdetection/issues/3048

hellock:
> Hi, Does that work for every model? It seems we're having some problems here: #3048

https://github.com/open-mmlab/mmdetection/blob/master/docs/install.md#install-with-cpu-only

SebastienLinker:
Thanks @hellock but it seems my code does not use those things. The `CPU only` option should be set during install? Can we use the same install of mmdetection for both CPU and GPU, by setting `CUDA_VISIBLE_DEVICES` environment variable? I am trying that for the purpose of unit tests

PR #2380, opened 2020-04-02
Change V2.0 coors

ZwwWayne:
This PR changes the codebase's coordinate system and has effects on all the operations related to anchors, boxes, and masks.

PR #2388, opened 2020-04-03
Remove unnecessary operation with -1 in AnchorGenerator

Mxbonn:
The AnchorGenerator does a -1 operation in 2 places.
In the case with for example SSD when using an input size of 300 and a scale factor of 1., the expected behaviour would be that the generated anchors produce bounding boxes with a width and a height of 30, which is not the case here (it is 28).

Other implementations such as the Tensorflow Object Detection framework don't have this behaviour.
https://github.com/tensorflow/models/blob/34a3581c0e09c293cd67b5ac2d64e8bc753cefce/research/object_detection/anchor_generators/grid_anchor_generator.py#L213: 
`return tf.concat([centers - .5 * sizes, centers + .5 * sizes], 1)`

### DIFF
--- a/mmdet/core/anchor/anchor_generator.py
+++ b/mmdet/core/anchor/anchor_generator.py
@@ -8,10 +8,10 @@ class AnchorGenerator(object):
         >>> self = AnchorGenerator(9, [1.], [1.])
         >>> all_anchors = self.grid_anchors((2, 2), device='cpu')
         >>> print(all_anchors)
-        tensor([[ 0.,  0.,  8.,  8.],
-                [16.,  0., 24.,  8.],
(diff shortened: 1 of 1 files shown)


Mxbonn:
I just looked at other pull requests and saw you just changed this in 2.0
https://github.com/open-mmlab/mmdetection/pull/2380

Is there any timeline for the release on 2.0 and do you suggest switching to the 2.0 preview branch in the meantime?

ZwwWayne:
Hi @Mxbonn ,
V2.0 should be released in the following one week. Since V2.0 changes frequently and heavily, we suggest not use it until it is merged into master and released.

PR #2400, opened 2020-04-06
Add ability to specify device ids for non-distributed training

melikovk:
The motivation for this pull request is to simultaneously train different models on the multi-gpu system
with each model trained on separate device (or devices).

Changes to be committed:
modified:   mmdet/apis/train.py
modified:   tools/train.py

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -200,7 +200,9 @@ def _non_dist_train(model,
             seed=cfg.seed) for ds in dataset
     ]
     # put model on gpus
-    model = MMDataParallel(model, device_ids=range(cfg.gpus)).cuda()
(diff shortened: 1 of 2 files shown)


mmdet/apis/train.py
```diff
@@ -200,7 +200,9 @@ def _non_dist_train(model,
             seed=cfg.seed) for ds in dataset
     ]
     # put model on gpus
-    model = MMDataParallel(model, device_ids=range(cfg.gpus)).cuda()
+    gpu_ids = range(cfg.gpus) if cfg.gpu_ids is None else cfg.gpu_ids
+    assert len(gpu_ids) == cfg.gpus
```
hellock, line 204: We may deprecate `gpus` if `gpu_ids` is specified.

melikovk: I have made changes to deprecate gpus

PR #2405, opened 2020-04-07
Refactor (assigner): add iou_calculator

ZwwWayne:
This PR adds an `iou_calculator` in the core to support potentially more IoU calculation while reusing the supported assigners.


mmdet/core/bbox/iou_calculators/registry.py
```diff
@@ -0,0 +1,3 @@
+from mmdet.utils import Registry
+
+IOUCALCULATOR = Registry('iou_calculator')
```
xvjiarui, line 3: We may rename to `IOU_CALCULATOR`.

mmdet/core/bbox/iou_calculators/registry.py
```diff
@@ -0,0 +1,3 @@
+from mmdet.utils import Registry
+
+IOU_CALCULATOR = Registry('iou_calculator')
```
hellock, line 3: `IOU_CALCULATORS = Registry('IoU calculator')`

mmdet/core/bbox/iou_calculators/iou2d_calculator.py
```diff
@@ -0,0 +1,24 @@
+from ..geometry import bbox_overlaps
```
hellock, line 1: There is only one method in `geometry.py`. We may consider moving it to this file.

PR #2410, opened 2020-04-07
Refactor (training cfgs): change default training settings for better…

ZwwWayne:
This PR changes the default training settings in MMDetection. Since V2.0, the codebase will have the following changes.

1. Change `nms_post` from 2000 to 1000 in Mask/Faster RCNN for better performance.
2. Change to use L1 loss rather than smooth l1 loss for regression in Mask/Faster RCNN for better performance.
3. Does not use gradient clip by default for a faster training process.
4. Change the warmup ratio from 1/3.0 to 1/1000 for a more stable training process after removing the gradient clip.
5. Use RoIAlign V2.0 for better-aligned feature pooling.
6. Change default sample_num in RoIAlign from 2 to 0.

A more detailed ablation of these changes will be in the documentation.


configs/_base_/schedules/schedule_1x.py
```diff
@@ -1,11 +1,11 @@
 # optimizer
 optimizer = dict(type='SGD', lr=0.02, momentum=0.9, weight_decay=0.0001)
-optimizer_config = dict(grad_clip=dict(max_norm=35, norm_type=2))
+optimizer_config = dict(grad_clip=None)
```
xvjiarui, line 3: We may also do the same for `schedule_2x.py` and `schedule_20e.py`.

configs/_base_/schedules/schedule_1x.py
```diff
@@ -1,11 +1,11 @@
...
 # learning policy
 lr_config = dict(
     policy='step',
     warmup='linear',
     warmup_iters=500,
-    warmup_ratio=1.0 / 3,
+    warmup_ratio=1.0 / 1000,
```
hellock, line 9: `warmup_ratio = 0.001`

mdv3101:
@ZwwWayne Should we apply the same changes for Cascade Mask RCNN to improve the performance?

ZwwWayne:
Hi @mdv3101 ,
The nms_pre/nms_post/max_num decides the number of bounding boxes before/after NMS.

We tried the same changes on Cascade methods and they are not improved, which means that the improvements that those tricks could bring have already be taken in by the current cascade methods' setting.

mdv3101:
@ZwwWayne As far as i understood, `max_num` sets a threshold (limit) for the  number of bounding boxes in both cases i.e. before and after NMS. Am I right? I am still confused about the use of this hyper parameter.

Also I was doing an experiment (for single class) in which I set `nms_across_level=True` and increased `nms_pre=12000` in training and `nms_pre=6000` in testing configuration. After training the model, when i run the evaluation, the model was able to detect only one object in a dataset having nearly 150 ground truth objects. Also if I set the `nms_across_level=False` in the testing config, there are bounding boxes all over the image.
Can you comment on this behavior?

ZwwWayne:
Hi @mdv3101 ,
For `nms_pre`, it filters out bboxs according to their score before nms. `nms_post` is the number of boxes output by nms. `max_num` is to further reduce the number of boxes as you can find their use [here](https://github.com/open-mmlab/mmdetection/blob/7dbee18dee2f34ccc4aacdd9d747c6b7126cac45/mmdet/core/post_processing/bbox_nms.py#L69).

For these hyperparameters, I have not tune very much of them. Usually boxes across levels are not sent to nms together. If the bounding boxes are too many, you can set the threshold to filter them out and lower down the `nms_pre`.

mdv3101:
Thanks for your reply @ZwwWayne . I'll try to tune these hyper-parameters.

c1aris:
Hi @ZwwWayne, may I ask that why setting `match_low_quality=True` for rcnn in`mask_rcnn_r50_fpn` but not in `mask_rcnn_r50_caffe_c4`? Is it a feature or a bug?

mdv3101:
Hi @ZwwWayne , nms_pre is generally set to 12000, but in most of the config files in mmdetection it is set to 2000. Any particular reason behind it?

Issue #2407, opened 2020-04-07
How can I save the results of running test.py as an image?


marload:
I am using mmdetection in linux using ssh. test.py --show is not available because it is used by command line. How can I save this result as an image?

ZwwWayne:
You may save the pickle results first and plot them in another script.

hellock:
Or you may use the [high-level APIs](https://github.com/open-mmlab/mmdetection/blob/master/docs/GETTING_STARTED.md#high-level-apis-for-testing-images) to do it.

marload:
Thx!

Issue #2240, opened 2020-03-11
Testing trained model, no saved images


tiberium24:
Hello,

I am trying to run this:

`./tools/dist_test.sh custom_configs/cascade_rcnn_r50_fpn_1x.py work_dirs/cascade_rcnn_r50_fpn_1x/epoch_25.pth 4 --format_only --options "outfile_prefix=./cascade_rcnn_r50_fpn_1x_results" `

No images with detection results are however saved to any output folder. Do you know why? 
It would be nice to have the detections results on images saved.

suntaiwu:
+1

hellock:
It will not save images by default. The above command saves the detection results in a json file. You can either save it by yourself or visualize the results online by specifying `--show`.

Hemantr05:
You can try using High-level APIs for testing images, it saves the detection as an image.

Issue #1405, opened 2019-09-16
How to save the output images?


sicarioakki:
Hey @hellock @tjsongzw @lindahua 
Can anyone guide me as to how to save the output images with detections?
Like i want to save all the output images in a directory once the inference code is run

Sunnyheye:
@sicarioakki  
You can refer to this python file: mmdet/apis/inference.py -> show_result() , which can save the images of detection result to specified folder.

DecentMakeover:
@sicarioakki if your issue is solved, please close the issue.

hellock:
Feel free to reopen it if there is any further questions.

scut-salmon:
Hi I found that the image is reshaped when saved, anyway I can do to save the original size of image

YijianLiu:
have you saved this? thanks！

reubencmoyo:
> have you saved this? thanks！

How do I save a video after visualising it?

sisse008:
This should be reopened. There is no show_result() in version 2.4

justinkeke91:
The code below can show and save image with result. 
--------------------------------------------------------------------------------------------------------------------------------

import mmcv
import os
import cv2
import numpy as np
from mmcv.runner import load_checkpoint
from mmdet.models import build_detector
from mmdet.apis import init_detector, inference_detector, show_result_pyplot
import matplotlib.pyplot as plt


config_file = '/home/en_ro/mmdata/mmdetection/configs/balloon/faster_rcnn_r50_fpn_1x_voc_cat0602.py'
checkpoint_file =  '/home/en_ro/mmdetection/work_dirs/faster_rcnn_r50_fpn_1x_voc_cat0602/epoch_100.pth'

model = init_detector(config_file,checkpoint_file)

img_dir = '/home/en_ro/mmdata/valdata_1/'
out_dir = 'results/'

if not os.path.exists(out_dir):
    os.mkdir(out_dir)
test_path = 'jpg0615.txt'
fp = open(test_path, 'r')
test_list = fp.readlines()

count = 0
imgs = [] 
img = '017-01_0034_o_S017-1_0035.JPG'
result = inference_detector(model,img)

show_result_pyplot(model, img, result, score_thr=0.6)# show  the image with result
model.show_result(img, result, out_file='0617.jpg')# save image with  result

ahmetnihat:
# Note : Show_result_pyplot was removed after mmdet version 2.0. OpenCV imshow is used instead.

[...]

PR #2414, opened 2020-04-08
Add option to save the result images of running tools/test

daavoo:
This P.R. adds a new optional argument to `tools/test.py`: 

`--images_out_dir`

If specified, detection results will be plotted on the images and saved to the specified directory. It is only applicable to single GPU testing and used for debugging and visualization. You **don't** need a GUI available in your environment for using this option.

This should be an alternative solution to some issues requesting this feature (#2407 #2240 #1405)

The `show_result` method of `BaseDetector` has been refactored to use `mmdet.apis.inference`. I'm not sure if this was the right approach because there was a TODO comment regarding the merge.


docs/GETTING_STARTED.md
```diff
@@ -43,23 +44,31 @@ python tools/test.py configs/faster_rcnn_r50_fpn_1x.py \
...
+4. Test Mask R-CNN with 8 GPUs, and evaluate the bbox and mask AP.
 
 ```shell
 ./tools/dist_test.sh configs/mask_rcnn_r50_fpn_1x.py \
     checkpoints/mask_rcnn_r50_fpn_1x_20181010-069fa190.pth \
     8 --out results.pkl --eval bbox segm
 ```
 
-4. Test Mask R-CNN on COCO test-dev with 8 GPUs, and generate the json file to be submit to the official evaluation server.
+5. Test Mask R-CNN on COCO test-dev with 8 GPUs, and generate the json file to be submit to the official evaluation server.
```
hellock, line 71: The following number should also be updated: 5 -> 6.

mmdet/models/detectors/base.py
```diff
@@ -148,11 +146,14 @@ def forward(self, img, img_metas, return_loss=True, **kwargs):
...
-        else:
-            bbox_result, segm_result = result, None
+    def show_result(self,
+                    data,
+                    result,
+                    dataset=None,
+                    score_thr=0.3,
+                    show=False,
+                    images_out_dir=None):
+        from mmdet.apis.inference import show_result as show_single_result
```
hellock, line 156: `mmdet.apis` contains high-level APIs exposed to basic users, and it should not be relied on by other modules. Different models may have different visualization methods. It would be better to call `detector.show_result()` in the inference API.

daavoo: I refactored `mmdet.apis` to use `detector.show_result` method.

I'm not sure if what I did was the right approach but, having in mind your point (Different models may have different visualization methods.), I:

- Made  `detector.show_result` expect a single image as input (instead of previous `data` dict). I could have added some conditionals and make it work with both single image and data but this felt cleaner. 
- Refactor single_gpu_test to adapt to the single image interface.

hellock:
Thanks for your contribution! The v2.0 branch has been updated. Could you rebase the PR?

daavoo:
> Thanks for your contribution! The v2.0 branch has been updated. Could you rebase the PR?

It should be updated now. But build is (randomly?, previous commit passes) failing with the following error:

`AttributeError: module 'mmdet.ops.grid_sampler.grid_sampler_ext' has no attribute 'grid_sampler_forward'`

Which also occurs in the lates v2.0 build:
https://travis-ci.com/github/open-mmlab/mmdetection/jobs/319300767

docs/GETTING_STARTED.md
```diff
@@ -113,15 +122,15 @@ model = init_detector(config_file, checkpoint_file, device='cuda:0')
 img = 'test.jpg'  # or img = mmcv.imread(img), which will only load it once
 result = inference_detector(model, img)
 # visualize the results in a new window
-show_result(img, result, model.CLASSES)
+model.module.show_result(img, result)
```
hellock, line 125: `model.module` is unnecessary because model is not a DataParallel or DistributedDataParallel.

mmdet/apis/inference.py
```diff
@@ -216,7 +146,7 @@ def show_result_pyplot(img,
         out_file (str, optional): If specified, the visualization result will
             be written to the out file instead of shown in a window.
     """
-    img = show_result(
-        img, result, class_names, score_thr=score_thr, show=False)
+    img = model.module.show_result(
+        img, result, score_thr=score_thr, show=False)
```
hellock, line 150:
```python
if hasattr(model, 'module'):
    model = model.module
img = model.show_result(img, result, score_thr=score_thr, show=False)
```

mmdet/apis/test.py
```diff
@@ -2,25 +2,45 @@
...
 import mmcv
 import torch
 import torch.distributed as dist
 from mmcv.runner import get_dist_info
 
+from mmdet.core import tensor2imgs
 
-def single_gpu_test(model, data_loader, show=False):
+
+def single_gpu_test(model, data_loader, show=False, images_out_dir=None):
```
hellock, line 15: I suggest using `out_dir` instead of `images_out_dir` for simplicity.

mmdet/apis/test.py
```diff
@@ -2,25 +2,45 @@
...
+            img_tensor = data['img'][0]
+            img_metas = data['img_metas'][0].data[0]
+            imgs = tensor2imgs(img_tensor, **img_metas[0]['img_norm_cfg'])
+            assert len(imgs) == len(img_metas)
+
+            for img, img_meta in zip(imgs, img_metas):
+                h, w, _ = img_meta['img_shape']
+                img_show = img[:h, :w, :]
+                out_file = '{}/{}'.format(images_out_dir,
+                                          Path(img_meta['filename']).name)
```
hellock, line 38: Use `out_file = str(Path(images_out_dir) / Path(img_meta['filename']).name)` in case that `images_out_dir` ends with a slash already.

mmdet/models/detectors/base.py
```diff
@@ -155,51 +155,51 @@ def forward(self, img, img_metas, return_loss=True, **kwargs):
...
             return self.forward_test(img, img_metas, **kwargs)
 
-    def show_result(self, data, result, dataset=None, score_thr=0.3):
+    def show_result(self,
+                    img,
+                    result,
+                    score_thr=0.3,
+                    wait_time=0,
+                    show=False,
+                    out_file=None):
```
hellock, line 164: Add a docstring here.

docs/GETTING_STARTED.md
```diff
@@ -159,9 +168,9 @@ async def main():
         result = await async_inference_detector(model, img)
 
     # visualize the results in a new window
-    show_result(img, result, model.CLASSES)
+    model.module.show_result(img, result)
     # or save the visualization results to image files
-    show_result(img, result, model.CLASSES, out_file='result.jpg')
+    model.module.show_result(img, result out_file='result.jpg')
```
hellock, line 173: `result` -> `result, `

mmdet/apis/test.py
```diff
@@ -2,25 +2,45 @@
...
     model.eval()
     results = []
     dataset = data_loader.dataset
     prog_bar = mmcv.ProgressBar(len(dataset))
     for i, data in enumerate(data_loader):
         with torch.no_grad():
-            result = model(return_loss=False, rescale=not show, **data)
+            result = model(
+                return_loss=False,
+                rescale=not (show or images_out_dir),
```
hellock, line 24: We can set `rescale=True` to get the results on the original image scale, and then resize the image to `ori_shape` for visualization after Line36.

hellock:
Yes there seems a random error in the CI, and we have not figured out the reason.

mmdet/apis/test.py
daavoo: Okey. That is a nice change. The current behaviour had caused me (and some others, I guess) problems in the past.

mmdet/apis/test.py
```diff
@@ -2,25 +2,43 @@
...
+            img_tensor = data['img'][0]
+            img_metas = data['img_metas'][0].data[0]
+            imgs = tensor2imgs(img_tensor, **img_metas[0]['img_norm_cfg'])
+            assert len(imgs) == len(img_metas)
+
+            for img, img_meta in zip(imgs, img_metas):
+                h, w, _ = img_meta['img_shape']
+                img_show = img[:h, :w, :]
+                out_file = '{}/{}'.format(out_dir,
+                                          Path(img_meta['filename']).name)
```
ZwwWayne, line 36: - `Path` seems to be unnecessary.
- The codebase prefers to use `import os.path as osp` in the header.
- ```out_file = osp.join(out_dir, img_meta['filename'])``` might be better.

mmdet/apis/test.py
```diff
@@ -2,25 +2,43 @@
 import pickle
 import shutil
 import tempfile
+from pathlib import Path
```
ZwwWayne, line 5: The codebase prefers to use `os.path` rather than `Path`.

tools/test.py
```diff
@@ -68,14 +68,17 @@ def parse_args():
         help='evaluation metrics, which depends on the dataset, e.g., "bbox",'
         ' "segm", "proposal" for COCO, and "mAP", "recall" for PASCAL VOC')
     parser.add_argument('--show', action='store_true', help='show results')
+    parser.add_argument(
+        '--out-dir',
```
ZwwWayne, line 72: It might be a little bit misleading since the `test.py` already have an option `--out` but `--out_dir` does totally different things. Would `--show_dir` be better?

mmdet/apis/test.py
daavoo: Copy that!

tools/test.py
daavoo: I guess that `show_dir` is a better option. I used images_out_dir befor but maybe it was too long.

mmdet/apis/inference.py
```diff
@@ -130,76 +128,8 @@ async def async_inference_detector(model, img):
...
-        score_thr=score_thr,
-        show=show,
-        wait_time=wait_time,
-        out_file=out_file)
-    if not (show or out_file):
-        return img
-
-
-def show_result_pyplot(img,
+def show_result_pyplot(model,
```
hellock, line 131: The docstring is not updated.

mmdet/apis/test.py
```diff
@@ -8,19 +8,36 @@
...
+            img_tensor = data['img'][0]
+            img_metas = data['img_metas'][0].data[0]
+            imgs = tensor2imgs(img_tensor, **img_metas[0]['img_norm_cfg'])
+            assert len(imgs) == len(img_metas)
+
+            for img, img_meta in zip(imgs, img_metas):
+                h, w, _ = img_meta['img_shape']
+                img_show = img[:h, :w, :]
+                img_show = mmcv.imresize(img_show,
+                                         img_meta['ori_shape'][:-1][::-1])
```
hellock, line 34: Explicitly specifying the shape may be helpful for users to understand the codes.
```python
h, w = img_meta['ori_shape'][:-1]
img_show = mmcv.imresize(img_show, (w, h))
```

mmdet/apis/test.py
```diff
@@ -8,19 +8,36 @@
...
+                out_file = osp.join(out_dir, img_meta['filename'])
```
hellock, line 35: If `out_dir` is None, an error will be raised.

mmdet/models/detectors/base.py
```diff
@@ -155,51 +155,85 @@ def forward(self, img, img_metas, return_loss=True, **kwargs):
...
+                    wait_time=0,
+                    out_file=None):
+        """Draw `result` over `img`.
+
+        Args:
+            img (str or Tensor): The image to be displayed.
+            result (Tensor or tuple): The results to draw over `img`
+                bbox_result or (bbox_result, segm_result).
+            score_thr (float, optional): Minimum score of bboxes to be shown.
+                Default 0.3
```
hellock, line 177: `Default: 0.3.`

hellock:
Could you fix the conflicts? It is ready to be merged then.

ZwwWayne:
Since the CI is down, I tried to compile this branch and CI locally, and it is fine. So this branch should be OK to be merged. @hellock

PR #2419, opened 2020-04-09
Add sampler and assigner registry

xvjiarui:
This PR added sampler and assigner registry.


mmdet/core/bbox/registry.py
```diff
@@ -0,0 +1,4 @@
+from mmdet.utils import Registry
+
+ASSIGNERS = Registry('assigner')
+SAMPLERS = Registry('sampler')
```
hellock, line 4: Rename the registies to `BBOX_ASSIGNERS` and `BBOX_SAMPLERS` to avoid ambiguity. There is also a registry for dataset sampler.

mmdet/core/anchor/anchor_target.py
```diff
@@ -1,6 +1,7 @@
 import torch
 
-from ..bbox import PseudoSampler, assign_and_sample, bbox2delta, build_assigner
+from ..bbox import assign_and_sample, bbox2delta, build_assigner
+from ..bbox.samplers.pseudo_sampler import PseudoSampler
```
hellock, line 4: `PseudoSampler` can also be imported from `..bbox`

mmdet/core/bbox/builder.py
```diff
@@ -0,0 +1,27 @@
+from mmdet.utils import build_from_cfg
+from .assigners.base_assigner import BaseAssigner
```
hellock, line 2: `from .assigners import BaseAssigner`

PR #2422, opened 2020-04-09
Add train_cfg/test_cfg in AnchorHead

xvjiarui:
This PR added `train_cfg` and `test_cfg` as class members in all anchor heads. And it also removed `cfg` in `loss` function.

### DIFF
--- a/mmdet/models/detectors/two_stage.py
+++ b/mmdet/models/detectors/two_stage.py
@@ -32,9 +32,9 @@ class TwoStageDetector(BaseDetector, RPNTestMixin):
 
         if rpn_head is not None:
             rpn_train_cfg = train_cfg.rpn if train_cfg is not None else None
-            rpn_head.update(train_cfg=rpn_train_cfg)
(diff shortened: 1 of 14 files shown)


mmdet/models/detectors/two_stage.py
```diff
@@ -31,6 +31,9 @@ def __init__(self,
             self.neck = builder.build_neck(neck)
 
         if rpn_head is not None:
+            rpn_train_cfg = train_cfg.rpn if train_cfg is not None else None
+            rpn_head.update(train_cfg=rpn_train_cfg)
+            rpn_head.update(test_cfg=test_cfg.rpn)
```
hellock, line 36: For a temporary solution, we can do this.
```python
rpn_train_cfg = train_cfg.rpn if train_cfg is not None else None
rpn_head_  = rpn_head.copy()
rpn_head_.update(train_cfg=rpn_train_cfg, test_cfg=test_cfg.rpn)
```
We may consider expose the `default_args` argument of `build_head()` in the future.

PR #2440, opened 2020-04-13
Merge non dist and dist train

daavoo:
I think that the `_dist_train` and `_non_dist_train` functions can be merged in order to reduce duplicate code and avoid potential divergences between the 2 modes.

It might be another reason that I'm not aware of to keep both functions separated  but after #2324 being merged I can't find one.


hellock:
Nice!

nnop:
How do you keep the logic not break without unit tests? Pure human review?

PR #2442, opened 2020-04-13
fix losses['loss_bbox'] bug

Gr4ph1c:
I think it is necessary to initialize 'losses['loss_bbox'] ', otherwise when 'pos_inds.any()' is False, losses will be missing the key('loss_bbox')

### DIFF
--- a/mmdet/datasets/xml_style.py
+++ b/mmdet/datasets/xml_style.py
@@ -6,7 +6,7 @@ import numpy as np
 
 from .custom import CustomDataset
 from .registry import DATASETS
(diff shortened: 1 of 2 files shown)


yhcao6:
Thanks for fixing the bug. In fact, we have fix this bug in v2.0. Please refer [here](https://github.com/open-mmlab/mmdetection/blob/a76bd6fd97791c41e4332b3b5552906e0d4b9d31/mmdet/models/bbox_heads/bbox_head.py#L141-L142).

PR #2446, opened 2020-04-13
Add configs and benchmarks.

yhcao6:


ZwwWayne:
Two things to remind:

1. RetinaNet should also use L1-loss in box regression for higher performance.

2. When adding readme for each config directory, the BibTeX should also be added like that in https://github.com/open-mmlab/mmdetection/pull/2438.

ZwwWayne:
Guided Anchoring methods need the gradient clip to stabilize the training.

Issue #2325, opened 2020-03-26
how to view AP for each class


Wenbin94:
in the latest version,how to view AP for each class

hellock:
https://github.com/open-mmlab/mmdetection/pull/2459

Issue #1903, opened 2020-01-01
How to show evaluation value of every category?


yuyijie1995:
Using the default mode,this evaluation value shows like this:
Average Precision  (AP) @[ IoU=0.50:0.95 \| area=   all \| maxDets=100 ] = 0.500
--
Average Precision  (AP) @[ IoU=0.50      \| area=   all \| maxDets=100 ] = 0.775
Average Precision  (AP) @[ IoU=0.75      \| area=   all \| maxDets=100 ] = 0.503
Average Precision  (AP) @[ IoU=0.50:0.95 \| area= small \| maxDets=100 ] = 0.312
Average Precision  (AP) @[ IoU=0.50:0.95 \| area=medium \| maxDets=100 ] = 0.410
Average Precision  (AP) @[ IoU=0.50:0.95 \| area= large \| maxDets=100 ] = 0.559
Average Recall     (AR) @[ IoU=0.50:0.95 \| area=   all \| maxDets=  1 ] = 0.516
Average Recall     (AR) @[ IoU=0.50:0.95 \| area=   all \| maxDets= 10 ] = 0.611
Average Recall     (AR) @[ IoU=0.50:0.95 \| area=   all \| maxDets=100 ] = 0.617
Average Recall     (AR) @[ IoU=0.50:0.95 \| area= small \| maxDets=100 ] = 0.428
Average Recall     (AR) @[ IoU=0.50:0.95 \| area=medium \| maxDets=100 ] = 0.562
Average Recall     (AR) @[ IoU=0.50:0.95 \| area= large \| maxDets=100 ] = 0.658
I want to get the every category's P and R . Is there a method to get the result except changing the code in coco_util.py ?

hellock:
https://mmdetection.readthedocs.io/en/latest/GETTING_STARTED.html#analyse-class-wise-performance

hsp2454:
Did you solve this problem?

Issue #1526, opened 2019-10-11
Claswise mAP


bond9779:
How to get the classwise mAP of a given dataset?

The below command gives the overall mAP:
`python3 tools/test.py --config/<config_name> checkpoint/<checkpoint_file_name> --eval bbox segm --show/--out`   

But is there a way to get the classwise mAP from this?

bond9779:
@hellock

ZwwWayne:
Hi @bond9779 ,

bond9779:
Thanks for the prompt reply @ZwwWayne :)
I will check the link provided by you

ZwwWayne:
> Thanks for the prompt reply @ZwwWayne :)
> I will check the link provided by you

Hi @bond9779 , I tried to run the script by myself and find that it does not show the class-wise mAP but shows other details that might help you to analyze.

So I added some code in the script and proposed a PR. You can check this link https://github.com/open-mmlab/mmdetection/pull/1549
It can show the class-wise mAP for both segm and bbox now by simply adding `--with_ap` option in the running command.

Hope it could help.

ZwwWayne:
The PR https://github.com/open-mmlab/mmdetection/pull/1549 is approved, so current master branch support to view classwise mAP now.

I am going to close this issue, feel free to reopen it if you have any further questions.

Cli98:
> The PR #1549 is approved, so current master branch support to view classwise mAP.
> 
> I am going to close this issue, feel free to reopen it if you have any further questions.

Why u need to PR this? coco provided this function.

PR #2459, opened 2020-04-15
V2.0 classwise

ZwwWayne:
Showing class-wise mAP could helps users better analyze the model and there are indeed some need according to https://github.com/open-mmlab/mmdetection/issues/2325, https://github.com/open-mmlab/mmdetection/issues/1903, https://github.com/open-mmlab/mmdetection/issues/1526, etc.

This PR adds the class-wise mAP evaluation on the COCO dataset back to the v2.0 branch after the https://github.com/open-mmlab/mmdetection/pull/2042. To use this API, add `--options "classwise=True"` when using the test script or add `classwise=True` in the evaluation config for training-time evaluation.

An example of testing Faster RCNN R50 is as the following picture.
image

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -59,7 +59,15 @@ python tools/test.py configs/pascal_voc/faster_rcnn_r50_fpn_1x_voc.py \
     8 --out results.pkl --eval bbox segm
 ```
 
-4. Test Mask R-CNN on COCO test-dev with 8 GPUs, and generate the json file to be submit to the official evaluation server.
+4. Test Mask R-CNN with 8 GPUs, and evaluate the **classwise** bbox and mask AP.
+
+```shell
+./tools/dist_test.sh configs/mask_rcnn_r50_fpn_1x_coco.py \
+    checkpoints/mask_rcnn_r50_fpn_1x_20181010-069fa190.pth \
+    8 --out results.pkl --eval bbox segm --options "classwise=True"
+```
+
+5. Test Mask R-CNN on COCO test-dev with 8 GPUs, and generate the json file to be submit to the official evaluation server.
 
 ```shell
 ./tools/dist_test.sh configs/mask_rcnn_r50_fpn_1x_coco.py \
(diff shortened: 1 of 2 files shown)


mmdet/datasets/coco.py
```diff
@@ -401,7 +403,38 @@ def evaluate(self,
...
+                        precision = precision[precision > -1]
+                        if precision.size:
+                            ap = np.mean(precision)
+                        else:
+                            ap = float('nan')
+                        results_per_category.append(
+                            ('{}'.format(nm['name']),
+                             '{:0.3f}'.format(float(ap * 100))))
+
+                    N_COLS = min(6, len(results_per_category) * 2)
```
xvjiarui, line 427: May rename to `num_cols`.

xvjiarui:
Should we support add this feature in coco dataset only or custom dataset?

mmdet/datasets/coco.py
ZwwWayne: sure

mmdet/datasets/coco.py
```diff
@@ -401,7 +403,38 @@ def evaluate(self,
...
+                    results_flatten = list(
+                        itertools.chain(*results_per_category))
+                    headers = ['category', 'AP'] * (N_COLS // 2)
+                    results_2d = itertools.zip_longest(
+                        *[results_flatten[i::N_COLS] for i in range(N_COLS)])
+                    table_data = [headers]
+                    table_data += [result for result in results_2d]
+                    table = AsciiTable(table_data)
+                    print(table.table)
```
hellock, line 436: `print_log()` may be used.

mmdet/datasets/coco.py
```diff
@@ -401,7 +403,40 @@ def evaluate(self,
...
+                        itertools.chain(*results_per_category))
+                    headers = ['category', 'AP'] * (num_columns // 2)
+                    results_2d = itertools.zip_longest(*[
+                        results_flatten[i::num_columns]
+                        for i in range(num_columns)
+                    ])
+                    table_data = [headers]
+                    table_data += [result for result in results_2d]
+                    table = AsciiTable(table_data)
+                    print_log(table.table)
```
hellock, line 438: `print_log(msg, logger=logger)`

docs/GETTING_STARTED.md
```diff
@@ -79,6 +79,14 @@ You will get two json files `mask_rcnn_test-dev_results.bbox.json` and `mask_rcn
 
 The generated png and txt would be under `./mask_rcnn_cityscapes_test_results` directory.
 
+6. Test Mask R-CNN with 8 GPUs, and evaluate the bbox and mask AP.
```
hellock, line 82: - This example may be put between 3 and 4
- `evaluate the **classwise** bbox and mask AP`

mytk2012:
According to your command line，we must apply dist_test.sh(added options) to reach classwise，while we use test.py,what can we do?

PR #2488, opened 2020-04-20
Add bypass_duplicate option in optimizer constructor

xvjiarui:
This PR did the following items.
1. Fixed config files that use `paramwise_options` to `paramwise_cfg`.
2. Added `bypass_duplicate` in `paramwise_cfg` to avoid adding the same parameter group twice.
3. Added `prefix` in `add_params`.


tests/test_config.py
```diff
@@ -54,6 +54,9 @@ def test_config_build_detector():
             test_cfg=config_mod.test_cfg)
         assert detector is not None
 
+        optimizer = build_optimizer(detector, config_mod.optimizer)
+        assert optimizer is not None
```
hellock, line 58: assert optimizer is an instance of torch.optim.Optimizer

mmdet/core/optimizer/default_constructor.py
```diff
@@ -93,6 +117,10 @@ def add_params(self, params, module):
             if not param.requires_grad:
                 params.append(param_group)
                 continue
+            if bypass_duplicate and not self._is_disjoint(params, param_group):
+                warnings.warn('{} is duplicate. It is skipped since bypass '
```
hellock, line 121: `bypass_`

PR #2509, opened 2020-04-22
Make 'lr' optional parameter

Mxbonn:
While probably 90+ % of the PyTorch optimizers have `lr` rate as field some exceptions exist.
I for example am often using the "Learned Optimizer List" optimizer (https://github.com/google-research/google-research/tree/master/opt_list), which does not have `lr`as field.
(I'm already preparing my code base for an upgrade to v2.0 and this is one of the things that work in v1.x but not in v2)


hellock:
Thanks!

PR #2516, opened 2020-04-23
Add Scripts for Regression benchmark

ZwwWayne:
This PR adds a script to filter out configs for the regression benchmark.

As the fast development of mmdetection, the regression benchmark is sometimes necessary to ensure the performance and compatibility of models. This PR adds the script to filter out necessary model configs for benchmark under different refactoring progress.


.dev_scripts/benchmark_filter.py
```diff
@@ -0,0 +1,116 @@
...
+    benchmark_configs = []
+    for cfg_root in benchmark_type:
+        cfg_dir = osp.join(config_dpath, cfg_root)
+        configs = os.scandir(cfg_dir)
+        for cfg in configs:
+            config_path = osp.join(cfg_dir, cfg.name)
+            if config_path in benchmark_pool:
+                benchmark_configs.append(config_path)
+
+    print(f'Totally find {len(benchmark_configs)} configs to benchmark')
```
hellock, line 110: find -> found

PR #2536, opened 2020-04-25
more consistent assigner

Johnson-Wang:
The assign function in every assigner is more consistent with the new label indexing (0-based) method. 

-  doc strings are modified for the consistency
- base, atss, max_iou, approx_max_iou and point assigners are all modified 
- the background_label will be passed to the assign function at all Anchorhead when it is called.


ZwwWayne:
Can you merge the newest V2.0 and see whether the changes pass the CI?

mmdet/core/bbox/assigners/approx_max_iou_assigner.py
```diff
@@ -63,19 +62,20 @@ def assign(self,
                approxs_per_octave,
                gt_bboxes,
                gt_bboxes_ignore=None,
-               gt_labels=None):
+               gt_labels=None,
+               background_label=-1):
```
ZwwWayne, line 66: It is a bit risky to enable the users to set `background_label` since in most of cases it is -1.
So let's just fill the default `assigned_labels` as -1 and do not allow the users to set it.

PR #2537, opened 2020-04-26
Convert code for ONNX export

drcut:
1. Convert the dynamic shapes into the static shapes
2. Convert some ops that are not supported by ONNX


drcut:
I will add more code for NMS convert and request for a merge as a whole

PR #2548, opened 2020-04-27
Fix loss_bbox in fovea_head.py when num_pos=0

anurag1paul:
During multi-GPU training, when num_pos == 0 on line 216 in fovea_head.py, it causes loss_bbox to be a tensor vector of shape [1] while when num_pos > 0, it causes loss_bbox to be a scalar of shape [], causing the gather function to fail. 

This pull request fixes this bug by unsqueezing loss_bbox when num_pos > 0.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -3,6 +3,6 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = setuptools
 known_first_party = mmdet
-known_third_party = asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
+known_third_party = PIL,asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
 no_lines_before = STDLIB,LOCALFOLDER
 default_section = THIRDPARTY
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -3,6 +3,6 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = setuptools
 known_first_party = mmdet
-known_third_party = PIL,asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
+known_third_party = asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
 no_lines_before = STDLIB,LOCALFOLDER
 default_section = THIRDPARTY
--- a/mmdet/models/anchor_heads/fovea_head.py
+++ b/mmdet/models/anchor_heads/fovea_head.py
@@ -223,6 +223,7 @@ class FoveaHead(nn.Module):
                 pos_bbox_targets,
                 pos_weights,
                 avg_factor=num_pos)
+            loss_bbox = loss_bbox.unsqueeze(0)
         else:
             loss_bbox = torch.tensor([0],
                                      dtype=flatten_bbox_preds.dtype,
--- a/mmdet/models/anchor_heads/fovea_head.py
+++ b/mmdet/models/anchor_heads/fovea_head.py
@@ -223,9 +223,8 @@ class FoveaHead(nn.Module):
                 pos_bbox_targets,
                 pos_weights,
                 avg_factor=num_pos)
-            loss_bbox = loss_bbox.unsqueeze(0)
         else:
-            loss_bbox = torch.tensor([0],
+            loss_bbox = torch.tensor(0,
                                      dtype=flatten_bbox_preds.dtype,
                                      device=flatten_bbox_preds.device)
         return dict(loss_cls=loss_cls, loss_bbox=loss_bbox)
--- a/mmdet/models/anchor_heads/fovea_head.py
+++ b/mmdet/models/anchor_heads/fovea_head.py
@@ -224,9 +224,9 @@ class FoveaHead(nn.Module):
                 pos_weights,
                 avg_factor=num_pos)
         else:
-            loss_bbox = torch.tensor(0,
+            loss_bbox = torch.tensor([0],
                                      dtype=flatten_bbox_preds.dtype,
-                                     device=flatten_bbox_preds.device)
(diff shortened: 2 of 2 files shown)


.isort.cfg
```diff
@@ -3,6 +3,6 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = setuptools
 known_first_party = mmdet
-known_third_party = asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
+known_third_party = PIL,asynctest,cityscapesscripts,cv2,matplotlib,mmcv,numpy,onnx,pycocotools,robustness_eval,roi_align,roi_pool,seaborn,six,terminaltables,torch,torchvision
```
ZwwWayne, line 6: It seems that we usually do not use PIL in our codebase, may remove this.

anurag1paul: Should we remove it from runtime requirements as well? 
https://github.com/open-mmlab/mmdetection/blob/c40cafb1e1347645c1289634a40cc26b7b58163c/requirements/runtime.txt#L5

I committed using an environment generated using requirements files and pre-commit check forced me to add this to commit before it accepted the commit.

ZwwWayne: No, we cannot remove it from the requirements because of the potential incompatibility of torchvision and PILLOW versions.

ZwwWayne: But the mmdet code indeed does not use PIL, it is wierd.

anurag1paul: I have removed PIL from .isort.cfg.

mmdet/models/anchor_heads/fovea_head.py
```diff
@@ -223,6 +223,7 @@ def loss(self,
                 pos_bbox_targets,
                 pos_weights,
                 avg_factor=num_pos)
+            loss_bbox = loss_bbox.unsqueeze(0)
```
ZwwWayne, line 226: Unsqueezing the `loss_bbox` when `num_pos>0` is a little bit risky since it is more commonly used. It seems that other detectors' losses also do not have such a pattern.
So maybe it is better and safer to squeeze the loss_bbox when `num_pos==0`?

anurag1paul: Squeezing when num_pos==0 would make loss a scalar, which sometimes causes pytorch to throw a warning "UserWarning: Was asked to gather along dimension 0, but all input tensors were scalars; will instead unsqueeze and return a vector."
Otherwise, I do not see why the author chose:
`loss_bbox = torch.tensor([0],
                                     dtype=flatten_bbox_preds.dtype,
                                     device=flatten_bbox_preds.device)`

and not
`loss_bbox = torch.tensor(0,
                                     dtype=flatten_bbox_preds.dtype,
                                     device=flatten_bbox_preds.device)`

ZwwWayne: However, in the codebases, the final losses are always a scalar. 
So we suggest changing the case`num_pos==0` and use either
```python
loss_bbox = torch.tensor([0], dtype=flatten_bbox_preds.dtype, device=flatten_bbox_preds.device).sum()
```
or 
```python
loss_bbox = torch.tensor(0, dtype=flatten_bbox_preds.dtype, device=flatten_bbox_preds.device)
```
Does this work for you?

anurag1paul: Sure, I will do that.

ZwwWayne: Thanks.

ZwwWayne:
Hi @anurag1paul ,
Thanks for your kind contribution. I notice that your latest commit changes the code from `loss_bbox = torch.tensor(0)` to `loss_bbox = torch.tensor([0]).sum()`, do you change it for any specific reason? For example, do you meet warning or bug when using `loss_bbox = torch.tensor(0)`?

anurag1paul:
With 'loss_bbox = torch.tensor(0)`, the build was Travis PR build was erroring out, so I tried your second suggestion and it was able to pass the Travis CI checks.

ZwwWayne:
> With 'loss_bbox = torch.tensor(0)`, the Travis PR build was erroring out, so I tried your second suggestion and it was able to pass the Travis CI checks.

OK, thank you.

anurag1paul:
> > With 'loss_bbox = torch.tensor(0)`, the Travis PR build was erroring out, so I tried your second suggestion and it was able to pass the Travis CI checks.
> 
> OK, thank you.
> But I check the CI, it seems that 'loss_bbox = torch.tensor(0)` cannot pass the CI only because of the format cannot pass yapf and it should works after you reformat the code?

Yeah, I checked. It does seem a yapf issue. I am having a lot of trouble committing as pre-commit keeps adding PIL to .isort.cfg. Can you help me in skipping "seed isort known_third_party"?

ZwwWayne:
> > > With 'loss_bbox = torch.tensor(0)`, the Travis PR build was erroring out, so I tried your second suggestion and it was able to pass the Travis CI checks.
> > 
> > 
> > OK, thank you.
> > But I check the CI, it seems that 'loss_bbox = torch.tensor(0)` cannot pass the CI only because of the format cannot pass yapf and it should works after you reformat the code?
> 
> Yeah, I checked. It does seem a yapf issue. I am having a lot of trouble committing as pre-commit keeps adding PIL to .isort.cfg. Can you help me in skipping "seed isort known_third_party"?

Add `-n` in your command should work. It seems that `xml_dataset` indeed use PIL, so you can also allow the pre-commit to add it. I missed that in the format comments.

anurag1paul:
> > > > With 'loss_bbox = torch.tensor(0)`, the Travis PR build was erroring out, so I tried your second suggestion and it was able to pass the Travis CI checks.
> > > 
> > > 
> > > OK, thank you.
> > > But I check the CI, it seems that 'loss_bbox = torch.tensor(0)` cannot pass the CI only because of the format cannot pass yapf and it should works after you reformat the code?
> > 
> > 
> > Yeah, I checked. It does seem a yapf issue. I am having a lot of trouble committing as pre-commit keeps adding PIL to .isort.cfg. Can you help me in skipping "seed isort known_third_party"?
> 
> Add `-n` in your command should work. It seems that `xml_dataset` indeed use PIL, so you can also allow the pre-commit to add it. I missed that in the format comments.

Thank you. I have added PIL to .isort-cfg to make life easier with pre-commit.

ZwwWayne:
Thanks very much.

openmmlab-v:
Hi @anurag1paul ！First of all, we want to express our gratitude for your significant PR in the mmdetection project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

PR #2553, opened 2020-04-27
Support class agnostic nms

Mxbonn:
Support class agnostic non maximum suppression, i.e. IoU thresholding happens over all bboxes, regardless of the predicted class.

I'm not sure if this is the best way to implement it. Ideally you want this to be a parameter that is defined within the `nms` dict in config files, but just like the `type` keyword it is not an actual parameter of the nms function. It is needed in the `batched_nms` function of which the parameters are not decided by a config buy whether it is called from rcnn, a bbox head, ...

(submitting this to the v2.0 branch as my code base already switched to this version)


mmdet/ops/nms/nms_wrapper.py
```diff
@@ -136,10 +136,14 @@ def batched_nms(bboxes, scores, inds, nms_cfg):
     Returns:
         tuple: kept bboxes and indice.
     """
-    max_coordinate = bboxes.max()
-    offsets = inds.to(bboxes) * (max_coordinate + 1)
-    bboxes_for_nms = bboxes + offsets[:, None]
     nms_cfg_ = nms_cfg.copy()
+    class_agnostic = nms_cfg_.pop('class_agnostic', False)
```
hellock, line 140: I suggest adding `class_agnostic` as an argument of `batched_nms()`, with the default value False.

Mxbonn: So `class_agnostic = nms_cfg_.pop('class_agnostic', class_agnostic)` where `class_agnostic` is an argument of `batched_nms()`?
Or would you get rid of putting it in the `nms_cfg`? Because I would like to keep it there as well.

yhcao6:
Currently when `class_agnostic`, the predicted bboxes will be expanded to alight with `num_classes`. So it is still necessary to offset the bboxes with max_coordinates. 
https://github.com/open-mmlab/mmdetection/blob/672fd5de50a16cda1553e176f6038518fadf4c6e/mmdet/core/post_processing/bbox_nms.py#L32-L35

Mxbonn:
@yhcao6, I'm not sure if I completely follow. Do you mean that I still need to offset it so that the background class (class 0) is skipped?

Mxbonn:
@hellock, @yhcao6 I addressed both of your comments.

Mxbonn:
Nevermind I dont think this works right, I'm a bit confused as to where the background filtering happens.
What exactly is does:
```python
 if multi_bboxes.shape[1] > 4: 
     bboxes = multi_bboxes.view(multi_scores.size(0), -1, 4) 
 else: 
     bboxes = multi_bboxes[:, None].expand(-1, num_classes, 4) 
``` 
do?

Mxbonn:
Okay. It now work as expected. Ideally there should be unit tests for nms. Still confused by @yhcao6 comment about having to expand with max_coordinates as the background class is already filtered  before `batched_nms` is called.

yhcao6:
> Nevermind I dont think this works right, I'm a bit confused as to where the background filtering happens.
> What exactly is does:
> 
> ```python
>  if multi_bboxes.shape[1] > 4: 
>      bboxes = multi_bboxes.view(multi_scores.size(0), -1, 4) 
>  else: 
>      bboxes = multi_bboxes[:, None].expand(-1, num_classes, 4) 
> ```
> 
> do?
> 
> Also https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/post_processing/bbox_nms.py#L16-L17
> Mentions that the 0th column is the background class but this is no longer correct since V2.0 right?

Yes, there is something wrong with the doc. In v2.0 the last column corresponds to the background class instead of the first column.

Mxbonn:
Okay, thought so, I added a commit that corrects this.

yhcao6:
> @yhcao6, I'm not sure if I completely follow. Do you mean that I still need to offset it so that the background class (class 0) is skipped?

For your better understanding, you can refer to the old `multiclass_nms`.
If a detector is agnostic, for example, the Retinanet, its bbox prediction is in shape (N, 4)
the for loop iterates over every positive class and doing nms separately, there are two important things: 
https://github.com/open-mmlab/mmdetection/blob/699cb9142328a8f2ecdb6ffa8a9c600bbd6659e0/mmdet/core/post_processing/bbox_nms.py#L34-L63

1. The predicted bbox is used repeatedly by every class 
2. The nms is done inside for-loop --> inside a class, the bboxes of different loop index will not suppress each other.

To satisfy the above criteria, the current `multiclass_nms` 
https://github.com/open-mmlab/mmdetection/blob/6603790b67d182fb7828b82ac519f3f0d1086231/mmdet/core/post_processing/bbox_nms.py#L32-L35.
firstly expanded predicted bbox in the shape of (N, 4) to (N, #num_classes, 4) so that each column corresponds to one class which corresponds to the criteria 1; In batched nms, we give a shift to the expanded bbox so that the bboxes in different column will not suppress each other which corresponds to criteria 2.

By the way, I git clone your repo and test a trained retinanet, I set the `class_agnostic=False`, it will yield the wrong result.
Or perhaps I misunderstand your pr but we should rename the parameter `class_agnostic` in your pr to another name, otherwise it will cause confusion, and you should give an explanation to clarify its usage in the docstring.

yhcao6:
> Okay, thought so, I added a commit that corrects this.

Thanks for the fixing.

Mxbonn:
@yhcao6 you're misunderstanding my pull request.
With class agnostic nms I mean what you call criteria 2.

> 2. The nms is done inside for-loop --> inside a class, the bboxes of different loop index will not suppress each other.

In class agnostic nms, bboxes of different indices CAN suppress each other.
While it might be confusing this is what is commonly called class agnostic nms (see also https://github.com/tensorflow/models/blob/master/research/object_detection/core/post_processing.py#L612)
Which is also how I defined it in the docstring:
```
        class_agnostic (bool): if true, nms is class agnostic,
            i.e. IoU thresholding happens over all bboxes,
            regardless of the predicted class
```

The class-agnostic part of retinanet has a different meaning, i.e. it means that the predicted bounding box coordinates are not class specific. Which is the same behavior for SSD.

See an example below to clarify:
`class_agnostic=False` (current behavior) for SSD
image
`class_agnostic=True`
image

Same can be seen with retinanet:
`class_agnostic=False` 
image
`class_agnostic=True`
image

Mxbonn:
@yhcao6 is it clear now?

yhcao6:
> @yhcao6 is it clear now?

Hi, mate, sorry for the late reply, I am clear now, class agnostic here indicates that the NMS will across all bboxes without considering their classes. But if it is, you can directly call the function of NMS instead of batched NMS right?

Mxbonn:
That's right but the goal of this PR is that you can specify class agnostic in the config file and not have to change the underlying function calls.

PR #2560, opened 2020-04-28
fix unmap

yhcao6:


mmdet/models/anchor_heads/anchor_head.py
```diff
@@ -231,7 +231,8 @@ def _get_targets_single(self,
         # map up to original set of anchors
         if unmap_outputs:
             num_total_anchors = flat_anchors.size(0)
-            labels = unmap(labels, num_total_anchors, inside_flags)
+            labels = unmap(labels, num_total_anchors, inside_flags,
+                           self.num_classes)  # fill bg label
```
hellock, line 235: To make it more clear, `fill=self.num_classes`.

PR #2570, opened 2020-04-30
Support MomentumUpdaterHook

daavoo:
Requires to bump mmcv version to 0.4.4 in requirements.


hellock:
It seems that the commit history of the branch has diverged from v2.0. You may checkout a new branch from the latest v2.0 and raise a PR.

daavoo:
> It seems that the commit history of the branch has diverged from v2.0. You may checkout a new branch from the latest v2.0 and raise a PR.

#2571

PR #2601, opened 2020-05-04
Add score threshold

igonro:
Add score threshold parameter in order to filter out all the detections below this threshold.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -71,7 +71,7 @@ Optional arguments:
 - `EVAL_METRICS`: Items to be evaluated on the results. Allowed values depend on the dataset, e.g., `proposal_fast`, `proposal`, `bbox`, `segm` are available for COCO, `mAP`, `recall` for PASCAL VOC. Cityscapes could be evaluated by `cityscapes` as well as all COCO metrics.
 - `--show`: If specified, detection results will be plotted on the images and shown in a new window. It is only applicable to single GPU testing and used for debugging and visualization. Please make sure that GUI is available in your environment, otherwise you may encounter the error like `cannot connect to X server`.
 - `--show-dir`: If specified, detection results will be plotted on the images and saved to the specified directory. It is only applicable to single GPU testing and used for debugging and visualization. You **don't** need a GUI available in your environment for using this option.
-- `--show_score_thr`: If specified, detections with score below this threshold will be removed.
+- `--show-score-thr`: If specified, detections with score below this threshold will be removed.
 
 
 Examples:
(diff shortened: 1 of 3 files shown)


mmdet/apis/test.py
```diff
@@ -11,7 +11,11 @@
 from mmdet.core import tensor2imgs
 
 
-def single_gpu_test(model, data_loader, show=False, out_dir=None):
+def single_gpu_test(model,
+                    data_loader,
+                    show=False,
+                    out_dir=None,
+                    score_thr=0.3):
```
hellock, line 18: During testing, we adopt the score threshold specified in the config file. Here the threshold is only used for visualization, and the variable name `score_thr` can be misleading. Renaming it to `show_score_thr` would be better.

daavoo: I would then set the default to None and use model.test_cfg.score_thr as default in order to be consistent between results and visualization by default.

mmdet/apis/test.py
```diff
@@ -11,7 +11,11 @@
 from mmdet.core import tensor2imgs
 
 
-def single_gpu_test(model, data_loader, show=False, out_dir=None):
+def single_gpu_test(model,
+                    data_loader,
+                    show=False,
+                    out_dir=None,
+                    show_score_thr=0.3):
```
daavoo, line 18:
```suggestion
                    show_score_thr=None):
```

mmdet/apis/test.py
daavoo: In mmdet/models/detectors/base.py:

```python
score_thr=None,
# ...
if score_thr is None:
    score_thr = self.test_cfg.score_thr
```

hellock: A higher score is usually preferred for visualization and we hardly use the same threshold as in the config. If we use the same threshold, the plotted image will be messy.

docs/getting_started.md
```diff
@@ -71,6 +71,7 @@ Optional arguments:
 - `EVAL_METRICS`: Items to be evaluated on the results. Allowed values depend on the dataset, e.g., `proposal_fast`, `proposal`, `bbox`, `segm` are available for COCO, `mAP`, `recall` for PASCAL VOC. Cityscapes could be evaluated by `cityscapes` as well as all COCO metrics.
 - `--show`: If specified, detection results will be plotted on the images and shown in a new window. It is only applicable to single GPU testing and used for debugging and visualization. Please make sure that GUI is available in your environment, otherwise you may encounter the error like `cannot connect to X server`.
 - `--show-dir`: If specified, detection results will be plotted on the images and saved to the specified directory. It is only applicable to single GPU testing and used for debugging and visualization. You **don't** need a GUI available in your environment for using this option.
+- `--show_score_thr`: If specified, detections with score below this threshold will be removed.
```
hellock, line 74: `--show-score-thr`

PR #2608, opened 2020-05-04
Add gradient clip back in cfgs

ZwwWayne:
This PR adds gradient clip to some models to avoid the loss going to Nan.

### DIFF
--- a/configs/foveabox/fovea_align_r50_fpn_gn-head_4x4_2x_coco.py
+++ b/configs/foveabox/fovea_align_r50_fpn_gn-head_4x4_2x_coco.py
@@ -6,3 +6,5 @@ model = dict(
 # learning policy
 lr_config = dict(step=[16, 22])
 total_epochs = 24
+optimizer_config = dict(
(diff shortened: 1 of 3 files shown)


Kaeseknacker:
Hey @ZwwWayne , can you explain why you added gradient clipping to those specific config files? I also had problems with INF losses when training a RetinaNet on my own data with MMDetection v2.0. It took me some time to find the difference between the config files before and after version 2.0 and that missing gradient clipping was probably the problem for INF losses.  
Isn't gradient clipping always a good idea?

ZwwWayne:
> Hey @ZwwWayne , can you explain why you added gradient clipping to those specific config files? I also had problems with INF losses when training a RetinaNet on my own data with MMDetection v2.0. It took me some time to find the difference between the config files before and after version 2.0 and that missing gradient clipping was probably the problem for INF losses.
> Isn't gradient clipping always a good idea?

If the loss of your model easily goes to be Nan at the beginning of training (e.g., the first 2000 iters), you can try gradient clip to stabilize the training. INF loss is a bit tricky, you might need to check your data and your code.

PR #2612, opened 2020-05-05
Add ori_filename to img_metas and use it in test show-dir

daavoo:


mmdet/apis/test.py
```diff
@@ -35,7 +35,8 @@ def single_gpu_test(model, data_loader, show=False, out_dir=None):
                 img_show = mmcv.imresize(img_show, (ori_w, ori_h))
 
                 if out_dir:
-                    out_file = osp.join(out_dir, img_meta['filename'])
+                    out_file = osp.join(out_dir,
+                                        osp.basename(img_meta['filename']))
```
hellock, line 39: For different datasets, we may want to save the images differently. Sometimes the datasets are categorized into several subfolders, and we want to keep it when saving visualization results, and sometimes we just want to save all images in `out_dir`. An argument may be added to control the behavior.

daavoo: The problem is that the `img_prefix` is added to `filename` at:
https://github.com/open-mmlab/mmdetection/blob/88d042a1e4555cb263d8811b78a157415b93d233/mmdet/datasets/pipelines/loading.py#L19

So currently, it will try to save the result at `{out_dir}/{img_prefix}/{filename}` and fail.

daavoo:
Resolves #2659

mmdet/apis/test.py
hellock: I see. Thanks for the fix!

hellock:
Unit tests still fail. You need to 1. pull the master branch, 2. fix [this](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/inference.py#L55) by adding `ori_filename`.

PR #2625, opened 2020-05-06
Add python path in scripts

ZwwWayne:
Add python path in scripts.


tools/slurm_test.sh
```diff
@@ -12,6 +12,7 @@ CPUS_PER_TASK=${CPUS_PER_TASK:-5}
 PY_ARGS=${@:5}
 SRUN_ARGS=${SRUN_ARGS:-""}
 
+PYTHONPATH=$(cd "$(dirname $0)/..";pwd):$PYTHONPATH \
```
xvjiarui, line 15: We may extract `PYTHONPATH=$(cd "$(dirname $0)/..";pwd):$PYTHONPATH` above and pass `$PYTHONPATH` here.

PR #2629, opened 2020-05-06
Allow different x and y strides in anchor heads.

Mxbonn:
Allow non square input sizes and different strides in the x and y direction.


mmdet/core/anchor/anchor_generator.py
```diff
@@ -74,8 +77,9 @@ def __init__(self,
                 f'{strides} and {centers}'
 
         # calculate base sizes of anchors
-        self.strides = strides
-        self.base_sizes = list(strides) if base_sizes is None else base_sizes
+        self.strides = [_pair(stride) for stride in strides]
+        self.base_sizes = [min(stride) for stride in strides
```
ZwwWayne, line 81: We might need a new unit test to test the new behavior.

Mxbonn: @ZwwWayne , I added a new test that verifies that the new behavior is equivalent to the old when passing square strides.

If you want another test, please let me know what kind of test.

ZwwWayne: I think the test case when the strides are different might be also necessary since that is what this refactoring for.

Mxbonn: You're right. I added one that tests the behavior of the strides.

mmdet/core/anchor/anchor_generator.py
```diff
@@ -10,14 +11,16 @@ class AnchorGenerator(object):
...
         ratios (list[float]): The list of ratios between the height and width
             of anchors in a single level.
         scales (list[int] | None): Anchor scales for anchors in a single level.
             It cannot be set at the same time if `octave_base_scale` and
             `scales_per_octave` are set.
-        base_sizes (list[int] | None): The basic sizes of anchors in multiple
-            levels. If None is given, strides will be used as base_sizes.
+        base_sizes (list[int] | list[tuple[int, int]] | None): The basic sizes
+            of anchors in multiple levels.
+            If None is given, strides will be used as base_sizes.
```
hellock, line 23: Explain in the documentation that how strides are used as base_sizes (especially when it is a list of tuple).

mmdet/core/anchor/anchor_generator.py
```diff
@@ -295,14 +301,16 @@ def __init__(self,
                  strides,
                  ratios,
                  basesize_ratio_range,
-                 input_size=300,
+                 input_size=(300, 300),
```
hellock, line 304: For SSD, only square size is supported.

tests/test_anchor.py
```diff
@@ -19,6 +19,27 @@ def test_standard_anchor_generator():
...
+    assert torch.all(torch.eq(anchors[0], expected_anchors))
+
+    # Different strides in x and y direction
+    self = AnchorGenerator([(10, 20)], [1.], [1.], [10])
+    anchors = self.grid_anchors([(2, 2)], device='cpu')
+
+    expected_anchors = torch.tensor([[-5., -5., 5., 5.], [5., -5., 15., 5.],
+                                     [-5., 15., 5., 25.], [5., 15., 15., 25.]])
+
+    assert torch.all(torch.eq(anchors[0], expected_anchors))
```
hellock, line 40: `assert torch.equal(anchors[0], expected_anchors)`

PR #2657, opened 2020-05-08
compute pure inference time instead

yhcao6:
The original inference time includes the time of data loading, model forward and post-processing. For the models of instance segmentation, the post-processing time also includes the time of RLE encoding. This PR excludes the time of data loading and RLE encoding to compute the pure inference time.

### DIFF
--- a/mmcv
+++ b/mmcv
@@ -1,0 +1,1 @@
+/Users/yuhangcao/Desktop/mmcv/mmcv
--- a/mmcv
+++ b/mmcv
@@ -1,1 +1,0 @@
-/Users/yuhangcao/Desktop/mmcv/mmcv
--- a/mmdet/apis/test.py
+++ b/mmdet/apis/test.py
@@ -55,7 +55,7 @@ def single_gpu_test(model,
 
         torch.cuda.synchronize()
         elapsed = time.perf_counter() - start_time
-        batch_size = len(data['img_meta']._data)
+        batch_size = data['img'][0].size(0)
         prog_bar.update(completed=batch_size, elapsed_time=elapsed)
 
         # encode mask results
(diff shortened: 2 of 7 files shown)


mmdet/apis/test.py
```diff
@@ -81,17 +117,28 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
...
 
         if rank == 0:
-            batch_size = (
-                len(data['img_meta']._data)
-                if 'img_meta' in data else data['img'][0].size(0))
-            for _ in range(batch_size * world_size):
-                prog_bar.update()
+            torch.cuda.synchronize()
+            elapsed = time.perf_counter() - start_time
+            batch_size = data['img'][0].size(0)
```
ZwwWayne, line 132: We need to keep using to obtain the `batch_size` in case downstream repos' inputs do not have the key `img`.
```python
batch_size = (
                len(data['img_meta']._data)
                if 'img_meta' in data else data['img'][0].size(0))
```

mmcv
```diff
@@ -0,0 +1 @@
+/Users/yuhangcao/Desktop/mmcv/mmcv
```
hellock, line 1: This file should be removed.

tools/benchmark.py
```diff
@@ -0,0 +1,100 @@
...
+        wrap_fp16_model(model)
+    checkpoint = load_checkpoint(model, args.checkpoint, map_location='cpu')
+    if args.fuse_conv_bn:
+        model = fuse_module(model)
+    # old versions did not save class info in checkpoints, this walkaround is
+    # for backward compatibility
+    if 'CLASSES' in checkpoint['meta']:
+        model.CLASSES = checkpoint['meta']['CLASSES']
+    else:
+        model.CLASSES = dataset.CLASSES
```
hellock, line 63: `model.CLASSES` is unused.

tools/benchmark.py
```diff
@@ -0,0 +1,100 @@
...
+
+    model = MMDataParallel(model, device_ids=[0])
+
+    model.eval()
+    dataset = data_loader.dataset
+
+    # the first several iterations may be very slow so skip them
+    num_warmup = min(5, len(dataset) - 1)
```
hellock, line 71: We may assume that the dataset length is larger than 200, otherwise an error may be raised.

PR #2671, opened 2020-05-09
Close multiprocessing Pool

### LINKED ISSUE #2670: Memory leak?
Ever since using V2.0 I've noticed that my RAM memory keeps increasing every epoch. Which indicates that there is a memory leak somewhere.
Is someone else noticing the same behavior?

Mxbonn:
Closes multiprocessing pool.
Potentially resolves #2670.


hellock:
Thanks for the fix.

PR #2680, opened 2020-05-11
Make FSAF loss more robust to no gt

Johnson-Wang:
Changed the default num_pos to num_neg when num_pos = 0. Otherwise, the loss would be very high when gt is absent in the batch.

### DIFF
--- a/mmdet/models/dense_heads/fsaf_head.py
+++ b/mmdet/models/dense_heads/fsaf_head.py
@@ -241,11 +241,12 @@ class FSAFHead(RetinaHead):
             min_levels=argmin)
         num_pos = torch.cat(pos_inds, 0).sum().float()
         acc = self.calculate_accuracy(cls_scores, labels_list, pos_inds)
+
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/fsaf_head.py
```diff
@@ -239,12 +239,15 @@ def loss(self,
...
-        num_pos = torch.cat(pos_inds, 0).sum().float().clamp(min=1e-3)
+        num_pos = torch.cat(pos_inds, 0).sum().float()
         acc = self.calculate_accuracy(cls_scores, labels_list, pos_inds)
         for i in range(len(losses_cls)):
-            losses_cls[i] /= num_pos
-            losses_bbox[i] /= num_pos
+            if num_pos == 0:
+                avg_factor = num_pos + float(num_total_neg)
+            else:
+                avg_factor = num_pos
```
hellock, line 248: The computation of `avg_factor` can be moved out of the for-loop.

PR #2754, opened 2020-05-18
Fix bug in nms cpu implementation

Mxbonn:
The cpu implementation does not return the suppressed detection in order of confidence.
That means that when  `multiclass_nms` has a limit on `max_num` it will not necessarily keep the detections with the highest score.

I have changed the `nms_test`, and committed this first to show how the previous behaviour fails and version in this PR fixes this.

### DIFF
--- a/mmdet/ops/nms/src/cpu/nms_cpu.cpp
+++ b/mmdet/ops/nms/src/cpu/nms_cpu.cpp
@@ -39,8 +39,7 @@ at::Tensor nms_cpu_kernel(const at::Tensor& dets, const float threshold) {
 
   for (int64_t _i = 0; _i < ndets; _i++) {
     auto i = order[_i];
-    if (suppressed[i] == 1)
-      continue;
+    if (suppressed[i] == 1) continue;
     keep[num_to_keep++] = i;
     auto ix1 = x1[i];
     auto iy1 = y1[i];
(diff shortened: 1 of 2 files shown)


Mxbonn:
(I think the first commit passed because it just pulls the branch and the fix was already pushed, you can confirm previous behaviour by running it locally)

hellock:
> (I think the first commit passed because it just pulls the branch and the fix was already pushed, you can confirm previous behaviour by running it locally)

Yes unit tests will fail without the second commit. Thanks for the fix!

mmdet/ops/nms/src/cpu/nms_cpu.cpp
```diff
@@ -23,20 +23,25 @@ at::Tensor nms_cpu_kernel(const at::Tensor& dets, const float threshold) {
...
   auto y1 = y1_t.data_ptr<scalar_t>();
   auto x2 = x2_t.data_ptr<scalar_t>();
   auto y2 = y2_t.data_ptr<scalar_t>();
   auto areas = areas_t.data_ptr<scalar_t>();
 
+  int64_t num_to_keep = 0;
+
   for (int64_t _i = 0; _i < ndets; _i++) {
     auto i = order[_i];
-    if (suppressed[i] == 1) continue;
```
hellock, line 39: The code is formatted with clang based on the google style. We may just leave it as is.

PR #2758, opened 2020-05-19
support self-defined batch process function

z-bingo:
The self-defined batch process function can be used with this version. The default one is still mmdet.apis.train.batch_processor.


hellock:
Thanks for the PR! It is now configurable via `model.train_step()`

PR #2774, opened 2020-05-20
fix bug when background label is 0

FreyWang:
fix bug when background label is 0, pos and neg sample cannot be mined


FreyWang:
I found that the background label cannot be 0 when calculating mean AP，modify the code will be hard to read, the best way may be restrict background label to be the num_class, removing background argument in HEAD @ZwwWayne 
   
`mean_ap.py:`
```
    for i in range(num_classes):
        # get gt and det bboxes of this class
        cls_dets, cls_gts, cls_gts_ignore = get_cls_results(
            det_results, annotations, i)
```

PR #2780, opened 2020-05-20
Add loss EMA normalizer in Retinanet

ZwwWayne:
This PR adds loss EMA normalizer for RetinaNets to improve the performance.


mmdet/models/dense_heads/anchor_head.py
```diff
@@ -400,6 +401,12 @@ def loss(self,
             concat_anchor_list.append(torch.cat(anchor_list[i]))
         all_anchor_list = images_to_levels(concat_anchor_list,
                                            num_level_anchors)
+        if self.loss_normalizer != -1:
+            avg_factor = (
```
Johnson-Wang, line 405: It seems that the implementation is different with that in [detectron2](https://github.com/facebookresearch/detectron2/blob/bf11a9b5bbc8e80cfe713c547203576a77a09065/detectron2/modeling/meta_arch/retinanet.py#L237). `self.loss_normalizer` is not updated here. Am i missing something?

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -52,6 +52,8 @@ def __init__(self,
                      loss_weight=1.0),
                  loss_bbox=dict(
                      type='SmoothL1Loss', beta=1.0 / 9.0, loss_weight=1.0),
+                 loss_normalizer=-1,
```
Johnson-Wang, line 55: The meaning of `loss_normalizer` and `loss_normalizer_momentum` should be reflected in docstring.

mmdet/models/dense_heads/anchor_head.py
ZwwWayne: No, it seems that I write a bug here.

ZwwWayne: Fixed

configs/retinanet/README.md
```diff
@@ -24,3 +24,23 @@
...
 | X-101-64x4d-FPN | pytorch |   2x    |    -     |       -        |  40.8  | [model](https://open-mmlab.s3.ap-northeast-2.amazonaws.com/mmdetection/v2.0/retinanet/retinanet_x101_64x4d_fpn_2x_coco/retinanet_x101_64x4d_fpn_2x_coco_20200131-bca068ab.pth) &#124; [log](https://open-mmlab.s3.ap-northeast-2.amazonaws.com/mmdetection/v2.0/retinanet/retinanet_x101_64x4d_fpn_2x_coco/retinanet_x101_64x4d_fpn_2x_coco_20200131_114833.log.json) |
+
+### Some tricks
+
+We also support exponential momentum average (EMA) over the number of positive samples as loss normalizer to improve the performance. This trick is introduced in [Detectron2](https://github.com/facebookresearch/detectron2/blob/master/detectron2/modeling/meta_arch/retinanet.py#L103) to stabilize the training by reducing the variance of foreground number. Adding the following two lines in the head config could enable this option.
```
hellock, line 30: We may replace `master` with the specific commit id, since the line number may change with further commits.

configs/retinanet/README.md
```diff
@@ -24,3 +24,23 @@
...
+
+```python
+loss_normalizer=100,
```
hellock, line 33: `loss_normalizer` has been removed.

configs/retinanet/retinanet_r50_fpn_loss-ema-norm_1x_coco.py
```diff
@@ -0,0 +1,35 @@
...
+            type='AnchorGenerator',
+            octave_base_scale=4,
+            scales_per_octave=3,
+            ratios=[0.5, 1.0, 2.0],
+            strides=[8, 16, 32, 64, 128]),
+        bbox_coder=dict(
+            type='DeltaXYWHBBoxCoder',
+            target_means=[.0, .0, .0, .0],
+            target_stds=[1.0, 1.0, 1.0, 1.0]),
+        loss_normalizer=100,
```
hellock, line 23: `loss_normalizer` has been removed.

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -27,6 +27,8 @@ class AnchorHead(nn.Module):
             num_classes if None is given.
         loss_cls (dict): Config of classification loss.
         loss_bbox (dict): Config of localization loss.
+        loss_normalizer_momentum (float): momentum that used for EMA of
+            loss normalizer
```
hellock, line 31: Explain what it means and the default value.

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -91,6 +94,12 @@ def __init__(self,
             self.sampler = build_sampler(sampler_cfg, context=self)
         self.fp16_enabled = False
 
+        assert 0 <= loss_normalizer_momentum <= 1, \
+            '"loss_normalizer_momentum" should be in range [0, 1], ' \
+            f'got {loss_normalizer_momentum}'
+        self.loss_normalizer_momentum = loss_normalizer_momentum
+        self.register_buffer('loss_normalizer', torch.tensor(100))
```
hellock, line 101: Leave a comment for this hardcoded value.

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -52,7 +54,6 @@ def __init__(self,
                      loss_weight=1.0),
                  loss_bbox=dict(
                      type='SmoothL1Loss', beta=1.0 / 9.0, loss_weight=1.0),
-                 loss_normalizer=-1,
                  loss_normalizer_momentum=0.9,
```
Johnson-Wang, line 57: The default value of `loss_normalizer_momentum` should be 0.0, so that the default setting is still using the real-time `num_total_samples`

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -27,6 +27,13 @@ class AnchorHead(nn.Module):
             num_classes if None is given.
         loss_cls (dict): Config of classification loss.
         loss_bbox (dict): Config of localization loss.
```
Johnson-Wang, line 29: It's a bit too tedious for an arg description. We may just change it to something like
"The momentum factor (from 0-1) used for exponential moving average (EMA) loss normalizer. 
Default: 0, no EMA "

configs/retinanet/README.md
```diff
@@ -24,3 +24,22 @@
...
+We also support exponential momentum average (EMA) over the number of positive samples as loss normalizer to improve the performance. This trick is introduced in [Detectron2](https://github.com/facebookresearch/detectron2/blob/bf11a9b5bbc8e80cfe713c547203576a77a09065/detectron2/modeling/meta_arch/retinanet.py#L108) to stabilize the training by reducing the variance of foreground number. Adding the following two lines in the head config could enable this option.
+
+```python
+loss_normalizer_momentum=0.9,
+```
+
+To enable a fair comparison when benchmarking with Detectron2, we use it in the benchmark configs. However, we do not recommend it as a default option since so many methods have proposed based on the original baseline. We compare the performance of models before/after adding EMA loss normalizer in the following table.
+
+|    Backbone     |  Style  |Multi-scale Training|EMA Normalizer |Lr schd | Mem (GB) | Inf time (fps) | box AP | Download  |
+| :-------------: | :-----: |Multi-scale Training|:-----:  |:-----: | :------: | :------------: | :----: | :-------: |
```
hellock, line 39: `Multi-scale Training` -> `:-----:`

configs/retinanet/README.md
```diff
@@ -24,3 +24,22 @@
...
+
+### Some tricks
+
+We also support exponential momentum average (EMA) over the number of positive samples as loss normalizer to improve the performance. This trick is introduced in [Detectron2](https://github.com/facebookresearch/detectron2/blob/bf11a9b5bbc8e80cfe713c547203576a77a09065/detectron2/modeling/meta_arch/retinanet.py#L108) to stabilize the training by reducing the variance of foreground number. Adding the following two lines in the head config could enable this option.
+
+```python
+loss_normalizer_momentum=0.9,
+```
+
+To enable a fair comparison when benchmarking with Detectron2, we use it in the benchmark configs. However, we do not recommend it as a default option since so many methods have proposed based on the original baseline. We compare the performance of models before/after adding EMA loss normalizer in the following table.
```
ppwwyyxx, line 36: We actually didn't observe significant improvements from this option (maybe at most 0.2AP). So we feel comfortable adding it by default and consider it as a bugfix to the original poorly design normalizer (it can fix the poor performance of RetinaNet in 1im/GPU case).

I guess the improvements listed in this table may come from the original incorrect impl. pointed out at https://github.com/open-mmlab/mmdetection/pull/2780#discussion_r428435677 , which effectively changed the scale of loss.

ZwwWayne: Yes, this option indeed only brings about 0.2~0.3 AP improvement. The [bug version](https://github.com/open-mmlab/mmdetection/pull/2780#discussion_r428435677) actually brings more gains but we are not going to use that since it is not very reasonable and may not be generalizable to other models.

PR #2787, opened 2020-05-21
Add upsample_cfg support in FPN

Johnson-Wang:
Small changes. 
1. Added upsample config support in FPN. 
2. Support three sources of the extra convs: inputs, laterals and outs. These three connections all exist in different codebases. This is to support model conversion to mmdet.


mmdet/models/necks/fpn.py
```diff
@@ -34,6 +35,14 @@ class FPN(nn.Module):
         norm_cfg (dict): Config dict for normalization layer. Default: None.
         act_cfg (str): Config dict for activation layer in ConvModule.
             Default: None.
+        upsample_cfg (dict): Config dict for interpolate layer.
+            Default: `dict(mode='nearest')`
+        extra_convs_source (str, optional): The source feature map of the extra
```
Johnson-Wang, line 40: `extra_convs_source` have some overlaps with `extra_convs_on_inputs`. But to keep compatible with previous configs, I have left the key `extra_convs_on_inputs` intact.

ZwwWayne:
Need unit tests to to test the API `upsample_cfg` and `extra_convs_source`.

tests/test_necks.py
```diff
@@ -0,0 +1,140 @@
...
+            feats[i].shape[2:] == outs[i].shape[2:]
+
+    # Tests for fpn with no extra convs (pooling is used instead)
+    fpn_model = FPN(
+        in_channels=in_channels,
+        out_channels=out_channels,
+        start_level=1,
+        add_extra_convs=False,
+        num_outs=5)
+    fpn_model(feats)
```
ZwwWayne, line 63: Maybe we need to further check the size of the output?

mmdet/models/necks/fpn.py
hellock: We can deprecate this option and remove it in future releases.

mmdet/models/necks/fpn.py
```diff
@@ -34,6 +35,14 @@ class FPN(nn.Module):
...
+            convs. `None` means it is specified by `extra_convs_on_inputs`.
+            Otherwise, only the following modes are allowed
+             'inputs': Last feature map of neck inputs (i.e. backbone feature).
+             'laterals':  Last feature map after lateral convs
+             'outputs': The last output feature map after fpn convs
```
hellock, line 45: I suggest using `input`, `lateral`, `output`.

mmdet/models/necks/fpn.py
```diff
@@ -63,7 +72,9 @@ def __init__(self,
                  no_norm_on_lateral=False,
                  conv_cfg=None,
                  norm_cfg=None,
-                 act_cfg=None):
+                 act_cfg=None,
+                 upsample_cfg=dict(mode='nearest'),
+                 extra_convs_source=None):
```
hellock, line 77: This argument may be placed after `add_extra_convs`.

mmdet/models/necks/fpn.py
hellock: We may use a single argument `add_extra_convs` to specific behaviors.
- False
- `'on_input'`
- `'on_lateral'`
- `'on_output'`

mmdet/models/necks/fpn.py
Johnson-Wang: Removed extra_convs_source according to the last comment.

mmdet/models/necks/fpn.py
```diff
@@ -22,10 +22,18 @@ class FPN(nn.Module):
...
-        extra_convs_on_inputs (bool): Whether to apply extra conv on
-            the original feature from the backbone. Default: False.
+        add_extra_convs (bool | str): If bool, it decides whether to add conv
+            layers on top of the original feature maps. Default to False.
+            If True, its actual mode is specified by `extra_convs_on_inputs`.
+            If str, it specifies the source feature map of the extra convs.
+            Only the following options are allowed
+             'on_input': Last feat map of neck inputs (i.e. backbone feature).
+             'on_lateral':  Last feature map after lateral convs.
+             'on_output': The last output feature map after fpn convs.
```
hellock, line 32: The format will not be correctly parsed in the doc.
```suggestion
            Only the following options are allowed
            
            - 'on_input': Last feat map of neck inputs (i.e. backbone feature).
            - 'on_lateral':  Last feature map after lateral convs.
            - 'on_output': The last output feature map after fpn convs.
```

hellock:
The module `necks` is missing in `api.rst`.

docs/api.rst
```diff
@@ -80,6 +80,11 @@ dense_heads
 .. automodule:: mmdet.models.dense_heads
     :members:
 
+necks
+^^^^^^^^^^^^
+.. automodule:: mmdet.models.necks
+    :members:
```
hellock, line 86: It may be moved before `dense_heads` following the conventional order.

mmdet/models/necks/fpn.py
```diff
@@ -22,10 +22,21 @@ class FPN(nn.Module):
...
             build the feature pyramid. Default: -1, which means the last level.
-        add_extra_convs (bool): Whether to add conv layers on top of the
-            original feature maps. Default: False.
-        extra_convs_on_inputs (bool): Whether to apply extra conv on
-            the original feature from the backbone. Default: False.
+        add_extra_convs (bool | str): If bool, it decides whether to add conv
+            layers on top of the original feature maps. Default to False.
+            If True, its actual mode is specified by `extra_convs_on_inputs`.
+            If str, it specifies the source feature map of the extra convs.
+            Only the following options are allowed
```
hellock, line 29: The docstring should follow the rst syntax. An empty line is needed before list, but not needed between items of the list.

mmdet/models/necks/fpn.py
```diff
@@ -22,10 +22,19 @@ class FPN(nn.Module):
...
+
+            - 'on_input': Last feat map of neck inputs (i.e. backbone feature).
+            - 'on_lateral':  Last feature map after lateral convs.
+            - 'on_output': The last output feature map after fpn convs.
+        extra_convs_on_inputs (bool, deprecated): Whether to apply extra convs
```
hellock, line 34: Since this argument is deprecated, we can update existing configs that are using it.

Johnson-Wang: Ok, by the way, I just want to make sure that even though this argument is deprecated, we should not delete it directly. Only changing the configs would suffice. Am I correct?

hellock: Yes. We would like to switch to the new argument in later versions, but also provide backward compatibility for existing forks and projects.

PR #2793, opened 2020-05-22
fix bug for coco person subset

xinghaochen:
('person') will be parsed as string.
It should be a tuple


ZwwWayne:
Hi @xinghaochen ,
Thanks for your bug fix, could you reformat the config to pass the CI? We will merge this PR as long as the code format passes the CI.

xinghaochen:
I have reformated the config, but one of the CI passed and another one is still queued.

PR #2797, opened 2020-05-22
Add cls weight in CrossEntropyLoss

ZwwWayne:
This PR adds `cls_weight` in CrossEntropyLoss to meet the needs when users want to specify the class weight in CrossEntropyLoss.


mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -6,9 +6,14 @@
 from .utils import weight_reduce_loss
 
 
-def cross_entropy(pred, label, weight=None, reduction='mean', avg_factor=None):
+def cross_entropy(pred,
+                  label,
+                  weight=None,
+                  reduction='mean',
+                  avg_factor=None,
+                  cls_weight=None):
```
hellock, line 14: `class_weight` is more clear and remains simple.

tests/test_loss.py
```diff
@@ -0,0 +1,31 @@
+import pytest
```
hellock, line 1: Rename this file to `test_losses.py`.

PR #2800, opened 2020-05-22
Support Img fields during transform

ZwwWayne:
This PR adds `img_fields` in loading during transform, which enables the similar transformation for multiple images.

### DIFF
--- a/tests/test_pipelines/test_transform.py
+++ b/tests/test_pipelines/test_transform.py
@@ -10,10 +10,12 @@ from mmdet.datasets.builder import PIPELINES
 
 
 def test_resize():
+    # test assertion if img_scale is a list
     with pytest.raises(AssertionError):
         transform = dict(type='Resize', img_scale=[1333, 800], keep_ratio=True)
         build_from_cfg(transform, PIPELINES)
 
+    # test assertion if len(img_scale) while ratio_range is not None
     with pytest.raises(AssertionError):
         transform = dict(
             type='Resize',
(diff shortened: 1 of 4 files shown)


tests/test_pipelines/test_transform.py
```diff
@@ -0,0 +1,163 @@
...
+        multiscale_mode='value',
+        keep_ratio=False)
+    resize_module = build_from_cfg(transform, PIPELINES)
+    results = resize_module(results)
+    assert np.equal(results['img'], results['img2']).all()
+    assert results['img_shape'] == (800, 1280, 3)
+
+
+def test_flip():
+    with pytest.raises(AssertionError):
```
hellock, line 64: Add a comment for each exception.

roachsinai:
Hi @ZwwWayne could you please show me a case that `the similar transformation for multiple images` is needed?

What's more, if the `img_fields` has multiple images, for loop here:

https://github.com/open-mmlab/mmdetection/blob/809f1b1d01004cb2e8c0ab16a3e93d3cb8f4cc66/mmdet/datasets/pipelines/transforms.py#L665-L670

shows each image has a corresponding `offset_h` and `offset_w`, but in for loop below:

https://github.com/open-mmlab/mmdetection/blob/809f1b1d01004cb2e8c0ab16a3e93d3cb8f4cc66/mmdet/datasets/pipelines/transforms.py#L681-L684

shows all bboxes's offset is same with only last image.

So, the second question is: are all bboxes's offset is set by the last image of `img_fileds`? If it set by design, again could you show a case about it?

Thanks a lot!

PR #2810, opened 2020-05-25
Fixed bugs in transform.py

Johnson-Wang:
The following two bugs are fixed in transform.py
1. In `RandomCrop`, it returns None if either of these two conditions are satisfied
    a) There is no gt bboxes in the cropped area. (wanted)
    b) There is no gt bboxes in the original image. (unwanted).
2. 'gt_bboxes_ignore', 'gt_label_ignore' and 'gt_masks_ignore' are not considered in `RandomCrop`, `MinIoURandomCrop` and `Expand` classes. 
unittests for random crop and min iou random crop are also written.


Johnson-Wang:
Sorry that I forgot to squash the commits before this pr, and there seems to be many commits. But only two files are changed.

ZwwWayne:
We can also add one of config in nas_fpn in [test_config.py::test_config_data_pipeline](https://github.com/open-mmlab/mmdetection/blob/master/tests/test_config.py#L102) and see whether the whole pipeline could pass.

Johnson-Wang:
If the crop does not have overlap with any bboxes, the RandomCrop object will return None. Therefore there are chances that the nas-fpn config will fail the unittest in normal training. Therefore, I gave up nas-fpn unittest in this PR. I have tested it locally and it is okay.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -373,35 +373,42 @@ def __call__(self, results):
...
         for key in results.get('seg_fields', []):
             results[key] = results[key][crop_y1:crop_y2, crop_x1:crop_x2]
 
-        # filter out the gt bboxes that are completely cropped
-        if 'gt_bboxes' in results:
-            gt_bboxes = results['gt_bboxes']
-            valid_inds = (gt_bboxes[:, 2] > gt_bboxes[:, 0]) & (
-                gt_bboxes[:, 3] > gt_bboxes[:, 1])
-            # if no gt bbox remains after cropping, just skip this image
-            if not np.any(valid_inds):
```
Johnson-Wang, line 395: In the original implementation, this function returns None if either of these two conditions are satisfied

There is no gt bboxes in the cropped area. (wanted)
There is no gt bboxes or there. (unwanted).

mmdet/datasets/pipelines/transforms.py
```diff
@@ -373,35 +373,42 @@ def __call__(self, results):
...
+        # if no gt bbox remains after cropping, just skip this image
+        if 'bbox_fields' in results and not valid_flag:
+            return None
 
         # crop semantic seg
         for key in results.get('seg_fields', []):
             results[key] = results[key][crop_y1:crop_y2, crop_x1:crop_x2]
 
-        # filter out the gt bboxes that are completely cropped
-        if 'gt_bboxes' in results:
```
Johnson-Wang, line 390: gt_bboxes_ignore is not considered in the original implementation.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -373,35 +373,42 @@ def __call__(self, results):
...
-                gt_bboxes[:, 3] > gt_bboxes[:, 1])
-            # if no gt bbox remains after cropping, just skip this image
-            if not np.any(valid_inds):
-                return None
-            results['gt_bboxes'] = gt_bboxes[valid_inds, :]
-            if 'gt_labels' in results:
-                results['gt_labels'] = results['gt_labels'][valid_inds]
-
-            # filter and crop the masks
-            if 'gt_masks' in results:
```
Johnson-Wang, line 402: gt_masks_ignore is not considered as well

mmdet/datasets/pipelines/transforms.py
```diff
@@ -614,12 +622,14 @@ def __init__(self, min_ious=(0.1, 0.3, 0.5, 0.7, 0.9), min_crop_size=0.3):
         self.min_crop_size = min_crop_size
 
     def __call__(self, results):
-        img, boxes, labels = [
```
Johnson-Wang, line 617: gt_bboxes_ignore not considered here.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -567,23 +574,24 @@ def __call__(self, results):
...
                 int(h * ratio), int(w * ratio), top, left)
 
-        # not tested
-        if 'gt_semantic_seg' in results:
-            assert self.seg_ignore_label is not None
-            gt_seg = results['gt_semantic_seg']
+        # expand segs
+        for key in results.get('seg_fields', []):
+            gt_seg = results[key]
             expand_gt_seg = np.full((int(h * ratio), int(w * ratio)),
```
ZwwWayne, line 591: np.full can directly set `dtype=gt_seg.dtype` , so no need to use astype.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -373,35 +373,42 @@ def __call__(self, results):
...
+
+            # mask fields, e.g. gt_masks and gt_masks_ignore
+            mask_key = key.replace('bboxes', 'masks')
+            if mask_key in results:
+                results[mask_key] = results[mask_key][
+                    valid_inds.nonzero()[0]].crop(
+                        np.asarray([crop_x1, crop_y1, crop_x2, crop_y2]))
+
+        # if no gt bbox remains after cropping, just skip this image
+        if 'bbox_fields' in results and not valid_flag:
```
ZwwWayne, line 405: We can add a `TODO: check whether we can keep the image no matter how gt_bbox would be`.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -373,35 +382,43 @@ def __call__(self, results):
...
-            results[key] = bboxes
+            valid_inds = (bboxes[:, 2] > bboxes[:, 0]) & (
+                bboxes[:, 3] > bboxes[:, 1])
+            # When there is no gt bbox, cropping is conducted.
+            # When the crop is valid, cropping is conducted.
+            if len(valid_inds) == 0 or valid_inds.any():
+                valid_flag = True
+            results[key] = bboxes[valid_inds, :]
+            # label fields. e.g. gt_labels and gt_labels_ignore
+            label_key = key.replace('bboxes', 'labels')
```
hellock, line 402: Some keys may not contain the substring `bboxes`, such as `proposals`. In this case `label_key` is just `key` and the next two lines will result in errors.

Johnson-Wang: Keen eyes! I have added bbox2label and bbox2mask key correspondence in RandomCrop and MinIouRandomCrop classes. Please have a look.

Johnson-Wang:
By the way, should I update the models and training logs of nas-fpn?

PR #2860, opened 2020-05-31
Add procedure to remove compiled shared objects

TonyLianLong:
Since there are some shared objects in `mmdet` directory, we probably need to remove them to get a build that is the latest.


hellock:
Thanks!

PR #2935, opened 2020-06-08
Implement FCOS training tricks

Chrisfsj2051:
Implement another three training tricks (except for center_sampling) used in the official repo (https://github.com/tianzhi0549/FCOS). By jointly using them, mAP gains of up to 4.9 (with DCN) and 1.3 (without DCN) will be brought.

In this implementation, the following results are achieved:
without DCN: mAP=38.6 (0.1 lower than the official repo)
      with DCN: mAP=42.5 (0.2 higher than the official repo)


mmdet/models/dense_heads/fcos_head.py
```diff
@@ -456,7 +485,6 @@ def centerness_target(self, pos_bbox_targets):
         # only calculate pos centerness targets, otherwise there may be nan
         left_right = pos_bbox_targets[:, [0, 2]]
         top_bottom = pos_bbox_targets[:, [1, 3]]
-        centerness_targets = (
-            left_right.min(dim=-1)[0] / left_right.max(dim=-1)[0]) * (
-                top_bottom.min(dim=-1)[0] / top_bottom.max(dim=-1)[0])
+        centerness_targets = (left_right.min(dim=-1)[0] / left_right.max(dim=-1)[0]) \
+                             * (top_bottom.min(dim=-1)[0] / top_bottom.max(dim=-1)[0])
```
yhcao6, line 489: flake 8 error: `line too long (86 > 79 characters)`

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -80,26 +91,31 @@ def _init_layers(self):
...
                     chn,
                     self.feat_channels,
                     3,
                     stride=1,
                     padding=1,
-                    conv_cfg=self.conv_cfg,
+                    conv_cfg=conv_cfg,
                     norm_cfg=self.norm_cfg,
-                    bias=self.norm_cfg is None))
+                    bias=True))
```
yhcao6, line 108: Originally most of the time `bias` will be False since FCOS uses `GN` in the head. Could I ask if `bias` has a great effect on the performance?

yhcao6:
Thanks for your contribution, please check the comments!

Chrisfsj2051:
Hi @yhcao6 and @Johnson-Wang , thanks for your comments!

The origin settings of FCOS in mmdetection and the official repo are slightly different. For example, the author set `nms_th` as 0.6 and `grad clip` is not utilized, while in mmdetection the `nms_th` is set as 0.5 and `grad clip` is used. Also, the official repo use `bias=True` in conv layer while the previous implementation in mmdetection does. Here is the [link](https://github.com/tianzhi0549/FCOS/blob/master/fcos_core/modeling/rpn/fcos/fcos.py) to the official implementation.

I implement and train it by strictly following the author's setting, which may lead to differences from the previous implementation in mmdetection. I suggest to keep the `bias` setting for now. Also, I will conduct some experiments to see if it will give significant difference, which will take about two days.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -15,6 +16,10 @@ class FCOSHead(nn.Module):
     The FCOS head does not use anchor boxes. Instead bounding boxes are
     predicted at each pixel and a centerness measure is used to supress
     low-quality predictions.
+    Here norm_on_bbox, centerness_on_reg, dcn_on_last_conv are training
+    tricks used in official repo, which will bring remarkable mAP gains
```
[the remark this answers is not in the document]
Chrisfsj2051: It is given at the end of this paragraph.

Johnson-Wang:
Please check your ci test.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -109,32 +133,44 @@ def _init_layers(self):
...
-        for reg_layer in self.reg_convs:
-            reg_feat = reg_layer(reg_feat)
-        # scale the bbox_pred of different level
-        # float to avoid overflow when enabling FP16
-        bbox_pred = scale(self.fcos_reg(reg_feat)).float().exp()
-        return cls_score, bbox_pred, centerness
+        cls_scores = []
+        bbox_preds = []
+        centernesses = []
+        for i, x in enumerate(feats):
```
Johnson-Wang, line 150: Since there is no special operation on a specific fpn layer, is it possible to revert to `forward_single` implementation?

Johnson-Wang: Except that there would be another argument (stride) to be passed to `forward_single`.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -436,9 +474,9 @@ def _get_target_single(self, gt_bboxes, gt_labels, points, regress_ranges,
 
         # condition2: limit the regression range for each location
         max_regress_distance = bbox_targets.max(-1)[0]
-        inside_regress_range = (
-            max_regress_distance >= regress_ranges[..., 0]) & (
-                max_regress_distance <= regress_ranges[..., 1])
+        inside_regress_range = \
```
Johnson-Wang, line 477: The original implementation actually comply more with the format standard.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -456,7 +494,7 @@ def centerness_target(self, pos_bbox_targets):
         # only calculate pos centerness targets, otherwise there may be nan
         left_right = pos_bbox_targets[:, [0, 2]]
         top_bottom = pos_bbox_targets[:, [1, 3]]
-        centerness_targets = (
-            left_right.min(dim=-1)[0] / left_right.max(dim=-1)[0]) * (
-                top_bottom.min(dim=-1)[0] / top_bottom.max(dim=-1)[0])
+        centerness_targets = \
```
Johnson-Wang, line 497: Same comment with above

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -144,33 +144,32 @@ def init_weights(self):
...
-            if self.norm_on_bbox:
-                bbox_pred = F.relu(bbox_pred)
-                if not self.training:
-                    bbox_pred *= self.strides[i]
-            else:
-                bbox_pred = bbox_pred.exp()
+        return multi_apply(self.forward_single, feats, self.scales,
+                           self.strides)
+
+    def forward_single(self, x, scale, stride):
```
Johnson-Wang, line 150: docstring is needed here.

mmdet/models/dense_heads/fcos_head.py
Chrisfsj2051: Great idea! Already updated in the latest commit.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -474,9 +473,9 @@ def _get_target_single(self, gt_bboxes, gt_labels, points, regress_ranges,
 
         # condition2: limit the regression range for each location
         max_regress_distance = bbox_targets.max(-1)[0]
-        inside_regress_range = \
-            (max_regress_distance >= regress_ranges[..., 0]) \
-            & (max_regress_distance <= regress_ranges[..., 1])
+        inside_regress_range = (
```
Johnson-Wang: The following format may be better
```
inside_regress_range = (
    (max_regress_distance >= regress_ranges[..., 0]) 
    & (max_regress_distance <= regress_ranges[..., 1]))
```

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -15,6 +16,10 @@ class FCOSHead(nn.Module):
     The FCOS head does not use anchor boxes. Instead bounding boxes are
     predicted at each pixel and a centerness measure is used to supress
     low-quality predictions.
+    Here norm_on_bbox, centerness_on_reg, dcn_on_last_conv are training
+    tricks used in official repo, which will bring remarkable mAP gains
+    of up to 4.9. Please see https://github.com/tianzhi0549/FCOS for
+    more detail.
```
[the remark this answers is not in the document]
Johnson-Wang: FCOSHead is based on nn.Module. Therefore, I am afraid that all args should be described in the docstring. Most of them can be copied from elsewhere though.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -109,31 +133,50 @@ def _init_layers(self):
...
-        return multi_apply(self.forward_single, feats, self.scales)
+        return multi_apply(self.forward_single, feats, self.scales,
+                           self.strides)
+    """
+    Args:
+        x (torch.Tensor): FPN feature maps of the specified stride
+        scale (mmcv.cnn.Scale): Factor to resize the bbox prediction.
+        stride (int): The corresponding stride for feature maps, only
+            used to normalize the bbox prediction when self.norm_on_bbox
+            is True.
```
Johnson-Wang, line 156: Returns are also needed

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -109,31 +133,50 @@ def _init_layers(self):
...
         bias_cls = bias_init_with_prob(0.01)
         normal_init(self.fcos_cls, std=0.01, bias=bias_cls)
         normal_init(self.fcos_reg, std=0.01)
         normal_init(self.fcos_centerness, std=0.01)
 
     def forward(self, feats):
-        return multi_apply(self.forward_single, feats, self.scales)
+        return multi_apply(self.forward_single, feats, self.scales,
+                           self.strides)
+    """
```
Johnson-Wang, line 150: Is this docstring placed in a wrong line?

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -109,31 +133,50 @@ def _init_layers(self):
...
+    Args:
+        x (torch.Tensor): FPN feature maps of the specified stride
```
Johnson-Wang: x (Tensor)

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -109,31 +133,50 @@ def _init_layers(self):
...
+        scale (mmcv.cnn.Scale): Factor to resize the bbox prediction.
```
Johnson-Wang, line 153: scale (:obj: \`mmcv.cnn.Scale\`): Learnable scale module to resize the bbox prediction.

Chrisfsj2051:
> The main implementation seems fine to me now. There is another work to do here. The unittest of fcos head was missing. It would be better to add unittest gradually in the future. Could you please add the unittest of fcos? (Special attention should be paid to corner cases such as bbox_ignore, empty gt_bbox et al. )

No problem.

hyz-xmaster:
Hi @Chrisfsj2051, it's great to see the implementation of improved FCOS. I was interested the performance of FCOS if you just add the dconv2 layer to the last conv layers without using the dconv backbone. Thanks.

Chrisfsj2051:
> Hi @Chrisfsj2051, it's great to see the implementation of improved FCOS. I was interested the performance of FCOS if you just add the dconv2 layer to the last conv layers without using the dconv backbone. Thanks.

Sure, I will try this later.

Chrisfsj2051:
Hi @yhcao6 and @Johnson-Wang , I retrain FCOS w/ or w/o using `bias` in conv layer. Here's some results:
|DCN   |  bias  | mAP |
| :--:     |  :--:    | :-:     |
| N       |  N     | 38.6 |
| N       |  Y      | 38.7 |
| N       |  N     | not finished |
| N       |  Y     | not finished |

Since `bias` doesn't gives significant improvement for now, maybe we can keep the origion setting in mmdetection (i.e. `bias=self.norm_cfg is None`) ?

hyz-xmaster:
> > Hi @Chrisfsj2051, it's great to see the implementation of improved FCOS. I was interested the performance of FCOS if you just add the dconv2 layer to the last conv layers without using the dconv backbone. Thanks.
> 
> Sure, I will try this later.

Cool! Thank you.

Chrisfsj2051:
Hi @Johnson-Wang  and @yhcao6 , thanks for your advices. Now I use a extra argument `conv_bias` to manually choose using bias or not. Also, the unit test for `FCOS` is added.

By the way, I noticed that `gt_bboxes_ignore=None` is not actually used in the original implementation  when calculate loss, is it a expected behavior？

Chrisfsj2051:
> > > Hi @Chrisfsj2051, it's great to see the implementation of improved FCOS. I was interested the performance of FCOS if you just add the dconv2 layer to the last conv layers without using the dconv backbone. Thanks.
> > 
> > 
> > Sure, I will try this later.
> 
> Cool! Thank you.

Get mAP=39.9 without using DCN in backbone.

hyz-xmaster:
> > > > Hi @Chrisfsj2051, it's great to see the implementation of improved FCOS. I was interested the performance of FCOS if you just add the dconv2 layer to the last conv layers without using the dconv backbone. Thanks.
> > > 
> > > 
> > > Sure, I will try this later.
> > 
> > 
> > Cool! Thank you.
> 
> Get mAP=39.9 without using DCN in backbone.

Amazing. Thanks a lot.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -15,13 +16,25 @@ class FCOSHead(nn.Module):
...
+    of up to 4.9. Please see https://github.com/tianzhi0549/FCOS for
+    more detail.
+
+    Args:
+        norm_on_bbox (bool): If true, normalize the regression targets
+            with FPN strides.
+        centerness_on_reg (bool): If true, position centerness on the
+            regress branch. Please refer to https://github.com/tianzhi0549/FCOS/issues/89#issuecomment-516877042.
+        dcn_on_last_conv (bool): If true, use dcn in the last layer of
+            towers.
```
Johnson-Wang, line 30: docstring of other arguments is missing.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -33,6 +46,10 @@ def __init__(self,
                                  (512, INF)),
                  center_sampling=False,
                  center_sample_radius=1.5,
+                 norm_on_bbox=False,
+                 centerness_on_reg=False,
+                 dcn_on_last_conv=False,
+                 conv_bias=False,
```
Johnson-Wang, line 52: We may directly make `conv_bias`  accept the same options as that in [mmcv](https://github.com/open-mmlab/mmcv/blob/67a26da917c553561b93aa0fc75faedc70781027/mmcv/cnn/bricks/conv_module.py#L34)

Johnson-Wang:
Still, please have a check on the ci.

tests/test_heads.py
```diff
@@ -2,11 +2,73 @@
...
+    empty_box_loss = empty_gt_losses['loss_bbox']
+    assert empty_cls_loss.item() > 0, 'cls loss should be non-zero'
+    assert empty_box_loss.item() == 0, (
+        'there should be no box loss when there are no true boxes')
+
+    # When truth is non-empty then both cls and box loss should be nonzero for
+    # random inputs
+    gt_bboxes = [
+        torch.Tensor([[23.6667, 23.8757, 238.6326, 151.8874]]),
+    ]
```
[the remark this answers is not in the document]
Chrisfsj2051: Unit test doesn't include gt_bboxes_ignore test, since the implemenration of FCOS doesn't support gt_bboxes_ignore (passed as argument but not actually used).

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -80,26 +126,35 @@ def _init_layers(self):
...
             chn = self.in_channels if i == 0 else self.feat_channels
+            if self.dcn_on_last_conv and \
+                    i == self.stacked_convs - 1:
+                conv_cfg = dict(type='DCNv2')
+            else:
+                conv_cfg = self.conv_cfg
+            if self.conv_bias == 'auto':
+                bias = False if self.norm_cfg is None else True
+            else:
+                bias = self.conv_bias
```
Johnson-Wang, line 138: These four lines can be directly deleted and `self.conv_bias` can be sent to `ConvModule` directly. Please refer to https://github.com/open-mmlab/mmcv/blob/b87e774f66ff638587902bd0db238026d9549259/mmcv/cnn/bricks/conv_module.py#L91

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -80,26 +126,31 @@ def _init_layers(self):
         self.reg_convs = nn.ModuleList()
         for i in range(self.stacked_convs):
             chn = self.in_channels if i == 0 else self.feat_channels
+            if self.dcn_on_last_conv and \
+                    i == self.stacked_convs - 1:
```
hellock, line 130: Use parentheses instead of back slashes.
`if (self.dcn_on_last_conv and i == self.stacked_convs - 1)` and let yapf to format it.

mmdet/models/dense_heads/fcos_head.py
```diff
@@ -15,13 +16,48 @@ class FCOSHead(nn.Module):
...
+        regress_ranges (tuple[tuple[int, int]]): Regress range of multiple
+            level points.
+        center_sampling (bool): If true, use center sampling.
+        center_sample_radius (float): Radius of center sampling.
+        norm_on_bbox (bool): If true, normalize the regression targets
+            with FPN strides.
+        centerness_on_reg (bool): If true, position centerness on the
+            regress branch. Please refer to https://github.com/tianzhi0549/FCOS/issues/89#issuecomment-516877042.
+        dcn_on_last_conv (bool): If true, use dcn in the last layer of
+            towers.
```
hellock, line 41: Default values of several arguments are not described.

Chrisfsj2051:
Hi @hellock , thanks for the comments! 
Default values for some arguments have been specified currently. However, I didn't describe the other default values (regress_ranges, loss_cls, etc.) since they might be too long to be properly shown in docstring.

hedes1992:
Hello,
    Thank you very much for your great work.
    I just find some bugs(maybe) in the implementation which will not disturb the training process, but just the in-consistency...
    In detail, in [fcos_head.py](https://github.com/open-mmlab/mmdetection/blob/14b5d8eb426ebdf102a2906626f610e25e91866b/mmdet/models/dense_heads/fcos_head.py#L232), the distance2bbox function just calculate the predicting bbox and gt bbox in the original image. But if self.norm_on_bbox is True, like [here](https://github.com/open-mmlab/mmdetection/blob/14b5d8eb426ebdf102a2906626f610e25e91866b/mmdet/models/dense_heads/fcos_head.py#L456), 
the "pos_bbox_targets" is normed by corresponding stride, but the "pos_points" is in the original image.
    In this way, the "pos_decoded_target_preds" is not the true gt bboxes... But this will not cause error, because the loss_bbox here is IoULoss(which just use the relative offset between pred and gt to calculate loss).
    Is my worry is correct?

PR #2936, opened 2020-06-08
fix a bug of logger in base detector

GT9505:
This PR mainly solves a bug of logger in the base detector. In the init_weights function of base detector, if pretrained is not none, a new 'root' logger will be generated to record log information. This PR changes the 'root' logger to the default 'mmdet' logger.

### DIFF
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -7,6 +7,7 @@ import torch.nn as nn
 from mmcv.utils import print_log
 
 from mmdet.core import auto_fp16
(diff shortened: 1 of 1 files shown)


mmdet/models/detectors/base.py
```diff
@@ -7,6 +7,7 @@
 from mmcv.utils import print_log
 
 from mmdet.core import auto_fp16
+from mmdet.utils import get_root_logger
```
hellock, line 10: Merge this line with Line7

GT9505: Done

PR #2938, opened 2020-06-08
Deprecate DistOptimizerHook

Johnson-Wang:
DistOptimizerHook is deprecated as its methods are all the same with its base class

### DIFF
--- a/mmdet/core/utils/dist_utils.py
+++ b/mmdet/core/utils/dist_utils.py
@@ -42,15 +42,5 @@ def allreduce_grads(params, coalesce=True, bucket_size_mb=-1):
 
 
 class DistOptimizerHook(OptimizerHook):
-
-    def __init__(self, grad_clip=None, coalesce=True, bucket_size_mb=-1):
-        self.grad_clip = grad_clip
-        self.coalesce = coalesce
-        self.bucket_size_mb = bucket_size_mb
-
-    def after_train_iter(self, runner):
-        runner.optimizer.zero_grad()
-        runner.outputs['loss'].backward()
-        if self.grad_clip is not None:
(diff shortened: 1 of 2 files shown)


mmdet/core/utils/dist_utils.py
```diff
@@ -42,15 +42,5 @@ def allreduce_grads(params, coalesce=True, bucket_size_mb=-1):
...
-        self.bucket_size_mb = bucket_size_mb
-
-    def after_train_iter(self, runner):
-        runner.optimizer.zero_grad()
-        runner.outputs['loss'].backward()
-        if self.grad_clip is not None:
-            self.clip_grads(runner.model.parameters())
-        runner.optimizer.step()
+    """Deprecated optimizer hook for distributed training"""
+    pass
```
hellock, line 46: We may raise a warning.

mmdet/core/utils/dist_utils.py
```diff
@@ -42,15 +43,10 @@ def allreduce_grads(params, coalesce=True, bucket_size_mb=-1):
...
-    def after_train_iter(self, runner):
-        runner.optimizer.zero_grad()
-        runner.outputs['loss'].backward()
-        if self.grad_clip is not None:
-            self.clip_grads(runner.model.parameters())
-        runner.optimizer.step()
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            '"DistOptimizerHook" is deprecated, please switch to'
+            '"mmcv.runner.OptimizerHook".', DeprecationWarning)
```
hellock, line 51: `DeprecationWarning` is ignored by default, we may just use a `UserWarning` by leaving the second argument empty.

PR #2943, opened 2020-06-08
add RandomCenterCropPad

tianyuandu:


mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,188 @@ def __call__(self, results):
...
+            if self.pad_mode[0] in ['logical-or']:
+                inp_h = h | self.pad_mode[1]
+                inp_w = w | self.pad_mode[1]
+            elif self.pad_mode[0] in ['size_divisor']:
+                divisor = self.pad_mode[1]
+                inp_h = int(np.ceil(h / divisor)) * divisor
+                inp_w = int(np.ceil(w / divisor)) * divisor
+            else:
+                raise NotImplementedError(
+                    f'Not Implemented for mode: {self.pad_mode[0]}')
```
ZwwWayne, line 1109: Error message should also tell what is expected.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,188 @@ def __call__(self, results):
...
+    Args:
+        crop_size (tuple | None): expected size after crop, final size will
+            computed according to ratio. Requires (h, w) in train mode, and
+            None in test mode.
+        ratios (tuple): random select a ratio from tuple and crop image to
+            (crop_size[0] * ratio) * (crop_size[1] * ratio).
+            Only available in train mode.
+        border (int): max distance from center select area to image border.
+            Only available in train mode.
+        mean (tuple): mean value of dataset.
```
ZwwWayne, line 956: Could we directly use keys in `img_norm_cfg` by putting this module behind normalize? Or directly use the config's `img_norm_cfg`. If we cannot, we should add some remind here because these hyper-parameters are easy to forget but could affect the performance.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,188 @@ def __call__(self, results):
     def __repr__(self):
         repr_str = self.__class__.__name__ + f'(transforms={self.transforms})'
         return repr_str
+
+
+@PIPELINES.register_module()
+class RandomCenterCropPad(object):
+    """Random center crop and random around padding for CornerNet.
+
```
ZwwWayne, line 946: Add a detail description in Notes, e.g., how it works, what is the difference between `RandomCenterCropPad` and `RandomCrop` or other modules. A figure in docstring like below should also help (if I correctly understand how it works).

```python
"""
:code-block:
                   +----------------------------+
                    |                  padded area        |
    +--------|----------------------------|----------+
     |              |         cropped area                |                 |
     |              |         +---------------+        |      image |
     |              |         |        center  .     |         |                 |
     |              |         |        range         |         |                 |
     |              |         +---------------+        |                 |
    +--------|----------------------------|----------+
                    |                  padded area        |
                   +---------------------------+
"""
    

```

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,188 @@ def __call__(self, results):
...
+            boxes = results['gt_bboxes']
+            while True:
+                scale = random.choice(self.ratios)
+                new_h = int(self.crop_size[0] * scale)
+                new_w = int(self.crop_size[1] * scale)
+                h_border = self._get_border(self.border, h)
+                w_border = self._get_border(self.border, w)
+
+                for i in range(50):
+                    ctx = random.randint(low=w_border, high=w - w_border)
```
ZwwWayne, line 1045: ctx -> center_x

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,188 @@ def __call__(self, results):
...
+            if 'scale_factor' in results:
+                scale_factor = results['scale_factor']
+                img = mmcv.imresize(
+                    img, (int(w * scale_factor), int(h * scale_factor)),
+                    return_scale=False)
+                h, w, c = img.shape
+            results['img_shape'] = img.shape
+
+            if self.pad_mode[0] in ['logical-or']:
+                inp_h = h | self.pad_mode[1]
```
ZwwWayne, line 1101: What does `inp` stands for?

mmdet/datasets/pipelines/transforms.py
tianyuandu: I think it's better to do Crop and Resize before Normalize. If the original image has big resolution, the Normalize op will cost lots of time.

To solve this problem, we could set invalid default values to mean, std and to_rgb, and ensure they are assigned with new value in __init__ function.

mmdet/datasets/pipelines/test_time_aug.py
```diff
@@ -23,13 +23,21 @@ class MultiScaleFlipAug(object):
 
     def __init__(self,
                  transforms,
-                 img_scale,
+                 img_scale=None,
+                 ratio=None,
                  flip=False,
```
ZwwWayne, line 28: Modify docstring for the key `ratio`.

mmdet/datasets/pipelines/test_time_aug.py
```diff
@@ -23,13 +23,21 @@ class MultiScaleFlipAug(object):
...
-        self.img_scale = img_scale if isinstance(img_scale,
-                                                 list) else [img_scale]
-        assert mmcv.is_list_of(self.img_scale, tuple)
+        assert img_scale or ratio
+        if img_scale is not None:
+            self.img_scale = img_scale if isinstance(img_scale,
+                                                     list) else [img_scale]
+            self.scale_key = 'scale'
+            assert mmcv.is_list_of(self.img_scale, tuple)
+        else:
```
ZwwWayne, line 37: Can` img_scale` and `ratio` be both set? We may need assertion here.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,253 @@ def __call__(self, results):
...
+
+    This operation generate random cropped image from original image and pad
+    it simultaneously. Different from `RandomCrop`, the output shape may not
+    equal to `crop_size` strictly. We choose a random value from `ratios` and
+    the output shape could be bigger or smaller than `crop_size`. Also the pad
+    in this operation is different with `Pad`, actually we use around padding
+    instead of right-bottom padding.
+
+    The relation between output image (padding image) and original image:
+                    output image
```
ZwwWayne, line 955: Use `:code-block` for the plotted figure, see the syntax of sphinx.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +938,253 @@ def __call__(self, results):
...
+                                    'RandomCenterCropPad only supports bbox.')
+
+                    # crop semantic seg
+                    for key in results.get('seg_fields', []):
+                        raise NotImplementedError(
+                            'RandomCenterCropPad only supports bbox.')
+                    return results
+        else:  # test mode
+            if 'scale_factor' in results:
+                scale_factor = results['scale_factor']
```
ZwwWayne, line 1157: We should use `Resize` pipeline to do the resize process.

mmdet/datasets/pipelines/test_time_aug.py
```diff
@@ -24,18 +25,20 @@ class MultiScaleFlipAug(object):
     def __init__(self,
                  transforms,
                  img_scale=None,
-                 ratio=None,
+                 scale_factor=None,
                  flip=False,
                  flip_direction='horizontal'):
         self.transforms = Compose(transforms)
-        assert img_scale or ratio
+        assert img_scale or scale_factor
```
ZwwWayne, line 32: assert (img_scale is None) ^ (scale_factor is None)

mmdet/datasets/pipelines/transforms.py
```diff
@@ -28,7 +28,9 @@ class Resize(object):
     This transform resizes the input image to some scale. Bboxes and masks are
     then resized with the same scale factor. If the input dict contains the key
     "scale", then the scale in the input dict is used, otherwise the specified
-    scale in the init method is used.
+    scale in the init method is used. If the input dict contains the key
+    "scale_factor" (if MultiScaleFlipAug not given img_scale but given ratio),
```
ZwwWayne, line 32: if ``MultiScaleFlipAug`` does not give `img_scale` but `ratio`

mmdet/datasets/pipelines/transforms.py
```diff
@@ -28,7 +28,9 @@ class Resize(object):
...
+    the actual scale we used will computed by image shape and scale_factor.
```
ZwwWayne, line 33: the actual scale will be computed by image shape and scale_factor.

mmdet/datasets/pipelines/test_time_aug.py
```diff
@@ -29,8 +29,7 @@ def __init__(self,
                  flip=False,
                  flip_direction='horizontal'):
         self.transforms = Compose(transforms)
-        assert img_scale or scale_factor
-        assert None in [img_scale, scale_factor]
+        assert (img_scale is None) ^ (scale_factor is None)
```
ZwwWayne, line 32: Is `scale_factor` tested and observed gains in the corresponding model? As you can find in line 61, the result is shallow copied. If the scale factor is [0.8, 1, 1.2], the behavior might not be what we expected, i.e., the shape of the image (assumed to have size 512) might become `512*0.8`, `512*0.8*1`, and `512*0.8*1*1.2`, rather than `512*0.8`, `512*1`, and `512*1.2`.

tianyuandu: This code works as expected in my env (python3.6). Due to the shallow copy, `id(results['img']) == id(_results['img'])` , but `mmcv.imrescale` will return a new img with different id in `Resize` op, so I think the shallow copy won't lead to bad effect. I also added an unittest for this case.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,247 @@ def __call__(self, results):
...
+            self.mean = mean[::-1]
+        else:
+            self.mean = mean
+        self.test_mode = test_mode
+        assert pad_mode[0] in ['logical_or', 'size_divisor']
+        self.pad_mode = pad_mode
+
+    def _get_border(self, border, size):
+        i = 1
+        while size - border // i <= border // i:
```
ZwwWayne, line 1063: Can we simplify this function by setting `i = ((2*border/size)//2 + 1)*2`

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,245 @@ def __call__(self, results):
...
+        x1 = min(center_x + target_w // 2, img_w)
+        y0 = max(0, center_y - target_h // 2)
+        y1 = min(center_y + target_h // 2, img_h)
+        patch = np.array((int(x0), int(y0), int(x1), int(y1)))
+
+        left, right = center_x - x0, x1 - center_x
+        top, bottom = center_y - y0, y1 - center_y
+
+        cropped_center_y, cropped_center_x = target_h // 2, target_w // 2
+        if image.dtype == np.uint8:
```
ZwwWayne, line 1087: Why `np.float32` is necessary here? Could we 1. remove this constriant 2. Use `to_float32=True` in loading.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,245 @@ def __call__(self, results):
...
+
+            cropped_img, border, _ = self._crop_image_and_paste(
+                img, [h // 2, w // 2], [target_h, target_w])
+            results['img'] = cropped_img
+            results['pad_shape'] = cropped_img.shape
+            results['border'] = border
+            return results
+
+    def __repr__(self):
+        repr_str = self.__class__.__name__
```
ZwwWayne, line 1186: Update `__repr__` function to match the arguments.

tests/test_pipelines/test_transform.py
```diff
@@ -290,3 +290,166 @@ def test_albu_transform():
...
     assert results['img'].dtype == np.float32
+
+
+def test_random_center_crop_pad():
+    # test assertion for invalid random center crop pad
+    with pytest.raises(AssertionError):
+        transform = dict(type='RandomCenterCropPad', crop_size=(-1, 0))
+        build_from_cfg(transform, PIPELINES)
+
+    with pytest.raises(AssertionError):
```
ZwwWayne, line 301: Add a comment on what error you are testing.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,264 @@ def __call__(self, results):
...
+                raise NotImplementedError(
+                    'RandomCenterCropPad only support two testing pad mode:'
+                    'logical-or and size_divisor.')
+
+            cropped_img, border, _ = self._crop_image_and_paste(
+                img, [h // 2, w // 2], [target_h, target_w])
+            results['img'] = cropped_img
+            results['pad_shape'] = cropped_img.shape
+            results['border'] = border
+            return results
```
hellock, line 1199: This method is too long. You may split it into two.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,264 @@ def __call__(self, results):
...
+        return border // i
+
+    def _filter_boxes(self, patch, boxes):
+        center = (boxes[:, :2] + boxes[:, 2:]) / 2
+        mask = (center[:, 0] > patch[0]) * (center[:, 1] > patch[1]) * (
+            center[:, 0] < patch[2]) * (
+                center[:, 1] < patch[3])
+        return mask
+
+    def _crop_image_and_paste(self, image, center, size):
```
hellock, line 1087: Add a docstring.

hellock:
Many docstrings are missing.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,330 @@ def __call__(self, results):
...
+            cropped_center_y - top, cropped_center_y + bottom,
+            cropped_center_x - left, cropped_center_x + right
+        ],
+                          dtype=np.float32)
+
+        return cropped_img, border, patch
+
+    def _train_aug(self, results):
+        """Around padding the original image without cropping.
+        The padding mode and value are from `test_pad_mode`.
```
hellock, line 1159: The docstring is copied from _test_aug?

mmdet/datasets/pipelines/transforms.py
```diff
@@ -938,3 +948,330 @@ def __call__(self, results):
...
+                    raise NotImplementedError(
+                        'RandomCenterCropPad only supports bbox.')
+                return results
+
+    def _test_aug(self, results):
+        """Around padding the original image without cropping.
+        The padding mode and value are from `test_pad_mode`.
+
+        Args:
+            results (dict): Image infomations in the augment pipeline.
```
hellock, line 1229: Add a blank line after.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -172,7 +175,14 @@ def _resize_seg(self, results):
 
     def __call__(self, results):
         if 'scale' not in results:
```
hellock, line 177: You may assert `scale` and `scale_factor` cannot be both set.

hellock:
Looks good to me now.

hellock:
@tianyuandu Please fix the linting error.

Issue #2933, opened 2020-06-08
the asynchronous num_pos issue in FCOS


hyz-xmaster:
**Describe the feature**
Fix the [num_pos](https://github.com/open-mmlab/mmdetection/blob/63e2d65291ad0957466932ccce60be00cd3289f7/mmdet/models/dense_heads/fcos_head.py#L185) bug in FCOS.

**Motivation**
As [the author of FCOS was aware](https://github.com/tianzhi0549/FCOS#training) that

> Now, training with 8 GPUs and 4 GPUs can have the same performance. Previous performance gap was because we did not synchronize num_pos between GPUs when computing loss.

I have also ran into this problem that there is an obvious performance gap (about 1 mAP) when training the FCOS with 8GPUs and 4 GPUs with mmdetection.  The problem is probably caused by the same asynchronous num_pos issue.

**Related resources**
The author of FCOS has fixed the problem by synchronizing 'num_pos' between GPUs. Please see the bug fix commit: https://github.com/tianzhi0549/FCOS/commit/a62aeca5f31eaeda37cfaf7ab66d2e3d84abba32

Thank you.

yhcao6:
Thank you for the reminder. We would appreciate it if you would like to create a PR to fix this problem. I also notice that ATSS has already adopted reduced num positive across all GPUs. https://github.com/open-mmlab/mmdetection/blob/e574b8a65b5dbf5058dd371843bcb4ae817deed0/mmdet/models/dense_heads/atss_head.py#L205-L206

hyz-xmaster:
well, I'd like to make my first PR, but you have to wait for a moment, because I need to first learn how to create a PR...

yhcao6:
never mind.

PR #2946, opened 2020-06-08
fix the num_pos bug in fcos

hyz-xmaster:
synchronize num_pos in fcos_head and change requires_grad=True for fcos

### DIFF
--- a/mmdet/models/dense_heads/fcos_head.py
+++ b/mmdet/models/dense_heads/fcos_head.py
@@ -1,4 +1,5 @@
 import torch
+import torch.distributed as dist
 import torch.nn as nn
 from mmcv.cnn import ConvModule, Scale, bias_init_with_prob, normal_init
 
@@ -6,6 +7,14 @@ from mmdet.core import distance2bbox, force_fp32, multi_apply, multiclass_nms
 from ..builder import HEADS, build_loss
 
 INF = 1e8
+
+
+def reduce_mean(tensor):
+    if not (dist.is_available() and dist.is_initialized()):
+        return tensor
+    tensor = tensor.clone()
+    dist.all_reduce(tensor.div_(dist.get_world_size()), op=dist.ReduceOp.SUM)
+    return tensor
 
 
 @HEADS.register_module()
@@ -180,9 +189,15 @@ class FCOSHead(nn.Module):
         pos_inds = ((flatten_labels >= 0)
                     & (flatten_labels < bg_class_ind)).nonzero().reshape(-1)
         num_pos = len(pos_inds)
+
+        # sync num_pos across all gpus
+        num_pos_avg_per_gpu = reduce_mean(
+            pos_inds.new_tensor([pos_inds.numel()])).item()
+        num_pos_avg_per_gpu = max(num_pos_avg_per_gpu, 1.0)
+
         loss_cls = self.loss_cls(
             flatten_cls_scores, flatten_labels,
-            avg_factor=num_pos + num_imgs)  # avoid num_pos is 0
+            avg_factor=num_pos_avg_per_gpu)
 
         pos_bbox_preds = flatten_bbox_preds[pos_inds]
         pos_centerness = flatten_centerness[pos_inds]
@@ -194,16 +209,24 @@ class FCOSHead(nn.Module):
             pos_decoded_bbox_preds = distance2bbox(pos_points, pos_bbox_preds)
             pos_decoded_target_preds = distance2bbox(pos_points,
                                                      pos_bbox_targets)
+            # average pos_centerness_targets from all gpus,
+            # which is used to normalize centerness-weighed reg loss
+            pos_centerness_targets_avg_per_gpu = \
+                reduce_mean(pos_centerness_targets.sum()).item()
+
             # centerness weighted iou loss
             loss_bbox = self.loss_bbox(
                 pos_decoded_bbox_preds,
                 pos_decoded_target_preds,
                 weight=pos_centerness_targets,
-                avg_factor=pos_centerness_targets.sum())
-            loss_centerness = self.loss_centerness(pos_centerness,
-                                                   pos_centerness_targets)
+                avg_factor=pos_centerness_targets_avg_per_gpu)
+            loss_centerness = self.loss_centerness(
+                pos_centerness,
+                pos_centerness_targets,
+                avg_factor=num_pos_avg_per_gpu)
         else:
             loss_bbox = pos_bbox_preds.sum()
+            reduce_mean(pos_centerness.new_tensor([0.0]))
             loss_centerness = pos_centerness.sum()
 
         return dict(
--- a/mmdet/models/dense_heads/fcos_head.py
+++ b/mmdet/models/dense_heads/fcos_head.py
@@ -196,8 +196,7 @@ class FCOSHead(nn.Module):
         num_pos_avg_per_gpu = max(num_pos_avg_per_gpu, 1.0)
 
         loss_cls = self.loss_cls(
-            flatten_cls_scores, flatten_labels,
-            avg_factor=num_pos_avg_per_gpu)
+            flatten_cls_scores, flatten_labels, avg_factor=num_pos_avg_per_gpu)
 
         pos_bbox_preds = flatten_bbox_preds[pos_inds]
         pos_centerness = flatten_centerness[pos_inds]
--- a/configs/fcos/fcos_center_r50_caffe_fpn_gn-head_1x_coco.py
+++ b/configs/fcos/fcos_center_r50_caffe_fpn_gn-head_1x_coco.py
@@ -1,2 +1,2 @@
-_base_ = './fcos_r50_caffe_fpn_gn-head_4x4_1x_coco.py'
+_base_ = './fcos_r50_caffe_fpn_gn-head_1x_coco.py'
 model = dict(bbox_head=dict(center_sampling=True, center_sample_radius=1.5))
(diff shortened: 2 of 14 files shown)


hyz-xmaster:
This PR is linked to this [issue](https://github.com/open-mmlab/mmdetection/issues/2933)

configs/fcos/fcos_r50_caffe_fpn_4x4_1x_coco.py
```diff
@@ -12,7 +12,7 @@
         num_stages=4,
         out_indices=(0, 1, 2, 3),
         frozen_stages=1,
-        norm_cfg=dict(type='BN', requires_grad=False),
+        norm_cfg=dict(type='BN', requires_grad=True),
```
yhcao6, line 15: Why do you modify this?

yhcao6: The `samples_per_gpu` and `workers_per_gpu` should change to 2 now.

yhcao6:
Thanks for your PR to fix the issue.

configs/fcos/fcos_r50_caffe_fpn_4x4_1x_coco.py
hyz-xmaster: Recently, I have read the mmdetection paper and noticed that the requires_grad=True is the default setting in mmdetection. I thought that this requirese_grad being set as False for FCOS only was due to the async num_pos issue. So I changed requires_grad=True after fixing the num_pos bug.

**But**, after running an initial experiment, I found that setting requires_grad=True hurt the performance for over 1 mAP.    I am running another experiment with requires=False to see the performance. If the performance is good, that means setting requires_grad=True is not a good choice and I will change it back to False. Otherwise, I need to debug the code further.

hellock: It depends on the pretrained backbone you use. In the paper we use pytorch style backbones and pretrained weights from torchvision. However, `open-mmlab://detectron/resnet50_caffe` is not a normal pretrained weight, which fused conv and bn layers and the weight/bias of bn layers are 0 and 1. Thus we should not set `requires_grad=True`.

hyz-xmaster: That makes sense and my experiment results also agree with this. I will change it back and do some minor changes to the over-length lines.

hyz-xmaster:
Any idea of fixing this format[ issue](https://travis-ci.com/github/open-mmlab/mmdetection/jobs/346428268):

--- mmdet/models/dense_heads/fcos_head.py	(original)
+++ mmdet/models/dense_heads/fcos_head.py	(reformatted)
@@ -196,9 +196,7 @@
 ```
num_pos_avg_per_gpu = max(num_pos_avg_per_gpu, 1.0)
 
         loss_cls = self.loss_cls(
-            flatten_cls_scores,
-            flatten_labels,
-            avg_factor=num_pos_avg_per_gpu)
+            flatten_cls_scores, flatten_labels, avg_factor=num_pos_avg_per_gpu)
 
         pos_bbox_preds = flatten_bbox_preds[pos_inds]
         pos_centerness = flatten_centerness[pos_inds]
```
The command "yapf -r -d --style .style.yapf mmdet/ tools/ tests/ configs/" failed and exited with 1 during .

**Comment by yhcao6** [comment]
> Any idea of fixing this format[ issue](https://travis-ci.com/github/open-mmlab/mmdetection/jobs/346428268):
> 
> --- mmdet/models/dense_heads/fcos_head.py (original)
> +++ mmdet/models/dense_heads/fcos_head.py (reformatted)
> @@ -196,9 +196,7 @@
> 
> ```
> num_pos_avg_per_gpu = max(num_pos_avg_per_gpu, 1.0)
> 
>         loss_cls = self.loss_cls(
> -            flatten_cls_scores,
> -            flatten_labels,
> -            avg_factor=num_pos_avg_per_gpu)
> +            flatten_cls_scores, flatten_labels, avg_factor=num_pos_avg_per_gpu)
> 
>         pos_bbox_preds = flatten_bbox_preds[pos_inds]
>         pos_centerness = flatten_centerness[pos_inds]
> ```
> 
> The command "yapf -r -d --style .style.yapf mmdet/ tools/ tests/ configs/" failed and exited with 1 during .

The error tells you how to reformat the code.
You should change
```
loss_cls = self.loss_cls(
             flatten_cls_scores,
             flatten_labels,
             avg_factor=num_pos_avg_per_gpu)
```
to 
```
loss_cls = self.loss_cls(
       flatten_cls_scores, flatten_labels, avg_factor=num_pos_avg_per_gpu)
```

**Comment by hyz-xmaster** [comment]
Thanks for the help. After a few attempts, I still can't solve this [issue](https://travis-ci.com/github/open-mmlab/mmdetection/jobs/346492181), could you help with this problem?

**Comment by hyz-xmaster** [comment]
Excellent!

**Comment by hellock** [comment]
@yhcao6 will benchmark the new settings.

File: mmdet/models/dense_heads/fcos_head.py [hunk 2]
```diff
@@ -194,16 +208,24 @@ def loss(self,
...
-            loss_centerness = self.loss_centerness(pos_centerness,
-                                                   pos_centerness_targets)
+                avg_factor=pos_centerness_targets_avg_per_gpu)
+            loss_centerness = self.loss_centerness(
+                pos_centerness,
+                pos_centerness_targets,
+                avg_factor=num_pos_avg_per_gpu)
         else:
             loss_bbox = pos_bbox_preds.sum()
+            reduce_mean(pos_centerness.new_tensor([0.0]))
```
ZwwWayne, line 228: We might need to send 1 for reduce in case all GPUS have no positive sample?

hyz-xmaster: Well, I almost have no knowledge in DDP and just copied [that line](https://github.com/tianzhi0549/FCOS/blob/5fd728bfea150359fae28acb527e6e89d43c27e3/fcos_core/modeling/rpn/fcos/loss.py#L279) from the author of FCOS. So if you think it necessary to modify it, please feel free to do it.

yhcao6:
Hi, @hyz-xmaster , I benchmark 6 models with the new setting, but the performance of single scale schedule all drops. I will re-benchmark these 4 models using the current setting.
| Model            | 4gpu | 8gpu |
|------------------|------|------|
| R-50 1x          | 36.6 | 36.3 |
| R-50 2x          | 36.9 | 36.5 |
| R-101 1x         | 39.2 | 38.9 |
| R-101 2x         | 39.1 | 38.5 |
| R-50 2x mstrain  | 38.7 | 38.9 |
| R-101 2x mstrain | 40.9 | 41.0 |

hyz-xmaster:
> Hi, @hyz-xmaster , I benchmark 6 models with the new setting, but the performance of single scale schedule all drops. I will re-benchmark these 4 models using the current setting.
> 
> Model	4gpu	8gpu
> R-50 1x	36.6	36.3
> R-50 2x	36.9	36.5
> R-101 1x	39.2	38.9
> R-101 2x	39.1	38.5
> R-50 2x mstrain	38.7	38.9
> R-101 2x mstrain	40.9	41.0

That's a bit weird. I have only run the improved version of FCOS using 2 GPUs (8 imgs per gpu) and got 38.6 mAP, which is consistent with the author's result. Maybe one way of figuring this issue out is to run the author's implementation with different GPU settings to see if there is still a performance gap. If there is no such gap, that means something is wrong.

hyz-xmaster:
> Hi, @hyz-xmaster , I benchmark 6 models with the new setting, but the performance of single scale schedule all drops. I will re-benchmark these 4 models using the current setting.
> 
> Model	4gpu	8gpu
> R-50 1x	36.6	36.3
> R-50 2x	36.9	36.5
> R-101 1x	39.2	38.9
> R-101 2x	39.1	38.5
> R-50 2x mstrain	38.7	38.9
> R-101 2x mstrain	40.9	41.0

Hi @yhcao6, 
   I just compared FCOS configs with ATSS and RepPoints's, and found that only FCOS is still using Caffe style pre-trained weights and its BN.requires_grad=False. I am not sure how this will impact the training with different GPUs. Do you have any insight ? Or replace the Caffe style weights with PyTorch style weights to see the results? Thanks.

hyz-xmaster:
> Hi, @hyz-xmaster , I benchmark 6 models with the new setting, but the performance of single scale schedule all drops. I will re-benchmark these 4 models using the current setting.
> 
> Model	4gpu	8gpu
> R-50 1x	36.6	36.3
> R-50 2x	36.9	36.5
> R-101 1x	39.2	38.9
> R-101 2x	39.1	38.5
> R-50 2x mstrain	38.7	38.9
> R-101 2x mstrain	40.9	41.0

One interesting thing I just noticed is that in both of the implementations of RetinaNet in [maskrcnn benchmark](https://github.com/facebookresearch/maskrcnn-benchmark/blob/57eec25b75144d9fb1a6857f32553e1574177daf/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L75) and [mmdetection](https://github.com/open-mmlab/mmdetection/blob/ca115f2d14b1a4235fdbe0e3f75fbe6e92149b35/mmdet/models/dense_heads/anchor_head.py#L353) this **num_pos synchronization trick** is not used at all. So is this synchronization correct  in theory and necessary in practice?

yhcao6:
@hyz-xmaster,
I found the reason leads to the performance drops. In the latest commits, the default warm-up policy of FCOS had been changed slightly. Now I open a PR(#3119) to restore them. The performance should be able to align with 4GPUS now on 1x training schedules.

hyz-xmaster:
> @hyz-xmaster,
> I found the reason leads to the performance drops. In the latest commits, the default warm-up policy of FCOS had been changed slightly. Now I open a PR(#3119) to restore them. The performance should be able to align with 4GPUS now on 1x training schedules.

Great to hear. The warmup_ratio and policy can have such a big impact on the performance? Is there any research on this?

PR #2958, opened 2020-06-09
fix flip test check

thangvubk:
``self.transforms`` is a object ``Compose`` which is not iterable.
It can be solved by changing ``self.transforms`` -> ``self.transforms.transform`` but I think checking the transform type is cleaner.

### DIFF
--- a/mmdet/datasets/pipelines/test_time_aug.py
+++ b/mmdet/datasets/pipelines/test_time_aug.py
@@ -4,7 +4,6 @@ import mmcv
 
 from ..builder import PIPELINES
 from .compose import Compose
-from .transforms import RandomFlip
 
 
 @PIPELINES.register_module()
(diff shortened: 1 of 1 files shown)


xvjiarui:
Hi @thangvubk 
Thanks for the fix.

mmdet/datasets/pipelines/test_time_aug.py
```diff
@@ -37,10 +36,10 @@ def __init__(self,
         if not self.flip and self.flip_direction != ['horizontal']:
             warnings.warn(
                 'flip_direction has no effect when flip is set to False')
-        if (self.flip and
-                not any([isinstance(_, RandomFlip) for _ in self.transforms])):
+        if (self.flip
+                and not any([t.type == 'RandomFlip' for t in transforms])):
```
xvjiarui, line 40: For `dict` indexing, we may use `t['type']` for general usage.

thangvubk: thanks. I have updated it.

PR #2959, opened 2020-06-09
Add ResNeSt backbone

xvjiarui:


sunpeng1996:
Could you provide the ResNeSt-pretrained Model for me, i,e, your ResNeSt50 and ResNeSt101... thx...

cpboost:
resnest能够带来全面的模型提升，使用官方版本的resnest在1.x修改使用在各个竞赛的使用上都有很棒的效果

mmdet/models/utils/res_layer.py
```diff
@@ -38,7 +38,7 @@ def __init__(self,
         if stride != 1 or inplanes != planes * block.expansion:
             downsample = []
             conv_stride = stride
-            if avg_down and stride != 1:
+            if avg_down:
```
ZwwWayne, line 41: Does this modification influence other models like resnet or regnet?

xvjiarui: No. `avg_down` is used in `ResNetV1d` only, which doesn't have any checkpoint.

mmdet/models/backbones/resnest.py
```diff
@@ -0,0 +1,314 @@
...
+            x = torch.sigmoid(x)
+        return x
+
+
+class SplitAttentionConv2d(nn.Module):
+    """Split-Attention Conv2d in ResNeSt.
+
+    Args:
+        in_channels (int): Same as nn.Conv2d.
+        out_channels (int): Same as nn.Conv2d.
```
ZwwWayne, line 44: Update docstrings according to the real arguments.

mmdet/models/backbones/resnest.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class Bottleneck(_Bottleneck):
+    """Bottleneck block for ResNeSt.
+
+    Args:
+        inplane (int): Input planes of this block.
+        planes (int): Middle planes of this block.
+        groups (int): Groups of conv2.
+        width_per_group (int): Width per group of conv2. 64x4d indicates
```
ZwwWayne, line 153: Update docstrings according to the real arguments.

mmdet/models/backbones/resnest.py
```diff
@@ -0,0 +1,314 @@
...
+            None, channels, inter_channels, 1, groups=self.groups)
+        self.norm1_name, norm1 = build_norm_layer(
+            norm_cfg, inter_channels, postfix=1)
+        self.add_module(self.norm1_name, norm1)
+        self.fc2 = build_conv_layer(
+            None, inter_channels, channels * radix, 1, groups=self.groups)
+        self.rsoftmax = RSoftmax(radix, groups)
+
+    @property
+    def norm0(self):
```
ZwwWayne, line 108: ResNet models tend to use `norm1` and `norm2`, we may make them consistent.

xvjiarui: It will make the norm name mismatch with pretrained checkpoint.

ZwwWayne (approved):
LGTM now. See whether @hellock has any comments.

PR #2963, opened 2020-06-09
use BaseDenseHead

thangvubk:
This is related to #1900. A BaseDenseHead is used to wrap train and test logics of dense heads, make them more flexible and easier to extend.
Here are some unit tests that I have performed on Faster R-CNN, RetinaNet, and RPN with R50-FPN:

|            Item            |                 Result                 |
|:--------------------------:|:--------------------------------------:|
|         Simple test        |       Reproduce model zoo results      |
|          Aug test          | Improved results compared to model zoo |
| Train (several iterations) |              Run normally              |

I do not have enough time and resource to rebenchmark the models.


mmdet/models/dense_heads/base_rpn_head.py
```diff
@@ -0,0 +1,43 @@
+from mmdet.core import merge_aug_proposals
+from .base_dense_head import BaseDenseHead
+
+
+class BaseRPNHead(BaseDenseHead):
```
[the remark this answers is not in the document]
thangvubk: Thank you for your feedback. I use ``BaseRPNHead`` to define ``simple_test`` and ``aug_test`` for all RPNs. But I agree it makes inheritance not very clear. I have found a better way is to let ``RPNHead`` inherit ``RPNTestMixin``

mmdet/models/dense_heads/rpn_test_mixin.py
```diff
@@ -13,20 +10,20 @@ class RPNTestMixin(object):
...
 
-        async def async_test_rpn(self, x, img_metas):
+        async def async_simple_test_rpn(self, x, img_metas):
             sleep_interval = self.rpn_head.test_cfg.pop(
                 'async_sleep_interval', 0.025)
             async with completed(
                     __name__, 'rpn_head_forward',
                     sleep_interval=sleep_interval):
-                rpn_outs = self.rpn_head(x)
+                rpn_outs = self.__call__(x)
```
hellock, line 19: Simply `self(x)`?

mmdet/models/dense_heads/rpn_test_mixin.py
```diff
@@ -13,20 +10,20 @@ class RPNTestMixin(object):
```
hellock, line 10: It may be more reasonable to rename it to `RPNHeadTestMixin`.

mmdet/models/dense_heads/rpn_test_mixin.py
```diff
@@ -1,10 +1,7 @@
-import logging
```
hellock, line 1: This file can be moved to `models/dense_heads/`.

mmdet/models/dense_heads/rpn_test_mixin.py
thangvubk: I agree

mmdet/models/dense_heads/rpn_test_mixin.py
thangvubk: Similar to ``BboxTestMixin`` and ``MaskTestMixin``, I think ``RPNTestMixin`` is reasonable.

mmdet/models/dense_heads/rpn_test_mixin.py
thangvubk: File is already moved to ``models/dense_heads/``

mmdet/models/detectors/two_stage.py
```diff
@@ -139,16 +138,11 @@ def forward_train(self,
...
-            rpn_losses = self.rpn_head.loss(
-                *rpn_loss_inputs, gt_bboxes_ignore=gt_bboxes_ignore)
-            losses.update(rpn_losses)
-
             proposal_cfg = self.train_cfg.get('rpn_proposal',
                                               self.test_cfg.rpn)
-            proposal_list = self.rpn_head.get_bboxes(
-                *rpn_outs, img_metas, cfg=proposal_cfg)
+            rpn_losses, proposal_list = self.rpn_head.forward_train(
+                x, img_metas, gt_bboxes, None, gt_bboxes_ignore, proposal_cfg)
```
hellock, line 144: Use keyword arguments to prevent bugs if there are future api changes.

PR #2964, opened 2020-06-09
Add fl num class check

### LINKED ISSUE #2903: focal loss won't throw any error when model's num_classes is inconsistent with class number in ann_file
https://github.com/open-mmlab/mmdetection/blob/e05e85834c3162a8981dc5169ffd04b1c44d67d6/mmdet/ops/sigmoid_focal_loss/src/cuda/sigmoid_focal_loss_cuda.cu#L37

When num_classes in bbox_head isn't  consistent with the max class number in ann_file,
this won't throw any error,  add a check like ``t < num_classes``  may avoid some mistakes

yhcao6:
Fix #2903


mmdet/ops/sigmoid_focal_loss/src/cuda/sigmoid_focal_loss_cuda.cu
```diff
@@ -8,11 +8,10 @@
 // cyfu@cs.unc.edu
 #include <ATen/ATen.h>
 #include <ATen/cuda/CUDAContext.h>
-
 #include <THC/THC.h>
+
```
hellock, line 12: Is this blank line needed?

PR #2972, opened 2020-06-10
fix the nan loss bug in dirty data when training SSD

FreyWang:
First thank you for your nice work!
I meet the problem when I train SSD in some unclear data.
https://github.com/open-mmlab/mmdetection/blob/8bc0b9ce6d7d0656a8476ee8f31fa0441f8c4319/mmdet/datasets/pipelines/transforms.py#L150 when the data is all right, np.clip will have no effect
But when gt box is out of image area in few dirty data, 
np.clip will lead to x1=x2 or y1=y2. then 
https://github.com/open-mmlab/mmdetection/blob/8bc0b9ce6d7d0656a8476ee8f31fa0441f8c4319/mmdet/core/bbox/coder/delta_xywh_bbox_coder.py#L78 will be zero and 
https://github.com/open-mmlab/mmdetection/blob/8bc0b9ce6d7d0656a8476ee8f31fa0441f8c4319/mmdet/core/bbox/coder/delta_xywh_bbox_coder.py#L83 will be nan, cause the L1 smooth loss be NAN.
@ZwwWayne 

Signed-off-by: wangxionghui.kk <wangxionghui.kk@bytedance.com>

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -147,8 +147,6 @@ class Resize(object):
         img_shape = results['img_shape']
         for key in results.get('bbox_fields', []):
             bboxes = results[key] * results['scale_factor']
-            bboxes[:, 0::2] = np.clip(bboxes[:, 0::2], 0, img_shape[1])
-            bboxes[:, 1::2] = np.clip(bboxes[:, 1::2], 0, img_shape[0])
             results[key] = bboxes
 
     def _resize_masks(self, results):


ZwwWayne:
Hi @FreyWang ,
Thanks for your bug report. This is indeed a problem that we did not realize before. 

However, it is not a good idea to remove the clip during transform, because this does not really fix the problem caused by out-of-image box and could cause some deeper bugs in other places, e.g., in the second stage of two-stage detectors where GT is added to help training. 

Would you like to keep the clip in the transform, but add checks in the datasets, where those out-of-image boxes should be fixed when they first show up.

FreyWang:
> Hi @FreyWang ,
> Thanks for your bug report. This is indeed a problem that we did not realize before.
> 
> However, it is not a good idea to remove the clip during transform, because this does not really fix the problem caused by out-of-image box and could cause some deeper bugs in other places, e.g., in the second stage of two-stage detectors where GT is added to help training.
> 
> Would you like to keep the clip in the transform, but add checks in the datasets, where those out-of-image boxes should be fixed when they first show up.

Yes it will be better to filter the bad case in dataset, our label team did not add width and height information to ann file in default and we omit the situation. On the other hand, assert statement can be add to transform to avoid the extreme case.

ZwwWayne:
> > Hi @FreyWang ,
> > Thanks for your bug report. This is indeed a problem that we did not realize before.
> > However, it is not a good idea to remove the clip during transform, because this does not really fix the problem caused by out-of-image box and could cause some deeper bugs in other places, e.g., in the second stage of two-stage detectors where GT is added to help training.
> > Would you like to keep the clip in the transform, but add checks in the datasets, where those out-of-image boxes should be fixed when they first show up.
> 
> Yes it will be better to filter the bad case in dataset, our label team did not add width and height information to ann file in default and we omit the situation. On the other hand, assert statement can be add to transform to avoid the extreme case.

Sorry I am not clear enough before. I mean we can check and filter out them in some places like [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/coco.py#L117) when the code of dataset load and parse annotations.

PR #2985, opened 2020-06-11
fix anchor target label

thangvubk:
If I am not wrong, it would be a bug when RPN is using (``background_label = 0``), leading to all samples are positive.
Not sure that RPN in model zoo is benchmarked with this bug or not.

### DIFF
--- a/mmdet/models/dense_heads/anchor_head.py
+++ b/mmdet/models/dense_heads/anchor_head.py
@@ -232,8 +232,10 @@ class AnchorHead(nn.Module):
         if unmap_outputs:
             num_total_anchors = flat_anchors.size(0)
             labels = unmap(
-                labels, num_total_anchors, inside_flags,
-                fill=self.num_classes)  # fill bg label
+                labels,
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Hi @thangvubk ,
The background_label does not affect the training because the filled part is weighted zero when calculating the loss. But this indeed could affect the performance in some situation. Thanks for fixing this bug.

thangvubk:
Hi @ZwwWayne. I don't think that the loss for filled are filtered out by label-weights since the ``fill`` value also overwrite negative labels (which have label-weights of 1). You can double-check that.

ZwwWayne:
See [here](https://github.com/open-mmlab/mmdetection/blob/3969e50370130d8c3bc0d738538f37cf1871b302/mmdet/core/utils/misc.py#L28) `fill` of unmap only affects the out-of-range parts but does not affect the assigned parts. And from the performance, the results are reasonable.

thangvubk:
Ah. I got it. Thank you :D

PR #2998, opened 2020-06-13
fix the name error for lvis validation image load(val2017)

tonysy:


xvjiarui:
Hi @tonysy 
Thanks for bringing this up. PR is approved.

PR #2999, opened 2020-06-13
Fixed nan loss by filtering out-of-frame gt_bboxes in coco.py

Jokoe66:
There may exist out-of-frame annotations in custom datasets. These abnormal annotations will cause nan loss. For example, such cases exist in the object365 datasets, and I found both Libra RCNN and Cascade RCNN diverged due to nan loss. After filtering these abnormal cases, these detectors converged. Even though this should be avoided by dataset annotators, it can be more robust of mmdetection to filter out-of-frame annotations.

### DIFF
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -113,12 +113,9 @@ class CocoDataset(CustomDataset):
             if ann.get('ignore', False):
                 continue
             x1, y1, w, h = ann['bbox']
-            x2 = x1 + w
-            y2 = y1 + h
-            if ((x1 < 0 or x1 >= img_info['width'] or y1 < 0
-                 or y1 >= img_info['height'])
-                    and (x2 < 0 or x2 >= img_info['width'] or y2 < 0
-                         or y2 >= img_info['height'])):
+            inter_w = max(0, min(x1 + w, img_info['width']) - max(x1, 0))
+            inter_h = max(0, min(y1 + h, img_info['height']) - max(y1, 0))
+            if inter_w * inter_h == 0:
                 continue
             if ann['area'] <= 0 or w < 1 or h < 1:
                 continue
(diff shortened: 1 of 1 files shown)


mmdet/datasets/coco.py
```diff
@@ -109,11 +109,17 @@ def _parse_ann_info(self, img_info, ann_info):
...
             if ann.get('ignore', False):
                 continue
             x1, y1, w, h = ann['bbox']
+            x2 = x1 + w
+            y2 = y1 + h
+            if ((x1 < 0 or x1 >= img_info['width'] or y1 < 0
+                 or y1 >= img_info['height'])
+                    and (x2 < 0 or x2 >= img_info['width'] or y2 < 0
+                         or y2 >= img_info['height'])):
+                continue
```
yhcao6, line 122: I have some questions about the condition statement,
1. The conditions seem are redundant. For example if `x1 >= img_info['width']` then x2 must `>= img_info['width']`, so we only need to check `x2 >= img_info['width']`.
2. Why you use `and` between the first condition and the second condition,  in my opinion, if one of two conditions satisfy then the image is out-of-frame.

Jokoe66: 1. Yeah, the conditions are redundant. But it is more readable. The first condition checks if the left-top point is out of image, and the second one checks the bottom-right point.
2. I mean to filter boxes that are fully outside the image (frame), but preserve those partially outside the image. In another word, when the boxes do not intersect with the image, then they are filtered. 

I think I should check the intersection instead. Current conditions are True (both corner points are out of frame) when the image is inside a box, and the box is filtered. But the box should be preserved since it intersects with the image.

openmmlab-v:
Hi @Jokoe66 ！First of all, we want to express our gratitude for your significant PR in the mmdetection project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

PR #3011, opened 2020-06-13
Modify mdconv doc

### LINKED ISSUE #2650: ModulatedDeformConv doesn't support different padding in H and W
**Describe the bug**
ModulatedDeformConv doesn't support different padding in H and W. 
PS: My project only require (Modulated)DeformConv function so that it is directly imported.

**Error traceback**
```
File "src/models/deform_conv.py", line 421, in forward
    self.groups, self.deformable_groups)
  File "src/models/deform_conv.py", line 143, in forward
    ModulatedDeformConvFunction._infer_shape(ctx, input, weight))
  File "src/models/deform_conv.py", line 181, in _infer_shape
    height_out = (height + 2 * ctx.padding -
TypeError: unsupported operand type(s) for +: 'int' and 'tuple'
```

**Bug fix**
Unlike DeformConv, ModulatedDeformConv only supports the same padding in  height and width. It only accepts padding parameter as int not tuple which is inconsistent with what is said in the code comments

xvjiarui:
Fixed #2650

### DIFF
--- a/mmdet/ops/dcn/deform_conv.py
+++ b/mmdet/ops/dcn/deform_conv.py
@@ -177,6 +177,7 @@ class ModulatedDeformConvFunction(Function):
         channels_out = weight.size(0)
         height, width = input.shape[2:4]
         kernel_h, kernel_w = weight.shape[2:4]
+        # TODO: support different padding/stride/dilation in height and width
         height_out = (height + 2 * ctx.padding -
                       (ctx.dilation * (kernel_h - 1) + 1)) // ctx.stride + 1
         width_out = (width + 2 * ctx.padding -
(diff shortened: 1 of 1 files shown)


mmdet/ops/dcn/deform_conv.py
```diff
@@ -408,9 +409,9 @@ def __init__(self, *args, **kwargs):
             self.deformable_groups * 3 * self.kernel_size[0] *
             self.kernel_size[1],
             kernel_size=self.kernel_size,
-            stride=_pair(self.stride),
-            padding=_pair(self.padding),
-            dilation=_pair(self.dilation),
+            stride=self.stride,
+            padding=self.padding,
+            dilation=self.dilation,
```
hellock, line 414: Are these changes necessary?

Issue #2931, opened 2020-06-07
Roadmap of MMDetection :fire: :fire: :fire:


hellock:
We keep this issue open to collect feature requests from users and hear your voice. Our release plan is also be available here.

You can either:
1. Suggest a new feature.
2. Vote for a feature request with :+1: or be against with :-1:. (Remember that developers are busy and cannot respond to all feature requests, so vote for your most favorable one!)
3. Tell us that you would like to help implement one of the features in the list. (This is the greatest things to hear about!)

V2.2:
- [ ] ResNeSt. (WIP)
- [ ] PointRend. (#2752)
- [ ] CornerNet. (WIP)
- [ ] Refactoring of Anchor-free detectors. (#2867)

mmeendez8:
It would be interesting to add batch inference support. I've seen a couple of issues related with this (#2703 #1833 #1659) and some details in source code and it looks like it has been left as future work (see https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/detectors/base.py#L118) so it could be a good moment.

I'd be glad to help with this too.

hiyyg:
How about adding the support for yolov4?

leehao178:
Hi authors
Can you add DOTA dataset support?
ex:more bbox parameters!
[https://captain-whu.github.io/DOTA/dataset.html](url)

Think you so much!!

zehuichen123:
Any plans about knowledge distillation for object detection, for example fitnet, so that users can customize their distillation loss functions based on their requirements.

hellock:
> I would be interested in adding support for EfficientNet (backbone, #764) and EfficientDet (detector, #1827) architectures along with trained models.
> 
> I will also be willing to contribute this feature myself.

Looking forward to it. Is there expected time for EfficientNet and EfficientDet? We need to have a plan whether to release it in V2.2 (June 30) or V2.3 (July 31).

hellock:
> It would be interesting to add batch inference support. I've seen a couple of issues related with this (#2703 #1833 #1659) and some details in source code and it looks like it has been left as future work (see https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/detectors/base.py#L118) so it could be a good moment.
> 
> I'd be glad to help with this too.

Thanks! We have already created a PR (#1833) to support batch inference a few months ago, and will continue working on it to make it compatible with V2.0. You are still welcome to contribute any other new features.

hellock:
[...]

Issue #1827, opened 2019-12-18
EfficientDet implementation


stevenwudi:
Do author plan to implement EfficientDet "EfficientDet: Scalable and Efficient Object Detection" into mmdet framework
That would be great!
Thanks

hellock:
PRs are welcome!

bilel-bj:
Is there any plan to implement it? It is currently the state of the art for object detection.
Thanks.

GoogleWiki:
Expecting EfficientDet too！

loveunk:
A Pytorch version: https://github.com/zylo117/Yet-Another-EfficientDet-Pytorch

sailfish009:
> A Pytorch version: https://github.com/zylo117/Yet-Another-EfficientDet-Pytorch

It is LGPL v3 license, apache 2.0 license version is here: https://github.com/rwightman/efficientdet-pytorch

mathmanu:
Looking forward for EfficientDet  and EfficientDetLite (which is the embedded friendly variant) as supported in https://github.com/rwightman/efficientdet-pytorch

ZwwWayne:
EfficientNet has been supported. PRs of EfficientDet is welcomed.

PR #3061, opened 2020-06-17
Add efficientnet

### LINKED ISSUE #764: Any plans for support efficientnet ？

daavoo:
#2931

Resolves #764 #1827

I tried to follow the approach of regnet implementations, migrating weights fron pycls.

I don't know which experiments make the most sense because there are not official results of efficientnets for object detection without.the efficientDet neck and head.


mmdet/models/backbones/efficientnet.py
```diff
@@ -0,0 +1,314 @@
...
+
+class MBConv(nn.Module):
+    """Mobile inverted Bottleneck block with Squeeze-and-Excitation (SE).
+
+    Args:
+        input_width (int): Number of input filters.
+        output_width (int): Number of output filters.
+        stride (int): stride of the first block.
+        exp_ratio (Sequence[int]): Expansion ratio..
+        kernel (Sequence[int]): Kernel size of the dwise conv.
```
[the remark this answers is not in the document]
daavoo: The docstring was wrongly copy-pasted

daavoo:
> Hi,
> Thanks a lot for your great work and patience. The general implementation detail of EfficientNet is okay to me except some minor comments. As to the larger purpose (EfficientDet), there are two concerns.
> 
> 1. Since mmdetection has imposed a stricter docstring requirement on nearly all classes and its submethods (see #3120 and #3125 ), it would be highly appreciated that docstring could be added to class methods in this implementation as well.
> 2. As for benchmarking this PR, I am afraid that reasonable results would not be attained until some training tricks are implemented as well. I list several tricks as follows. By the way, is it possible to run a voc experiment at first?
> 
> * [weight decay = 4e-5](https://github.com/google/automl/blob/2311ca096fc547ffb481644726e3906a8448e032/efficientdet/hparams_config.py#L236)
> * [multi-scale training](https://github.com/google/automl/blob/2311ca096fc547ffb481644726e3906a8448e032/efficientdet/hparams_config.py#L184)
> * [cosing lr](https://github.com/google/automl/blob/2311ca096fc547ffb481644726e3906a8448e032/efficientdet/hparams_config.py#L262)
> * [exponential moving average](https://github.com/google/automl/blob/2311ca096fc547ffb481644726e3906a8448e032/efficientdet/hparams_config.py#L263)

Thanks for the review. I will try to run an experiment in VOC applying the suggested training tricks

daavoo:
> Is there any progress on that or was it abandoned?

I don't have access to the hardware required to run the experiments in COCO with proper batch size.
I'm going to update the configs with the recommended training tricks.

daavoo:
I managed to sneak a COCO training on my machines with the config pushed in this branch (but only batch 32); I will comment the results tomorrow.

@Johnson-Wang I added all the training to the config tricks you suggested (with exception of E.M.A) and tried to follow the preprocessing used in the original implementation. 

The only Pytorch repository that seemed to reproduce the original results (with efficientDet) says:

> The official training code uses EMA weight averaging by default, it's not clear there is a point in doing this with the cosine LR schedule, I find the non-EMA weights end up better than EMA in the last 10-20% of training epochs.

daavoo:
Training with previous commit (e8ee996) batch size 32:

`Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.297`

This is with 24 epochs (2x_scheduler). Official code trains for 300 epochs but I can't launch a training that long.

I realized that the learning rate scheduler in the official code is being called by iter instead of by epoch so I updated the config accordingly.

Johnson-Wang:
Hi, thanks for your work. I wonder how about the results using the current config. I think running a 2x model is acceptable with constrained resources, but it would be better to run the baseline with the same training budget using the official code as well. By the way, EMA is available in mmcv now, We can add a line `custom_hooks = [dict(type='EMAHook')]`  in the config to activate it.

daavoo:
> Hi, thanks for your work. I wonder how about the results using the current config. I think running a 2x model is acceptable with constrained resources, but it would be better to run the baseline with the same training budget using the official code as well. By the way, EMA is available in mmcv now, We can add a line `custom_hooks = [dict(type='EMAHook')]` in the config to activate it.

I trained for 80 epochs:

`Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.324`

In the original paper 34.6 for Efficient**D**et-D0 (512), which is presumably the closest result (as far as I know there is not published result of using only EfficientNet as backgone). 
In the ablation study, the biFPN (the part that is missing in this experiment
) improves 4AP with respect to the baseline.

I don't know if this numbers are good enough to assume that the implementation is correct.

Johnson-Wang:
Sorry for the late reply. 
Here is a crude ablation for B0 model from another source (confidential). 
Baseline setting: EfficientDet B0 trained for 32 epoch. 

|   model  | mAP |
| ----         | ---   |
| baseline | 31.8 |
| Remove bifpn | 32.5|
|Remove ema | 28.8 |
|Remove multiscal training | 31.4| 
|weight decay 1e-4 | 27.3 |

Maybe you can roughly compare your results with this one, but please do not take this result seriously. Here are some of my thoughts based on this ablation,
1. Ema and weight decay should be aligned with the settings in the official TF baseline, since it may have huge impact on the results for EfficientDet B0. 
2. BiFPN could bring negative effect for B0 and the one without it could end up with higher results. 
3. It would be trivial to compare with the baseline with different settings. I recommend running the model in TF using the same training budget to check its validity. 
4. Direct model weights conversion is thought necessary since it would be convenient for you or other users to check the correctness of the pipeline.

mathmanu:
The head in EfficientDet is almost similar to that of RetinaNet - so RetinaNet head should be fine. But the neck in EfficientDet is BiFPN which may contribute to some accuracy improvement over FPN.
 
Looking forward for this PR to be merged.

daavoo:
> The head in EfficientDet is almost similar to that of RetinaNet with only small differences - so RetinaNet head should be fine from an accuracy point of view. But the neck in EfficientDet is BiFPN which may contribute to some accuracy improvement over FPN.
> 
> Looking forward for this PR to be merged.

I don't fully understand what changes should be added to the P.R.

mmdet/models/utils/se_block.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, w_in, w_se, with_swish=True):
+        super(SE, self).__init__()
+        self.avg_pool = nn.AdaptiveAvgPool2d((1, 1))
+        activation = Swish if with_swish else nn.ReLU
+        self.f_ex = nn.Sequential(
+            nn.Conv2d(w_in, w_se, 1, bias=True),
+            activation(),
+            nn.Conv2d(w_se, w_in, 1, bias=True),
+            nn.Sigmoid(),
```
ZwwWayne, line 24: clean comma

configs/efficientnet/README.md
```diff
@@ -0,0 +1,39 @@
+# EfficientNet: Rethinking Model Scaling for Convolutional Neural Networks
```
ZwwWayne, line 1: May indicate to the users that this is an experimental feature

ZwwWayne:
Hi @daavoo,
Thanks for your kind contribution! We plan to merge this PR lately (maybe after a few weeks).

Johnson-Wang:
Hi Sorry for the late update. I have benchmarked the model with the config in this PR and ends up with mAP=33.6::

    Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.336
    Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=1000 ] = 0.525
    Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=1000 ] = 0.350
    Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.140
    Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.384
    Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.505
    Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.491
    Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=300 ] = 0.491
    Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=1000 ] = 0.491
    Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.240
    Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.557
    Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.706

It has around 1AP gain over the 80e training schedule setting and only 0.8AP lower than efficient-det-d0 reported in the paper. However, the fpn channels used in efficient-d0 is only 64 (instead of 256 used in standard retinanet). Therefore, a straightforward comparison is still missing. This is also the main reason that we were a bit reluctant to merge this PR. 

For now, I think a more realistic plan could be like the following. 
The efficientnet-d3+FPN performance was provided in the original paper, and can serve as a solid baseline. I wonder if it is possible for @daavoo to provide a efficient-d3 config and we may help benchmark the full performance.

daavoo:
> Hi Sorry for the late update. I have benchmarked the model with the config in this PR and only change the total epoch from 80 to 300, and ends up with mAP=33.6::
> 
> ```
> Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.336
> Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=1000 ] = 0.525
> Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=1000 ] = 0.350
> Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.140
> Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.384
> Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.505
> Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.491
> Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=300 ] = 0.491
> Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=1000 ] = 0.491
> Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.240
> Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.557
> Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.706
> ```
> 
> It has around 1AP gain over the 80e training schedule setting and only 0.8AP lower than efficient-det-d0 reported in the paper. However, the fpn channels used in efficient-d0 is only 64 (instead of 256 used in standard retinanet). Therefore, a straightforward comparison is still missing. This is also the main reason that we were a bit reluctant to merge this PR.
> 
> For now, I think a more realistic plan could be like the following.
> The efficientnet-d3+FPN performance was provided in the original paper, and can serve as a solid baseline. I wonder if it is possible for @daavoo to provide a efficient-d3 config and we may help benchmark the full performance.

I guess I could extend this P.R. to include the biFPN part in order to make a straightforward comparison with the paper results using EfficientDet.

mmdet/models/backbones/efficientnet.py
```diff
@@ -0,0 +1,313 @@
...
+                 norm_cfg=dict(type='BN', requires_grad=True),
+                 norm_eval=True):
+        super().__init__()
+        self.out_indices = out_indices
+        self.conv_cfg = conv_cfg
+        self.norm_cfg = norm_cfg
+
+        self.stage_depths, self.stage_widths = self.arch_settings[scale]
+
+        self._make_stem_layer(3, base_channels)
```
qingswu, line 236:
```suggestion
        self._make_stem_layer(in_channels, base_channels)
```

ZwwWayne:
supported in https://github.com/open-mmlab/mmdetection/pull/7514

PR #3097, opened 2020-06-22
GFocal

implus:
*  add Generalized Focal loss (GFL), including QualityFocalLoss(QFL) and DistributionFocalLoss(DFL)
*  add GFL backbone, head, configs, and pretrained models and logs


configs/gfl/gfl_r50_fpn_1x_coco.py
```diff
@@ -0,0 +1,57 @@
+_base_ = [
+    '../_base_/datasets/coco_detection.py',
+    '../_base_/schedules/schedule_1x.py', '../_base_/default_runtime.py'
+]
+model = dict(
+    type='GFL',
+    pretrained='torchvision://resnet50',
+    backbone=dict(
```
ZwwWayne, line 8: The backbone and neck can inherit from RetinaNet in _base_.

mmdet/core/bbox/transforms.py
```diff
@@ -124,3 +124,27 @@ def distance2bbox(points, distance, max_shape=None):
...
+    Args:
+        points (Tensor): Shape (n, 2), [x, y].
+        bbox (Tensor): Shape (n, 4), "xyxy" format
+        max_dis (float): Upper bound of the distance.
+        eps (float): a small value to ensure target < max_dis, instead <=
+
+    Returns:
+        Tensor: Decoded distances.
+    """
+    L = points[:, 0] - bbox[:, 0]
```
ZwwWayne, line 141: Try to avoid single-letter variable, change `L, T, R, B` to `left, top, right, bottom`.

mmdet/models/losses/gfocal_loss.py
```diff
@@ -0,0 +1,107 @@
...
+    # all goes to 0
+    pred_sigmoid = pred.sigmoid()
+    pt = pred_sigmoid
+    zerolabel = pt.new_zeros(pred.shape)
+    loss = F.binary_cross_entropy_with_logits(
+        pred, zerolabel, reduction='none') * pt.pow(beta)  # pt = pt.abs()
+    # find positive positions and quality labels
+    label = label - 1
+    pos = (label >= 0).nonzero().squeeze(1)
+    a = pos
```
ZwwWayne, line 24: Assign `a = pos` seems to be unnecessary.

mmdet/models/losses/gfocal_loss.py
```diff
@@ -0,0 +1,107 @@
...
+    loss = weight_reduce_loss(loss, weight, reduction, avg_factor)
+    return loss
+
+
+def distribution_focal_loss(pred,
+                            label,
+                            weight=None,
+                            reduction='mean',
+                            avg_factor=None):
+    disl = label.long()
```
ZwwWayne, line 39: 1. disl -> long_label
2. change `disr` to a more clear name

configs/gfl/gfl_r50_fpn_1x_coco.py
implus: got it~

mmdet/core/bbox/transforms.py
implus: Got it, I will quickly modify that part~

mmdet/models/losses/gfocal_loss.py
implus: Thanks for suggestions, I will update this part soon~

mmdet/models/losses/gfocal_loss.py
implus: Got it, one more question, should I close this pull request first, then update related code to perform a new pull request?

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,493 @@
...
+    def anchor_center(self, anchors):
+        anchors_cx = (anchors[:, 2] + anchors[:, 0]) / 2
+        anchors_cy = (anchors[:, 3] + anchors[:, 1]) / 2
+        return torch.stack([anchors_cx, anchors_cy], dim=-1)
+
+    def iou_target(self, pred, target, eps=1e-6):
+        # overlap
+        lt = torch.max(pred[:, :2], target[:, :2])
+        rb = torch.min(pred[:, 2:], target[:, 2:])
+        wh = (rb - lt + 1).clamp(min=0)
```
ZwwWayne, line 149: 1. MMDetection implements IoU calculation [here](https://github.com/open-mmlab/mmdetection/blob/e7537a4b394728eb02133e27b5aadcd8caec8268/mmdet/core/bbox/iou_calculators/iou2d_calculator.py#L38).
2. If this function is necessary, in MMDetection V2.0, width and height is calculated by `xmax - xmin`, `+1` is not necessary. Here we might need to varify the influence of removing `+1` when calculating `iou`

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,493 @@
...
+        anchors = anchors.reshape(-1, 4)
+        cls_score = cls_score.permute(0, 2, 3,
+                                      1).reshape(-1, self.cls_out_channels)
+        bbox_pred = bbox_pred.permute(0, 2, 3,
+                                      1).reshape(-1, 4 * (self.reg_max + 1))
+        bbox_targets = bbox_targets.reshape(-1, 4)
+        labels = labels.reshape(-1)
+        label_weights = label_weights.reshape(-1)
+
+        # adjust label to fit Quality Focal Loss
```
ZwwWayne, line 172: Can we change the Quality Focal Loss rather than change the label assign here for more consistent loss calculation?

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if len(pos_inds) > 0:
+            pos_bbox_targets = bbox_targets[pos_inds]
+            pos_bbox_pred = bbox_pred[pos_inds]
+            pos_anchors = anchors[pos_inds]
+            pos_anchor_centers = self.anchor_center(pos_anchors) / stride[0]
+
+            weight_targets = cls_score.detach().sigmoid()
+            weight_targets = weight_targets.max(dim=1)[0][pos_inds]
+            pos_bbox_pred_ltrb = self.integral(pos_bbox_pred)
```
ZwwWayne, line 187: Can we use `xxx_corners` rather than abbr. `xxx_ltrb`

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,493 @@
...
+
+        avg_factor = sum(avg_factor)
+        avg_factor = reduce_mean(avg_factor).item()
+        losses_bbox = list(map(lambda x: x / avg_factor, losses_bbox))
+        losses_dfl = list(map(lambda x: x / avg_factor, losses_dfl))
+        return dict(
+            loss_cls=losses_cls, loss_bbox=losses_bbox, loss_dfl=losses_dfl)
+
+    @force_fp32(apply_to=('cls_scores', 'bbox_preds'))
+    def get_bboxes(self,
```
ZwwWayne, line 277: It seems that this function is very similar to that in `AnchorHead`, try to remove it if they are the same.

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,493 @@
...
+            valid_flag_list,
+            gt_bboxes,
+            img_metas,
+            gt_bboxes_ignore_list=gt_bboxes_ignore,
+            gt_labels_list=gt_labels,
+            label_channels=label_channels)
+        if cls_reg_targets is None:
+            return None
+
+        (anchor_list, labels_list, label_weights_list, bbox_targets_list,
```
ZwwWayne, line 250: What is the difference between the `anchor_list` returned by the `self.get_target()` here and the `anchor_list` in line 235?

mmdet/models/losses/gfocal_loss.py
```diff
@@ -0,0 +1,107 @@
...
+                       reduction='mean',
+                       avg_factor=None):
+    # all goes to 0
+    pred_sigmoid = pred.sigmoid()
+    scale_factor = pred_sigmoid
+    zerolabel = scale_factor.new_zeros(pred.shape)
+    loss = F.binary_cross_entropy_with_logits(
+        pred, zerolabel, reduction='none') * scale_factor.pow(beta)
+    # find positive positions and quality labels
+    label = label - 1
```
ZwwWayne, line 22: We can use a more consistent way to obtain positive and negative labels here.

mmdet/models/dense_heads/gfl_head.py
implus: anchor_list in 235 is image-wise, whilst anchor_list returned by `self.get_target()` is fpn-level-wise. The implementation is the same with line 187-202 in `https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/dense_heads/atss_head.py`

mmdet/models/dense_heads/gfl_head.py
implus: This part is modified to be consistent with new label rules in MMDet v2.0:
FG cat_id: [0~num_classes-1];  BG cat_id: num_classes

mmdet/models/dense_heads/gfl_head.py
implus: `iou_target` is replaced with `bbox_overlaps`.

I would do a quick experiment to check the influence of removing `+1`.

mmdet/models/dense_heads/gfl_head.py
implus: got it~

mmdet/models/dense_heads/gfl_head.py
implus: removed~

mmdet/models/dense_heads/gfl_head.py
implus: The R50_1x_coco experiment shows the performance without `+1` remains exactly the same with the one with `+1`, i.e., 40.2 vs 40.2 AP.

mmdet/models/losses/gfocal_loss.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def quality_focal_loss(pred,
+                       label,
+                       score,
+                       weight=None,
+                       beta=2.0,
+                       reduction='mean',
+                       avg_factor=None):
+    # all goes to 0
```
liaopeiyuan, line 15: Could use a bit more explanation to what this means (the formula in your paper is very intuitive)

mmdet/models/losses/gfocal_loss.py
```diff
@@ -0,0 +1,107 @@
...
+from .utils import weight_reduce_loss
+
+
+def quality_focal_loss(pred,
+                       label,
+                       score,
+                       weight=None,
+                       beta=2.0,
+                       reduction='mean',
+                       avg_factor=None):
```
hellock, line 14: The decorator `@weighted_loss` is designed for adding weight and reduction support for losses, e.g., https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/iou_loss.py#L10

`quality_focal_loss` and `distribution_focal_loss` can be simplified via this decorator.

hellock: Add a docstring for each method.

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,451 @@
...
+        return tensor
+    tensor = tensor.clone()
+    dist.all_reduce(tensor.div_(dist.get_world_size()), op=dist.ReduceOp.SUM)
+    return tensor
+
+
+class Integral(nn.Module):
+    """A fixed layer for calculating integral result from distribution
+
+    This layer calculates the target location by \\sum{P(y_i) * y_i},
```
hellock, line 26:
```
:math:`\sum{P(y_i) * y_i}`
```

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,451 @@
...
+    Generalized Focal Loss: Learning Qualified and Distributed Bounding Boxes
+    for Dense Object Detection
+
+    GFL head structure is similar with ATSS, however GFL uses
+    1) joint representation for classification and localization quality, and
+    2) flexible General distribution for bounding box locations,
+    which are supervised by
+    Quality Focal Loss (QFL) and Distribution Focal Loss (DFL), respectively
+
+    https://arxiv.org/abs/2006.04388
```
hellock, line 58: `Args` are missing.

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,451 @@
...
+
+        return (anchors, labels, label_weights, bbox_targets, bbox_weights,
+                pos_inds, neg_inds)
+
+    def get_num_level_anchors_inside(self, num_level_anchors, inside_flags):
+        split_inside_flags = torch.split(inside_flags, num_level_anchors)
+        num_level_anchors_inside = [
+            int(flags.sum()) for flags in split_inside_flags
+        ]
+        return num_level_anchors_inside
```
hellock, line 451: These days we are adding missing docstrings in the codebase to achieve a high coverage. There are still many methods in this class without docstrings.

mmdet/core/bbox/transforms.py
```diff
@@ -124,3 +124,27 @@ def distance2bbox(points, distance, max_shape=None):
         x2 = x2.clamp(min=0, max=max_shape[1])
         y2 = y2.clamp(min=0, max=max_shape[0])
     return torch.stack([x1, y1, x2, y2], -1)
+
+
+def bbox2distance(points, bbox, max_dis=None, eps=0.1):
+    """Decode bounding box based to distances.
```
ZwwWayne, line 130: `based on` or `to`?

mmdet/models/dense_heads/gfl_head.py
```diff
@@ -0,0 +1,451 @@
...
+
+            nms_pre = cfg.get('nms_pre', -1)
+            if nms_pre > 0 and scores.shape[0] > nms_pre:
+                max_scores, _ = scores.max(dim=1)
+                _, topk_inds = max_scores.topk(nms_pre)
+                anchors = anchors[topk_inds, :]
+                bbox_pred = bbox_pred[topk_inds, :]
+                scores = scores[topk_inds, :]
+
+            bboxes = distance2bbox(
```
ZwwWayne, line 294: It seems that the main difference of the `_get_bboxes_single` from the original one lies here. We can implement a `DistanceBboxCoder` to avoid rewriting this function.

implus: There are actually more differences. E.g., we need to use `self.integral` and `stride` to obtain the final bbox coordinates.

mmdet/core/bbox/transforms.py
implus: based on ~

mmdet/models/losses/gfocal_loss.py
implus: Thanks for advices. I would add more clear comments in this functions~

mmdet/models/losses/gfocal_loss.py
implus: `@weighted_loss` and docstring are added.

mmdet/models/dense_heads/gfl_head.py
implus: Got it~

mmdet/models/dense_heads/gfl_head.py
implus: Docstrings are complete now~

hellock:
LGTM. Waiting for the benchmarking results.

ZwwWayne:
> LGTM. Waiting for the benchmarking results.

For now, I have finished training some models, which produce the same AP as reported.
gfl_r50_fpn_1x_coco.py: 40.2, gfl_r50_fpn_mstrain_2x_coco.py: 42.9.
So I think it is fine to merge this PR today and we can upload the complete benchmarked models in the next batch.

PR #3119, opened 2020-06-24
Fix fcos warm up

yhcao6:


hiyyg:
In the new FCOS version, the warmup is "linear" schedule.
https://github.com/aim-uofa/AdelaiDet/commit/4d2dba204bb2c43cd0c34a2bb56b115faa424cd1

PR #3132, opened 2020-06-26
Fix a bug in tools/publish_model.py

XiaohangZhan:
`rstrip('.pth')` removes all characters in {'.', 'p', 't', 'h'} at the right. For example, if `out_file="imagenet.pth"`, then `out_file.rstrip('.pth')` is `"imagene"` where "t" is also removed.


hellock:
Thanks for the fix!

Issue #3023, opened 2020-06-15
DataLoader seems to ignore samples without objects


aabramovrepo:
Dear All,

   am I right that `DataLoader` ignores samples without objects? I have just noticed that the data set length shown in `INFO` during training is slightly lower than the actual data set length. Just for my understanding, how can I force `DataLoader` to load all images (I'm training `Faster RCNN` on data in `PASCAL VOC` format)? Setting `filter_empty_gt` to False does not make a difference.

Thanks!

Best,
Alexey

Shallowte:
Using MSCOCO dataset, I think `filter_empty_gt` works fine when first loading the data, then the data is filtered by `get_subset_by_classes` found in mmdetection/mmdet/datasets/coco.py. 

Adding a condition `if self.filter_empty_gt` to the function this way

    def get_subset_by_classes(self):

        ids = set()
        if self.filter_empty_gt:
            for i, class_id in enumerate(self.cat_ids):
                ids |= set(self.coco.catToImgs[class_id])
            self.img_ids = list(ids)

        data_infos = []
        for i in self.img_ids:
            info = self.coco.loadImgs([i])[0]
            info['filename'] = info['file_name']
            data_infos.append(info)
        return data_infos

seems to do the trick on my side for MSCOCO, it might give you an idea for your case. 

Cheers!

aabramovrepo:
@hellock thanks for your reply, I will check it out.

PR #3133, opened 2020-06-26
Fix coco loader empty images

mmeendez8:
This PR is related with https://github.com/open-mmlab/mmdetection/issues/3023 I followed @Shallowte idea and added a regression test. So when `filter_empty_gt` is `False` all images will be loaded independently of their annotations.
I am not sure if this is the best approach (it might be more efficient to filter images [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/coco.py#L40) instead ), just let me know how this could be improved

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,7 +104,6 @@ venv.bak/
 .mypy_cache/
 
 mmdet/version.py
-data/
 .vscode
 .idea
 .DS_Store
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -81,10 +81,11 @@ class CocoDataset(CustomDataset):
             ids (list[int]): integer list of img ids
         """
 
-        ids = set()
-        for i, class_id in enumerate(self.cat_ids):
-            ids |= set(self.coco.cat_img_map[class_id])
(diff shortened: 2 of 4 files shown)


mmeendez8:
I cannot really understand why tests are failing with:
`AttributeError: 'CocoDataset' object has no attribute 'img_ids'`
It should be fullfilled in after calling `load_annotations` (https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/coco.py#L40)

Am I missing something? This is working fine locally

ZwwWayne:
Hi @mmeendez8 ,
Are you using the newest `cocoapi` and the same environment as those in CI? I do not find any bug after reading the code yet. It is weird, but we cannot merge this PR if it cannot pass CI.

mmeendez8:
Hi @ZwwWayne ,
I have installed a fresh version of mmdet following installation instructions and everything works fine. I think there must be something wrong with Continuous Integration. I review the code and run locally multiple times but can't figure out what's going on.

ZwwWayne:
This PR is closed because it is continued with more refactoring in  #3695.

PR #3153, opened 2020-06-30
Add allow_negative_crop in RandomCrop

Johnson-Wang:
1. Add allow_negative_crop in RandomCrop
2. Resolve a TODO: check whether we can keep the image regardless of the crop.
3. Find out the reason of a sudden boost of nas-fpn performance using the RandomCrop in #2810. In #2810, the allow_negative_crop was implicitly set to True when there is `gt_bboxes_ignore` flag and `gt_bboxes_ignore` is an empty tensor. 
I will keep `gt_bboxes_ignore=False` in this version and provide the performance with  `gt_bboxes_ignore=True` in the nas-fpn README. 

### Conclusion after re-running the experiments
The performance is very close for coco dataset with `gt_bbox_ignore=True` or `gt_bbox_ignore=False`. Both of these are higher than that reported in the model zoo

| | before #2810  | #2810   | `gt_bbox_ignore=True`   |  `gt_bbox_ignore=False` | 
|---|---|---|---|---|
| retinanet_fpn_crop640  | 37.4  | 37.9  | 37.8  | 37.9  |
| retinanet_nas_fpn_crop640  | 40.2  | 40.5  | 40.6  | 40.7  |

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -469,8 +469,8 @@ class RandomCrop(object):
 
     Args:
         crop_size (tuple): Expected size after cropping, (h, w).
-        allow_negative_crop (bool): Whether to allow a crop whose center is not
-            located in any bbox. Default to False.
+        allow_negative_crop (bool): Whether to allow a crop that does not
+            contain any bbox area. Default to False.
 
     Notes:
         - If the image is smaller than the crop size, return the original image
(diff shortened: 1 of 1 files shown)


mmdet/datasets/pipelines/transforms.py
```diff
@@ -469,20 +469,23 @@ class RandomCrop(object):
...
 
     Notes:
         - If the image is smaller than the crop size, return the original image
         - The keys for bboxes, labels and masks must be aligned. That is,
           `gt_bboxes` corresponds to `gt_labels` and `gt_masks`, and
           `gt_bboxes_ignore` corresponds to `gt_labels_ignore` and
           `gt_masks_ignore`.
-        - If there are gt bboxes in an image and the cropping area does not
-          have intersection with any gt bbox, this image is skipped.
+        - If the crop is not located in any bbox and `allow_negative_crop` is
```
hellock, line 481: The description `the crop is not located in any bbox` is unclear.

PR #3155, opened 2020-06-30
More robust Accuracy module

Johnson-Wang:
1. Make mmdet.models.Accuracy class more robust to empty predictions. 
2. Add score threshold for Accuracy calculation
2. Add unittest for Accuracy module
3. change accuracy to pos_recall in FSAF and use Accuracy module to calculate it.


mmdet/models/losses/accuracy.py
```diff
@@ -1,15 +1,17 @@
...
     Args:
-        pred (torch.Tensor): The model prediction.
-        target (torch.Tensor): The target of each prediction
+        pred (torch.Tensor): The model prediction, shape (N, num_class)
+        target (torch.Tensor): The target of each prediction, shape (N, )
         topk (int | tuple[int], optional): If the predictions in ``topk``
             matches the target, the predictions will be regarded as
             correct ones. Defaults to 1.
+        thresh (float, optional): If not None, threshold for pred with scores
+            under this value being incorrect. Default to None.
```
yhcao6, line 14: This sentence is a little confusing. Is it better to say this?
`If not None, predictions with scores under this threshold are to be considered incorrect`

mmdet/models/losses/accuracy.py
```diff
@@ -39,15 +50,18 @@ def accuracy(pred, target, topk=1):
...
 
-    def __init__(self, topk=(1, )):
+    def __init__(self, topk=(1, ), thresh=None):
         """Module to calculate the accuracy
 
         Args:
             topk (tuple, optional): The criterion used to calculate the
                 accuracy. Defaults to (1,).
+            thresh (float, optional): If not None, threshold for prediction
+                scores under this value being incorrect. Default to None.
```
yhcao6, line 60: Need to also modify here.

Issue #3008, opened 2020-06-13
corner pooling and `torch.cummax`


Scalsol:
Hi, I notice that you are adding corner pooling module as a new feature. But in PyTorch 1.5, a new function named `torch.cummax` is introduced, which I think behaves just like corner pooling. As a known fact that corner pooling runs at a slow speed, would `torch.cummax` runs faster?

hellock:
Thanks for the reference. @tianyuandu may have a look.

tianyuandu:
Hi @Scalsol , thanks for your suggestion. I found the output of `torch.cummax` and our corner pool is same, but the grad is mismatching.

Here is the verification script:
```python
import torch

from mmdet.ops import CornerPool


if __name__ == '__main__':
    input_tensor = torch.tensor([[[[0, 0, 0, 0, 0], [2, 1, 3, 0, 2], 
                                   [5, 4, 1, 1, 6], [0, 0, 0, 0, 0], 
                                   [0, 0, 0, 0, 0]]]]).float().cuda()
    cum_tensor = input_tensor.detach().requires_grad_()
    pool_tensor = input_tensor.detach().requires_grad_()

    pool = CornerPool('bottom')
    pool_out = pool(pool_tensor)
    pool_loss = pool_out.sum()
    pool_loss.backward()

    cum_out = torch.cummax(cum_tensor, dim=2)[0]
    cum_loss = cum_out.sum()
    cum_loss.backward()

    print("pool_out:")
    print(pool_out)
    print("cum_out:")
    print(cum_out)
    print("pool_tensor's grad:")
    print(pool_tensor.grad)
    print("cum_tensor's grad:")
    print(cum_tensor.grad)
```

and the output is:
```python
pool_out:
tensor([[[[0., 0., 0., 0., 0.],
          [2., 1., 3., 0., 2.],
          [5., 4., 3., 1., 6.],
          [5., 4., 3., 1., 6.],
          [5., 4., 3., 1., 6.]]]], device='cuda:0',
       grad_fn=<BottomPoolFunctionBackward>)
cum_out:
tensor([[[[0., 0., 0., 0., 0.],
          [2., 1., 3., 0., 2.],
          [5., 4., 3., 1., 6.],
          [5., 4., 3., 1., 6.],
          [5., 4., 3., 1., 6.]]]], device='cuda:0', grad_fn=<CummaxBackward>)
pool_tensor's grad:
tensor([[[[1., 1., 1., 2., 1.],
[...]

PR #3158, opened 2020-06-30
use torch.cummax in corner_pool for torch 1.5+

tianyuandu:
#3008


mmdet/ops/corner_pool/corner_pool.py
```diff
@@ -92,10 +93,29 @@ class CornerPool(nn.Module):
...
+        'right': [3, False],
+        'top': [2, True],
+    }
+
     def __init__(self, mode):
         super(CornerPool, self).__init__()
         assert mode in self.pool_functions
-        self.corner_pool = self.pool_functions[mode]
+        if torch.__version__ >= '1.5.0':
+            self.corner_pool = self.cummax_dim_flip[mode]
```
hellock, line 107: This is quite weird. I suggest saving `mode` and get dim and flip in `forward()`.

mmdet/ops/corner_pool/corner_pool.py
```diff
@@ -92,10 +93,27 @@ class CornerPool(nn.Module):
         'top': TopPoolFunction,
     }
 
+    cummax_dim_flip = {
+        'bottom': [2, False],
+        'left': [3, True],
+        'right': [3, False],
+        'top': [2, True],
```
hellock, line 100: It makes more sense to use tuple instead of list.

hellock:
This change should also be submitted to mmcv, since we will switch to use all ops from mmcv.

hellock:
It has been merged to mmcv.

PR #3187, opened 2020-07-03
Fix initialization bug in DetectoRS

joe-siyuan-qiao:
Dear developers,

In the previous merge of DetectoRS, I noticed that ``DetectoRS with Cascade + ResNet-50`` isn't very close to what I reported. So I did a quick examination offline to find out what the reasons might be. It turns out that the issue is due to the init_weights function in RFP which alters the default initialization of SAC layers in self.rfp_modules when RFP and SAC are used at the same time. This bug affects the following performances reported in README.md.

| Method | Detector |
|:------:|:--------:|
| DetectoRS | Cascade + ResNet-50 |
| DetectoRS | HTC + ResNet-50 |

It could also lead to NaN when larger backbones are used. So I create this PR to fix this issue.

Thanks a lot for your help!

Cheers!


mmdet/models/necks/rfp.py
```diff
@@ -95,7 +95,10 @@ def __init__(self,
             bias=True)
 
     def init_weights(self):
-        super().init_weights()
+        for convs in [self.lateral_convs, self.fpn_convs]:
```
xvjiarui, line 98: We may add comments about this to remind other developers.

xvjiarui:
Hi @joe-siyuan-qiao 
Thanks for the fix. We will benchmark these two models again to see if the performance will be the same as reported.

PR #3192, opened 2020-07-03
fix collect env doc

yhcao6:


dreamerlin:
It seems that `reimplementation_questions.md` should also be fixed.

PR #3208, opened 2020-07-05
Fix gt priority tiny bug

simonJJJ:
The original method calculates gt priority has a tiny bug, it actually outputs the rank, not the priority.
E.g. areas=tensor([200., 300., 500., 400., 100.])
The sort_idx would be tensor([2, 3, 1, 0, 4]). But the priority is actually tensor([3, 2, 0, 1, 4]) (Due to the gt areas). So I use an torch.argsort() method to get it.


Johnson-Wang:
Hi, thanks for the contribution. Have you run the baseline after fixing this?

simonJJJ:
> Hi, thanks for the contribution. Have you run the baseline after fixing this?

Nope. I also want to see the performance gap due to the fix.

Johnson-Wang:
I have tried to re-benchmark the performance with the fix. The AP of Resnet50 does not change (still 37.4), while that of Resnet50 with 0.2-0.5 treated as ignored decreases a bit (from 37.0 to 36.0). Performance of other backbone is still pending.

Johnson-Wang:
Update on the performance. 
|                   | R50-ignore range 0.2-0.5 | R50 | R101|
|----------------|-----------------------------------|-------|---------|
|Before fix    |       37.0                            |37.4 |  39.3  |
|After fix       |      36.0                             | 37.4|  39.4  |

PR #3221, opened 2020-07-06
Clean background_labels in the dense heads

ZwwWayne:
This PR unifies the label assign strategies in DenseHeads. Original codebase faces a dilemma that the normal CrossEntropyLoss could only be used for RPN head if `use_sigmoid=True` because the RPN head assigns positive samples with label 1 and 0 for those negative ones. However, in other places of the code, following the new convention, label in [0, num_classes-1] are positive samples and label equals to num_classes are negative samples. This introduces bugs in some situations, e.g., one-class detection using sigmoid. 

This PR fixes this issue by assigning 0 to positive samples and 1 to negative samples in RPN, which is consistent with the situation for single-class classification. Thus, the `background_label` is no longer necessary and we uniformly use `num_classes` to indicate background labels in the whole codebase.

**Note**: The PR has no effect on the pre-trained models in the model zoo but will affect the model training of many detectors. Two-stage detectors whose RPN uses softmax will be affected because the order of category is changed. We benchmarked related models to ensure the fix will not introduce new bugs. The performance of SABL, FCOS, FreeAnchor, Guided Anchoring, FSAF, PAA, ATSS, RepPoints, RetinaNet, GFL, SSD, Mask R-CNN, GHM, PISA, and Dynamic R-CNN are verified as normal as before.

Resolve https://github.com/open-mmlab/mmdetection/issues/3629


mmdet/models/dense_heads/fsaf_head.py
```diff
@@ -124,7 +124,7 @@ def _get_targets_single(self,
             pos_gt_inds[pos_inds] = sampling_result.pos_assigned_gt_inds
             if gt_labels is None:
                 # only rpn gives gt_labels as None, this time FG is 1
```
Johnson-Wang, line 126: change the comments as well

mmdet/models/dense_heads/anchor_head.py
```diff
@@ -250,8 +240,8 @@ def _get_targets_single(self,
             bbox_targets[pos_inds, :] = pos_bbox_targets
             bbox_weights[pos_inds, :] = 1.0
             if gt_labels is None:
-                # only rpn gives gt_labels as None, this time FG is 1
-                labels[pos_inds] = 1
+                # Foreground is the first class since v2.5.0
```
Johnson-Wang, line 243: \# Only rpn gives gt_labels as None. Foreground is the first class since v2.5.0.

mmdet/models/dense_heads/atss_head.py
```diff
@@ -609,7 +609,7 @@ def _get_target_single(self,
             bbox_targets[pos_inds, :] = pos_bbox_targets
             bbox_weights[pos_inds, :] = 1.0
             if gt_labels is None:
-                labels[pos_inds] = 1
+                labels[pos_inds] = 0
```
Johnson-Wang, line 612: Add comments here as well.

Johnson-Wang (approved):
Overall speaking, it is okay to me. But full regression benchmark should be run before merging.

hellock:
Task linked: [CU-4hu3jh Clean BG labels (#3221)](https://app.clickup.com/t/4hu3jh)

PR #3224, opened 2020-07-06
Fix sampling_result classmethod typo

simonJJJ:
A tiny typo in sampling_reult.py where sample classmethod initializes a RandomSampler instance.


hellock:
Thanks for the fix!

PR #3234, opened 2020-07-07
Fix bug of ordered list in getting_started

ZwwWayne:
This PR fixes some tweaks in the docs.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -318,7 +318,14 @@ CUDA_VISIBLE_DEVICES=4,5,6,7 PORT=29501 ./tools/dist_train.sh ${CONFIG_FILE} 4
 
 If you use launch training jobs with Slurm, there are two ways to specify the ports.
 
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -317,24 +316,35 @@ CUDA_VISIBLE_DEVICES=0,1,2,3 PORT=29500 ./tools/dist_train.sh ${CONFIG_FILE} 4
 CUDA_VISIBLE_DEVICES=4,5,6,7 PORT=29501 ./tools/dist_train.sh ${CONFIG_FILE} 4
 ```
 
-If you use launch training jobs with Slurm, you need to modify the config files (usually the 6th line from the bottom in config files) to set different communication ports.
+If you use launch training jobs with Slurm, there are two ways to specify the ports.
```
hellock, line 319: Option 2 is simpler, we can put it first.

PR #3232, opened 2020-07-07
clean op

yhcao6:

PR #3244, opened 2020-07-08
Fixed bugs about cascade_roi_head.

Jokoe66:
* Fixed the bug of refining bg rois at training when reg_class_agnostic is False.

* Fixed the bug of potentially infering roi label to be bg at inference.


ZwwWayne:
Hi @Jokoe66 ,
Thanks for your contribution. It seems that the modification changes the training logic if `self.reg_class_agnostic=False`. Have you validate the performance?

Jokoe66:
Hi, I don't validate it yet due to gpu&nbsp;limit.
Besides, I think the training logic of class-aware cascade roi head needs further modification. Under current modification, class-aware cascade roi head does not refine negative rois at training, but do it at inference by regarding them as their closest objects. I think it's better to keep the behavior at training consistent with that at inference. So I will change the gt_label of rois used to refine rois at training to what will be at inference.

------------------ Original ------------------

mmdet/ops/conv_ws.py
```diff
@@ -20,7 +19,6 @@ def conv_ws_2d(input,
     return F.conv2d(input, weight, bias, stride, padding, dilation, groups)
 
 
-@CONV_LAYERS.register_module(name='ConvWS')
```
Jokoe66, line 23: It seems this module has been registered elsewhere, and the last build failed. I am not clear why earlier builds succeeded.

hellock: This is due to the version of mmcv. v0.6.2 works for the current version and v1.0 will work after #3232 is merged.

Jokoe66: OK. I have merged latest commits into this branch.

ZwwWayne:
The code looks good to me now. I will benchmark some models to validate the modification.

Jokoe66:
> The code looks good to me now. I will benchmark some models to validate the modification.

Sure. Thanks.

ZwwWayne:
The performance seems OK. I ran cascade mask r50 three times and obtained 41.1/35.9, 41.2/35.9, and 41.1/35.8 box/mask mAP.

PR #3333, opened 2020-07-17
Fixed BC of stem_channels

xvjiarui:


mmdet/models/backbones/resnet.py
```diff
@@ -306,7 +306,8 @@ class ResNet(nn.Module):
 
     Args:
         depth (int): Depth of resnet, from {18, 34, 50, 101, 152}.
-        stem_channels (int): Number of stem channels. Default: 64.
+        stem_channels (int | None): Number of stem channels. If not specified,
+            will be the same as `base_channels`. Default: None.
```
Johnson-Wang, line 310: It will be the same as ...

xvjiarui:
> It seems good to me in regards to the code. It's better to check configs that need to set `stem_channels` explicitly. BTW, how come I didn't find any configs with `stem_channels` arguments?

The default value works for all configs. I just did a global search and didn't find any config need modification.

Issue #3048, opened 2020-06-17
TypeError: object of type 'DataContainer' has no len()


shliang0603:
@ZwwWayne I test in mmdetection2.0, I have only CPU. My test commad following:
>python tools/test.py ./configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py ./work_dirs/latest.pth  --show-dir result --eval mAP

 I got the following output：
```python
which: no hipcc in (/HDD/anaconda3/envs/mmdetection/bin:/HDD/anaconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
[                                                  ] 0/169, elapsed: 0s, ETA:Traceback (most recent call last):
  File "tools/test.py", line 149, in <module>
    main()
  File "tools/test.py", line 127, in main
    args.show_score_thr)
  File "/HDD/project/mmdetection_hat/mmdet/apis/test.py", line 25, in single_gpu_test
    result = model(return_loss=False, rescale=True, **data)
  File "/HDD/anaconda3/envs/mmdetection/lib/python3.7/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "/HDD/anaconda3/envs/mmdetection/lib/python3.7/site-packages/torch/nn/parallel/data_parallel.py", line 143, in forward
    return self.module(*inputs, **kwargs)
  File "/HDD/anaconda3/envs/mmdetection/lib/python3.7/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "/HDD/project/mmdetection_hat/mmdet/core/fp16/decorators.py", line 51, in new_func
    return old_func(*args, **kwargs)
  File "/HDD/project/mmdetection_hat/mmdet/models/detectors/base.py", line 149, in forward
    return self.forward_test(img, img_metas, **kwargs)
  File "/HDD/project/mmdetection_hat/mmdet/models/detectors/base.py", line 130, in forward_test
    return self.simple_test(imgs[0], img_metas[0], **kwargs)
  File "/HDD/project/mmdetection_hat/mmdet/models/detectors/two_stage.py", line 187, in simple_test
    proposal_list = self.simple_test_rpn(x, img_metas)
  File "/HDD/project/mmdetection_hat/mmdet/models/detectors/test_mixins.py", line 29, in simple_test_rpn
[...]

PR #3391, opened 2020-07-23
HRNet keypoints head (Facial landmarks)

SebastienLinker:
This PR is a re implementation of HR Net keypoints detection (Facial landmarks): https://github.com/HRNet/HRNet-Facial-Landmark-Detection
Some ideas from Mask RCNN tentative: https://github.com/powerlic/mmdetection/commit/e7773ca738c2ff35cf7d38af117464d445a6edbc?diff=unified

The MSE loss is calculated on the heatmaps
Outputs should be a N*3 vector for N keypoints, with 2 coordinates and the confidence level for each keypoint

## Current status

Trained and tested on WFLW dataset: https://wywu.github.io/projects/LAB/WFLW.html
Note that a small sigma (on the heatmap) will make it more difficult to adapt
COCO dataset has a visibility flag for each keypoint, we added this possibility and the input data should be a N*3 array similar to MSCOCO input

## TODO

- [x] Keypoints head
- [x] Check results on non square images, non square heatmaps
- [x] Double check heatmaps decoding, we may have a 1 or 2 pixels errors when going back to the original image
- [x] Data augmentation
- [x] Output confidence level
- [x] Visibility flag
- [ ] Test on CPU, see: https://github.com/open-mmlab/mmdetection/issues/3048
- [x] Unit tests
- [x] Asyncio interface
- [ ] Unit tests for data augmenters
- [x] Normalize keypoints errors, interocular distance similar to HRNet paper

## Ideas

- [x] Heeatmaps: We can add more flexibility and specify a decoding function in the config file
- [ ] BBox RoI extractor is not used in the original repo of HRNet facial landmarks, but we might need it in another project, can be added here following some ROI extractors already implemented
- [x] Re use necks, similarly to the implementation of other HRNets in mmdetection, different from the original implementation
- [ ] Use MS COCO dataset (could be used to test RoI extractor and allow comparison with more methods)
- [ ] Dataset WFLW: Similar to the original HRNet paper, we focus on a small area of the image. This is done by calling `CenterOnFace` pipeline. What we can do is convert the dataset and load a small image already centered on the face, with the proper annotation.
- [ ] RetinaFace: Which combines 5 keypoints and face bbox: https://github.com/deepinsight/insightface/tree/master/RetinaFace

## Notes

- The network used in HRNet keypoints detection does not have any neck, while the one used in for bboxes and masks has. (I included 2 examples, with and without neck under `config` folder, accuracy is similar) The original ones is available here: https://github.com/HRNet/HRNet-Object-Detection/blob/master/mmdet/models/backbones/hrnet.py#L442
- This model does not output any bbox, few changes had to be made in the global structure:
    - Some assertions have been removed to allow a model without any keypoint
    - The output can be either a dictionary with the following keys (`bbox`, `mask`, `keypoints`, `heatmaps`), or the original format: a numpy array (bboxes only) or a tuple (np.array, list of dicts) for masks
    - Test function in `mmdet/apis/test.py` had to be modified accordingly to handle and display keypoints (TODO: Check whether mmcv can handle keypoints directly)
- The WFLW dataset has 98 keypoints per image. This number is a bit high and it may be slow to converge especially if you train from scratch (although it eventually converges). You can reduce the number of keypoints to get faster convergence. This should be done both in the config file (`model/roi_head/keypoint_head/num_keypoints`) and the `WflwDataset` code under `load_annotations`
- At train time, heatmaps are generated in the dataloader to speedup. However you need to use `keep_ratio=False` if the batch size is > 1, or it might create errors due to the concatenation of heatmaps of different sizes (depends on your original image size)
- I noticed you guys recently started a new `mmpose` project, but we started working on that some time ago. Feel free to move it to `mmpose` if you think it should go there.

### DIFF
--- a/configs/_base_/datasets/wflw.py
+++ b/configs/_base_/datasets/wflw.py
@@ -1,0 +1,64 @@
+dataset_type = 'WflwDataset'
+data_root = '/mnt/data-home/sebastien/data/public/WFLW/'
+
+img_scale = (384, 384)
+img_norm_cfg = dict(
+    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
+train_pipeline = [
+    dict(type='LoadImageFromFile', to_float32=True),
+    dict(type='FocusOnFace', out_size=384),
+    dict(type='LoadAnnotations', with_bbox=True, with_keypoint=True),
+    dict(type='Resize', img_scale=img_scale, keep_ratio=False),
+    dict(type='PhotoMetricDistortion'),
+    dict(type='Corrupt', corruption='gaussian_noise'),
+    dict(type='Corrupt', corruption='jpeg_compression'),
+    dict(type='RandomFlip', flip_ratio=0.),
+    dict(type='Normalize', **img_norm_cfg),
+    dict(type='Pad', size_divisor=32),
+    dict(type='KeypointsToHeatmaps', sigma=3),
+    dict(type='DefaultFormatBundle'),
+    dict(
+        type='Collect',
+        keys=['img', 'gt_bboxes', 'gt_labels', 'gt_keypoints', 'heatmaps']),
+]
+test_pipeline = [
+    dict(type='LoadImageFromFile'),
+    dict(type='FocusOnFace', out_size=384),
+    dict(
+        type='MultiScaleFlipAug',
+        img_scale=img_scale,
+        flip=False,
+        transforms=[
+            dict(type='Resize', img_scale=img_scale, keep_ratio=False),
+            dict(type='RandomFlip', flip_ratio=0.),
+            dict(type='Normalize', **img_norm_cfg),
+            dict(type='Pad', size_divisor=32),
+            dict(type='ImageToTensor', keys=['img']),
+            dict(type='Collect', keys=['img', 'proposals']),
+        ])
+]
+data = dict(
+    samples_per_gpu=10,
+    workers_per_gpu=20,
+    train=dict(
+        type=dataset_type,
+        ann_file=data_root +
+        'WFLW_annotations/list_98pt_rect_attr_train_test/' +
+        'list_98pt_rect_attr_train.txt',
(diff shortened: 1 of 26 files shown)


joihn:
Hello, any news ?

SebastienLinker:
> Hello, any news ? A merge would be helpfull to use a keypoint regression in combination with the latest models from the zoo :)

Well it is more than one year old now, I am not sure how much work is needed to make it compatible with latest mmdet.
Sorry I don't really have time to keep it up to date. Feel free to try and update if necessary :)

23119841:
> > Hello, any news ? A merge would be helpfull to use a keypoint regression in combination with the latest models from the zoo :)
> 
> Well it is more than one year old now, I am not sure how much work is needed to make it compatible with latest mmdet. Sorry I don't really have time to keep it up to date. Feel free to try and update if necessary :)

Thanks for your contribution.  Is it possible to use this keypoints header for the swin transformer which based on mmdection?

SebastienLinker:
> > > Hello, any news ? A merge would be helpfull to use a keypoint regression in combination with the latest models from the zoo :)
> > 
> > 
> > Well it is more than one year old now, I am not sure how much work is needed to make it compatible with latest mmdet. Sorry I don't really have time to keep it up to date. Feel free to try and update if necessary :)
> 
> Thanks for your contribution. Is it possible to use this keypoints header for the swin transformer which based on mmdection?

I guess you can do some refactoring and get a keypoints head compatible with the swin transformer

23119841:
> > > > Hello, any news ? A merge would be helpfull to use a keypoint regression in combination with the latest models from the zoo :)
> > > 
> > > 
> > > Well it is more than one year old now, I am not sure how much work is needed to make it compatible with latest mmdet. Sorry I don't really have time to keep it up to date. Feel free to try and update if necessary :)
> > 
> > 
> > Thanks for your contribution. Is it possible to use this keypoints header for the swin transformer which based on mmdection?
> 
> I guess you can do some refactoring and get a keypoints head compatible with the swin transformer

Thanks for your advice!

PR #3392, opened 2020-07-23
Imresize backend

Johnson-Wang:
Enable flexible backend for imresize


hellock:
There are also other transforms like `SegRescale` which needs to be updated.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -599,7 +616,10 @@ def __call__(self, results):
         for key in results.get('seg_fields', []):
             if self.scale_factor != 1:
                 results[key] = mmcv.imrescale(
-                    results[key], self.scale_factor, interpolation='nearest')
+                    results[key],
+                    self.scale_factor,
+                    interpolation='nearest',
+                    backend=self.backend)
```
hellock, line 622: `self.backend` is undefined.

hellock:
We need to update the minimum version of mmcv in requirements and readme.

Johnson-Wang:
Since there is currently no version requirement for mmcv in requirements and readme, I can wait for the next version bump of mmdetection to merge this pr.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -72,6 +76,7 @@ def __init__(self,
             # mode 2: given multiple scales or a range of scales
             assert multiscale_mode in ['value', 'range']
 
+        self.backend = backend
```
ZwwWayne, line 79: We need to add an assertion here to check the backend is in `['cv2', 'pillow']`.

Johnson-Wang: I think it might be better to just leave it as it is here, snce it has been correctly checked in `mmcv.imresize` function. It also allows future extension of adding new imresize backends by just modifying codes in mmcv.

PR #3395, opened 2020-07-24
Add custom-defined hooks in train api

Johnson-Wang:
- Allows users to define hooks in configs
- Reformatted.


mmdet/apis/train.py
```diff
@@ -121,6 +122,14 @@ def train_detector(model,
         eval_hook = DistEvalHook if distributed else EvalHook
         runner.register_hook(eval_hook(val_dataloader, **eval_cfg))
 
+    # user-defined hooks
+    if cfg.get('hooks', None):
```
hellock, line 126: Rename the field `hooks` to `additional_hooks` or `custom_hooks`?

PR #3414, opened 2020-07-26
add pat change

yhcao6:


mmdet/core/fp16/hooks.py
```diff
@@ -127,7 +127,8 @@ def patch_norm_fp32(module):
     """
     if isinstance(module, (nn.modules.batchnorm._BatchNorm, nn.GroupNorm)):
         module.float()
-        if isinstance(module, nn.GroupNorm) or torch.__version__ < '1.3':
+        if (isinstance(module, nn.GroupNorm) or torch.__version__ < '1.3'
```
hellock, line 130: The condition `torch.__version__ < '1.3'` can be removed since we only support pytorch 1.3+.

mmdet/models/roi_heads/roi_extractors/single_level_roi_extractor.py
```diff
@@ -57,6 +57,9 @@ def forward(self, feats, rois, roi_scale_factor=None):
         num_levels = len(feats)
         roi_feats = feats[0].new_zeros(
             rois.size(0), self.out_channels, *out_size)
+        # TODO: remove this when parrots supports
+        if torch.__version__ == 'parrots':
+            roi_feats.requires_grad = True
```
hellock, line 62: Does parrots support:
```python
roi_feats = feats[0].new_zeros(
    rois.size(0), self.out_channels, *out_size, requires_grad=True)
```

PR #3447, opened 2020-07-31
Add SOLO for instance segmentation

WXinlong:
Dear mmdet,

This PR adds support for SOLO(v1), including vanilla head and decoupled head.
The pretrained models can be directly downloaded from the[ SOLO repo](https://github.com/WXinlong/SOLO). 
Can you help review and benchmark this PR?
Thanks.:)


hellock:
Thanks for the nice work! We will have a review soon. You may fix the linting errors.

configs/solo/decoupled_solo_r50_fpn_8gpu_1x.py
```diff
@@ -0,0 +1,126 @@
+# model settings
```
ZwwWayne, line 1: For config filenames, we usually use `8x2` rather than `8gpu` to also indicate the batch size as indicated [here](https://github.com/open-mmlab/mmdetection/blob/master/docs/config.md). If the config uses the default setting, i.e., 8 GPUS with 2 batch size per GPU, we can omit `8x2` or `8gpu` for simplicity.

configs/solo/decoupled_solo_r50_fpn_8gpu_1x.py
```diff
@@ -0,0 +1,126 @@
...
+test_cfg = dict(
+    nms_pre=500,
+    score_thr=0.1,
+    mask_thr=0.5,
+    update_thr=0.05,
+    kernel='gaussian',  # gaussian/linear
+    sigma=2.0,
+    max_per_img=100)
+# dataset settings
+dataset_type = 'CocoDataset'
```
ZwwWayne, line 51: It seems the data pipeline, dataset, schedule, and runtime configs are very similar to the default setting. These configs could inherit the setting of the dataset/schedule from `_base_` configs for simplicity.

mmdet/models/dense_heads/decoupled_solo_head.py
```diff
@@ -0,0 +1,505 @@
...
+from scipy import ndimage
+
+def points_nms(heat, kernel=2):
+    # kernel must be 2
+    hmax = nn.functional.max_pool2d(
+        heat, (kernel, kernel), stride=1, padding=1)
+    keep = (hmax[:, :, :-1, :-1] == heat).float()
+    return heat * keep
+
+def dice_loss(input, target):
```
ZwwWayne, line 22: `dice_loss` is duplicated in `decoupled_solo_head` and `solo_head`. It can be moved into `mmdet.losses` as a `DiceLoss` module to reduce duplicated codes.

mmdet/models/dense_heads/decoupled_solo_head.py
```diff
@@ -0,0 +1,505 @@
...
+        self.loss_cate = build_loss(loss_cate)
+        self.ins_loss_weight = loss_ins['loss_weight']
+        self.conv_cfg = conv_cfg
+        self.norm_cfg = norm_cfg
+        self.train_cfg = train_cfg
+        self.test_cfg = test_cfg
+        self._init_layers()
+
+    def _init_layers(self):
+        norm_cfg = dict(type='GN', num_groups=32, requires_grad=True)
```
ZwwWayne, line 82: `norm_cfg` here should be arguments that can be set through configs. Can we use `self.norm_cfg`?

mmdet/models/dense_heads/decoupled_solo_head.py
```diff
@@ -0,0 +1,505 @@
...
+                 seg_feat_channels=256,
+                 stacked_convs=4,
+                 strides=(4, 8, 16, 32, 64),
+                 base_edge_list=(16, 32, 64, 128, 256),
+                 scale_ranges=((8, 32), (16, 64), (32, 128), (64, 256), (128, 512)),
+                 sigma=0.4,
+                 num_grids=None,
+                 cate_down_pos=0,
+                 background_label=None,
+                 loss_ins=None,
```
ZwwWayne, line 50: To unify the name of losses with other heads. It might be better to use `loss_mask` and `loss_cls`.

mmdet/models/detectors/single_stage_seg.py
```diff
@@ -0,0 +1,110 @@
+import torch.nn as nn
+
+from mmdet.core import bbox2result
+from ..builder import DETECTORS, build_backbone, build_head, build_neck
+from .base import BaseDetector
+
+
+@DETECTORS.register_module()
+class SingleStageSegDetector(BaseDetector):
```
ZwwWayne, line 9: Is it possible that we do not write a new `SingleStageSegDetector` but modify the existing `SingleStageDetector` to have a `mask_head`? It could make the `SingleStageDetector` more general and it seems we only need to add a few lines of code for `mask_head`.

mmdet/models/dense_heads/decoupled_solo_head.py
```diff
@@ -0,0 +1,505 @@
...
+    target = target.contiguous().view(target.size()[0], -1).float()
+
+    a = torch.sum(input * target, 1)
+    b = torch.sum(input * input, 1) + 0.001
+    c = torch.sum(target * target, 1) + 0.001
+    d = (2 * a) / (b + c)
+    return 1-d
+
+@HEADS.register_module()
+class DecoupledSOLOHead(BaseDenseSegHead):
```
ZwwWayne, line 33: Is it possible to make `DecoupledSOLOHead` inherit from `SOLOHead` to reduce some duplicated codes?

configs/solo/decoupled_solo_r50_fpn_1x_coco.py
```diff
@@ -0,0 +1,69 @@
...
+optimizer_config = dict(grad_clip=dict(max_norm=35, norm_type=2))
+# learning policy
+lr_config = dict(
+    policy='step',
+    warmup='linear',
+    warmup_iters=500,
+    warmup_ratio=1.0 / 3,
+    step=[9, 11])
+# runtime settings
+device_ids = range(8)
```
ZwwWayne, line 67: We do not need to write `device_ids` and `work_dir` in the config, this will be generated by the training scripts automatically.

configs/solo/decoupled_solo_r50_fpn_1x_coco.py
```diff
@@ -0,0 +1,69 @@
...
+work_dir = './work_dirs/decoupled_solo_release_r50_fpn_8gpu_1x'
+find_unused_parameters = True
```
ZwwWayne, line 69: `find_unused_parameters=True` sometimes will slow down the training and is only necessary when there are some parameters that will not be trained during training sometimes. If this option is not necessary for SOLO, we can omit this setting.

mmdet/core/post_processing/matrix_nms.py
```diff
@@ -0,0 +1,58 @@
...
+
+def matrix_nms(seg_masks, cate_labels, cate_scores,
+               kernel='gaussian', sigma=2.0, sum_masks=None):
+    """Matrix NMS for multi-class masks.
+
+    Args:
+        seg_masks (Tensor): shape (n, h, w)
+        cate_labels (Tensor): shape (n), mask labels in descending order
+        cate_scores (Tensor): shape (n), mask scores in descending order
+        kernel (str):  'linear' or 'gauss'
```
ZwwWayne, line 12: 'linear' or 'gaussian'. The docstring should be consistent with the argument.

WXinlong: fixed.:)

configs/solo/decoupled_solo_r50_fpn_1x_coco.py
WXinlong: updated.:)

mmdet/core/post_processing/matrix_nms.py
```diff
@@ -0,0 +1,58 @@
...
+    # matrix nms
+    if kernel == 'gaussian':
+        decay_matrix = torch.exp(-1 * sigma * (decay_iou ** 2))
+        compensate_matrix = torch.exp(-1 * sigma * (compensate_iou ** 2))
+        decay_coefficient, _ = (decay_matrix / compensate_matrix).min(0)
+    elif kernel == 'linear':
+        decay_matrix = (1-decay_iou)/(1-compensate_iou)
+        decay_coefficient, _ = decay_matrix.min(0)
+    else:
+        raise NotImplementedError
```
ZwwWayne, line 54: NotImplementedError -> NotImplementedError(f'{kernel} is not supported in matrix nms!')

ZwwWayne:
Since MMDetection V2.0, we have more strict requirements on the unit tests for newly added modules. Since SOLO head is very important, could you add unit tests to test the SOLOHead and DecoupledSOLOHead？

ZwwWayne:
ping

ZwwWayne:
Continued in #5166

OpenMMLab-Assistant-007:
Hi
@WXinlong 
We are grateful for your efforts in helping improve this open-source project during your personal time. 

PR #3457, opened 2020-08-01
Add ops warpper

ZwwWayne:
This PR adds `mmdet.ops` back for back compatibility. In `mmdet.ops`, wrappers of `mmcv.ops` are added thus the projects dependent on mmdet do not need to change the origin codes.

### DIFF
--- a/mmdet/ops/__init__.py
+++ b/mmdet/ops/__init__.py
@@ -1,3 +1,6 @@
+# This file is added for back-compatibility. Thus, downstream codebase
+# could still use and import mmdet.ops.
+
 # yapf: disable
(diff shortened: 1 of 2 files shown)


mmdet/ops/__init__.py
```diff
@@ -0,0 +1,29 @@
+# yapf: disable
```
hellock, line 1: We may add a comment to indicate that this file is added for BC.

pecanjk:
for the latest mmcv 1.1.5 and mmdet v2.4.0. the mmdet.ops will import from mmcv.ops import (ContextBlock, Conv2d, ConvTranspose2d, ConvWS2d, and mmcv.ops has no ContextBlock. It still gives import error. How to solve this? Do I need manually change mmdet v2.4.0 mmdet.ops __init__.py file?

PR #3466, opened 2020-08-02
Remove git hash in __version__

ZwwWayne:


setup.py
```diff
@@ -66,16 +66,17 @@ def write_version_py():
 __version__ = '{}'
 short_version = '{}'
 version_info = ({})
+version_hash = "{}"
```
hellock, line 69: We do not need to write the version file dynamically. Version number can be hard coded and the git hash can be computed on the fly.

mmdet/utils/version_utils.py
```diff
@@ -0,0 +1,28 @@
+import os
+import subprocess
+
+
+def get_git_hash():
```
hellock, line 5: This method can be moved to mmcv.

mmdet/version.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Open-MMLab. All rights reserved.
+
+__version__ = '2.3.0rc0'
```
hellock, line 3: `short_version` and `version_info`

mmdet/version.py
```diff
@@ -0,0 +1,19 @@
...
+short_version = '2.3.0rc0'
+
+
+def parse_version_info():
+    version_info = []
+    for x in short_version.split('.'):
```
hellock, line 9: `short_version` is deprecated

mmdet/utils/collect_env.py
```diff
@@ -53,7 +53,7 @@ def collect_env():
     env_info['OpenCV'] = cv2.__version__
 
     env_info['MMCV'] = mmcv.__version__
-    env_info['MMDetection'] = mmdet.__version__
+    env_info['MMDetection'] = mmdet.__version__ + '+' + get_git_hash()
```
hellock, line 56: We may also use `get_git_hash()[:7]`

mmdet/version.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Open-MMLab. All rights reserved.
+
+__version__ = '2.3.0rc0'
+short_version = '2.3.0rc0'
```
hellock, line 4: `short_version = __version__`

PR #3475, opened 2020-08-03
fix: correct hrnet forward func

royxue:
According to the original hrnet implementation
according to https://github.com/leoxiaobin/deep-high-resolution-net.pytorch/blob/master/lib/models/pose_hrnet.py#L257

### DIFF
--- a/mmdet/models/backbones/hrnet.py
+++ b/mmdet/models/backbones/hrnet.py
@@ -184,7 +184,7 @@ class HRModule(nn.Module):
 
         x_fuse = []
         for i in range(len(self.fuse_layers)):
-            y = 0
+            y = x[0] if i == 0 else self.fuse_layers[i][0](x[0])
             for j in range(self.num_branches):
                 if i == j:
                     y += x[j]


xvjiarui:
Hi @royxue 
In the original implementation, `j` index starts from `1`, while ours starts from `0`. So it should be the same.

royxue:
@xvjiarui  Hi, Jerry
For result, it should be the same;
The reason why it's better to keep the same as original implementation is, when convert this HRNet to Caffe, current implementation will causing key missing error;
I think it might cause by this y initial value

ZwwWayne:
Since the implementation is the same, this PR will not be merged. The key missing error should be caused by other reasons. If it is confirmed that changing the y initial value is necessary. Feel free to reopen this PR and add a unit test to make sure the implementations in this PR is the same as the original one.

OpenMMLab-Assistant-007:
Hi
@royxue 
We are grateful for your efforts in helping improve this open-source project during your personal time. 

PR #3521, opened 2020-08-10
Implement Cutout

Chrisfsj2051:
Please note that this is slightly different from the current version in mmdet-projects. 
To be more specific, `n_holes` can be given as a tuple to specify the range and `cutout_shape` can be given as float value in order to be compatible with STAC.


mmdet/datasets/pipelines/transforms.py
```diff
@@ -1530,3 +1530,67 @@ def __repr__(self):
...
+
+    Args:
+        n_holes (int | tuple[int, int]): Number of regions to be dropped.
+            If it is given as a list, number of n_holes will be randomly
+            selected from `n_holes[0]` to `n_holes[1]`.
+        cutout_shape (tuple[int, int] | list[tuple[int, int]] |
+            tuple[float, float] | list[tuple[float, float]]): The candidate
+            shape of dropped regions. It can be
+                1. `tuple[int, int]`: use a fixed cutout shape.
+                2. `tuple[float, float]`: use a fixed ratio of image shape.
```
hellock, line 1547: I suggest adding an argument `cutout_ratio` for this case. `cutout_shape` and `cutout_ratio` cannot be both specified.

Johnson-Wang: I agree adding another `cutout_ratio` argument since `cutout_shape` now is too flexible and may be confusing.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -1530,3 +1530,67 @@ def __repr__(self):
...
+                3. `list[tuple[int, int]]`: random choose shape from list.
+                4. `list[tuple[int, int]]`: random choose ratio from list.
+            Default: (8, 8).
+        fill_in (tuple[int, int, int]): The value of pixel to fill in the
```
Johnson-Wang, line 1551: Does each component have to be `int` type?

mmdet/datasets/pipelines/transforms.py
```diff
@@ -1530,3 +1530,81 @@ def __repr__(self):
...
+class CutOut(object):
+    """CutOut operation.
+
+    Randomly drop some regions of image used in
+    `Cutout <https://arxiv.org/abs/1708.04552>`_.
+
+    Args:
+        n_holes (int | tuple[int, int]): Number of regions to be dropped.
+            If it is given as a list, number of n_holes will be randomly
+            selected from `n_holes[0]` to `n_holes[1]`.
```
hellock, line 1545: Indicate whether the interval is closed or open.

hellock:
More unit tests can be added to increase the coverage.

hellock:
Generally LGTM.

Chrisfsj2051:
Hi @hellock , the coverage has been improved. It seems like every statement (except for `__repr__` method) have been covered.

Chrisfsj2051:
Is there anything wrong with the latest CI? It failed to install CUDA.

Issue #3220, opened 2020-07-06
Support Evaluation over ConcatDataset


mmeendez8:
When combining multiple datasets for validation, evaluation function does not work.
`AttributeError: 'ConcatDataset' object has no attribute 'evaluate'
`

**Motivation**
Usually you download datasets for different sources and you keep them in the same format (say COCO). You can now train with all data using concat method but it would be very useful to validate on both too.

**Related resources**
None

**Additional context**

I am not sure about which would be the best approach... maybe edit `ConcatDataset` wrapper to merge different datasets first and then call parent class constructor. 

I am willing to implement this feature myself.

mmeendez8:
Any update with this?

PR #3522, opened 2020-08-10
Concat eval

ZwwWayne:
This PR supports to evaluate concatenated datasets.
Fix https://github.com/open-mmlab/mmdetection/issues/3220


mmdet/datasets/dataset_wrappers.py
```diff
@@ -50,6 +61,59 @@ def get_cat_ids(self, idx):
             sample_idx = idx - self.cumulative_sizes[dataset_idx - 1]
         return self.datasets[dataset_idx].get_cat_ids(sample_idx)
 
+    def evaluate(self, results, metric, logger=None, **kwargs):
```
Johnson-Wang, line 64: Better add a unittest to prevent possible bugs.

docs/tutorials/new_dataset.md
```diff
@@ -249,7 +249,7 @@ You may refer to [source code](../../mmdet/datasets/dataset_wrappers.py) for det
 
 ### Concatenate dataset
 
-There two ways to concatenate the dataset.
+There three ways to concatenate the dataset.
```
Johnson-Wang, line 252: There are ...

docs/tutorials/new_dataset.md
```diff
@@ -278,6 +287,27 @@ There two ways to concatenate the dataset.
...
+        train=dataset_A_train,
+        val=dict(
+            type='ConcatDataset',
+            datasets=[dataset_A_val, dataset_B_val],
+            separate_eval=False))
+    ```
+    This manner allows users to evaluate all the datasets as a single one by setting `separate_eval=False`.
+
+**Note:**
+This option `separate_eval=False` assumes the dataset uses `self.data_infos` during evaluation. Therefore, COCO datasets does notsupport this behavior since COCO datasets do not fully rely on `self.data_infos` during evaluation. Combining different type ofdatasets and evaluating them as a whole is not tested thus is not suggested.
```
Johnson-Wang, line 310: Check grammars, e.g. `Coco datasets does notsupport` --> `Coco dataset does not support`.

mmdet/datasets/dataset_wrappers.py
```diff
@@ -50,6 +62,60 @@ def get_cat_ids(self, idx):
...
+                    f'{len(results_per_dataset)} images now',
+                    logger=logger)
+
+                eval_results_per_dataset = dataset.evaluate(
+                    results_per_dataset, **kwargs)
+                dataset_idx += 1
+                for k, v in eval_results_per_dataset.items():
+                    total_eval_results.update({f'{dataset_idx}_{k}': v})
+
+            return total_eval_results
```
Johnson-Wang, line 103: Now the result is a dict of dict. Chances are that the TextLogger.average may fail if TextLogger.average is called. Please double check that.

docs/tutorials/new_dataset.md
Johnson-Wang, line 310: Add a note that repeat dataset or Concat(Repeat) currently do not support evaluation.

mmdet/datasets/dataset_wrappers.py
ZwwWayne: total_eval_results is still a dict that has similar keys as eval_results_per_dataset. It has similar data structure as those of normal datasets. The unit tests should validate that.

docs/tutorials/new_dataset.md
```diff
@@ -278,6 +287,28 @@ There two ways to concatenate the dataset.
...
+        val=dict(
+            type='ConcatDataset',
+            datasets=[dataset_A_val, dataset_B_val],
+            separate_eval=False))
+    ```
+    This manner allows users to evaluate all the datasets as a single one by setting `separate_eval=False`.
+
+**Note:**
+1. The option `separate_eval=False` assumes the datasets use `self.data_infos` during evaluation. Therefore, COCO datasets do not support this behavior since COCO datasets do not fully rely on `self.data_infos` for evaluation. Combining different types of ofdatasets and evaluating them as a whole is not tested thus is not suggested.
+2. Evaluating `ClassBalancedDataset` and `RepeatDataset` is not supported thus evaluating concatenated datasets of these types is also not supported.
```
hellock, line 311: Can we raise a clear error for these datasets?

mmdet/datasets/dataset_wrappers.py
```diff
@@ -19,9 +21,19 @@ class ConcatDataset(_ConcatDataset):
         datasets (list[:obj:`Dataset`]): A list of datasets.
     """
 
-    def __init__(self, datasets):
+    def __init__(self, datasets, separate_eval=True):
```
hellock, line 24: Add docstring for the new argument.

mmdet/datasets/dataset_wrappers.py
```diff
@@ -50,6 +65,65 @@ def get_cat_ids(self, idx):
             sample_idx = idx - self.cumulative_sizes[dataset_idx - 1]
         return self.datasets[dataset_idx].get_cat_ids(sample_idx)
 
+    def evaluate(self, results, metric='bbox', logger=None, **kwargs):
```
Johnson-Wang, line 68: An additional comments: This `metric` can be deleted and let `**kwargs` set it.

PR #3534, opened 2020-08-11
Add config to runner meta

Johnson-Wang:
'config' is added to runner meta at first.


tools/train.py
```diff
@@ -109,7 +109,8 @@ def main():
     logger.info('Environment info:\n' + dash_line + env_info + '\n' +
                 dash_line)
     meta['env_info'] = env_info
-
+    meta['config_dict'] = dict(cfg)
+    meta['config_file'] = args.config
```
hellock, line 113: Better to use the absolute path.

tools/train.py
```diff
@@ -156,7 +156,6 @@ def main():
         # checkpoints as meta data
         cfg.checkpoint_config.meta = dict(
             mmdet_version=__version__ + get_git_hash()[:7],
-            config=cfg.pretty_text,
```
Johnson-Wang, line 159: `meta['config']` is already added in line 133.

Johnson-Wang, line 159: `meta['config']` is added in line 133.

PR #3537, opened 2020-08-11
Add eval-option flag for testing

xvjiarui:


tools/test.py
```diff
@@ -54,7 +54,15 @@ def parse_args():
...
-        '--options', nargs='+', action=DictAction, help='arguments in dict')
+        '--options',
+        nargs='+',
+        action=DictAction,
+        help='custom options to overwrite config')
+    parser.add_argument(
+        '--eval-options',
+        nargs='+',
+        action=DictAction,
+        help='custom options for evaluation')
```
hellock, line 65: The description is hard to be digested by users.

Johnson-Wang: Are there any options that cannot be merged by the config?

xvjiarui: Basically speaking, any options could be merged by config. However, in practice, we just need to change the existing keys.

tools/test.py
```diff
@@ -54,7 +55,23 @@ def parse_args():
...
+        nargs='+',
+        action=DictAction,
+        help='override some settings in the used config, the key-value pair '
+        'in xxx=yyy format will be merged into config file.')
+    parser.add_argument(
+        '--options',
+        nargs='+',
+        action=DictAction,
+        help='custom options for evaluation, the key-value pair in xxx=yyy '
+        'format will be kwargs for dataset.evaluate() function (deprecate)')
```
Johnson-Wang, line 68: Add `change to --eval-options instead`

tools/test.py
```diff
@@ -142,11 +169,14 @@ def main():
...
             mmcv.dump(outputs, args.out)
-        kwargs = {} if args.options is None else args.options
+        kwargs = {} if args.eval_options is None else args.eval_options
         if args.format_only:
             dataset.format_results(outputs, **kwargs)
         if args.eval:
-            dataset.evaluate(outputs, args.eval, **kwargs)
+            eval_kwargs = cfg.get('evaluation', {}).copy()
+            eval_kwargs.pop('interval', None)
+            eval_kwargs.update(dict(metric=args.eval, **kwargs))
```
Johnson-Wang, line 178: It's a bit hard-coded, and it may also trigger some bugs if there are some other arguments specified in `evaluation=dict()` in the config.

tools/train.py
```diff
@@ -45,7 +46,15 @@ def parse_args():
...
-        '--options', nargs='+', action=DictAction, help='arguments in dict')
+        '--options',
+        nargs='+',
+        action=DictAction,
+        help='arguments in dict (deprecated), change to --cfg-options instead')
+    parser.add_argument(
+        '--cfg-options',
+        nargs='+',
+        action=DictAction,
+        help='arguments in dict')
```
hellock, line 57: The description is less informative. Refer to `tools/test.py`.

tools/test.py
Johnson-Wang: In this case, I suggest adding poping `interval`, `tmpdir`, `start` and `gpu_collect`   keys. `start` is an additional key to be added in EvalHook (https://github.com/open-mmlab/mmdetection/pull/3550)

tools/test.py
```diff
@@ -142,11 +170,16 @@ def main():
...
             mmcv.dump(outputs, args.out)
-        kwargs = {} if args.options is None else args.options
+        kwargs = {} if args.eval_options is None else args.eval_options
         if args.format_only:
             dataset.format_results(outputs, **kwargs)
         if args.eval:
-            dataset.evaluate(outputs, args.eval, **kwargs)
+            eval_kwargs = cfg.get('evaluation', {}).copy()
+            # hard-code way to remove EvalHook args
+            for key in ['interval', 'tmpdir', 'start', 'gpu_collect']:
```
Johnson-Wang, line 179: Add a note in EvaluationHook in case of its further extension.

Johnson-Wang (approved):

hellock (approved):

PR #3543, opened 2020-08-12
fix bug of gaussian_target, update unittest of heatmap

tianyuandu:
1.  Fixed the bug in `gen_gaussian_target` while multiple gts in one class but only got one peak in the heat map.
2. Expand to 3 gt_bboxes in `test_corner_head_encode_and_decode_heatmap` and two of them are the same class.
3. Replace `expand` with `repeat` in `corner_head.py` to avoid shallow-copy cases, which may generate unexpected results in `2.`

| codebase | batchsize | mAP |
| :-: | :-: | :-: |
| Current | 16 x 6 | 41.2 |
| Current | 10 x 5 | still running |
| Last benchmark | 16 x 6 | 41.1 |
| Last benchmark | 10 x 5 | 41.3 |
| Paper | 10 x 5 | 40.5 |

### DIFF
--- a/mmdet/models/dense_heads/corner_head.py
+++ b/mmdet/models/dense_heads/corner_head.py
@@ -216,8 +216,14 @@ class CornerHead(BaseDenseHead):
         """Initialize weights of the head."""
         bias_init = bias_init_with_prob(0.1)
         for i in range(self.num_feat_levels):
+            self.tl_heat[i][-1].conv.reset_parameters()
             self.tl_heat[i][-1].conv.bias.data.fill_(bias_init)
+            self.br_heat[i][-1].conv.reset_parameters()
             self.br_heat[i][-1].conv.bias.data.fill_(bias_init)
+            self.tl_emb[i][-1].conv.reset_parameters()
+            self.br_emb[i][-1].conv.reset_parameters()
+            self.tl_off[i][-1].conv.reset_parameters()
+            self.br_off[i][-1].conv.reset_parameters()
 
     def forward(self, feats):
         """Forward features from the upstream network.
--- a/mmdet/models/dense_heads/corner_head.py
+++ b/mmdet/models/dense_heads/corner_head.py
@@ -216,6 +216,9 @@ class CornerHead(BaseDenseHead):
         """Initialize weights of the head."""
         bias_init = bias_init_with_prob(0.1)
         for i in range(self.num_feat_levels):
+            # Due to the initial parameters are different between nn.Conv2d and
+            # ConvModule, and our experiment shows using the initial parameter
+            # from nn.Conv2d could increase the final mAP about 0.2%.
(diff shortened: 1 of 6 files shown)


mmdet/models/utils/gaussian_target.py
```diff
@@ -54,7 +54,7 @@ def gen_gaussian_target(heatmap, center, radius, k=1):
     masked_heatmap = heatmap[y - top:y + bottom, x - left:x + right]
     masked_gaussian = gaussian_kernel[radius - top:radius + bottom,
                                       radius - left:radius + right]
-    out_heatmap = torch.zeros_like(heatmap)
+    out_heatmap = heatmap
```
ZwwWayne, line 57: Will this change the input `heatmap`? Is this behavior expected or not?

hellock:
ping @tianyuandu

mmdet/models/utils/gaussian_target.py
tianyuandu: The `torch.max` will change the input `heatmap`, and we would also return the changed `heatmap`, so we didn't clone the `heatmap` here to save cost and this behavior is expected.

tianyuandu: The `torch.max` will change the input `heatmap`, and we would also return the changed `heatmap`, so we didn't clone the `heatmap` here to save cost and this behavior is expected.

mmdet/models/dense_heads/corner_head.py
```diff
@@ -216,8 +216,14 @@ def init_weights(self):
         """Initialize weights of the head."""
         bias_init = bias_init_with_prob(0.1)
         for i in range(self.num_feat_levels):
+            self.tl_heat[i][-1].conv.reset_parameters()
```
ZwwWayne, line 219: Add comments to tell the motivation.

mmdet/models/dense_heads/corner_head.py
```diff
@@ -768,7 +774,7 @@ def _gather_feat(self, feat, ind, mask=None):
             feat (Tensor): Gathered feature.
         """
         dim = feat.size(2)
-        ind = ind.unsqueeze(2).expand(ind.size(0), ind.size(1), dim)
```
ZwwWayne, line 771: May also add comments on the consequences if we use `expand` rather than `repeat` to remind the future refactoring to avoid expand.

tianyuandu:
| codebase | batchsize | mAP |
| :-: | :-: | :-: |
| Current | 16 x 6 | 41.2 |
| Current | 10 x 5 | still running |
| Last benchmark | 16 x 6 | 41.1 |
| Last benchmark | 10 x 5 | 41.3 |
| Paper | 10 x 5 | 40.5 |

ZwwWayne (approved):

hellock:
@tianyuandu The linting failed.

PR #3550, opened 2020-08-13
Add init_eval to evaluation hook

Johnson-Wang:


mmdet/core/evaluation/eval_hooks.py
```diff
@@ -10,16 +10,24 @@ class EvalHook(Hook):
     Attributes:
         dataloader (DataLoader): A PyTorch dataloader.
         interval (int): Evaluation interval (by epochs). Default: 1.
+        init_eval (bool): Evaluation before run. It is useful when loading a
+          checkpoint. Default: False
     """
 
-    def __init__(self, dataloader, interval=1, **eval_kwargs):
+    def __init__(self, dataloader, interval=1, init_eval=False, **eval_kwargs):
```
hellock, line 17: We can add an argument `start`. When it is set to 0, we will perform evaluation before the first epoch. It also enables users to skip the first several epochs.

mmdet/core/evaluation/eval_hooks.py
```diff
@@ -10,17 +10,31 @@ class EvalHook(Hook):
...
+    def __init__(self, dataloader, interval=1, start=None, **eval_kwargs):
         if not isinstance(dataloader, DataLoader):
             raise TypeError('dataloader must be a pytorch DataLoader, but got'
                             f' {type(dataloader)}')
         self.dataloader = dataloader
         self.interval = interval
+        self.start = start
         self.eval_kwargs = eval_kwargs
 
+    def before_run(self, runner):
```
hellock, line 30: Better to use `before_train_epoch` instead of `before_run`. There may be custom hooks that is required before running the evaluation.

Johnson-Wang: I thought about that too. But adding it to `before_train_epoch` would cause two repeated evaluations. Besides, logger hook level is "VERY_LOW" and chances are rare that one would put a custom hook about dataset handling after this one.

mmdet/core/evaluation/eval_hooks.py
```diff
@@ -10,17 +10,35 @@ class EvalHook(Hook):
...
+        self.start = start
         self.eval_kwargs = eval_kwargs
+        self.initial_epoch_flag = True
+
+    def before_train_epoch(self, runner):
+        """Evaluate the model only at the start of training."""
+        if not self.initial_epoch_flag:
+            return
+        if self.start is not None and runner.epoch >= self.start:
+            self.after_train_epoch(runner)
```
hellock, line 36: In `self.after_train_epoch`, the condition `self.every_n_epochs(runner, self.interval)` may not be satisfied in this case.

Johnson-Wang: Then it would depend on the interval to decide. For example, if interval=2 , start=1, and the resuming epoch =1, it is logically correct not to run evaluation before the training epoch.

PR #3577, opened 2020-08-19
Add include_bkg in ClassBalancedDataset

Johnson-Wang:


mmdet/datasets/dataset_wrappers.py
```diff
@@ -126,11 +126,14 @@ class ClassBalancedDataset(object):
             no oversampling. For categories with ``f_c < oversample_thr``, the
             degree of oversampling following the square-root inverse frequency
             heuristic above.
+        include_bkg (bool): Whether to include pure background images as an
```
ZwwWayne, line 129: How about `include_empty` or just `include_background`? `bkg` is a little bit hard to guess.

ZwwWayne (approved):

PR #3608, opened 2020-08-23
[Feature] Support RandomFlip with horizontal/vertical/diagonal 

### LINKED ISSUE #3474: Data augmentation
Hi, I just had 2 very quick questions on data augmentation. As I have data that does not have default orientation, I wanted to implement both horizontal and vertical random flips in the data augmentation pipeline, and just wanted to check if the best way to do this was to add both these lines in the ```train_pipeline```:
```
dict(type='RandomFlip', flip_ratio=0.5, direction='horizontal'),
dict(type='RandomFlip', flip_ratio=0.5, direction='vertical'),
```
Also, just wanted to ask, when implementing random cropping, whether it is standard practice to perform cropping first followed by resizing. Thanks!

xvjiarui:
Fixed #3474

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -318,7 +318,7 @@ class RandomFlip(object):
 
         direction = direction if isinstance(direction, list) else [direction]
         for d in direction:
-            assert d in ['horizontal', 'vertical', 'horizontal+vertical']
+            assert d in ['horizontal', 'vertical', 'diagonal']
         self.direction = direction
         assert len(self.flip_ratio) == len(self.direction)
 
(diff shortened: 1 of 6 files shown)


mmdet/datasets/pipelines/transforms.py
```diff
@@ -302,17 +302,29 @@ class RandomFlip(object):
...
+        flip_ratio (float | list[float], optional): The flipping probability.
+            Default: None.
+        direction(str | list[str], optional): The flipping direction. Options
+            are 'horizontal', 'vertical', 'diagonal'. Default: 'horizontal'.
+            If input is a list, the length must equal ``flip_ratio``. Each
+            element in ``flip_ratio`` indicates the flip probability of
+            corresponding direction.
     """
 
     def __init__(self, flip_ratio=None, direction='horizontal'):
```
hellock, line 314: We need to support 3 cases:
- flip_ratio is a float, direction is a string.
- flip_ratio is a float, direction is a list of strings.
- flip_ratio is a list of float, direction is a list of strings.

hellock (approved):

PR #3643, opened 2020-08-28
Add augmentations which only transform image (e.g. Constrast, Equalize, Color, Brightness).

v-qjqs:
This PR implements augmentations which only transform image (e.g. Sharpness, Equalize, Color) in process.


mmdet/datasets/pipelines/__init__.py
```diff
@@ -1,3 +1,5 @@
+from .aug_only_img import (BrightnessTransform, ColorTransform,
```
xvjiarui, line 1: How about `img_augment`?

mmdet/datasets/pipelines/aug_only_img.py
```diff
@@ -0,0 +1,196 @@
...
+    def __call__(self, results):
+        """Call function for Color transformation.
+
+        Args:
+            results (dict): Result dict from loading pipeline.
+
+        Returns:
+            dict: Colored results.
+        """
+        if np.random.rand() > self.prob:
```
xvjiarui, line 51:
```suggestion
        if np.random.rand() < self.prob:
```

xvjiarui: Same for others.

mmdet/datasets/pipelines/aug_only_img.py
```diff
@@ -0,0 +1,196 @@
...
+            return results
+        self._adjust_color_img(results, self.factor)
+        return results
+
+    def __repr__(self, ):
```
xvjiarui, line 56:
```suggestion
    def __repr__(self):
```

xvjiarui: Same for others.

xvjiarui (approved):

xvjiarui:
@v-qjqs 
There is conflict now. Please resolve the conflict.

PR #3649, opened 2020-08-29
Port autoscale learning rate feature from mmclassification

anthracene:
The --autoscale-lr  switch for training in mmclassification is really handy. This pull request copies the code for that feature from mmclassification

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -64,6 +64,10 @@ def parse_args():
         default='none',
         help='job launcher')
     parser.add_argument('--local_rank', type=int, default=0)
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


anthracene:
After submitting this I think I realized that this feature is broken in mmclassification and so this pull request which uses the same implementation is also broken: I think the point at which it tries to scale the learning rate is after a separate process has already been launched for each GPU, so it always thinks it's running on a single GPU and scales accordingly. I think this would be a really useful feature so you don't have to manually keep the learning rate adjust for the number of GPUs, but I think the implementation that I took from mmclassification isn't the way to do it.

PR #3656, opened 2020-08-31
Add Shear augmentation.

v-qjqs:
This PR implements Shear augmentation, and clean up the irrelevant commits.


mmdet/core/mask/structures.py
```diff
@@ -498,6 +520,14 @@ def crop_and_resize(self,
             resized_masks.append(resized_mask)
         return PolygonMasks(resized_masks, *out_shape)
 
+    def shear(self,
```
ZwwWayne, line 523: We also need to implement shear operation for polygon masks.

mmdet/core/mask/structures.py
```diff
@@ -297,6 +298,27 @@ def expand(self, expanded_h, expanded_w, top, left):
                           left:left + self.width] = self.masks
         return BitmapMasks(expanded_mask, expanded_h, expanded_w)
 
+    def shear(self,
+              shear_matrix,
+              out_shape,
+              fill_val=0,
+              flags=cv2.INTER_NEAREST,
```
ZwwWayne, line 305: 1. We should hide the cv2 options inside mmcv, see [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/image/geometric.py#L29).
2. For shear transformation, we need to first implement one operator in mmcv, which warps all the operations and cv2 options. Then call `mmcv.imgshear` in mmdet for all the places.

v-qjqs: Hi @ZwwWayne, so a PR which implements mmcv.imshear is required before this PR?

mmdet/core/mask/structures.py
v-qjqs: Hi, shear operation for polygon masks will be supported in the future version.

mmdet/datasets/pipelines/shear.py
```diff
@@ -0,0 +1,254 @@
...
+            min_x, min_y, max_x, max_y = np.split(
+                results[key], results[key].shape[-1], axis=-1)
+            coordinates = np.stack([[min_x, min_y], [max_x, min_y],
+                                    [min_x, max_y],
+                                    [max_x, max_y]])  # [4, 2, nb_box, 1]
+            coordinates = coordinates[..., 0].transpose(
+                (2, 1, 0)).astype(np.float32)  # [nb_box, 2, 4]
+            new_coords = np.matmul(shear_matrix[None, :, :],
+                                   coordinates)  # [nb_box, 2, 4]
+            min_x, min_y = np.min(
```
ZwwWayne, line 141: The following format looks more clear and better.
```python
min_x = np.min(new_coords[:, 0, :], axis=-1)
min_y = np.min(new_coords[:, 1, :], axis=-1)
max_x = np.max(new_coords[:, 0, :], axis=-1)
max_y = np.max(new_coords[:, 1, :], axis=-1)
```

mmdet/datasets/pipelines/shear.py
```diff
@@ -0,0 +1,254 @@
...
+                     magnitude,
+                     direction='horizontal',
+                     fill_val=0,
+                     interpolation='bilinear'):
+        """Shear the masks."""
+        h, w, c = results['img_shape']
+        for key in results.get('mask_fields', []):
+            masks = results[key]
+            if isinstance(masks, PolygonMasks):
+                raise NotImplementedError
```
ZwwWayne, line 167: We need to implement it for polygon mask.

tests/test_data/test_shear.py
```diff
@@ -0,0 +1,191 @@
...
+        dtype=results['gt_semantic_seg'].dtype)
+    check_shear(results_gt, results_sheared)
+
+    # test PolygonMasks with shear vertically, magnitude=-1
+    results = construct_toy_data(poly2mask=False)
+    results_sheared = shear_module(
+        copy.deepcopy(results), random_negative_prob=1.)
+    gt_masks = [[np.array([1, 0, 2, 0, 2, 0, 1, 0], dtype=np.float)]]
+    results_gt['gt_masks'] = PolygonMasks(gt_masks, 2, 4)
+    check_shear(results_gt, results_sheared)
```
ZwwWayne, line 191: For a similar mask, the results after shear augmentation on PolygonMasks and BitMapMasks should be the same, we could also check this case.

mmdet/datasets/pipelines/shear.py
```diff
@@ -0,0 +1,247 @@
...
+        self,
+        level,
+        img_fill_val=128,
+        seg_ignore_label=255,
+        prob=0.5,
+        direction='horizontal',
+        max_shear_magnitude=0.3,
+    ):
+        assert isinstance(level, (int, float)), \
+            'The level must be type int or float.'
```
ZwwWayne, line 67: Make the bug information more clear
```python
f'The level must be type int or float, got {type(level)}'
```

mmdet/datasets/pipelines/shear.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    def __init__(
+        self,
+        level,
+        img_fill_val=128,
+        seg_ignore_label=255,
+        prob=0.5,
+        direction='horizontal',
+        max_shear_magnitude=0.3,
```
ZwwWayne, line 64: This comma is unnecessary.

mmdet/datasets/pipelines/shear.py
```diff
@@ -0,0 +1,247 @@
...
+            if label_key in results:
+                results[label_key] = results[label_key][valid_inds]
+            # mask fields, e.g. gt_masks and gt_masks_ignore
+            mask_key = bbox2mask.get(key)
+            if mask_key in results:
+                results[mask_key] = results[mask_key][valid_inds]
+
+    def __call__(self,
+                 results,
+                 random_negative_prob=0.5,
```
ZwwWayne, line 204: The input of __call__ should be only results. Make `random_negative_prob` and `interpolation` arguments of self.

mmdet/core/mask/structures.py
```diff
@@ -498,6 +556,39 @@ def crop_and_resize(self,
...
+            if direction == 'horizontal':
+                shear_matrix = np.stack([[1, magnitude],
+                                         [0, 1]]).astype(np.float32)
+            elif direction == 'vertical':
+                shear_matrix = np.stack([[1, 0], [magnitude,
+                                                  1]]).astype(np.float32)
+            for poly_per_obj in self.masks:
+                sheared_poly = []
+                for p in poly_per_obj:
+                    p = p.copy()
```
ZwwWayne, line 579: `copy` operation seems to be unnecessary here.

ZwwWayne (approved):

mmdet/datasets/pipelines/auto_augment.py
```diff
@@ -1,10 +1,40 @@
...
+import mmcv
 import numpy as np
 
 from ..builder import PIPELINES
 from .compose import Compose
 
+_MAX_LEVEL = 10
+
+
+def level_to_value(level, max_value):
```
hellock, line 12: Will this method be used by multiple transforms?

mmdet/datasets/pipelines/auto_augment.py
```diff
@@ -1,10 +1,40 @@
...
+    """Map from level to values based on max_value."""
+    return (level / _MAX_LEVEL) * max_value
+
+
+def random_negative(value, random_negative_prob):
```
hellock, line 17: and this

mmdet/datasets/pipelines/auto_augment.py
v-qjqs: Yeah, it will be used for another aug like translate, rotate, etc.

mmdet/datasets/pipelines/auto_augment.py
v-qjqs: Yeah, it will be used for another aug like translate, rotate, etc.

PR #3576, opened 2020-08-18
Yolo 2.0 solve conflicts

sudo-rm-covid19:
Fix yolo assigner as discussed #3083 . Since this is the first time I work with mmdet v2, please pay special attention to the possible changes to the conventions. During editing, I found that there is a inconsistency in the definition of strides in functions of "single_level_grid_anchors" (where stride[0] is along x-axis) and "valid_flags" (where stride[0] is along y-axis) in anchor_generator.py. I assume stirde[0] is along x-axis in my codes.

PR #3685, opened 2020-09-04
fix valid flag stride order

yhcao6:
Fix valid flag stride issue mentioned in #3576


mmdet/core/anchor/anchor_generator.py
```diff
@@ -242,8 +242,8 @@ def single_level_grid_anchors(self,
         Args:
             base_anchors (torch.Tensor): The base anchors of a feature grid.
             featmap_size (tuple[int]): Size of the feature maps.
-            stride (tuple[int], optional): Stride of the feature map.
-                Defaults to (16, 16).
+            stride (tuple[int], optional): Stride of the feature map in order
```
ZwwWayne, line 245: We may also update the docstring in the AnchorGenerator's arguments at around line 15.

ZwwWayne (approved):

PR #3687, opened 2020-09-04
Supports for Translate augmentation.

v-qjqs:
This PR re-implemts Translate augmentation based on mmcv.imtranslate.


mmdet/core/mask/structures.py
```diff
@@ -297,6 +297,39 @@ def expand(self, expanded_h, expanded_w, top, left):
                           left:left + self.width] = self.masks
         return BitmapMasks(expanded_mask, expanded_h, expanded_w)
 
+    def translate(self,
```
yhcao6, line 300: add an abstract method in the base class.

yhcao6 (approved):

mmdet/core/mask/structures.py
```diff
@@ -498,6 +554,30 @@ def crop_and_resize(self,
             resized_masks.append(resized_mask)
         return PolygonMasks(resized_masks, *out_shape)
 
+    def translate(self,
+                  out_shape,
+                  offset,
+                  direction='horizontal',
+                  fill_val=0,
+                  interpolation='bilinear'):
+        """Translate the PolygonMasks."""
```
yhcao6, line 563: We can set `fill_val` and `interpolation` here to be None. Report a warning when the user pass other values when translating a polygon mask since it doesn't support.

mmdet/datasets/pipelines/translate.py
```diff
@@ -0,0 +1,234 @@
...
+                         interpolation='bilinear'):
+        """Translate masks horizontally or vertically."""
+        h, w, c = results['img_shape']
+        for key in results.get('mask_fields', []):
+            masks = results[key]
+            if isinstance(masks, PolygonMasks):
+                results[key] = masks.translate((h, w), offset, direction)
+            elif isinstance(masks, BitmapMasks):
+                results[key] = masks.translate((h, w), offset, direction,
+                                               fill_val, interpolation)
```
yhcao6, line 151: No need if-else here since the interface is the same among polygon and bitmap.

mmdet/datasets/pipelines/auto_augment.py
```diff
@@ -532,3 +532,204 @@ def __repr__(self):
...
+
+    def __init__(self,
+                 level,
+                 prob=0.5,
+                 img_fill_val=128,
+                 seg_ignore_label=255,
+                 direction='horizontal',
+                 max_translate_offset=250.,
+                 random_negative_prob=0.5,
+                 interpolation='bilinear',
```
hellock, line 573: interpolation is not necessary since offset is an integer?

yhcao6 (approved):

PR #3702, opened 2020-09-07
fix atss no gt

jshilong:
Fixed two bug in the ATSS model, each occurred when a single image did not contain any gt boxe and the total batch did not contain any gt  boxe respectively


mmdet/models/dense_heads/atss_head.py
```diff
@@ -9,6 +9,8 @@
 from ..builder import HEADS, build_loss
 from .anchor_head import AnchorHead
 
+eps = 1e-12
```
ZwwWayne, line 12: We'd better use EPS to avoid conflict in case there some places also use eps as a variable.

hellock:
ping @jshilong

ZwwWayne (approved):

Issue #3563, opened 2020-08-16
A possible solution to the training stuck problem 


Wei-ucas:
When training Maskrcnn with my own datasets on multi-GPUs, I found if I set batchsize (image per GPU) small (1 or 2), after some iterations the training program always stuck without any output and the GPU usage keeps 100%. It seems some people met the same problem but no solution work.

I have found the reason and fix the problem, if you meet the same problem, you can try this solution.
in standard_roi_head.py, line 169,
```
            if pos_rois.shape[0] == 0:
                return dict(loss_mask=None)
```
and line 130,
```
            if mask_results['loss_mask'] is not None:
                losses.update(mask_results['loss_mask'])
```
It means that if there is no positive sample, the losses with not have the key "loss_mask". When training on multi-GPUs,  the losses from different GPUs will be reduced together in detectors/base.py line 201:
```
        for loss_name, loss_value in log_vars.items():
            # reduce loss when distributed training
            if dist.is_available() and dist.is_initialized():
                loss_value = loss_value.data.clone()
                dist.all_reduce(loss_value.div_(dist.get_world_size()))
```
So if the loss dict from different GPUs have different keys, the program will be stuck.

There are two methods to solve the problem:
1. check your train set, find the image without any positive ground truth and delete it;
2. change the loss_mask to handle the pos_rois.shape[0] == 0 situation.

This is the MaskRcnn model, maybe other models also have this problem making training stuck, so try to check the loss may be helpful.

ZwwWayne:
Thanks for your report, would you like to create a PR to solve it?

PR #3713, opened 2020-09-08
[Fix]: fix mask rcnn training stuck problem when there is no positive rois

ZwwWayne:
This PR fixes the training stuck problem in Mask R-CNN and Cascade Mask R-CNN when there is no positive ROIs using the wrappers of Conv/Pooling operators. This PR requires a new version of MMCV that merges https://github.com/open-mmlab/mmcv/pull/602

Fix https://github.com/open-mmlab/mmdetection/issues/3563


hellock:
Task linked: [CU-49tawu Fix training stuck problem (#3563)](https://app.clickup.com/t/49tawu)

hellock (approved):

PR #3714, opened 2020-09-08
fix res2net+dcn test bug

yuzhj:
I rewrite init_weights in the res2net.py to fix the test bug(AttributeError: 'Bottle2neck' object has no attribute 'conv2')

### DIFF
--- a/mmdet/models/backbones/res2net.py
+++ b/mmdet/models/backbones/res2net.py
@@ -1,5 +1,4 @@
 import math
-
 import torch
 import torch.nn as nn
 import torch.utils.checkpoint as cp
--- a/mmdet/models/backbones/res2net.py
+++ b/mmdet/models/backbones/res2net.py
@@ -1,5 +1,4 @@
 import math
-
 import torch
 import torch.nn as nn
 import torch.utils.checkpoint as cp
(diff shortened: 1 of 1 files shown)


yuzhj:
the format problem is so difficult!!!!!!!!!!

mmdet/models/backbones/res2net.py
```diff
@@ -313,3 +317,34 @@ def make_res_layer(self, **kwargs):
...
+            for m in self.modules():
+                if isinstance(m, nn.Conv2d):
+                    kaiming_init(m)
+                elif isinstance(m, (_BatchNorm, nn.GroupNorm)):
+                    constant_init(m, 1)
+
+            if self.dcn is not None:
+                for m in self.modules():
+                    if isinstance(m, Bottle2neck):
+                        for n in m.convs:
```
xvjiarui, line 341: We may add comments here.
Like `dcn in Res2Net bottleneck is in ModuleList`.

xvjiarui:
> the format problem is so difficult!!!!!!!!!!

Hi @yuzhj 
Thanks for your contribution. You may check [contribute guide](https://github.com/open-mmlab/mmdetection/blob/master/.github/CONTRIBUTING.md) for reference.

mmdet/models/backbones/res2net.py
yuzhj: > We may add comments here.
> Like `dcn in Res2Net bottleneck is in ModuleList`.

ok, i will add comments

xvjiarui (approved):

PR #3723, opened 2020-09-09
Add dynamic loss scale support (from apex)

halostorm:


yhcao6:
It may be better to support both the constant loss scaling and dynamic loss scaling. Since the dynamic loss scaling will introduce extra computation cost and the constant loss scaling can apply to most cases.

halostorm:
@yhcao6  good suggestion! I have update it. pass (loss_scale='dynamic') will activate dynamic loss scale

yhcao6:
There are some duplicate codes that are not used, for example, the `scale_gradient` and `backward` methods which the original `FP16OptimizerHook` has already implemented.
It may be better to merge the class `LossScalar` and `DynamicLossScalar` and rename it as `LossScalar` but add a param `policy` to specify whether we use `dynamic` or `static`.
Please add a docstring to every method and class.

yhcao6:
Hi, @halostorm,
I just reformat some doc and codes, please check that. 
May I ask if you have verified the code by training a model with dynamic loss scaling?
Would you like to add some unit tests for the newly added `LossScaler`? You can add them in tests/test_fp16.py

yhcao6:
Duplicate with https://github.com/open-mmlab/mmcv/pull/585

PR #3731, opened 2020-09-09
Upgrade pycocotools

xvjiarui:


mmdet/datasets/lvis.py
```diff
@@ -330,10 +331,13 @@ def evaluate(self,
         """
 
         try:
+            import lvis
+            assert lvis.__version__ >= '10.5.3'
             from lvis import LVISResults, LVISEval
         except ImportError:
-            raise ImportError('Please follow config/lvis/README.md to '
-                              'install open-mmlab forked lvis first.')
+            raise ImportError('Please run pip install mmlvis to install '
```
hellock, line 338: When the version is lower than 10.5.3, an AssertionError is raised.

hellock (approved):

PR #3741, opened 2020-09-11
[fix] typo

jd730:
fix typo in the comment:
my also contain -> may also contain


shinya7y:
The same typo could be fixed at the same time.
https://github.com/open-mmlab/mmdetection/blob/039ad4dd64edaa5efe69f00574a0c24240adac97/mmdet/core/post_processing/merge_augs.py#L17

jd730:
Hi @shinya7y, I fixed it now. 
Hi @hellock, I signed it.

hellock (approved):

PR #3695, opened 2020-09-05
[Refactor] refactor get_subset_by_classes in dataloader for training with empty-GT images

ZwwWayne:
This PR fixes the issue of training with empty-GT images when specifying classes, continuing on https://github.com/open-mmlab/mmdetection/pull/3133.

In the original implementation, `get_subset_by_classes` is not related to the flag `self.filter_empty_gt`. This brings ambiguous behavior and potential bugs when specifying dataset classes. More specifically, `CocoDataset` filters empty gt images in two places, one is in `_filter_img`, the other one is in `get_subset_by_classes`. Other datasets only filter empty gt images in `get_subset_by_classes`, but this will be triggered by specifying classes rather than `self.filter_empty_gt=True`. 

The goal of this PR is to make the behavior of `filter empty GT` independent to `specifying classes`. After this PR, if `filter_empty_gt=False`, no matter whether the classes are specified in a dataset, the dataset will use all the images in the annotations. If `filter_empty_gt=True`, no matter whether the classes are specified, the dataset will check the images and filter out images containing no GT boxes.

This PR also changes the logic during test_mode. Previously, the dataset will filter out empty-GT images no matter whether it is in test mode. After this PR, it only filters out images in training, i.e., if `test_mode=True`, it will test all the images given in the annotations. This behavior is more commonly need as mentioned in #3545, and is more consistent with before that the dataset will not filter out small images in test mode. Therefore, the users should be responsible for the data filtering/cleaning process for the test dataset.

Fix https://github.com/open-mmlab/mmdetection/issues/3023, Fix #2965, Fix #3545

PR #3744, opened 2020-09-12
only call `get_subset_by_classes` during training

jonbakerfish:
Only filter data during training but not in testing. Because test data is not labeled.


ZwwWayne:
Hi @jonbakerfish ,
This PR duplicates #3695. Could you check whether #3695 meets your need?

jonbakerfish:
Yes, it did fix the bug. Thanks.

PR #3754, opened 2020-09-14
fix rpn transforming bug in two stage networks

yuzhj:
I think rpn has no relation to num_classes in two stage networks.
the origin code will cause the size mismatch for rpn_head.rpn_reg.weight/bias between the old model and new model


hellock:
ping @ZwwWayne

jingege315:
It works! Thank you!

yuzhj:
@hellock @ZwwWayne This PR is helpful in upgrading model from v1.x to v2.x, I guess. Look forward to a review. Many thanks.

ZwwWayne (approved):

PR #1502, opened 2019-10-07
Fix fp16 training error of dpool

yhcao6:
Fix #1494

PR #3766, opened 2020-09-15
Change all fp16 operation to using MMCV

RyanXLi:
Training and inference with config 'configs/fp16/faster_rcnn_r50_fpn_fp16_1x_coco.py' has been tested to have the same mAP.


yhcao6:
Can you check the formatter of your code editor? It seems this PR introduces many unnecessary changes.

RyanXLi:
@yhcao6 The docformatter problem has been fixed. Thanks.

yhcao6 (approved):

hellock:
Please check if the fp16 module in mmdet is updated and diverged from mmcv.

RyanXLi:
I have checked and the fp16 in mmdet in current master is identical to mmcv. However, [#1502](https://github.com/open-mmlab/mmdetection/pull/1502) may need to be moved to mmcv.

PR #3778, opened 2020-09-16
[enhance]: Update tutorials for changing training settings

ZwwWayne:
This PR adds tutorials for changing training settings, including optimization, learning rate scheduler.


hellock:
Task linked: [CU-49ta06 Creating hooks](https://app.clickup.com/t/49ta06)

docs/tutorials/customize_runtime.md
```diff
@@ -157,6 +157,22 @@ We support many other learning rate schedule [here](https://github.com/open-mmla
...
+```python
+workflow = [('train', 1)]
+```
+which means running 1 epoch for training.
+Sometimes user may want to check some metrics (e.g. loss, accuracy) about the model on the validate set.
+In such case, we can set the workflow as
+```python
+[('train', 1), ('val', 1)]
+```
+so that 1 epochs for training and 1 epoch for validation will be ran iteratively.
```
ZwwWayne, line 173: 1. Do we need to clarify whether the model will be trained on validation set, or the workflow simply performs inference and calculate the loss?
2. In such case, what is the relation ship between this workflow and `total_epochs`, e.g., if `total_epochs=12`, will it train model on training set by 12 epochs, or perform train and val workflows by 6 epochs, respectively?

Johnson-Wang: Speaking of that, I think we may also add some description about the equivalence about [('train', 1), ('val", 1)] and [('train', 1) but using evaluation hook.

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,316 @@
...
+
+which means running 1 epoch for training.
+Sometimes user may want to check some metrics (e.g. loss, accuracy) about the model on the validate set.
+In such case, we can set the workflow as
+
+```python
+[('train', 1), ('val', 1)]
+```
+
+so that 1 epoch for training and 1 epoch for validation will be ran iteratively.
```
Johnson-Wang, line 177: will be run

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,316 @@
...
+custom_hooks = [
+    dict(type='MyHook', a=a_value, b=b_value, priority='NORMAL')
+]
+```
+
+### Modify default runtime hooks
+
+There are some common hooks that are not registerd through `custom_hooks`, they are
+
+- log_config
```
Johnson-Wang, line 275: Describe the default priorities of these hooks so that the users may know what priority to use in their own hooks.

hellock (approved):

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,319 @@
...
+
+```python
+custom_hooks = [
+    dict(type='MyHook', a=a_value, b=b_value, priority='NORMAL')
+]
+```
+
+### Modify default runtime hooks
+
+There are some common hooks that are not registerd through `custom_hooks`, they are
```
dreamerlin, line 275: typo here `registered`

Issue #3785, opened 2020-09-17
Test mask rcnn failed


yanmeizhao:
I only have one gpu, so my test mask rcnn script is :

python tools/test.py configs/mask_rcnn/mask_rcnn_r50_fpn_1x_coco.py checkpoints/mask_rcnn_r50_fpn_1x_coco_20200205-d4b0c5d6.pth --out results.pkl --eval bbox segm

but it run failed and reported error as below:
![image](https://user-images.githubusercontent.com/13143336/93432685-7531d780-f8f8-11ea-8307-043677814079.png)

My environment info as follows:
pytorch: 1.6.0+cu101
torchvision: 0.7.0+cu101
gcc: 5.4.0
cuda: 10.1
Addition：I test/train FasterRcnn successful, it's not any error, but MaskRcnn will crash when I test or train

I don't know why, is there anyone can help me ? Thank you in advance！

RyanXLi:
Hi! I do not seem to be able to reproduce your problem. Please provide more information by following the template of reporting an error.

yanmeizhao:
> Hi! I do not seem to be able to reproduce your problem. Please provide more information by following the template of reporting an error.

Sorry for not following the template，now my new description is as follows：
My target is to test maskrcnn by one gpu
1. What command or script did you run?
python tools/test.py configs/mask_rcnn/mask_rcnn_r50_fpn_1x_coco.py \
    checkpoints/mask_rcnn_r50_fpn_1x_coco_20200205-d4b0c5d6.pth \
    --show-dir mask_rcnn_r50_fpn_1x_results

2. Did you make any modifications on the code or config? Did you understand what you have modified?
I'm not change any thing, only make a data dir and checkpoints dir to place dataset and your trained models

3. What dataset did you use?
coco2017

4. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.
![image](https://user-images.githubusercontent.com/13143336/93550676-7e757f80-f99e-11ea-82d3-0b2aade3f54e.png)
5. You may add addition that may be helpful for locating the problem
I install pytorch by pip

6. If applicable, paste the error trackback here.
![image](https://user-images.githubusercontent.com/13143336/93550774-bed4fd80-f99e-11ea-8a47-1e1960c19d51.png)

If there is anything else I shoud add, please let me know，thank you for your attention！

RyanXLi:
[...]

PR #3791, opened 2020-09-18
Fix deprecated Conv2d from mmcv.ops

RyanXLi:
Solves [#3785 ](https://github.com/open-mmlab/mmdetection/issues/3785)


ZwwWayne (approved):

hellock:
The minimum version requirement of mmcv should be updated.

PR #3818, opened 2020-09-22
[docs] Add loss convention

xvjiarui:


hellock:
Task linked: [CU-4bw1wp [doc] Code Convention (loss)](https://app.clickup.com/t/4bw1wp)

docs/conventions.md
```diff
@@ -0,0 +1,64 @@
...
+                losses['loss_bbox'] = bbox_pred.sum() * 0
+        return losses
+```
+`bbox_head.loss()` will be called during model forward.
+The returned dict contains `'loss_bbox'`, `'loss_cls'`, `'acc'` .
+Only `'loss_bbox'`, `'loss_cls'` will be used during back propagation,
+`'acc'` will only be used as a metric to monitor training process.
+
+By default, only values whose keys contain `'loss'` will be back propagated.
+This behavior could be changed by modifying ``BaseDetector.train_step()``.
```
RyanXLi, line 64: May I ask, why are there 2 backticks on the ``BaseDetector.train_step()``?

RyanXLi (approved):

docs/conventions.md
hellock: Two backticks in rst equals to one backticks in markdown. There is no need to use 2 backticks here.

ZwwWayne (approved):

ZwwWayne (approved):

PR #3824, opened 2020-09-23
added yolo v4 network

RyanXLi:


xvjiarui:
Lint doesn't pass.

mmdet/models/backbones/darknet.py
```diff
@@ -52,6 +53,47 @@ def forward(self, x):
...
+    """
+
+    def __init__(self,
+                 in_channels,
+                 conv_cfg=None,
+                 norm_cfg=dict(type='BN', requires_grad=True),
+                 act_cfg=dict(type='LeakyReLU', negative_slope=0.1)):
+        super(ResBlockV4, self).__init__()
+        # assert in_channels % 2 == 0  # ensure the in_channels is even
+        # half_in_channels = in_channels // 2
```
xvjiarui, line 80: Remove unused code.

mmdet/models/backbones/darknet.py
```diff
@@ -52,6 +53,47 @@ def forward(self, x):
         return out
 
 
+class ResBlockV4(nn.Module):
```
xvjiarui, line 56: Duplicate with `ResBlock`?

mmdet/models/backbones/darknet.py
```diff
@@ -96,6 +138,7 @@ def __init__(self,
                  out_indices=(3, 4, 5),
                  frozen_stages=-1,
                  conv_cfg=None,
+                 csp_on=False,
```
xvjiarui, line 141: Docstring.

mmdet/models/backbones/darknet.py
```diff
@@ -197,3 +244,50 @@ def make_conv_res_block(in_channels,
             model.add_module('res{}'.format(idx),
                              ResBlock(out_channels, **cfg))
         return model
+
+
+class Csp_conv_res_block(nn.Module):
```
xvjiarui, line 249: Docstring.

mmdet/models/necks/yolo_neck.py
```diff
@@ -11,31 +11,36 @@
...
-            Default: dict(type='BN', requires_grad=True)
-        act_cfg (dict): Config dict for activation layer.
-            Default: dict(type='LeakyReLU', negative_slope=0.1).
+        Let out_channels = n, the DetectionBlock normally contains 5 ConvModules,
+        Their sizes are 1x1xn, 3x3x2n, 1x1xn, 3x3x2n, and 1x1xn respectively.
+        If the spp is on, the DetectionBlock contains 6 ConvModules and
+        3 pooling layers, sizes are 1x1xn, 3x3x2n, 1x1xn,
+        5x5 maxpool, 9x9 maxpool, 13x13 maxpool, 1x1xn, 3x3x2n, 1x1xn.
+        The input channel is arbitrary (in_channels)
+    yolo
```
xvjiarui, line 20: Format.

mmdet/models/necks/yolo_neck.py
```diff
@@ -11,31 +11,36 @@
...
+        Args:
+            in_channels (int): The number of input channels.
+            out_channels (int): The number of output channels.
+            conv_cfg (dict): Config dict for convolution layer. Default: None.
+            spp_on (bool): whether to integrate a spp module. Default: False.
+            spp_pooler_sizes (tuple): A set of sizes for spatial pyramid pooling.
```
xvjiarui, line 26: Could we use SPP when `spp_pooler_sizes` is not None?

xvjiarui: We may rename `spp_pooler_sizes` to `spp_scales`.

mmdet/models/necks/yolo_spp_pan_neck.py
```diff
@@ -0,0 +1,202 @@
...
+        5x5 maxpool, 9x9 maxpool, 13x13 maxpool, 1x1xn, 3x3x2n, 1x1xn.
+        The input channel is arbitrary (in_channels)
+    yolo
+        Args:
+            in_channels (int): The number of input channels.
+            out_channels (int): The number of output channels.
+            conv_cfg (dict): Config dict for convolution layer. Default: None.
+            spp_on (bool): whether to integrate a spp module. Default: False.
+            spp_pooler_sizes (tuple): A set of sizes for spatial pyramid pooling.
+                Default: (5, 9, 13).
```
xvjiarui, line 27: Similar.

xvjiarui:
Unit test is missing.

mmdet/models/necks/yolo_neck.py
RyanXLi: > Could we use SPP when spp_pooler_sizes is not None?
Yes, the code should be able to work with custom spp_scales.

mmdet/models/necks/yolo_spp_pan_neck.py
RyanXLi: Should we merge yolo_head and yolo_spp_pan_head? I think keeping them separated would make both file clearer and cause less confusion. They are sufficiently different but it's still a bit of duplicated code. Diff is [here](https://www.diffchecker.com/nLNnV4QN).

mmdet/models/backbones/darknet.py
xvjiarui:
```suggestion
                 with_csp=False,
```

mmdet/models/backbones/darknet.py
```diff
@@ -197,3 +211,68 @@ def make_conv_res_block(in_channels,
...
```
xvjiarui, line 216: Could this function be merged in `self.make_conv_res_block()`?

RyanXLi: I don't think so because this v4 conv_res_block has shortcut (therefore requiring a customized `forward` function) while the original `self.make_conv_res_block()` does not. However, the original `self.make_conv_res_block()` could probably be merged into this new class if necessary.

mmdet/models/necks/yolo_neck.py
```diff
@@ -75,6 +96,10 @@ class YOLOV3Neck(nn.Module):
         num_scales (int): The number of scales / stages.
         in_channels (int): The number of input channels.
         out_channels (int): The number of output channels.
+        spp_on (bool): Whether the spatial pyramid pooling is enabled.
+            Default: False.
+        spp_scales (tuple): A set of sizes for spatial pyramid pooling.
```
Johnson-Wang, line 101: Same comments as above.

mmdet/models/necks/yolo_neck.py
```diff
@@ -94,13 +121,18 @@ def __init__(self,
         self.num_scales = num_scales
         self.in_channels = in_channels
         self.out_channels = out_channels
+        self.spp_on = spp_on
 
         # shortcut
         cfg = dict(conv_cfg=conv_cfg, norm_cfg=norm_cfg, act_cfg=act_cfg)
 
+        # If spp is enabled, a spp block is added into the first DetectionBlock
```
Johnson-Wang, line 129: The comment is a bit ambiguous as it only contains one condition.

mmdet/models/necks/yolo_spp_pan_neck.py
```diff
@@ -0,0 +1,201 @@
+# Copyright (c) 2019 Western Digital Corporation or its affiliates.
```
Johnson-Wang, line 1: Can we support a universal yolo neck?

RyanXLi:
@Johnson-Wang Thanks for the suggestions. I have edited this PR according to the reviews and please have a look when it is convenient for you.

Johnson-Wang (approved):

goodloop:
so, is there any plan to merge to master branch? or just it will stay as a branch?

PR #3833, opened 2020-09-24
Clean outdated comments

shinya7y:
The comments like `# recompute feats to save memory` are so confusing.
What does `recompute feats` mean?

Old codes before the commit below recompute feats by calling `self.extract_feats(imgs)` 2-3 times.
https://github.com/open-mmlab/mmdetection/commit/c29a789c46014cc1ae181f1bbb2e8ec01c44a01c#diff-37358d90bcb60a9f015ed7c0ffcf6dd1


hellock:
Yes we used to extract features more than once to save memory. Thanks for the fix!

hellock (approved):

PR #3857, opened 2020-09-29
Added `generate_inputs_and_wrap_model` function for pytorch2onnx

RyanXLi:


LorenzoQ5:
On Sep 29, 2020 2:19 AM, Ryan Li <notifications@github.com> wrote:



You can view, comment on, or merge this pull request online at:
  https://github.com/open-mmlab/mmdetection/pull/3857

Commit Summary
added `generate_inputs_and_wrap_model` function for pytorch2onnxundo debugging changes

File Changes

    M
    tools/pytorch2onnx.py
    (94)
  

Patch Links:
https://github.com/open-mmlab/mmdetection/pull/3857.patchhttps://github.com/open-mmlab/mmdetection/pull/3857.diff

tools/pytorch2onnx.py
```diff
@@ -87,6 +87,101 @@ def pytorch2onnx(model,
...
+    function before exporting.
+    This function will:
+    (1) generate corresponding inputs which are used to execute the model.
+    (2) Wrap the model's forward function. For example, the MMDet models'
+    forward function has a parameter `return_loss:bool`. As we want to set
+    it as False while export API supports neither bool type or kwargs. So
+    we have to replace the forward like:
+    `model.forward = partial(model.forward, return_loss=False)`
+
+    Input:
```
ZwwWayne, line 104: Use standard docstring format.

tools/pytorch2onnx.py
```diff
@@ -87,6 +87,101 @@ def pytorch2onnx(model,
...
+    opset_version = 11
+    register_extra_symbolics(opset_version)
+
+    return model, tensor_data
+
+
+def _preprocess_example_input(input_config):
+    """Prepare an example input image for `generate_inputs_and_wrap_model`.
+
+    Input:
```
ZwwWayne, line 146: Use standard docstring format

tools/pytorch2onnx.py
```diff
@@ -17,32 +17,25 @@
     raise NotImplementedError('please update mmcv to version>=v1.0.4')
 
 
-def pytorch2onnx(model,
+def pytorch2onnx(config_path,
```
drcut, line 20: I think we'd better implement this function using ```generate_inputs_and_wrap_model```

tools/pytorch2onnx.py
```diff
@@ -82,9 +75,122 @@ def pytorch2onnx(model,
...
         assert np.allclose(
-            pytorch_result[0][:, 4], onnx_results[:, 4]
-        ), 'The outputs are different between Pytorch and ONNX'
-        print('The numerical values are same between Pytorch and ONNX')
+            pytorch_result[0][0][0][:4], onnx_results[0]
+            [:4]), 'The outputs are different between Pytorch and ONNX'
+        print('The numerical values are the same between Pytorch and ONNX')
+
+
+def generate_inputs_and_wrap_model(config_path, checkpoint_path, input_config):
```
drcut, line 83: In our design, each framework (MMDet, MMSeg...) should provide this function and AutoSDK will import the same function (at least the same name) while dealing with different framework. So I highly suggest to put this function into the framework, not just defined it ```tools/pytorch2onnx.py```

drcut (changes requested):
I have looked through the code and it really makes sense! In other words, I believe it's exactly what we need. However, we'd better put some function into the framework instead of just putting them in an extension folder (```tools```) as they will be needed by AutoSDK

tools/pytorch2onnx.py
RyanXLi: Thanks for the suggestion! I use the 2 helper functions instead of the `generate_inputs_and_wrap_model` because here we may want to keep the original model to verify whether the output is identical. If we use `generate_inputs_and_wrap_model` it may introduce some repetitive code to accomplish that.

tools/pytorch2onnx.py
RyanXLi: Thanks and I have added the `generate_inputs_and_wrap_model` and its required helper function in `mmdet.core.export`.

ZwwWayne (approved):

PR #3870, opened 2020-10-01
[Enhance]: Convert mask to bool before using it as img's index for robustness and speedup

LabMemNo003:
The mask returned by YOLACT has dtype 'uint8', which will extremely slow down the speed when using the mask as image's index.
``` bash
python tools/test.py configs/yolact/yolact_r101_1x8_coco.py checkpoints/yolact_r101_1x8_coco_20200908-4cbe9101.pth --show-dir ./result/

[                                                  ] 4/5000, 0.1 task/s, elapsed: 72s, ETA: 90141s
```

Requesting all models to return masks with dtype 'bool' is another way to solve this problem, but this PR is the easiest way to fix it, and gives more robustness to BaseDetector.show_result().

### DIFF
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -316,6 +316,7 @@ class BaseDetector(nn.Module, metaclass=ABCMeta):
                 i = int(i)
                 color_mask = color_masks[labels[i]]
                 mask = segms[i]
+                mask = mask if mask.dtype == np.bool else np.array(mask, dtype=bool)
                 img[mask] = img[mask] * 0.5 + color_mask * 0.5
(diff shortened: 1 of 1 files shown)


mmdet/models/detectors/base.py
```diff
@@ -316,6 +316,8 @@ def show_result(self,
                 i = int(i)
                 color_mask = color_masks[labels[i]]
                 mask = segms[i]
+                if mask.dtype != np.bool:
+                    mask = np.array(mask, dtype=bool)
```
hellock, line 320: The above 3 lines can be written as:
`mask = segms[i].astype(bool)`

ZwwWayne (approved):

PR #3930, opened 2020-10-14
[enhance]: Use RoIAlign implemented in MMCV for inference in CPU mode

RunningLeon:
Since RoIAlign in mmcv.ops is supported with aligned=False in CPU mode, there's no need to use the op from torchvision in CPU mode.

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -4,7 +4,7 @@ import matplotlib.pyplot as plt
 import mmcv
 import numpy as np
 import torch
-from mmcv.ops import RoIAlign, RoIPool
+from mmcv.ops import RoIPool
 from mmcv.parallel import collate, scatter
 from mmcv.runner import load_checkpoint
 
@@ -109,7 +109,7 @@ def inference_detector(model, img):
     else:
         # Use torchvision ops for CPU mode instead
         for m in model.modules():
-            if isinstance(m, (RoIPool, RoIAlign)):
+            if isinstance(m, RoIPool):
                 if not m.aligned:
                     # aligned=False is not implemented on CPU
                     # set use_torchvision on-the-fly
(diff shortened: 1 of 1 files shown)


mmdet/apis/inference.py
```diff
@@ -109,7 +109,7 @@ def inference_detector(model, img):
     else:
         # Use torchvision ops for CPU mode instead
         for m in model.modules():
-            if isinstance(m, (RoIPool, RoIAlign)):
+            if isinstance(m, RoIPool):
```
ZwwWayne, line 112: See [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/roi_pool.py#L71) RoiPool does not have the attribute `aligned`, so it will cause a bug here. We should skip the if condition in line 113.

RunningLeon: OK. I will skip if condition in [line 113](https://github.com/open-mmlab/mmdetection/blob/d6b1386711b41c21c009f630ab7b5c011a166af6/mmdet/apis/inference.py#L113) and create another PR to add aligned attribute to RoIPool.

mmdet/apis/inference.py
```diff
@@ -107,14 +107,10 @@ def inference_detector(model, img):
...
+        # RoIPool is not supported for CPU mode
         for m in model.modules():
-            if isinstance(m, (RoIPool, RoIAlign)):
-                if not m.aligned:
-                    # aligned=False is not implemented on CPU
-                    # set use_torchvision on-the-fly
-                    m.use_torchvision = True
-        warnings.warn('We set use_torchvision=True in CPU mode.')
+            assert not isinstance(
+                m, RoIPool), 'RoIPool is not supported currently.'
```
ZwwWayne, line 113: 'RoIPool is not supported currently.' -> 'CPU inference with RoIPool is not supported currently.'

ZwwWayne (approved):

PR #3955, opened 2020-10-19
Support DETR

ZwwWayne:


mmdet/models/dense_heads/transformer_head.py
```diff
@@ -0,0 +1,655 @@
...
+        self.num_classes = num_classes
+        self.cls_out_channels = num_classes + 1
+        self.in_channels = in_channels
+        self.num_fcs = num_fcs
+        self.train_cfg = train_cfg
+        self.test_cfg = test_cfg
+        self.use_sigmoid_cls = use_sigmoid_cls
+        self.embed_dims = embed_dims
+        self.num_query = test_cfg['max_per_img']
+        self.background_label = num_classes
```
ZwwWayne, line 148: Clean `self.background_label`, use `self.num_classes` instead.

mmdet/models/dense_heads/transformer_head.py
```diff
@@ -0,0 +1,655 @@
...
+            all_bbox_preds (Tensor): Sigmoid outputs from the regression
+                head with normalized coordinate format (cx, cy, w, h).
+                Shape [nb_dec, bs, num_query, 4].
+        """
+        # construct binary masks which used for the transformer.
+        # NOTE following the official DETR repo, non-zero values representing
+        # ignored positions, while zero values means valid positions.
+        batch_size = x.size(0)
+        input_img_h, input_img_w = img_metas[0]['input_img_shape']
+        masks = torch.ones((batch_size, input_img_h, input_img_w)).to(x.device)
```
ZwwWayne, line 237: `torch.ones((batch_size, input_img_h, input_img_w)).to(x.device)` -> `x.new_ones((batch_size, input_img_h, input_img_w))`

mmdet/models/dense_heads/transformer_head.py
```diff
@@ -0,0 +1,655 @@
...
+        for img_id in range(batch_size):
+            img_h, img_w, _ = img_metas[img_id]['img_shape']
+            masks[img_id, :img_h, :img_w] = 0
+        masks = masks.to(x.dtype)
```
ZwwWayne, line 241: This should be unnecessary if we use `x.new_ones` as above.

hellock (approved):

PR #3956, opened 2020-10-19
Better ONNX support for YoloV3

dsuess:
This PR tries to improve the ONNX export, specifically for YoloV3:
- With the standard-config you can now export a YoloV3 model to ONNX, which can be imported by onnx-runtime:
- The created ONNX file can be imported by `tensorrt >= 7.1` you disable postprocessing (confidence thresholding, NMS) by using the following `test_cfg`:
```
test_cfg = dict(
    nms_pre=0,
    min_bbox_size=0,
    score_thr=-1,
    conf_thr=-1,
    nms=None
)
```

Unfortunately, the  post processing can't be done in TensorRT due to missing support for the necessary indexing-operation (indexing with a boolean tensor).

### DIFF
--- a/mmdet/models/dense_heads/yolo_head.py
+++ b/mmdet/models/dense_heads/yolo_head.py
@@ -257,9 +257,11 @@ class YOLOV3Head(BaseDenseHead, BBoxTestMixin):
             # (h, w, num_anchors*num_attrib) -> (h*w*num_anchors, num_attrib)
             pred_map = pred_map.permute(1, 2, 0).reshape(-1, self.num_attrib)
 
-            pred_map[..., :2] = torch.sigmoid(pred_map[..., :2])
-            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i],
-                                               pred_map[..., :4], stride)
+            xy = torch.sigmoid(pred_map[..., :2])
+            xywh = torch.cat((xy, pred_map[..., 2:4]), axis=-1)
+            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i], xywh,
+                                               stride)
+
             # conf and cls
             conf_pred = torch.sigmoid(pred_map[..., 4]).view(-1)
             cls_pred = torch.sigmoid(pred_map[..., 5:]).view(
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/yolo_head.py
```diff
@@ -257,20 +257,23 @@ def _get_bboxes_single(self,
             # (h, w, num_anchors*num_attrib) -> (h*w*num_anchors, num_attrib)
             pred_map = pred_map.permute(1, 2, 0).reshape(-1, self.num_attrib)
 
-            pred_map[..., :2] = torch.sigmoid(pred_map[..., :2])
-            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i],
-                                               pred_map[..., :4], stride)
+            xy = torch.sigmoid(pred_map[..., :2])
```
RunningLeon, line 260: Is this really necessary to work around this way? The old version works fine with me when converting to ONNX.

mmdet/models/dense_heads/yolo_head.py
```diff
@@ -257,20 +257,23 @@ def _get_bboxes_single(self,
...
             cls_pred = torch.sigmoid(pred_map[..., 5:]).view(
                 -1, self.num_classes)  # Cls pred one-hot.
 
             # Filtering out all predictions with conf < conf_thr
             conf_thr = cfg.get('conf_thr', -1)
-            conf_inds = conf_pred.ge(conf_thr).nonzero().flatten()
-            bbox_pred = bbox_pred[conf_inds, :]
-            cls_pred = cls_pred[conf_inds, :]
-            conf_pred = conf_pred[conf_inds]
+            if conf_thr >= 0:
```
RunningLeon, line 272: Here it's better to have this condition as conf_thr > 0.

mmdet/models/dense_heads/yolo_head.py
```diff
@@ -257,20 +257,23 @@ def _get_bboxes_single(self,
...
+                conf_inds = conf_pred > conf_thr
```
RunningLeon, line 273: In my experience with torch=1.6.0, it works as following:
`# conf_inds = conf_pred.ge(conf_thr).nonzero().flatten() # not work`
`mask = conf_pred >= conf_thr`
`conf_inds = mask.nonzero(as_tuple=False).squeeze(1)`

RunningLeon (changes requested):
Thanks for your contribution. Pls. response to the comments.

mmdet/models/dense_heads/yolo_head.py
dsuess: Yeah, this is the most critical part that TensorRT didn't like. I think the reason is that it's an inplace-write on a slice, which doesn't seem to be supported (agreed, it works in ONNX/onnx-runtime though)

mmdet/models/dense_heads/yolo_head.py
dsuess: I'll give it a shot. Actually, I think indexing with a boolean tensor and indexing with indicies from `nonzero` get converted to the same ONNX ops, so the main advantage of your code is that the nonzero-indices get re-used for multiple indexing operations.

RunningLeon: Totally agree. Bool indexing works with 1D, when it comes to 2D tensor, it may fail (my experience). See following warning: `UserWarning: Exporting aten::index operator with indices of type Byte. Only 1-D indices are supported. In any other case, this will produce an incorrect ONNX graph.
"Exporting aten::index operator with indices of type Byte`

RunningLeon (changes requested):
Pls. provide sound proof for in-placing influence on exported onnx with TensorRT as backend.
Thanks a lot.

ZwwWayne:
This PR is continued on https://github.com/open-mmlab/mmdetection/pull/4083. So this PR is closed.

PR #4025, opened 2020-10-29
Fix show dir test robustness

daavoo:
Minor fixes of  #4000

- Don't raise AssertionError if `show-dir` is specified.
- Fix TypeError when joining `show_dir` with `corruption_severity` (int)


hellock (approved):

hellock:
Thanks for the fix!

PR #4069, opened 2020-11-05
add tpfp func in evaluating AP

Johnson-Wang:


mmdet/core/evaluation/mean_ap.py
```diff
@@ -270,6 +270,7 @@ def eval_map(det_results,
              iou_thr=0.5,
              dataset=None,
              logger=None,
+             tpfp_func=None,
```
hellock, line 273: Similar to `collate_fn`, we may rename it to `tpfp_fn`.

hellock (approved):

PR #4082, opened 2020-11-10
Clamp neg probability in freeanchor

ZwwWayne:
This PR tries to resolve the training instability of FreeAnchor when the probabilities are very close to 1 or 0.

### DIFF
--- a/mmdet/models/dense_heads/free_anchor_retina_head.py
+++ b/mmdet/models/dense_heads/free_anchor_retina_head.py
@@ -5,6 +5,8 @@ from mmdet.core import bbox_overlaps
 from ..builder import HEADS
 from .retina_head import RetinaHead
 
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/free_anchor_retina_head.py
```diff
@@ -260,6 +260,10 @@ def negative_bag_loss(self, cls_prob, box_prob):
             Tensor: Negative bag loss in shape (num_img, num_anchors, num_classes).
         """  # noqa: E501, W605
         prob = cls_prob * (1 - box_prob)
```
Johnson-Wang, line 262: Adding one line `prob = prob.clamp(min=EPS, max=1-EPS)` already works.

Johnson-Wang (approved):

PR #4083, opened 2020-11-10
[Enhance]: Better support yolov3 pytorch2onnx

RunningLeon:
Better support yolov3 pytorch2onnx with TensorRT as backend. PR(#3956)

### DIFF
--- a/mmdet/models/dense_heads/yolo_head.py
+++ b/mmdet/models/dense_heads/yolo_head.py
@@ -257,11 +257,9 @@ class YOLOV3Head(BaseDenseHead, BBoxTestMixin):
             # (h, w, num_anchors*num_attrib) -> (h*w*num_anchors, num_attrib)
             pred_map = pred_map.permute(1, 2, 0).reshape(-1, self.num_attrib)
 
-            xy = torch.sigmoid(pred_map[..., :2])
-            xywh = torch.cat((xy, pred_map[..., 2:4]), axis=-1)
-            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i], xywh,
-                                               stride)
-
+            pred_map[..., :2] = torch.sigmoid(pred_map[..., :2])
+            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i],
+                                               pred_map[..., :4], stride)
             # conf and cls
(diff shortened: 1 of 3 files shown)


mmdet/models/dense_heads/yolo_head.py
```diff
@@ -257,20 +257,24 @@ def _get_bboxes_single(self,
             # (h, w, num_anchors*num_attrib) -> (h*w*num_anchors, num_attrib)
             pred_map = pred_map.permute(1, 2, 0).reshape(-1, self.num_attrib)
 
-            pred_map[..., :2] = torch.sigmoid(pred_map[..., :2])
-            bbox_pred = self.bbox_coder.decode(multi_lvl_anchors[i],
-                                               pred_map[..., :4], stride)
+            xy = torch.sigmoid(pred_map[..., :2])
```
ZwwWayne, line 260: We may add comments here to indicate why this change is necessary.

mmdet/models/dense_heads/yolo_head.py
```diff
@@ -257,20 +257,24 @@ def _get_bboxes_single(self,
...
             cls_pred = torch.sigmoid(pred_map[..., 5:]).view(
                 -1, self.num_classes)  # Cls pred one-hot.
 
             # Filtering out all predictions with conf < conf_thr
             conf_thr = cfg.get('conf_thr', -1)
-            conf_inds = conf_pred.ge(conf_thr).nonzero().flatten()
-            bbox_pred = bbox_pred[conf_inds, :]
-            cls_pred = cls_pred[conf_inds, :]
-            conf_pred = conf_pred[conf_inds]
+            if conf_thr > 0:
```
ZwwWayne, line 272: We may add comments here to indicate why this change is necessary. The previous code seems to be more general.

mmdet/models/dense_heads/yolo_head.py
RunningLeon: Pls. check here(#3956) for more detailed info.

mmdet/models/dense_heads/yolo_head.py
RunningLeon: Pls. check here(#3956) for more detailed info.

ZwwWayne (approved):

PR #4087, opened 2020-11-11
[Enhance]: Support some detection models exportable to onnx

RunningLeon:
1. Support retinanet, ssd, fsaf and faster-rcnn exportable to onnx.
2. Add visualization to `pytorch2onnx` tool.
3. Modify `batch_nms` to be exportable to onnx.


mmdet/core/export/pytorch2onnx.py
```diff
@@ -118,12 +118,14 @@ def preprocess_example_input(input_config):
     input_path = input_config['input_path']
     input_shape = input_config['input_shape']
     one_img = mmcv.imread(input_path)
+    one_img = mmcv.imresize(one_img, input_shape[2:][::-1])
+    show_img = one_img.copy()
```
ZwwWayne, line 122: `show_img` is not normalized while `one_img` is normalized. And line 139 pass `show_img` for pytorch2onnx function. Is this expected behavior?

RunningLeon: Yes, `show_img` is for visualization, no need to be normalized.

ZwwWayne (approved):

Issue #4080, opened 2020-11-09
question about faster_rcnn_r50_fpn_1x_coco-person.py config


mrbulb:
in this config, the number of dataset class is set to one classs, but `num_classes` still is 80 in config
https://github.com/open-mmlab/mmdetection/blob/d3cf38d91c454b1a6881e8c36c1e4a66dc5521b8/configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco-person.py#L2

may be the following code should be add to this config?
```python
model = dict(roi_head=dict(bbox_head=dict(num_classes=1)))
```

ZwwWayne:
Yes I forget to update `num_classes` accordingly. Would you like to create a PR for that?

mrbulb:
> Yes I forget to update `num_classes` accordingly. Would you like to create a PR for that?

@ZwwWayne  I have created a PR https://github.com/open-mmlab/mmdetection/pull/4088 to fix this problem 😄

ZwwWayne:
Great! Thanks!

PR #4088, opened 2020-11-11
[fix-bug-pr] fix bug for faster rcnn config

mrbulb:
`num_classes` in model should be the same as the number of dataset class


mrbulb:
This PR is related to this issue: https://github.com/open-mmlab/mmdetection/issues/4080

ZwwWayne (approved):

xianzhongma:
where to find the  pre-trained model for only "person" class

PR #4140, opened 2020-11-18
[Docs] Add model statistics and polish some titles in configs README

dreamerlin:
mmdet


ZwwWayne:
There are some weird items like `Results and Models` and `common settings`. Shall we skip these two items?

ZwwWayne:
> There are some weird items like `Results and Models` and `common settings`. Shall we skip these two items?

It seems that some readme.md of datasets is not correctly named. We can fix it later after merging this PR.

dreamerlin:
mmdet2

ZwwWayne (approved):

PR #4153, opened 2020-11-20
fix yolo neck issue

ceykmc:
1. module "conv{i}" input channels is out_channels[i-1], not in_channels[i]
    based on the code: tmp = conv(out) in forward function.
2. module "detect1" input channels is in_channels[0], whichi is correct.
    module "detect{i+1}" input channels should be in_channels[i] + out_c


xvjiarui:
Hi @ceykmc 
Thanks for your contribution.
We are using `in_channels[i]` as the input channel for simple meaning, although there is a constraint that `in_channels[i]==out_channels[i-1]`

ceykmc:
Hi @xvjiarui 
Thanks for replying.
If `in_channels[i]==out_channels[i-1]` constraint, i think the code is right.
I just want to use another backbone, shufflenetv2 for example, which not satisfy this constraint.
we told the user this constraint in the document, or release this constraint. both are ok.

PR #4163, opened 2020-11-23
Fix incorrect gpu_id with distributed training mode

v-qjqs:
This PR fix the incorrect gpu_id shown in the training log, in the case of distributed training mode.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -8,7 +8,7 @@ import warnings
 import mmcv
 import torch
 from mmcv import Config, DictAction
-from mmcv.runner import init_dist
+from mmcv.runner import get_dist_info, init_dist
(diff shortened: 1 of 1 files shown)


tools/train.py
```diff
@@ -114,6 +114,9 @@ def main():
     else:
         distributed = True
         init_dist(args.launcher, **cfg.dist_params)
+        # re-set gpu_ids with distributed training mode
+        _, world_size = get_dist_info()
```
ZwwWayne, line 118: Should they be put in line 109?

v-qjqs: The args.gpus defaultly is set to None when distributed training mode.

ZwwWayne (approved):

PR #4210, opened 2020-12-01
Update gfl_head.py

ace7chan:
There is a error in pytorch1.6 version when call reduce_mean in this head: input tensor is "int" format, can not call div_ direct. so, change it to float format.

### DIFF
--- a/mmdet/models/dense_heads/gfl_head.py
+++ b/mmdet/models/dense_heads/gfl_head.py
@@ -345,7 +345,7 @@ class GFLHead(AnchorHead):
          bbox_weights_list, num_total_pos, num_total_neg) = cls_reg_targets
 
         num_total_samples = reduce_mean(
-            torch.tensor(num_total_pos).cuda()).item()
+            torch.tensor(num_total_pos).float().cuda()).item()
         num_total_samples = max(num_total_samples, 1.0)
 
         losses_cls, losses_bbox, losses_dfl,\
--- a/mmdet/models/dense_heads/gfl_head.py
+++ b/mmdet/models/dense_heads/gfl_head.py
@@ -345,7 +345,7 @@ class GFLHead(AnchorHead):
          bbox_weights_list, num_total_pos, num_total_neg) = cls_reg_targets
 
         num_total_samples = reduce_mean(
-            torch.tensor(num_total_pos).float().cuda()).item()
+            torch.tensor(num_total_pos, dtype=torch.float, device=device)).item()
         num_total_samples = max(num_total_samples, 1.0)
 
         losses_cls, losses_bbox, losses_dfl,\


mmdet/models/dense_heads/gfl_head.py
```diff
@@ -345,7 +345,7 @@ def loss(self,
          bbox_weights_list, num_total_pos, num_total_neg) = cls_reg_targets
 
         num_total_samples = reduce_mean(
-            torch.tensor(num_total_pos).cuda()).item()
+            torch.tensor(num_total_pos).float().cuda()).item()
```
hellock, line 348: `torch.tensor(num_total_pos, dtype=torch.float, device=device)`

hellock:
Could you fix the linting error? Then it can be merged into the master branch.

hellock:
@ZwwWayne may take over this PR.

shinya7y:
Changing integer division to float division breaks backward compatibility, because it changes loss balance by eliminating rounding error.
A possible solution is adding an argument like [this fix](https://github.com/shinya7y/UniverseNet/commit/122e6b6bca10e564f478ed713fff0f7a8df59dda#diff-46684aaa9343774704db6d90236b2513f19b3a71f8c7732a75ec52dbeaac956a).
Another solution is to rebenchmark on some datasets and notify in the Changelog.

The original ATSS uses [float division](https://github.com/sfzhang15/ATSS/blob/master/atss_core/modeling/rpn/atss/loss.py). Using integer division is a bug as an ATSS reimplementation.
The original GFL uses [integer division](https://github.com/implus/GFocal/blob/master/mmdet/models/anchor_heads/gfl_head.py), following the bug.

ZwwWayne (approved):

b03505036:
But there is still a bug on ATSS loss

pytorch = 1.6 

File "/apdcephfs/private_yiminou/work/official/nfl/mmdetection/mmdet/models/dense_heads/atss_head.py", line 272, in loss
    torch.tensor(num_total_pos).cuda()).item()
  File "/apdcephfs/private_yiminou/work/official/nfl/mmdetection/mmdet/core/utils/dist_utils.py", line 68, in reduce_mean
    dist.all_reduce(tensor.div_(dist.get_world_size()), op=dist.ReduceOp.SUM)
RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.

OpenMMLab-Assistant-007:
Hi!
@ace7chan 
First of all, we want to express our gratitude for your significant PR in the OpenMMLab project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA

If you have WeChat account，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

PR #4233, opened 2020-12-05
add focal_loss in hungarian_assigner.py

kellenf:
some need  to review, this is a base_code

### DIFF
--- a/mmdet/core/bbox/assigners/hungarian_assigner.py
+++ b/mmdet/core/bbox/assigners/hungarian_assigner.py
@@ -42,13 +42,16 @@ class HungarianAssigner(BaseAssigner):
                  bbox_weight=1.,
                  iou_weight=1.,
                  iou_calculator=dict(type='BboxOverlaps2D'),
-                 iou_mode='giou'):
+                 iou_mode='giou',
+                 focal_loss=dict(alpha=0.25, gamma=2.0),
+                 ):
         # defaultly giou cost is used in the official DETR repo.
         self.iou_mode = iou_mode
         self.cls_weight = cls_weight
         self.bbox_weight = bbox_weight
         self.iou_weight = iou_weight
         self.iou_calculator = build_iou_calculator(iou_calculator)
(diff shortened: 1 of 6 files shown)


mmdet/core/bbox/assigners/hungarian_assigner.py
```diff
@@ -42,13 +42,16 @@ def __init__(self,
                  bbox_weight=1.,
                  iou_weight=1.,
                  iou_calculator=dict(type='BboxOverlaps2D'),
-                 iou_mode='giou'):
+                 iou_mode='giou',
+                 focal_loss=dict(alpha=0.25, gamma=2.0),
```
ZwwWayne, line 46: comma is unnecessary.

mmdet/core/bbox/assigners/hungarian_assigner.py
```diff
@@ -116,14 +119,26 @@ def assign(self,
...
+        else:
+            cls_score = cls_pred.softmax(-1)
+            cls_cost = -cls_score[:, gt_labels]  # [num_bboxes, num_gt]
 
         # regression L1 cost
         img_h, img_w, _ = img_meta['img_shape']
         factor = torch.Tensor([img_w, img_h, img_w,
                                img_h]).unsqueeze(0).to(gt_bboxes.device)
         gt_bboxes_normalized = gt_bboxes / factor
+        # MODYFY: xyxy p1-norm in sparse-RCNN CODE which need to review
```
ZwwWayne, line 141: We may also use build loss for bbox distance.

mmdet/core/bbox/assigners/hungarian_assigner.py
kellenf: Haved updated, use `build_match_cost` to build all loss, please review

mmdet/core/bbox/assigners/hungarian_assigner.py
kellenf: updated

PR #4250, opened 2020-12-08
Fix img_norm in hrnet when using fcos detector

### LINKED ISSUE #4232: Different mean values of image normalization for HRNet
I find the different configs of HRNet experiments use different mean values for image normalization.

For example,
[fcos_hrnetv2p_w32_gn-head_4x4_1x_coco.py](https://github.com/open-mmlab/mmdetection/blob/master/configs/hrnet/fcos_hrnetv2p_w32_gn-head_4x4_1x_coco.py) uses the `[102.9801, 115.9465, 122.7717]` as the mean value,
while [fcos_hrnetv2p_w32_gn-head_mstrain_640-800_4x4_2x_coco.py](https://github.com/open-mmlab/mmdetection/blob/master/configs/hrnet/fcos_hrnetv2p_w32_gn-head_mstrain_640-800_4x4_2x_coco.py) uses the `[103.530, 116.280, 123.675]` as the mean value.

Which one is correct?

v-qjqs:
This PR fix #4232, fix different img_norm in hrnet when using fcos.


configs/hrnet/fcos_hrnetv2p_w32_gn-head_4x4_1x_coco.py
```diff
@@ -1,4 +1,6 @@
 _base_ = '../fcos/fcos_r50_caffe_fpn_gn-head_4x4_1x_coco.py'
+img_norm_cfg = dict(
```
ZwwWayne, line 2: This will not change the real img_norm_cfg used during training and testing because img_norm_cfg here is an intermediate variable, if you modify img_norm_cfg, you should also modify `train_pipeline` and `test_pipeline`

hellock:
There are linting errors.

ZwwWayne (approved):

PR #4251, opened 2020-12-08
Support returning indices of kept bboxes when using nms

OceanPang:


mmdet/core/post_processing/bbox_nms.py
```diff
@@ -22,11 +23,12 @@ def multiclass_nms(multi_bboxes,
         max_num (int): if there are more than max_num bboxes after NMS,
             only top max_num will be kept.
         score_factors (Tensor): The factors multiplied to scores before
-            applying NMS
+            applying NMS.
+        return_inds (bool): Whether return the indices of kept bboxes.
```
ZwwWayne, line 27: （bool, optional): xxx, Defaults to False.

ZwwWayne (approved):

PR #4257, opened 2020-12-09
Fix bug in distributed sampler

LXXXXR:


mmdet/datasets/samplers/distributed_sampler.py
```diff
@@ -18,7 +18,9 @@ def __iter__(self):
             indices = torch.arange(len(self.dataset)).tolist()
 
         # add extra samples to make it evenly divisible
-        indices += indices[:(self.total_size - len(indices))]
+        # in case that indices is shorter than half of total_size
+        indices = (indices *
+                   int(self.total_size / len(indices) + 1))[:self.total_size]
```
hellock, line 23: It may be easier to understand with `math.ceil`?

hellock (approved):

hellock:
@ZwwWayne can have a double check.

ZwwWayne (approved):

PR #4272, opened 2020-12-09
Fix detr config for higher performance

ZwwWayne:
This PR changes the default config of DETR from 4 images per GPU to 2 images per GPU for higher performance and match the log and checkpoints.


hellock:
CI fails.

hellock (approved):

PR #4294, opened 2020-12-13
Support to show F1 score in eval_map

ZwwWayne:


mmdet/core/evaluation/mean_ap.py
```diff
@@ -452,18 +456,20 @@ def print_map_summary(mean_ap,
...
         mean_ap = [mean_ap]
 
-    header = ['class', 'gts', 'dets', 'recall', 'ap']
+    header = ['class', 'gts', 'dets', 'recall', 'f1', 'ap']
     for i in range(num_scales):
         if scale_ranges is not None:
             print_log(f'Scale range {scale_ranges[i]}', logger=logger)
         table_data = [header]
         for j in range(num_classes):
+            f1_score = 2 * recalls[i, j] * precisions[i, j] \
```
Johnson-Wang, line 465: What if there is no prediction results at all?

PR #4300, opened 2020-12-14
Fix empirical attention in  resnext backbone error

### LINKED ISSUE #4228: Empirical attention with ResNext backbone
In mmdetectiom v1 I used ResNext backbone with empirical attention 0010_dcn but in v2 I got an error.
Any ideas how to fix this?

My updated config file:
```
_base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
model = dict(

    pretrained='open-mmlab://resnext50_32x4d',
    backbone=dict(
        type='ResNeXt',
        depth=50,
        groups=32,
        base_width=4,
        num_stages=4,
        out_indices=(0, 1, 2, 3),
        frozen_stages=1,
        norm_cfg=dict(type='BN', requires_grad=True),
        style='pytorch',
        plugins=[
            dict(
                cfg=dict(
                    type='GeneralizedAttention',
                    spatial_range=-1,
                    num_heads=8,
                    attention_type='0010',
                    kv_stride=2),
                stages=(False, False, True, True),
                position='after_conv2')
        ],
        dcn=dict(type='DCN', deform_groups=1, fallback_on_stride=False),
        stage_with_dcn=(False, True, True, True)))
```

Error:
```
RuntimeError: Given groups=1, weight of size [256, 256, 1, 1], expected input[2, 512, 41, 73] to have 256 channels, but got 512 channels instead
```

hhaAndroid:
Fixed #4228


hellock:
1. There are linting erros.
2. Better to add some unit tests.

mmdet/models/backbones/resnext.py
```diff
@@ -83,6 +83,39 @@ def __init__(self,
             bias=False)
         self.add_module(self.norm3_name, norm3)
 
+        if self.with_plugins:
+            self.after_conv1_plugin_names = self.make_block_plugins(
+                width, self.after_conv1_plugins)
+            self.after_conv2_plugin_names = self.make_block_plugins(
+                width, self.after_conv2_plugins)
+            self.after_conv3_plugin_names = self.make_block_plugins(
+                width * self.expansion, self.after_conv3_plugins)
```
xvjiarui, line 92: These are implemented in parent class. They could be removed here.

xvjiarui, line 92:
```suggestion
                self.planes * self.expansion, self.after_conv3_plugins)
```

mmdet/models/backbones/resnext.py
```diff
@@ -83,6 +83,39 @@ def __init__(self,
...
+        """
+        assert isinstance(plugins, list)
+        plugin_names = []
+        for plugin in plugins:
+            plugin = plugin.copy()
+            name, layer = build_plugin_layer(
+                plugin,
+                in_channels=in_channels,
+                postfix=plugin.pop('postfix', ''))
+            # fix duplicate plugin. When multiple same name plugins are added, only the last one is kept
```
xvjiarui, line 112: If the user defines two duplicated plugins, there will be no error.

xvjiarui: I suggest deleting all plugins defined in `Bottleneck` manually in `BottleneckX`.

xvjiarui:
Please merge master to pass CI.

xvjiarui (approved):

PR #4303, opened 2020-12-14
Fix #3950, avoid zero or too small value for beta in dynamic rcnn

### LINKED ISSUE #3950: Beta update assert error in dynamic rcnn
When the beta updates, new beta will be the median of the beta_history. It might be 0.0 but the beta used in loss function should be greater than 0.
The assert error occurred.
[https://github.com/open-mmlab/mmdetection/blob/5d07dda8719b0c67c11ad015410bc111e9f90200/mmdet/models/roi_heads/dynamic_roi_head.py#L144](url)
The code to fix it could be as follows:
`if np.abs(np.median(self.beta_history)) < 1e-15:
            new_beta = self.bbox_head.loss_bbox.beta
  else:
            new_beta = min(self.train_cfg.dynamic_rcnn.initial_beta,
                       np.median(self.beta_history))`
I am willing to create a PR to fix it as well.

v-qjqs:
This PR fix #3950, avoid zero or too small value for beta in dynamic rcnn


mmdet/models/roi_heads/dynamic_roi_head.py
```diff
@@ -128,7 +128,7 @@ def _bbox_forward_train(self, x, sampling_results, gt_bboxes, gt_labels,
         bbox_results.update(loss_bbox=loss_bbox)
         return bbox_results
 
-    def update_hyperparameters(self):
+    def update_hyperparameters(self, eps=1e-15):
```
ZwwWayne, line 131: Use EPS=1e-15 as that in atss_head or FCOS head.

ZwwWayne (approved):

PR #4317, opened 2020-12-17
modify the counting way of ckpts

congee524:
modify the counting way of ckpts

### DIFF
--- a/docs/stat.py
+++ b/docs/stat.py
@@ -10,6 +10,7 @@ files = sorted(glob.glob('../configs/*/README.md'))
 stats = []
 titles = []
 num_ckpts = 0
+num_configs = 0
 
 for f in files:
     url = osp.dirname(f.replace('../', url_prefix))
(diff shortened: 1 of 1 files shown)


congee524:
some ckpt files (including some log file) were stored in google drive, so the original method has some problems

congee524:
mmdet_modelzoo_statics

ZwwWayne:
Does it seem that the number of configs is fewer than the number of checkpoints?

congee524:
`fsaf, groie, legacy_1.x` don't have the same number of configs and ckpt, and there are some models share the same config.

But now, there is no need to count config, I will remove later.

ZwwWayne (approved):

docs/stat.py
```diff
@@ -10,6 +10,7 @@
 stats = []
 titles = []
 num_ckpts = 0
+num_configs = 0
```
ZwwWayne, line 13: Delete this if this variable is not used.

ZwwWayne (approved):

PR #4325, opened 2020-12-18
Delete mmdet.ops

hhaAndroid:
delete mmdet.ops because it has been moved to mmcv


ZwwWayne:
The directory of `mmdet.ops` should also be deleted.

ZwwWayne (approved):

PR #4337, opened 2020-12-21
install sklearn for PAA

luuuyi:
Normally, **sklearn** is not a basic package.
But it's necessary for method PPA. https://github.com/open-mmlab/mmdetection/blob/6eb3aad0f91b6e57a5c0f1198f0737f3cc6b2296/mmdet/models/dense_heads/paa_head.py#L11


luuuyi:
https://github.com/open-mmlab/mmdetection/blob/6eb3aad0f91b6e57a5c0f1198f0737f3cc6b2296/mmdet/models/dense_heads/paa_head.py#L11-L14

ZwwWayne:
Since sklearn is only used by PAAHead，we do not want to install this package in the default installation. Therefore, would you like to add sklearn in [this file](https://github.com/open-mmlab/mmdetection/blob/master/requirements/optional.txt)?

v-qjqs (commented):
Hi @luuuyi, it's better to move sklearn to [optional.txt](https://github.com/open-mmlab/mmdetection/blob/master/requirements/optional.txt)

ZwwWayne:
Done in https://github.com/open-mmlab/mmdetection/pull/4339

PR #4359, opened 2020-12-27
Update ATSS for PyTorch 1.6+

shinya7y:
The same fix as GFL. https://github.com/open-mmlab/mmdetection/pull/4210
I haven't checked AP difference caused by eliminating rounding error.


hellock (approved):

hellock:
The regression benchmark will be done.

Issue #4112, opened 2020-11-14
LVIS maskrcnn training bug: TypeError: can't pickle _thread.RLock objects


nemonameless:
Thanks for your error report and we appreciate it a lot.

**Describe the bug**
training on COCO dataset is ok, but when I train on LVIS meet this bug, It seems long time no updating the code about LVIS training?

**Reproduction**
1. What command or script did you run?
```
CUDA_VISIBLE_DEVICES=6,7  bash ./tools/dist_train.sh configs/lvis/mask_rcnn_r50_fpn_sample1e-3_mstrain_1x_lvis_v1.py 2 --work-dir work_dirs/mask_rcnn_r50_fpn_sample1e-3_mstrain_1x_lvis_v1 --no-validate
```
2. Did you make any modifications on the code or config? No
3. What dataset did you use? LVIS

**Environment**

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.
```
sys.platform: linux
Python: 3.6.10 |Anaconda, Inc.| (default, May  8 2020, 02:54:21) [GCC 7.3.0]
CUDA available: False
GCC: gcc (GCC) 5.2.0
PyTorch: 1.4.0
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - Intel(R) Math Kernel Library Version 2019.0.4 Product Build 20190411 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v0.21.1 (Git Hash 7d2fd500bc78936d1d648ca713b901012f470dbc)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
[...]

PR #4382, opened 2021-01-02
Remove custom_classes in LVIS

xvjiarui:


hellock:
Does this PR solves #4112?

hellock (approved):

Issue #3957, opened 2020-10-19
some samples are dropped after `collect_results_xpu` is called in distributed test


sleepfin:
**Checklist**
1. I have searched related issues but cannot get the expected help.
2. The bug has not been fixed in the latest version.

**Describe the bug**
When I start a distributed training with evaluation, I find that if the total number of samples cannot be divided equally on each GPU, some samples are dropped which cause ERROR.

This happend not only in mmdetection, also in mmclassification. For codes in function `collect_results_gpu` or `collect_results_cpu` (In mmdetection/mmdet/apis/test.py)

```python
        # sort the results
        ordered_results = []
        for res in zip(*part_list):
            ordered_results.extend(list(res))
```

Before sort the results, the samples in `part_recv_list` are complete. But after sort the results, samples in `ordered_results` can be incomplete.

I write a demo to explain this:

```python
result_GPU0 = [0, 4, 8, 12]  # one more samples than others
result_GPU1 = [1, 5, 9]
result_GPU2 = [2, 6, 10]
result_GPU3 = [3, 7, 11]

result_all_GPUs = [result_GPU0, result_GPU1, result_GPU2, result_GPU3]

result_reordered = []

for x in zip(*result_all_GPUs):
    result_reordered.extend(x)

print(result_reordered)
```

You can see sample No.12 is missing in `result_reordered`

**Reproduction**
1. What command or script did you run?
2. Did you make any modifications on the code or config? Did you understand what you have modified?
No
3. What dataset did you use?
flowers 5 classification, I split it to train and eval, which has 369 samples in eval part

**Environment**

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.
2. You may add addition that may be helpful for locating the problem, such as
    - How you installed PyTorch [e.g., pip, conda, source]
    - Other environment variables that may be related (such as `$PATH`, `$LD_LIBRARY_PATH`, `$PYTHONPATH`, etc.)

**Error traceback**
If applicable, paste the error trackback here.
```
A placeholder for trackback.
```

**Bug fix**
If you have already identified the reason, you can provide the information here. If you are willing to create a PR to fix it, please also leave a comment here and that would be much appreciated!

If this is a BUG, my suggestion is to use zip_longest, which solved my problem

```python
        # sort the results
        import itertools
        ordered_results = []
        for res in itertools.zip_longest(*part_list):
            ordered_results.extend(filter(lambda x: x is not None, list(res)))
```

[...]

PR #4386, opened 2021-01-03
fix bug of missing results in multi_gpu_test

ZwwWayne:
Resolves https://github.com/open-mmlab/mmdetection/issues/3957

### DIFF
--- a/mmdet/apis/test.py
+++ b/mmdet/apis/test.py
@@ -1,3 +1,4 @@
+import itertools
 import os.path as osp
 import pickle
 import shutil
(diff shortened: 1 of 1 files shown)


mmdet/apis/test.py
```diff
@@ -148,8 +149,8 @@ def collect_results_cpu(result_part, size, tmpdir=None):
             part_list.append(mmcv.load(part_file))
         # sort the results
         ordered_results = []
-        for res in zip(*part_list):
-            ordered_results.extend(list(res))
+        for res in itertools.zip_longest(*part_list):
+            ordered_results.extend(filter(lambda x: x is not None, list(res)))
```
hellock, line 153: `list(res)` is no longer needed, `res` is enough.

hellock: The same fix needs to be applied to `collect_results_gpu`.

hellock (approved):

PR #4389, opened 2021-01-04
Replace imshow_det_bboxes visualization backend

hhaAndroid:
Replace imshow_det_bboxes visualization backend


mmdet/models/detectors/base.py
```diff
@@ -273,7 +274,7 @@ def show_result(self,
                     bbox_color='green',
                     text_color='green',
                     thickness=1,
-                    font_scale=0.5,
+                    font_size=10,
```
ZwwWayne, line 277: We should keep both the original argument and the new argument in case downstream codebases use this function. Also, add a deprecated warning for `font_scale`.

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,107 @@
...
+        assert bboxes.shape[1] == 5
+        scores = bboxes[:, -1]
+        inds = scores > score_thr
+        bboxes = bboxes[inds, :]
+        labels = labels[inds]
+
+    bbox_color = mmcv.color_val(bbox_color)
+    text_color = mmcv.color_val(text_color)
+
+    bbox_color = [color / 255 for color in bbox_color[::-1]]
```
ZwwWayne, line 60: Encapsulate a function to do the similar thing that converts mmcv.color_val to matplotlib color_val

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,107 @@
...
+            bbox_int[1],
+            '%s' % label_text,
+            color=text_color,
+            fontsize=font_size)
+
+    p = PatchCollection(
+        polygons, facecolor='none', edgecolors=color, linewidths=thickness)
+    ax.add_collection(p)
+
+    if out_file is not None:
```
ZwwWayne, line 96: Add unit tests with demo images if necessary.

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,117 @@
...
+        dir_name = osp.abspath(osp.dirname(out_file))
+        mmcv.mkdir_or_exist(dir_name)
+        plt.savefig(out_file)
+    if show:
+        if wait_time == 0:
+            plt.show()
+        else:
+            plt.show(block=False)
```
ZwwWayne, line 114: since there is also an argument `block`, should we keep `block` in the API of `show_result_pyplot`.

mmdet/apis/inference.py
```diff
@@ -162,9 +161,8 @@ def show_result_pyplot(model,
                        img,
                        result,
                        score_thr=0.3,
-                       fig_size=(15, 10),
```
ZwwWayne, line 165: Try not BC-breaking the API. If necessary, please add deprecated warning.

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,117 @@
...
+        polygons.append(Polygon(np_poly))
+        color.append(bbox_color)
+        label_text = class_names[
+            label] if class_names is not None else f'cls {label}'
+        if len(bbox) > 4:
+            label_text += f'|{bbox[-1]:.02f}'
+        ax.text(
+            bbox_int[0],
+            bbox_int[1],
+            '%s' % label_text,
```
ZwwWayne, line 98: Use f-string

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,117 @@
...
+
+    for bbox, label in zip(bboxes, labels):
+        bbox_int = bbox.astype(np.int32)
+        poly = [[bbox_int[0], bbox_int[1]], [bbox_int[0], bbox_int[3]],
+                [bbox_int[2], bbox_int[3]], [bbox_int[2], bbox_int[1]]]
+        np_poly = np.array(poly).reshape((4, 2))
+        polygons.append(Polygon(np_poly))
+        color.append(bbox_color)
+        label_text = class_names[
+            label] if class_names is not None else f'cls {label}'
```
ZwwWayne, line 92: cls -> class

hhaAndroid:
opencv backend (old):
image

matplotlib backend(new):
image

mmdet/apis/inference.py
```diff
@@ -175,13 +175,21 @@ def show_result_pyplot(model,
...
     if hasattr(model, 'module'):
         model = model.module
-    img = model.show_result(img, result, score_thr=score_thr, show=False)
-    plt.figure(figsize=fig_size)
-    plt.imshow(mmcv.bgr2rgb(img))
-    plt.title(title)
-    plt.tight_layout()
-    plt.show(block=block)
+    if block:
+        wait_time = 0
```
ZwwWayne, line 185: Either deprecate `block` with warning or assert the consistency between `block` and `wait_time`.

hhaAndroid:
opencv backend (old):
image

matplotlib backend(new):
image

also you can specify the mask color
image

hiyyg:
@hhaAndroid How about putting the text inside the bbox? Otherwise if the bbox is on the top of the image, the text might not be shown.

hhaAndroid:
> @hhaAndroid How about putting the text inside the bbox? Otherwise if the bbox is on the top of the image, the text might not be shown.

hhaAndroid:
@hiyyg 
I try it
image

Doesn't seem to work very well，due to the size and the distance of the object, the y coordinate of the text is not easy to set

hiyyg:
@hhaAndroid Try a strategy similar to this? https://github.com/facebookresearch/detectron2/blob/master/detectron2/utils/visualizer.py#L702

hhaAndroid:
@hiyyg  Thank you，Let me see

hhaAndroid:
@hiyyg  I found that it can be achieved by 
`verticalalignment="top"
horizontalalignment="left"`

image

ZwwWayne:
Also involves @MeowZheng to review

hellock:
Better to add a filling color for text regions.

hhaAndroid:
> Better to add a filling color for text regions.

image

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,156 @@
...
+        img (str or ndarray): The image to be displayed.
+        bboxes (ndarray): Bounding boxes (with scores), shaped (n, 4) or
+            (n, 5).
+        labels (ndarray): Labels of bboxes.
+        segms (ndarray or None): Masks, shaped (n,h,w) or None
+        class_names (list[str]): Names of each classes.
+        score_thr (float): Minimum score of bboxes to be shown.  Default: 0
+        bbox_color (str or tuple or :obj:`Color`): Color of bbox lines.
+           Default: 'green'
+        text_color (str or tuple or :obj:`Color`): Color of texts.
```
MeowZheng, line 47: "BGR tuple (int)" may help to better understand than "tuple"

MeowZheng (approved):

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Args:
+        img (str or ndarray): The image to be displayed.
+        bboxes (ndarray): Bounding boxes (with scores), shaped (n, 4) or
+            (n, 5).
+        labels (ndarray): Labels of bboxes.
+        segms (ndarray or None): Masks, shaped (n,h,w) or None
+        class_names (list[str]): Names of each classes.
+        score_thr (float): Minimum score of bboxes to be shown.  Default: 0
+        bbox_color (str or BGR tuple (int) or :obj:`Color`):
```
ZwwWayne, line 45: This is only about type. So we should keep tuple(int) and add the order of color in the description like:
```python

bbox_color (str or tuple(int) or :obj:`Color`): Color of bbox lines. The tuple of color should be in BGR order. Default: 'green'
```

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,157 @@
...
+           The tuple of color should be in BGR order. Default: 'green'
+        mask_color (None or str or tuple(int) or :obj:`Color`):
+           Color of masks. The tuple of color should be in BGR order.
+           Default: None
+        thickness (int): Thickness of lines. Default: 2
+        font_scale (float): Font scales of texts. Default: 0.5
+        font_size (int): Font size of texts. Default: 13
+        show (bool): Whether to show the image. Default: True
+        win_name (str): The window name. Default: ''
+        fig_size (tuple): Figure size of the pyplot figure. Default: (15,10)
```
ZwwWayne, line 57: (15, 10)

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,157 @@
...
+
+import matplotlib.pyplot as plt
+import mmcv
+import numpy as np
+from matplotlib.collections import PatchCollection
+from matplotlib.patches import Polygon
+
+
+def color_val_matplotlib(color):
+    """Convert opencv color value to matplotlib (BGR->RGB->Norm)"""
```
ZwwWayne, line 12: May use standard docstring

mmdet/core/visualization/image.py
```diff
@@ -0,0 +1,157 @@
...
+        fig_size (tuple): Figure size of the pyplot figure. Default: (15,10)
+        wait_time (float): Value of waitKey param. Default: 0.
+        out_file (str or None): The filename to write the image. Default: None
+
+    Returns:
+        ndarray: The image with bboxes drawn on it.
+    """
+    warnings.warn('"font_scale" will be deprecated in v2.9.0,'
+                  'Please use "font_size"')
+    assert bboxes.ndim == 2
```
ZwwWayne, line 66: should add assertion messages for assert.

mmdet/apis/inference.py
```diff
@@ -164,7 +163,8 @@ def show_result_pyplot(model,
                        score_thr=0.3,
                        fig_size=(15, 10),
                        title='result',
-                       block=True):
+                       block=True,
```
ZwwWayne, line 166: The argument block does not actually work because it is not passed to `model.show_result`.

tests/test_visualization.py
```diff
@@ -0,0 +1,58 @@
...
+from mmdet.core import visualization as vis
+
+
+def test_color():
+    assert vis.color_val_matplotlib(mmcv.Color.blue) == (0., 0., 1.)
+    assert vis.color_val_matplotlib('green') == (0., 1., 0.)
+    assert vis.color_val_matplotlib((1, 2, 3)) == (3 / 255, 2 / 255, 1 / 255)
+    assert vis.color_val_matplotlib(100) == (100 / 255, 100 / 255, 100 / 255)
+    assert vis.color_val_matplotlib(np.zeros(3, dtype=np.int)) == (0., 0., 0.)
+    with pytest.raises(TypeError):
```
ZwwWayne, line 19: Add a comment above each `pytest.raises` to indicate what you are testing.

ZwwWayne (approved):

PR #4398, opened 2021-01-05
Support multi ious eval

yhcao6:


Johnson-Wang (approved):

mmdet/datasets/voc.py
```diff
@@ -59,24 +59,28 @@ def evaluate(self,
...
-            mean_ap, _ = eval_map(
-                results,
-                annotations,
-                scale_ranges=None,
-                iou_thr=iou_thr,
-                dataset=ds_name,
-                logger=logger)
-            eval_results['mAP'] = mean_ap
+            mean_aps = []
+            for iou_thr in iou_thrs:
```
ZwwWayne, line 70: It seems to be a little bit weird that we put the logic of processing multiple iou_thr outside the eavl_map function. This forces the users to write similar logic every time we call eval_map. Should we put the logic into eval_map?

ZwwWayne (approved):

ZwwWayne:
Please fix the unit test

PR #4408, opened 2021-01-07
Add batch inference docs about deprecate ImageToTensor

hhaAndroid:
Add batch inference docs about deprecate ImageToTensor

### DIFF
--- a/docs/1_exist_data_model.md
+++ b/docs/1_exist_data_model.md
@@ -236,6 +236,46 @@ data = dict(train=dict(...), val=dict(...), test=dict(samples_per_gpu=2, ...))
 
 Or you can set it through `--cfg-options` as `--cfg-options data.test.samples_per_gpu=2`
 
+### Deprecate ImageToTensor
+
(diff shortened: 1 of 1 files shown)


docs/1_exist_data_model.md
```diff
@@ -236,6 +236,46 @@ data = dict(train=dict(...), val=dict(...), test=dict(samples_per_gpu=2, ...))
 
 Or you can set it through `--cfg-options` as `--cfg-options data.test.samples_per_gpu=2`
 
+### Deprecate ImageToTensor
```
ZwwWayne, line 239: Change the title to be Batch Inference, and add the following content as notice.

docs/1_exist_data_model.md
```diff
@@ -236,6 +236,8 @@ data = dict(train=dict(...), val=dict(...), test=dict(samples_per_gpu=2, ...))
 
 Or you can set it through `--cfg-options` as `--cfg-options data.test.samples_per_gpu=2`
 
+### Batch Inference
```
ZwwWayne, line 239: The title should not be put here as the examples are not for batch inference.

ZwwWayne (approved):

PR #4410, opened 2021-01-07
Fix cpu device inference error

hhaAndroid:
Fix cpu device inference error


hhaAndroid:
I tested all models by running the demo/image.demo.py in CPU mode. So far, only the above errors are found in the cascade_rpn  model Until the commit id is **4824dc0c0c0b3cb701376314db9c2672973c140f**

ZwwWayne (approved):

Issue #2001, opened 2020-01-19
How to implement SWA（Stochastic Weight Averaging）in mmdetection？


clw5180:
just as above

ZwwWayne:
You can implement it [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/train.py#L172) by importing the SWA since it is already implemented in `[torchcontrib](https://github.com/pytorch/contrib)` as indicated by the [blog](https://pytorch.org/blog/stochastic-weight-averaging-in-pytorch/).

zr526799544:
> You can implement it [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/train.py#L172) by importing the SWA since it is already implemented in [torchcontrib](https://github.com/pytorch/contrib) as indicated by the [blog](https://pytorch.org/blog/stochastic-weight-averaging-in-pytorch/).
follow the blog I find it seems that I need add opt.swap_swa_sgd() at the end of training before save the model. so that it seems I have to motify the Runner in mmcv. But It doesn't work. Could you give me some advise?

ZwwWayne:
By saying **it does not work** you mean? Maybe the SWA could not improve the detection performance or maybe your implementation is not right. I am afraid that I cannot do much help here.

zr526799544:
> By saying **it does not work** you mean what? Maybe the SWA could not improve the detection performance or maybe your implementation is not right. I am afraid that I cannot do much help here.

My implementation is not right, I can not find a exactly right hook time to insert ‘’opt.swap_swa_sgd()‘’. I can not find end_of_train in /mmcv/runner/hooks/hook.py

ZwwWayne:
> > By saying **it does not work** you mean what? Maybe the SWA could not improve the detection performance or maybe your implementation is not right. I am afraid that I cannot do much help here.
> 
> My implementation is not right, I can not find a exactly right hook time to insert ‘’opt.swap_swa_sgd()‘’. I can not find end_of_train in /mmcv/runner/hooks/hook.py

For now, maybe you can use `every_n_epochs`, and set the epochs manually, for example, set the interval to be 12, and train the model for 12 epoch.

zr526799544:
> > > By saying **it does not work** you mean what? Maybe the SWA could not improve the detection performance or maybe your implementation is not right. I am afraid that I cannot do much help here.
> > 
> > 
[...]

PR #4414, opened 2021-01-07
swa object detection

hyz-xmaster:
Recently, we found the simple technique, [Stochastic Weights Averaging](https://arxiv.org/abs/1803.05407) (SWA),  was very effective in object detection. Through extensive experiments, we discovered a good policy of applying it to object detection, which can consistently achieve ~1.0 AP improvement over various object detectors on the COCO benchmark.  The policy is surprisingly simple: **training your detector for an extra 12 epochs using cyclical learning rates and then averaging these 12 checkpoints as your final detection model**. This does not incur any inference cost and is really worth trying. To find full details, please have a look at our report at [https://arxiv.org/abs/2012.12645](https://arxiv.org/abs/2012.12645).

Therefore, we implemented the SWA training with MMDetection so that the community (#2001) can benefit from this simple technique. The idea of the implementation can be found [here](https://github.com/hyz-xmaster/swa_object_detection#instructions). 

I create this pull request also as the discussion thread, according to the Contributing Guideline. If you think this PR is not suitable, I can close it then.


ZwwWayne:
Hi @hyz-xmaster ,
Thanks for your kind PR. We will review it ASAP.

Johnson-Wang:
Hi thanks for your contribution. This work is impressive. There are a few comments after an overview of this PR. 
- I note that the improvement of 1x training + 1x finetuning with swa would increase the performance by about 1% AP. However, the performance of a direct 2x training is actually comparable to that. I wonder if it is possible to show the comparison given the same training budget. For example, the 1x training + 1x swa finetuning should be compared to direct 2x training, and 2x training + 1x swa finetuning should be compared to direct 3x training. 
- There are some changes in `mmdet/apis/train.py`. This could cause multiple compatibility issues for other algorithms and downstream repos. I suggest reformatting SWAHook or adding a new feature in [mmcv runners](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/epoch_based_runner.py).

hyz-xmaster:
> Hi thanks for your contribution. This work is impressive. There are a few comments after an overview of this PR.
> 
> * I note that the improvement of 1x training + 1x finetuning with swa would increase the performance by about 1% AP. However, the performance of a direct 2x training is actually comparable to that. I wonder if it is possible to show the comparison given the same training budget. For example, the 1x training + 1x swa finetuning should be compared to direct 2x training, and 2x training + 1x swa finetuning should be compared to direct 3x training.
> * There are some changes in `mmdet/apis/train.py`. This could cause multiple compatibility issues for other algorithms and downstream repos. I suggest reformatting SWAHook or adding a new feature in [mmcv runners](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/epoch_based_runner.py).

Hi, thanks for your comments. 

* Good point. In fact, we have considered this, so we use the Mask RCNN-2x instead of the Mask RCNN-1x as our baseline detector for ablation experiments in our report. If you have a close look at our report, you can find the implicit comparison of 1x training + 1x swa with 2x training, and the former performs better than the latter. Even though we do not conduct the comparison between 2x training + 1x swa and 3x training, which is time-consuming, we believe the former will still have a better performance than the latter. 
* For the implementation, I have thought a lot about the flexibility and  the amount of changes to the codebase, and the current implementation is what I can think of. MMCV runner is a more basic function so I did not work on it. For the reimplementation, I hope you can help with it, because I am not a professional coder and do not have a clear overview of the whole code frame. In contrast, you pros may have much better ideas of the implementation. So look forward to your shot.😀

Johnson-Wang:
The [ema hook](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/ema.py) may help you to restructure your code. We may target on wrapping all pipelines in a single hook class.

Issue #4199, opened 2020-11-29
ValueError: need at least one array to concatenate


jevenail:
I tried it on windows 10 platform,and when I want to train my dataset,error occurs .And here is the log.
How can I solve my problem?

'tail' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
'gcc' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
fatal: not a git repository (or any of the parent directories): .git
2020-11-29 16:17:53,347 - mmdet - INFO - Environment info:
------------------------------------------------------------
sys.platform: win32
Python: 3.6.12 |Anaconda, Inc.| (default, Sep  9 2020, 00:29:25) [MSC v.1916 64 bit (AMD64)]
CUDA available: True
GPU 0: GeForce RTX 2060 with Max-Q Design
CUDA_HOME: C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.1
NVCC: Not Available
GCC: n/a
PyTorch: 1.6.0+cu101
PyTorch compiling details: PyTorch built with:
  - C++ Version: 199711
  - MSVC 192628806
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191125 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v1.5.0 (Git Hash e2ac1fac44c5078ca927cb9b90e1b3066a0b2ed0)
  - OpenMP 2019
  - CPU capability usage: AVX2
  - CUDA Runtime 10.1
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_37,code=compute_37
  - CuDNN 7.6.4
  - Magma 2.5.2
[...]

PR #4430, opened 2021-01-12
Customize dataset doc

v-qjqs:
This PR enhance #4199, add more detailed docs for training customized dataset with existing COCO format.


docs/tutorials/customize_dataset.md
```diff
@@ -49,36 +49,90 @@ There are three necessary keys in the json file:
...
+# explicitly over-write all the `num_classes` field from default 80 to 5.
+model = dict(bbox_head=dict(num_classes=5))
+```
+
+#### 2. Check the annotations of the customized dataset
+
+Assuming your customized dataset is COCO format, make sure you have the correct annotations in the customized dataset:
+
+1. The length for `categories` field should exactly equal the number of classes (e.g. 5 in this example) in your config file.
+2. The `name` in `categories` field should have exactly the same order with `classes` field in your config file.
```
ZwwWayne, line 108: Can we make sure this claim is correct?

docs/tutorials/customize_dataset.md
```diff
@@ -49,36 +49,90 @@ There are three necessary keys in the json file:
...
         classes=classes,
-        ann_file='path/to/your/test/data',
-        ...))
-...
+        ann_file='path/to/your/test/annotation_data',
+        img_prefix='path/to/your/test/image_data'))
+
+# 2. model settings
+# explicitly over-write all the `num_classes` field from default 80 to 5.
+model = dict(bbox_head=dict(num_classes=5))
```
ZwwWayne, line 100: Also show examples of two-stage detectors including bboxhead and mask head.

docs/tutorials/customize_dataset.md
v-qjqs: I tested my customized config based on a retinanet_r50_fpn model.
I guess at least the length for `categories` field should no less than the number of classes.

v-qjqs: At least the above are recommended and safe way to train the customized dataset with COCO format.

v-qjqs: Actually there is unnecessary for the `name` in `categories` field to have exactly the same order with `classes` field in the config file, as well as there should be exactly the same elements between them.
MMDetection automatically maps the uncontinuous `id` in `categories` to the continuous label indices. So it is the string order of `name` and `id` in `categories` field that matters, while the string order of `classes` in config does not matters.

docs/tutorials/customize_dataset.md
```diff
@@ -49,36 +49,107 @@ There are three necessary keys in the json file:
...
+    # explicitly over-write all the `num_classes` field from default 80 to 5.
+    mask_head=dict(num_classes=5)))
+```
+
+#### 2. Check the annotations of the customized dataset
+
+Assuming your customized dataset is COCO format, make sure you have the correct annotations in the customized dataset:
+
+1. The length for `categories` field in annotations should exactly equal the tuple length of `classes` fields in your config, meaning the number of classes (e.g. 5 in this example).
+2. The `classes` fields in your config file should have exactly the same elements with the `name` in `categories` of annotations. MMDetection automatically maps the uncontinuous `id` in `categories` to the continuous label indices. So it is the string order of `name` in `categories` field that matters, while the string order of `classes` in config does not matters.
```
ZwwWayne, line 125: The string order of `classes` in config matters because during visualization it may use the class order in config. Check this.

docs/tutorials/customize_dataset.md
```diff
@@ -49,36 +49,107 @@ There are three necessary keys in the json file:
...
+3. The `category_id` in `annotations` field should be valid, meaning all values in `category_id` should belong to `id` in `categories`.
```
ZwwWayne, line 126: ...meaning all... -> ..., i.e., all...

docs/tutorials/customize_dataset.md
```diff
@@ -49,36 +49,107 @@ There are three necessary keys in the json file:
 - `annotations`: contains the list of instance annotations.
 - `categories`: contains the list of categories names and their ID.
 
-After the data pre-processing, the users need to further modify the config files to use the dataset.
-Here we show an example of using a custom dataset of 5 classes, assuming it is also in COCO format.
+After the data pre-processing, there are two steps for users to train the customized new dataset with existing format (e.g. COCO format):
+
+1. modify the config file for using the customized dataset.
```
ZwwWayne, line 54: Modify

docs/tutorials/customize_dataset.md
v-qjqs: Yeah, the string order of `classes` in config matters because it is used during the visualization of predicted bounding boxes.
So the `name` in `categories` field should have exactly the same order with `classes` field in your config file.

ZwwWayne (approved):

PR #4455, opened 2021-01-14
fix corner case error in loss_bbox function within gfl_head

ceykmc:
`if not torch.any(weight_targets) > 0`
invoke GIoULoss forward function,
`if weight is not None and not torch.any(weight > 0): return (pred * weight).sum()  # 0`

in this situation, weight_targets ndim is 1, and pos_decode_bbox_pred ndim is 2
they can not multiply

### DIFF
--- a/mmdet/models/dense_heads/gfl_head.py
+++ b/mmdet/models/dense_heads/gfl_head.py
@@ -272,7 +272,8 @@ class GFLHead(AnchorHead):
             loss_bbox = self.loss_bbox(
                 pos_decode_bbox_pred,
                 pos_decode_bbox_targets,
-                weight=weight_targets,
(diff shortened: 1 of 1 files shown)


shinya7y:
Fixing [iou_loss.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/iou_loss.py) may be better to avoid similar issues in other heads.

E.g., replacing 5 lines
``` python
return (pred * weight).sum()  # 0
```
with
``` python
if pred.dim() == weight.dim() + 1:
    weight = weight.unsqueeze(1)
return (pred * weight).sum()  # 0
```

ceykmc:
agree with you

ceykmc:
agree with you
maybe i should create an issue

Issue #4421, opened 2021-01-08
Batch inference when calling model() with list of ndarrays?


dave-andersen:
**Describe the feature**
At the moment, the inference API assumes that if you pass in an ndarray, it represents a single image.

Passing in a list of ndarrays causes the code to incorrectly treat them as if they were filenames;
passing in a 4D tensor (i.e., img with a batch dimension) doesn't get split before being passed
to non-batch functions in the pipeline.

This is in conflict with the API comments, which indicates that imgs can be a list[str/ndarray].

Relevant code is mmdet/apis/inference.py 
in `inference_detector()`

Is fixing this desirable, e.g., would a patch be accepted (or is there already a fix in the pipeline)?

ZwwWayne:
Hi @dave-andersen ,
Thanks for your kind suggestion. The fix is desirable as the current models support the batched inference thus this feature should also be available for `inference_detector`.  Would you like to create a PR for that?

dave-andersen:
Yep. I've sent PR #4462 for this.

PR #4462, opened 2021-01-15
Support batching in synchronous inference API

dave-andersen:
This addresses issue #4421 for the synchronous API, allowing the caller to pass in either a single image (existing functionality) or a list of images. To avoid changing the API, the code assumes the caller means to use the passed in images as a single batch and does not chunk internally.


ZwwWayne:
Hi @dave-andersen ,
Thanks for your kind contribution. Could you also add a unit test to test the functionality of the batch inference of the demo?

ZwwWayne (approved):

PR #4468, opened 2021-01-18
add onnx simplify option to pytorch2onnx tool

RunningLeon:
This PR simply add `simplify` option to pytorch2onnx tool. Users could use this feature to simplify the exported onnx models.

### DIFF
--- a/docs/tutorials/pytorch2onnx.md
+++ b/docs/tutorials/pytorch2onnx.md
@@ -91,7 +91,7 @@ Notes:
 
 ## Reminders
 
-- `mmcv.onnx.simplify` feature is based on [onnx-simplifier](https://github.com/daquexian/onnx-simplifier). If you want to try it, please refer to [onnx in `mmcv`](https://github.com/open-mmlab/mmcv/blob/master/docs/onnx.md) and [onnxruntime op in `mmcv`](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md) for more information.
+- `mmcv.onnx.simplify` feature is based on [onnx-simplifier](https://github.com/daquexian/onnx-simplifier). If you want to try it, please refer to [onnx in `mmcv`](https://mmcv.readthedocs.io/en/latest/onnx.html) and [onnxruntime op in `mmcv`](https://mmcv.readthedocs.io/en/latest/onnxruntime_op.html) for more information.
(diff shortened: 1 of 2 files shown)


docs/tutorials/pytorch2onnx.md
```diff
@@ -90,6 +91,7 @@ Notes:
 
 ## Reminders
 
+- `mmcv.onnx.simplify` feature is based on [onnx-simplifier](https://github.com/daquexian/onnx-simplifier). If you want to try it, please refer to [onnx in `mmcv`](https://github.com/open-mmlab/mmcv/blob/master/docs/onnx.md) and [onnxruntime op in `mmcv`](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md) for more information.
```
ZwwWayne, line 94: May use the documentation in readthedocs.

tools/pytorch2onnx.py
```diff
@@ -20,7 +20,8 @@ def pytorch2onnx(config_path,
                  verify=False,
                  normalize_cfg=None,
                  dataset='coco',
-                 test_img=None):
+                 test_img=None,
+                 do_simplify=False):
```
ZwwWayne, line 24: do_simplify -> simplify

tools/pytorch2onnx.py
```diff
@@ -52,10 +53,31 @@ def pytorch2onnx(config_path,
         opset_version=opset_version)
 
     model.forward = orig_model.forward
+
+    # simplify onnx model
+    if do_simplify:
+        from packaging import version
```
ZwwWayne, line 59: Seems to be new dependency. We may try not to use it by following [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/__init__.py).

tools/pytorch2onnx.py
```diff
@@ -52,10 +53,31 @@ def pytorch2onnx(config_path,
...
-        from mmdet.core import get_classes
+        from mmdet.core import get_classes, bbox2result
         from mmdet.apis import show_result_pyplot
+
+        ort_custom_op_path = ''
+        try:
+            from mmcv.ops import get_onnxruntime_op_path
+            ort_custom_op_path = get_onnxruntime_op_path()
+        except (ImportError, ModuleNotFoundError):
+            pass
```
ZwwWayne, line 79: If we pass here, what would happen? Should we raise an warning?

RunningLeon: Added a warning. It's Ok to pass if the input model does not have a custom op from `mmcv`.

ZwwWayne (approved):

PR #4480, opened 2021-01-19
dirty data check

yuzhj:
there may exists some dirty data in our own datasets, that will lead to training failure


ZwwWayne:
Hi @yuzhj ,
Modifying the dataset might not be so general. We suggest implement an annotation filter pipeline and use that pipeline in your config. In this way, you only need to implement the pipeline once and apply it to all the datasets.

PR #4487, opened 2021-01-20
add parameter check

tangyanf:
add parameter check in BBoxTestMixin->merge_aug_bboxes()
when img_meta do not have 'flip_direction' the program will crash，and the default parameter in bbox_mapping_back（flip_direction='horizontal'） will not be used。

### DIFF
--- a/mmdet/models/dense_heads/dense_test_mixins.py
+++ b/mmdet/models/dense_heads/dense_test_mixins.py
@@ -25,7 +25,7 @@ class BBoxTestMixin(object):
             scale_factor = img_info[0]['scale_factor']
             flip = img_info[0]['flip']
             flip_direction = (img_info[0]['flip_direction']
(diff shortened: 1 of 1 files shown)


xvjiarui:
Hi @tangyanf 
Could you please describe in which case there will be no flip direction?

ZwwWayne:
@tangyanf ,
Thanks for your kind contribution. Would you like to sign the CLA first thus we could merge the PR?

tangyanf:
> Hi @tangyanf
> Could you please describe in which case there will be no flip direction?

Hi, I think it should be checked whether the user forgot to set the flip direction, if not set, the default value of func bbox_mapping_back() will be used.

tangyanf:
> @tangyanf ,
> Thanks for your kind contribution. Would you like to sign the CLA first thus we could merge the PR?

ok

PR #4525, opened 2021-01-25
Fix Missing documentation

### LINKED ISSUE #4482: Missing documentation: Create datasets
There is this documentation: https://github.com/open-mmlab/mmdetection/blob/master/docs/1_exist_data_model.md#prepare-datasets

But it's missing crucial steps so the documentation is barely useable.

Please extend the documentation by 

1) Providing URLs to the datasets. They are hard to find (unecessary step for users) and in a few cases unclear which version/which format is being used.

2) Something seems off with the Pascal VOC dataset. From the folder structure you expect both the 2012 and 2007 dataset to exist, also `https://github.com/open-mmlab/mmdetection/blob/master/configs/_base_/datasets/voc0712.py` looks like that. Aren't previous content datasets included in the newer ones? (EDIT: Just figured it out by reading the history table on the dataset page → would be great to document this tribal knowledge)

hhaAndroid:
resolve #4482

### DIFF
--- a/docs/1_exist_data_model.md
+++ b/docs/1_exist_data_model.md
@@ -145,7 +145,7 @@ This section will show how to test existing models on supported datasets.
 
 ### Prepare datasets
 
-Public datasets like Pascal VOC and COCO are available from official websites or mirrors.
+Public datasets like [Pascal VOC](http://host.robots.ox.ac.uk/pascal/VOC/index.html)  or [Pascal VOC Mirror](https://pjreddie.com/projects/pascal-voc-dataset-mirror/) and [COCO](https://cocodataset.org/#download) are available from official websites or mirrors. Note:  In the detection task, Pascal VOC 2012  is an extension of Pascal VOC 2007 without overlap, usually  together use it.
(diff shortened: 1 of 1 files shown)


docs/1_exist_data_model.md
```diff
@@ -145,7 +145,7 @@ This section will show how to test existing models on supported datasets.
 
 ### Prepare datasets
 
-Public datasets like Pascal VOC and COCO are available from official websites or mirrors.
+Public datasets like [Pascal VOC](http://host.robots.ox.ac.uk/pascal/VOC/index.html) or [Pascal VOC Mirror](https://pjreddie.com/projects/pascal-voc-dataset-mirror/) and [COCO](https://cocodataset.org/#download) are available from official websites or mirrors. Note: In the detection task, Pascal VOC 2012 is an extension of Pascal VOC 2007 without overlap, and we usually use them together.
```
hellock, line 148: We may just list the official website here.

ZwwWayne (approved):

PR #4526, opened 2021-01-25
Support batch inference 

hhaAndroid:
follow up of #4462 
 reslove #4421 

update some code and add unit test


jovenwayfarer:
Hello! So the new inference makes all images in the list as one batch? How to use this code to inference millions of images for example? I want to use batching for to do it. Thanks!

hhaAndroid:
> Hello! So the new inference makes all images in the list as one batch? How to use this code to inference millions of images for example? I want to use batching for to do it. Thanks!

Yes,  `inference_detector` has implemented batch inference, you can achieve your needs based on this API

jovenwayfarer:
@hhaAndroid Thanks! Have you tested speed vs single image inference on the last set of images?

hhaAndroid:
> @hhaAndroid Thanks! Have you tested speed vs single image inference on the large set of images?

Not yet.

ZwwWayne (approved):

reap195014:
Do you support exporting ONNX model with multi-batch? Currently the ONNX model exported by pytorch2onnx.py only has batch dimention in backbone.

hhaAndroid:
> Do you support exporting ONNX model with multi-batch? Currently the ONNX model exported by pytorch2onnx.py only has batch dimention in backbone.

Only one image is needed when exporting ONNX model, but the ONNX model exported by pytorch2onnx.py supports the batch inference

reap195014:
> > Do you support exporting ONNX model with multi-batch? Currently the ONNX model exported by pytorch2onnx.py only has batch dimention in backbone.
> 
> Only one image is needed when exporting ONNX model, but the ONNX model exported by pytorch2onnx.py supports the batch inference

Thank you for replying. I exported faster rcnn with shape 16x3x800x1216 input, but the ONNX model output boxes is still in shape 100x5. Shouldn't it be 16x100x5 for batch inference? Do you have any tutorial for ONNX batch inference?

hhaAndroid:
@reap195014  Sorry, there is an error in my reply. The batch inference is not supported for now since the exported ONNX model includes post-processing, It will remove the batch dimension.  We will support batch inference in the future.

ZwwWayne:
This PR will be merged after resolving conflicts.

ZwwWayne (approved):

ZwwWayne (approved):

Issue #4527, opened 2021-01-25
Questions about TridentNet


HAOCHENYE:
In section 4.2 of the paper:
**Similar to SNIP [38], we define a valid range [li; ui] for each branch i. During training, we only select the proposals and ground truth boxes whose scales fall in the corresponding valid range of each branch.**
However, I have not found any relevant implemention in TridentRoIHead about that TridentNet's branches can seperately regress different scale bboxes. Have I missed any details?

jshilong:
Actuall, we implement TridentNet-Fast as it in [Detectron2](https://github.com/facebookresearch/detectron2/tree/master/projects/TridentNet),
@Chrisfsj2051, please add a comment to readme of `TridentNet`.

Chrisfsj2051:
Hi @jshilong, please refer to https://github.com/open-mmlab/mmdetection/pull/4533

PR #4533, opened 2021-01-26
Update README.md in TridentNet

Chrisfsj2051:
resolves https://github.com/open-mmlab/mmdetection/issues/4527


configs/tridentnet/README.md
```diff
@@ -2,6 +2,8 @@
 
 ## Introduction
 
+Please note that similar to [Detectron2](https://github.com/facebookresearch/detectron2/tree/master/projects/TridentNet), we haven't implemented the Scale-aware Training Scheme in section 4.2 of the paper.
```
ZwwWayne, line 5: Put the note in the results and models section after the table.
The note should start with `**Note**`.

configs/tridentnet/README.md
```diff
@@ -1,7 +1,6 @@
 # Scale-Aware Trident Networks for Object Detection
 
 ## Introduction
-
```
ZwwWayne, line 4: Could you also add a tag here as `[ALGORITHM]` as that in [here](https://github.com/open-mmlab/mmdetection/tree/master/configs/cascade_rcnn).

ZwwWayne (approved):

PR #4534, opened 2021-01-26
[WIP] dla backbone

yuzhj:
limited by the poor performance of the GPU, I only verified the performance on the pascal voc dataset.
|algorithm|backbone|Mem (GB)|mAP|
|------| ------ |-----|-----|
|faster-rcnn|dla60|2.6|0.808|
|faster-rcnn|resnet50|2.6|0.795|

It's just a preliminary version now.

### DIFF
--- a/configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
+++ b/configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
@@ -16,5 +16,3 @@ model = dict(
         in_channels=[128, 256, 512, 1024],
         out_channels=256,
         num_outs=5))
-
-find_unused_parameters = True
--- a/configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
+++ b/configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
@@ -1,0 +1,20 @@
+_base_ = '../cascade_rcnn/cascade_mask_rcnn_r50_fpn_20e_coco.py'
+
+model = dict(
+    pretrained='open-mmlab://dla60',
+    backbone=dict(
+        type='DLANet',
+        depth=60,
+        num_stages=4,
+        out_indices=(0, 1, 2, 3),
+        frozen_stages=1,
+        norm_cfg=dict(type='BN', requires_grad=True),
+        norm_eval=True,
+        style='pytorch'),
+    neck=dict(
+        type='FPN',
+        in_channels=[128, 256, 512, 1024],
+        out_channels=256,
+        num_outs=5))
+
+find_unused_parameters = True
--- a/configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
+++ b/configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
@@ -16,5 +16,3 @@ model = dict(
         in_channels=[128, 256, 512, 1024],
         out_channels=256,
         num_outs=5))
-
-find_unused_parameters = True
--- a/configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
+++ b/configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
@@ -1,0 +1,20 @@
+_base_ = '../faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py'
+
+model = dict(
+    pretrained='open-mmlab://dla60',
+    backbone=dict(
+        type='DLANet',
+        depth=60,
+        num_stages=4,
+        out_indices=(0, 1, 2, 3),
+        frozen_stages=1,
+        norm_cfg=dict(type='BN', requires_grad=True),
+        norm_eval=True,
+        style='pytorch'),
+    neck=dict(
+        type='FPN',
+        in_channels=[128, 256, 512, 1024],
+        out_channels=256,
+        num_outs=5))
+
+find_unused_parameters = True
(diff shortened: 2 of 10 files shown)


configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
```diff
@@ -0,0 +1,20 @@
...
+        norm_cfg=dict(type='BN', requires_grad=True),
+        norm_eval=True,
+        style='pytorch'),
+    neck=dict(
+        type='FPN',
+        in_channels=[128, 256, 512, 1024],
+        out_channels=256,
+        num_outs=5))
+
+find_unused_parameters = True
```
ZwwWayne, line 20: This should be unnecessary as only the backbone and neck are changed. Otherwise, we should clean the code of implementation.

mmdet/models/backbones/dla.py
```diff
@@ -0,0 +1,640 @@
...
+            out = cp.checkpoint(_inner_forward, x, identity)
+        else:
+            out = _inner_forward(x, identity)
+
+        out = self.relu(out)
+
+        return out
+
+
+class Bottleneck(nn.Module):
```
ZwwWayne, line 103: It seems that BottleNeck and BasicBlock are not changed, can we simply use them from those in resnet.py?

mmdet/models/backbones/dla.py
```diff
@@ -0,0 +1,640 @@
...
+            x2 = self.tree2(x1)
+            x = self.root(x2, x1, *children)
+        else:
+            children.append(x1)
+            x = self.tree2(x1, children=children)
+        return x
+
+
+@BACKBONES.register_module()
+class DLANet(nn.Module):
```
ZwwWayne, line 460: Can DALNet inherit from ResNet to reduce some code in initialization?

ZwwWayne:
Thanks for the kind contribution. Usually, unit tests are necessary to check the implementation. Could you follow the [examples](https://github.com/open-mmlab/mmdetection/blob/master/tests/test_models/test_backbones.py) to add unit tests for DLA?

yuzhj:
> Thanks for the kind contribution. Usually, unit tests are necessary to check the implementation. Could you follow the [examples](https://github.com/open-mmlab/mmdetection/blob/master/tests/test_models/test_backbones.py) to add unit tests for DLA?

I'm sorry for the late reply.I will add unit test in the near feature

mmdet/models/backbones/dla.py
yuzhj: @ZwwWayne the arguments of forward function is different

configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
yuzhj: @ZwwWayne There exist two arguments in the forward function of BottleNeck and BasicBlock. If I delete this line, the procedure will throw errors.

yuzhj:
This is the [address](https://drive.google.com/drive/folders/1GGvXBOCCvmbBOByWuCjvTiTpx9CGczh_?usp=sharing) where the pretrained models are stored

yuzhj:
Here is the [address](https://drive.google.com/drive/folders/1GGvXBOCCvmbBOByWuCjvTiTpx9CGczh_?usp=sharing) where the pretrained models are stored

configs/dla/faster_rcnn_dla_60_fpn_1x_coco.py
yuzhj: There existed unused parameters which were found difficultly, so I added this line to make the procedure run normally. Now I have solved this problem

mmdet/models/backbones/dla.py
yuzhj: The 'arch_settings' between DLANet and ResNet is different, so inheriting from ResNet has some problems.

mmdet/models/backbones/dla.py
yuzhj: I have reduced the codes of BasicBlock and Bottleneck of DLANet by inheriting from the BasicBlock and Bottleneck of ResNet

configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
```diff
@@ -0,0 +1,18 @@
+_base_ = '../cascade_rcnn/cascade_mask_rcnn_r50_fpn_20e_coco.py'
+
+model = dict(
+    pretrained='open-mmlab://dla60',
```
hhaAndroid, line 4: Is this from a third-party library?

ZwwWayne: we need to upload model checkpoints if necessary.

yuzhj: > Is this from a third-party library?
Yes, but I did some modification to match the mmdet-style codes. [Here](https://github.com/ucbdrive/dla) is the official implement

hhaAndroid:
Hi @yuzhj Are you interested in training directly on centernet? If you are not free, I will continue to do your job？

FDInSky:
Could I get the pretrained weight using 'open-mmlab://dla60' ?

configs/dla/cascade_mask_rcnn_dla_60_fpn_20e_coco.py
FDInSky: Could I get the pretrained weight using 'open-mmlab://dla60' now?

FDInSky:
open-mmlab://dla60 现在这个权重上传了吗？

lyxlynn:
Is there DLA-34 backbone？

PR #4573, opened 2021-02-03
[Feature] Create gif for demo

hhaAndroid:
Create gif for demo


demo/create_result_gif.py
```diff
@@ -0,0 +1,155 @@
+import argparse
+import os
+import os.path as osp
+
+import imageio
```
ZwwWayne, line 5: Is this a new dependency? can we use API in cv2?

hhaAndroid: I did not find the cv2 to support gif creation, But I found that PIL supports, and the gif generated by PIL does not work well even if the `quality` parameter is added, the effect is the same

`
images=[Image.fromarray(frame) for frame in frames]
images[0].save(gif_name,save_all=True,append_images=images, duration=duration,loop=0,quality=95)
`

image

ZwwWayne (approved):

PR #4582, opened 2021-02-04
add bn buffer sync in eval_hook

jshilong:
I modified the eval_hook.py to support broadcast the buffer of batch_norm in rank 0 to other ranks , so as to ensure that the test results in the log can be consistent with individual test results with the saved checkpoint.

### DIFF
--- a/mmdet/core/evaluation/eval_hooks.py
+++ b/mmdet/core/evaluation/eval_hooks.py
@@ -230,7 +230,11 @@ class DistEvalHook(EvalHook):
         self.gpu_collect = gpu_collect
 
     def after_train_epoch(self, runner):
-
+        # Synchronization of BatchNorm's buffer (running_mean
+        # and running_var) is not supported in the DDP of pytorch,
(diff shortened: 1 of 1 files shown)


mmdet/core/evaluation/eval_hooks.py
```diff
@@ -219,10 +225,20 @@ def __init__(self,
             save_best=save_best,
             rule=rule,
             **eval_kwargs)
+        self.broadcast_bn_buffer = broadcast_bn_buffer
         self.tmpdir = tmpdir
         self.gpu_collect = gpu_collect
 
     def after_train_epoch(self, runner):
+
```
ZwwWayne, line 233: This empty line is unnecessary

ZwwWayne (approved):

ZwwWayne:
NAS-FPN has been benchmarked in this PR and achieves mAP 37.7 (37.9 is reported). The result seems to be acceptable so the PR is merged.

PR #4597, opened 2021-02-07
Add paper conference link and MMCV install docs

hhaAndroid:
Add paper conference link and MMCV install docs


hhaAndroid:
The following papers are not sure whether the conference information is correct：

`
VFNet，
DetectoRS，
Weight Standardization，
Mixed Precision (FP16) Training
`

ZwwWayne (approved):

PR #4614, opened 2021-02-10
Dev/mask doctests

Erotemic:
I've been working on understanding the structure and API of the MaskHead in MaskRCNN. As a byproduct, I've written a few doctests and updated documentation on aspects that were initially unclear to me.

There are no functionality changes in this PR. This just adds a few new tests and updates documentation in a few places. I've run each of the new doctests and they do pass on my machine. (Lets see what the CI says)

Please double check that I've understood everything correctly and my documentation / example usages make sense.

### DIFF
--- a/mmdet/core/mask/mask_target.py
+++ b/mmdet/core/mask/mask_target.py
@@ -18,6 +18,26 @@ def mask_target(pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list,
 
     Returns:
         list[Tensor]: Mask target of each image.
+
+    Example:
+        >>> from mmdet.core.mask.mask_target import *  # NOQA
+        >>> import mmdet
+        >>> import mmcv
+        >>> from mmdet.core.mask import BitmapMasks
+        >>> H, W = 14, 14
+        >>> cfg = mmcv.Config({'mask_size': 14})
+        >>> pos_proposals_list = [torch.Tensor([
+        >>>     [176.2425,   5.5929, 190.9414, 204.9541],
+        >>>     [ 57.3241, 133.6170, 197.3850, 195.3102],
+        >>>     [ 74.8448, 186.4010, 176.0314, 203.7681],
+        >>>     [ 25.9790,  32.6989, 122.4416,  81.8580],
+        >>>     [  0.0000,   0.0000, 120.1398,  79.8232],
+        >>> ])]
+        >>> pos_assigned_gt_inds_list = [torch.Tensor([0, 1, 2, 1, 1]).long()]
+        >>> rng = np.random.RandomState(0)
+        >>> gt_masks_list = [BitmapMasks((rng.rand(3, H, W)), height=H, width=W)]
+        >>> mask_targets = mask_target(
+        >>>     pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list, cfg)
     """
     cfg_list = [cfg for _ in range(len(pos_proposals_list))]
     mask_targets = map(mask_target_single, pos_proposals_list,
@@ -40,6 +60,34 @@ def mask_target_single(pos_proposals, pos_assigned_gt_inds, gt_masks, cfg):
 
     Returns:
         Tensor: Mask target of each positive proposals in the image.
+
+    Example:
+        >>> from mmdet.core.mask.mask_target import *  # NOQA
+        >>> import mmdet
+        >>> import mmcv
+        >>> from mmdet.core.mask import BitmapMasks
+        >>> H, W = 32, 32
+        >>> cfg = mmcv.Config({'mask_size': 8})
+        >>> rng = np.random.RandomState(0)
+        >>> # Masks for each ground truth box (relative to the image)
+        >>> #gt_masks_data = (rng.rand(3, H, W) > 0.5).astype(np.float)
+        >>> gt_masks_data = rng.rand(3, H, W)
+        >>> gt_masks = BitmapMasks(gt_masks_data, height=H, width=W)
+        >>> # Predicted positive boxes
+        >>> pos_proposals = torch.FloatTensor([
+        >>>     [ 16.2,   5.5, 19.9, 20.9],
+        >>>     [ 17.3,  13.6, 19.3, 19.3],
+        >>>     [ 14.8,  16.4, 17.0, 23.7],
+        >>>     [  0.0,   0.0, 16.0, 16.0],
+        >>>     [  4.0,   0.0, 20.0, 16.0],
+        >>> ])
+        >>> # For each predicted proposal, its assignment to a gt mask
+        >>> pos_assigned_gt_inds = torch.LongTensor([0, 1, 2, 1, 1])
+        >>> # It seems that the masks must be in the space of the box
+        >>> # proposals.
+        >>> mask_targets = mask_target_single(
+        >>>     pos_proposals, pos_assigned_gt_inds, gt_masks, cfg)
+        >>> print(mask_targets)
     """
     device = pos_proposals.device
     mask_size = _pair(cfg.mask_size)
--- a/mmdet/core/mask/mask_target.py
+++ b/mmdet/core/mask/mask_target.py
@@ -20,24 +20,39 @@ def mask_target(pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list,
         list[Tensor]: Mask target of each image.
 
     Example:
-        >>> from mmdet.core.mask.mask_target import *  # NOQA
-        >>> import mmdet
         >>> import mmcv
+        >>> import mmdet
         >>> from mmdet.core.mask import BitmapMasks
-        >>> H, W = 14, 14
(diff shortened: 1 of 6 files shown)


mmdet/core/mask/mask_target.py
```diff
@@ -18,6 +18,26 @@ def mask_target(pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list,
 
     Returns:
         list[Tensor]: Mask target of each image.
+
+    Example:
+        >>> from mmdet.core.mask.mask_target import *  # NOQA
+        >>> import mmdet
+        >>> import mmcv
+        >>> from mmdet.core.mask import BitmapMasks
```
hellock, line 26: We may follow the order of isort to put `import ...` before `from ... import ...`.

hellock:
Thanks for this quite useful PR!

mmdet/core/mask/mask_target.py
```diff
@@ -18,6 +18,26 @@ def mask_target(pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list,
...
+        >>>     [ 57.3241, 133.6170, 197.3850, 195.3102],
+        >>>     [ 74.8448, 186.4010, 176.0314, 203.7681],
+        >>>     [ 25.9790,  32.6989, 122.4416,  81.8580],
+        >>>     [  0.0000,   0.0000, 120.1398,  79.8232],
+        >>> ])]
+        >>> pos_assigned_gt_inds_list = [torch.Tensor([0, 1, 2, 1, 1]).long()]
+        >>> rng = np.random.RandomState(0)
+        >>> gt_masks_list = [BitmapMasks((rng.rand(3, H, W)), height=H, width=W)]
+        >>> mask_targets = mask_target(
+        >>>     pos_proposals_list, pos_assigned_gt_inds_list, gt_masks_list, cfg)
```
hellock, line 40: We may validate the shape of mask_targets?

mmdet/core/mask/mask_target.py
```diff
@@ -40,6 +60,34 @@ def mask_target_single(pos_proposals, pos_assigned_gt_inds, gt_masks, cfg):
...
+    Example:
+        >>> from mmdet.core.mask.mask_target import *  # NOQA
+        >>> import mmdet
+        >>> import mmcv
+        >>> from mmdet.core.mask import BitmapMasks
+        >>> H, W = 32, 32
+        >>> cfg = mmcv.Config({'mask_size': 8})
+        >>> rng = np.random.RandomState(0)
+        >>> # Masks for each ground truth box (relative to the image)
+        >>> #gt_masks_data = (rng.rand(3, H, W) > 0.5).astype(np.float)
```
hellock, line 73: This line is redundant?

mmdet/core/mask/mask_target.py
```diff
@@ -40,6 +60,34 @@ def mask_target_single(pos_proposals, pos_assigned_gt_inds, gt_masks, cfg):
...
+        >>>     [  0.0,   0.0, 16.0, 16.0],
+        >>>     [  4.0,   0.0, 20.0, 16.0],
+        >>> ])
+        >>> # For each predicted proposal, its assignment to a gt mask
+        >>> pos_assigned_gt_inds = torch.LongTensor([0, 1, 2, 1, 1])
+        >>> # It seems that the masks must be in the space of the box
+        >>> # proposals.
+        >>> mask_targets = mask_target_single(
+        >>>     pos_proposals, pos_assigned_gt_inds, gt_masks, cfg)
+        >>> print(mask_targets)
```
hellock, line 90: Instead of printing, we may use assertions.

Erotemic: Done.

mmdet/core/mask/mask_target.py
Erotemic: Yes, I was debugging and I forgot to delete it.

mmdet/core/mask/mask_target.py
Erotemic: It was a good idea to do this, because the way I had the test written it was slightly ambiguous what the shape was supposed to be. It should now be much clearer.

mmdet/core/mask/structures.py
```diff
@@ -205,6 +206,24 @@ class BitmapMasks(BaseInstanceMasks):
...
+        >>> # demo crop_and_resize
+        >>> num_boxes = 5
+        >>> bboxes = np.array([[0, 0, 30, 10.0]] * num_boxes)
+        >>> out_shape = (14, 14)
+        >>> inds = torch.randint(0, len(self), size=(num_boxes,))
+        >>> device = 'cpu'
+        >>> interpolation = 'bilinear'
+        >>> new = self.crop_and_resize(
+        ...     bboxes, out_shape, inds, device, interpolation)
+        >>> assert len(new) == num_boxes
```
hellock, line 226: We may also check the mask shape:
```python
assert new.height, new.width == out_shape
```

mmdet/core/mask/structures.py
```diff
@@ -490,6 +509,28 @@ class PolygonMasks(BaseInstanceMasks):
...
+
+    Example:
+        >>> from mmdet.core.mask.structures import *  # NOQA
+        >>> masks = [
+        >>>     [ np.array([0, 0, 10, 0, 10, 10., 0, 10, 0, 0]) ]
+        >>> ]
+        >>> height, width = 16, 16
+        >>> self = PolygonMasks(masks, height, width)
+
+        >>> # demo translate
```
hellock, line 521: We may also add this to BitmapMasks.

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -100,10 +100,10 @@ def mask_cross_entropy(pred,
     """Calculate the CrossEntropy loss for masks.
 
     Args:
-        pred (torch.Tensor): The prediction with shape (N, C), C is the number
-            of classes.
+        pred (torch.Tensor): The prediction with shape (N, C, *), C is the
+            number of classes. The trailing * is spatio-temporal dimensions.
```
hellock, line 104: The trailing * indicates arbitrary shape

mmdet/models/roi_heads/mask_heads/fcn_mask_head.py
```diff
@@ -159,12 +177,38 @@ def get_seg_masks(self, mask_pred, det_bboxes, det_labels, rcnn_test_cfg,
                 it will be converted to numpy array outside of this method.
             det_bboxes (Tensor): shape (n, 4/5)
             det_labels (Tensor): shape (n, )
-            img_shape (Tensor): shape (3, )
             rcnn_test_cfg (dict): rcnn testing config
-            ori_shape: original image size
+            ori_shape (Tuple): original image height and width, shape (2,)
```
hellock, line 181: Two arguments `scale_factor` and `rescale` are missing.

mmdet/core/mask/structures.py
Erotemic: I don't want to add too many examples of method calls to the class-level docstring. One or two common use-cases is usually ok. When possible its better if each method has its own docstring and example 

To facilitate this I like to make a `@classmethod` called random, so creating a test instance to play with is a 1-liner. I added those random classmethods to BitmapMask and Polygon (fortunately I've written code for random polygons before, because its somewhat complex). I think I'm starting to go beyond the scope of this PR, so I don't want to add too much more. We can address any concerns with existing code, but for new code we should do a new PR.

mmdet/models/roi_heads/mask_heads/fcn_mask_head.py
Erotemic: I find these two arguments confusing. It seems out of the scope of this PR to add those, but I gave it a shot. Please directly edit the docs or make a suggested edit (I believe maintainers can directly edit the branch) if they aren't to your liking.

I wrote a very literal description of what they are doing because I don't understand the intuition for when they should be used.

mmdet/core/mask/structures.py
Erotemic: Done

mmdet/models/losses/cross_entropy_loss.py
Erotemic: Fixed.

mmdet/core/mask/structures.py
hellock: Agree.

mmdet/core/mask/structures.py
```diff
@@ -810,6 +897,121 @@ def to_tensor(self, dtype, device):
         ndarray_masks = self.to_ndarray()
         return torch.tensor(ndarray_masks, dtype=dtype, device=device)
 
+    @classmethod
+    def random(cls, num_masks=3, H=32, W=32, n_verts=5, dtype=np.float32,
```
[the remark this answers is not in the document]
Erotemic: good point.

hellock (approved):

PR #4615, opened 2021-02-10
Add mask_soft config option to allow non-binary masks

### LINKED ISSUE #4606: Soft Segmentation Masks
Currently in the segmentation heads it appears that masks are always given as binary targets.

For some problems, it may be more natural to instead express the ground truth as either a weighted mask, or more simply a mask of fractional "soft" values.

From what I can tell, this might be as simple as conditionally executing `resized_masks = (targets >= 0.5).cpu().numpy()` in the `crop_and_resize` function of `BitmapMasks` (in mmdet/core/mask/structures.py) based on the `cfg` passed to `mask_target` (in mmdet/core/mask/mask_target.py).

Erotemic:
Closes #4606

When using mmdetection as a library for MaskRCNN there is almost nothing preventing the ground truth masks from being fractional labels instead of hard binary labels. The only issue seemed to be in `BitmapMasks.crop_and_resize` where after a resize the interpolated values were always rounded to 0 or 1.

To prevent this I simply added a keyword argument `rounded=True`, which if set to `False` will prevent this rounding step. I also added a small bit of logic to `mask_target` such that it looks for the configuration option `mask_soft` (defaulted to `False`), and it sets `rounded = not mask_soft`.

In the case for mask Polygons I do expose the `rounded` parameter for API consistency, but I raise a ValueError if it is false because soft masks don't make much sense for Polygon targets.

Thus I was able to simply add `roi_head.train_cfg['mask_soft'] = True` to my MaskRCNN training configuration and the mask loss seems to be computed correctly against these soft targets.

Again, I'm using mmdetection as a library to define models (i.e. I'm not using my own training loop and data loading scripts), so I don't think this feature would be exposed to mmdet script users without further changes, but this should be the core fundamental change necessary to add soft mask targets to MaskHeads.

### DIFF
--- a/mmdet/core/mask/structures.py
+++ b/mmdet/core/mask/structures.py
@@ -323,7 +323,7 @@ class BitmapMasks(BaseInstanceMasks):
                         out_shape,
                         inds,
                         device='cpu',
-                        interpolation='bilinear'):
+                        interpolation='bilinear', rounded=True):
         """See :func:`BaseInstanceMasks.crop_and_resize`."""
         if len(self.masks) == 0:
             empty_masks = np.empty((0, *out_shape), dtype=np.uint8)
@@ -345,7 +345,10 @@ class BitmapMasks(BaseInstanceMasks):
                 0, inds).to(dtype=rois.dtype)
             targets = roi_align(gt_masks_th[:, None, :, :], rois, out_shape,
                                 1.0, 0, 'avg', True).squeeze(1)
-            resized_masks = (targets >= 0.5).cpu().numpy()
+            if rounded:
+                resized_masks = (targets >= 0.5).cpu().numpy()
+            else:
+                resized_masks = targets.cpu().numpy()
         else:
             resized_masks = []
         return BitmapMasks(resized_masks, *out_shape)
@@ -633,11 +636,17 @@ class PolygonMasks(BaseInstanceMasks):
                         out_shape,
                         inds,
                         device='cpu',
-                        interpolation='bilinear'):
+                        interpolation='bilinear',
+                        rounded=True):
         """see :func:`BaseInstanceMasks.crop_and_resize`"""
         out_h, out_w = out_shape
         if len(self.masks) == 0:
             return PolygonMasks([], out_h, out_w)
+
+        if not rounded:
+            raise ValueError(
+                'Polygon are always binary, '
+                'setting rounded=False is unsupported')
 
         resized_masks = []
         for i in range(len(bboxes)):
--- a/mmdet/core/mask/mask_target.py
+++ b/mmdet/core/mask/mask_target.py
@@ -43,6 +43,7 @@ def mask_target_single(pos_proposals, pos_assigned_gt_inds, gt_masks, cfg):
     """
     device = pos_proposals.device
     mask_size = _pair(cfg.mask_size)
+    mask_soft = cfg.get('mask_soft', False)
     num_pos = pos_proposals.size(0)
     if num_pos > 0:
         proposals_np = pos_proposals.cpu().numpy()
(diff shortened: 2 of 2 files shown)


mmdet/core/mask/structures.py
```diff
@@ -323,7 +323,7 @@ def crop_and_resize(self,
                         out_shape,
                         inds,
                         device='cpu',
-                        interpolation='bilinear'):
+                        interpolation='bilinear', rounded=True):
```
hellock, line 326: rounded -> binarize

hellock: The docstring of the base class needs to be updated.

hellock:
Thanks for this nice feature. Could you also fix the linting errors?

ZwwWayne:
Kindly ping @Erotemic

ZwwWayne:
The PR looks good to me now. We will benchmark it in May and will merge it for v2.13.0

ZwwWayne (approved):

RangiLyu:
I have benchmarked mask_rcnn_r50_fpn with `soft_mask_target=True` on COCO. 
The bbox eval results are
```
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.382
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=1000 ] = 0.590
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=1000 ] = 0.416
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.223
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.417
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.495
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.522
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=300 ] = 0.522
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=1000 ] = 0.522
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.326
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.566
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.660
```
And the segm eval results are

```
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.345
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=1000 ] = 0.556
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=1000 ] = 0.368
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.185
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.377
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.469
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.474
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=300 ] = 0.474
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=1000 ] = 0.474
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.276
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.516
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.623
```

Erotemic:
@RangiLyu Can you link to the comparable results where `soft_mask_target=False` so it is clear if there is any difference?

Erotemic:
Thanks @RangiLyu for pulling up these numbers. The difference is likely due to non-determinism due to cudnn, parallelism, data loading, etc. The only place where the objective is changing is on boundaries in the mask truth (there will be a few fractional values along the boundary due to interpolation).  I imagine if you ran both experiments multiple times, you would find the distribution of APs would be difficult to separate (i.e. p > 0.5). But then again, perhaps fuzzy edges do cause a real 0.2 drop in the score.

But it seems likely there are no significant regressions, so this seems ready for a merge.

ZwwWayne:
Thanks @Erotemic & @RangiLyu

PR #4619, opened 2021-02-11
use isinstance method to get loading pipeline

jingege315:
Original code:
```python
loading_pipeline_cfg = []
for cfg in pipeline:
    if cfg['type'].startswith('Load'):
        loading_pipeline_cfg.append(cfg)
```

It is not suitable for getting loading pipeline to use `startswith` method. It is possible to exist a pipeline whose string start with 'Load' but which isn't a loading pipeline. It is also possible to exist a pipeline inherited from `LoadImageFromFile` class or `LoadAnnotations` class whose string is not start with 'Load', such as `AnimalLoadImageFromFile`.

New code:
```python
loading_pipeline_cfg = []
for cfg in pipeline:
    obj_cls = PIPELINES.get(cfg['type'])
    if obj_cls is not None and isinstance(obj_cls, (LoadImageFromFile, LoadAnnotations)):
        loading_pipeline_cfg.append(cfg)
```

### DIFF
--- a/mmdet/datasets/utils.py
+++ b/mmdet/datasets/utils.py
@@ -101,9 +101,8 @@ def get_loading_pipeline(pipeline):
     loading_pipeline_cfg = []
     for cfg in pipeline:
         obj_cls = PIPELINES.get(cfg['type'])
-        if obj_cls is None:
-            continue
-        if isinstance(obj_cls, (LoadImageFromFile, LoadAnnotations)):
+        # TODO：use more elegant way to distinguish loading modules
+        if obj_cls is not None and obj_cls in (LoadImageFromFile, LoadAnnotations):
             loading_pipeline_cfg.append(cfg)
     assert len(loading_pipeline_cfg) == 2, \
         'The data pipeline in your config file must include ' \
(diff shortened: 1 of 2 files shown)


mmdet/datasets/utils.py
```diff
@@ -98,7 +100,10 @@ def get_loading_pipeline(pipeline):
     """
     loading_pipeline_cfg = []
     for cfg in pipeline:
-        if cfg['type'].startswith('Load'):
+        obj_cls = PIPELINES.get(cfg['type'])
+        if obj_cls is None:
+            continue
+        if isinstance(obj_cls, (LoadImageFromFile, LoadAnnotations)):
```
hhaAndroid, line 106: Cannot use instance, because it is not an instance relationship.  ->  `obj_cls in (LoadImageFromFile, LoadAnnotations)`, and above add 'TODO：use more elegant way to distinguish loading modules'

hhaAndroid: like this：

    loading_pipeline_cfg = []
    for cfg in pipeline:
        obj_cls = PIPELINES.get(cfg['type'])
        # TODO：use more elegant way to distinguish loading modules
        if obj_cls is not None and obj_cls in (LoadImageFromFile, LoadAnnotations):
            loading_pipeline_cfg.append(cfg)

ZwwWayne (approved):

ZwwWayne (approved):

PR #4655, opened 2021-02-22
[WIP] reimplement tinaface

wedlight:
get_subset_by_classes, xxyxy2xywh, _det2json, results2json are newly added functions.
The format_results function has been rewritten.
The evaluate function has been modified. (Big changes)


mmdet/datasets/custom.py
```diff
@@ -85,7 +85,11 @@ def __init__(self,
                                               self.proposal_file)
         # load annotations (and proposals)
         self.data_infos = self.load_annotations(self.ann_file)
-
+
```
hhaAndroid, line 88: You should not modify the custom.py code, you can think of a more reasonable way

PR #4691, opened 2021-02-28
Add parser argument 'cfg-options' for some tools

jingege315:
It is necessary to override some settings in the used config. But I find it is not containing parser argument 'cfg-options' in some tools. So I add it.

### DIFF
--- a/mmdet/core/export/pytorch2onnx.py
+++ b/mmdet/core/export/pytorch2onnx.py
@@ -6,7 +6,10 @@ import torch
 from mmcv.runner import load_checkpoint
 
 
-def generate_inputs_and_wrap_model(config_path, checkpoint_path, input_config):
+def generate_inputs_and_wrap_model(config_path,
+                                   checkpoint_path,
+                                   input_config,
+                                   cfg_options=None):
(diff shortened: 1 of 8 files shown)


ZwwWayne (approved):

hhaAndroid:
@ZwwWayne [pytorch2onnx](https://github.com/open-mmlab/mmdetection/blob/3cb1a70d45b9928c8e12e37b58f0ebfcf625b4b9/tools/deployment/pytorch2onnx.py#L153) does it also need to add the `cfg-options` parameter ?

jingege315:
> @ZwwWayne [pytorch2onnx](https://github.com/open-mmlab/mmdetection/blob/3cb1a70d45b9928c8e12e37b58f0ebfcf625b4b9/tools/deployment/pytorch2onnx.py#L153) does it also need to add the `cfg-options` parameter ?

We can modify the 'build_model_from_cfg' method to achieve it

jingege315:
Can you get this PR merged? @ZwwWayne

ZwwWayne (approved):

PR #4716, opened 2021-03-05
Fix RPN show_result error

hhaAndroid:
Fix RPN show_result error

### DIFF
--- a/mmdet/models/dense_heads/embedding_rpn_head.py
+++ b/mmdet/models/dense_heads/embedding_rpn_head.py
@@ -1,7 +1,5 @@
-import mmcv
 import torch
 import torch.nn as nn
-from mmcv import tensor2imgs
 
 from mmdet.models.builder import HEADS
 from ...core import bbox_cxcywh_to_xyxy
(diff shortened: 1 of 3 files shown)


mmdet/models/dense_heads/embedding_rpn_head.py
```diff
@@ -101,6 +101,7 @@ def simple_test_rpn(self, img, img_metas):
         """Forward function in testing stage."""
         return self._decode_init_proposals(img, img_metas)
 
+    # TODO: is right ?
```
jshilong, line 104: Just delete it, `embedding_rpn_head` is not a `Detector` which should not have `show_result`.

mmdet/models/detectors/rpn.py
```diff
@@ -138,17 +138,6 @@ def aug_test(self, imgs, img_metas, rescale=False):
...
-        img_tensor = data['img'][0]
-        img_metas = data['img_metas'][0].data[0]
-        imgs = tensor2imgs(img_tensor, **img_metas[0]['img_norm_cfg'])
-        assert len(imgs) == len(img_metas)
-        for img, img_meta in zip(imgs, img_metas):
-            h, w, _ = img_meta['img_shape']
-            img_show = img[:h, :w, :]
-            mmcv.imshow_bboxes(img_show, result, top_k=top_k)
+    def show_result(self, data, result, dataset=None, top_k=20, **kwargs):
+        """Show RPN proposals on the image."""
```
ZwwWayne, line 142: Why do not we need convert tensor images to numpy anymore? We should add comments to notify that.

ZwwWayne (approved):

PR #4717, opened 2021-03-05
Fix TridentNet batch inference bug

Chrisfsj2051:
In the previous version, it will raise a bug when the batch size in inference is greater than 1.

### DIFF
--- a/mmdet/models/roi_heads/trident_roi_head.py
+++ b/mmdet/models/roi_heads/trident_roi_head.py
@@ -23,6 +23,24 @@ class TridentRoIHead(StandardRoIHead):
         self.test_branch_idx = test_branch_idx
         super(TridentRoIHead, self).__init__(**kwargs)
 
+    def merge_trident_bboxes(self, trident_det_bboxes, trident_det_labels):
+        if trident_det_bboxes.numel() == 0:
+            det_bboxes = trident_det_bboxes.new_zeros((0, 5))
+            det_labels = trident_det_bboxes.new_zeros((0,), dtype=torch.long)
+        else:
+            nms_bboxes = trident_det_bboxes[:, :4]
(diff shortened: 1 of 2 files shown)


mmdet/models/roi_heads/trident_roi_head.py
```diff
@@ -39,36 +58,25 @@ def simple_test(self,
...
-            det_bboxes, keep = batched_nms(nms_bboxes, nms_scores, nms_inds,
-                                           nms_cfg)
-            det_labels = trident_det_labels[keep]
-            if self.test_cfg['max_per_img'] > 0:
-                det_labels = det_labels[:self.test_cfg['max_per_img']]
-                det_bboxes = det_bboxes[:self.test_cfg['max_per_img']]
-
-        det_bboxes, det_labels = [det_bboxes], [det_labels]
-
+        det_results = [
```
hhaAndroid, line 65: This looks complicated and it is better not to use the list comprehensions.

mmdet/models/roi_heads/trident_roi_head.py
```diff
@@ -39,36 +58,25 @@ def simple_test(self,
...
+            self.merge_trident_bboxes(
+                torch.cat(det_bboxes_list[i * num_branch:(i + 1) *
+                                          num_branch]),
+                torch.cat(det_labels_list[i * num_branch:(i + 1) *
+                                          num_branch]),
+            ) for i in range(len(img_metas) // num_branch)
+        ]
+        det_bboxes, det_labels = ([item[i] for item in det_results]
```
hhaAndroid, line 73: `det_bboxes, det_labels = [det_result for det_result in zip(*det_results)]`. I think it might be better.

hhaAndroid (approved):

ZwwWayne (approved):

PR #4764, opened 2021-03-14
Update anchor_head.py

howardchina:
remove redundent description. img_meta (dict) repeated twice in line 195 and line 198.


jshilong:
Thanks for your correction.

jshilong (approved):

PR #4806, opened 2021-03-22
Use MMCV's EvalHook in MMDetection

hhaAndroid:
refer: https://github.com/open-mmlab/mmcv/pull/739


ZwwWayne:
1. CI failed.
2. Need to keep the module in mmdet.core for at least a month and add warnings to avoid BC-breaking.

mmdet/core/evaluation/eval_hooks.py
```diff
@@ -54,6 +54,10 @@ def __init__(self,
                  save_best=None,
                  rule=None,
                  **eval_kwargs):
+        warnings.warn(
```
ZwwWayne, line 57: We should let this module inherits from the EvalHook supported in MMCV, and only override the `__init__` function by adding the warnings. The other functions can use those in MMCV.

xvjiarui (approved):

tests/test_data/test_datasets/test_common.py
```diff
@@ -275,8 +275,10 @@ def test_evaluation_hook(EvalHookParam):
     assert evalhook.evaluate.call_count == 3  # before epoch1 and after e1 & e2
 
     runner = _build_demo_runner()
-    with pytest.warns(UserWarning):
-        evalhook = EvalHookParam(dataloader, start=-2)
+    with pytest.raises(ValueError):
```
ZwwWayne, line 278: add comments to indicate what error you are testing.

ZwwWayne (approved):

PR #4810, opened 2021-03-22
[Feature]: Add mmdet2torchserve tool

### LINKED ISSUE #4661: Add ability to serve models
**Describe the feature**

There are libaries such as Tensorflow Serve and Pytorch Serve. It would be useful is MMDetection had the ability to serve models, maybe inside a container, through an API endpoint.

**Motivation**
It would allow us to easily use models trained using MMDetection in other applications for inference.

**Related resources**
If there is an official code release or third-party implementations, please also provide the information here, which would be very helpful.

**Additional context**
At the moment I have a script which calls an MMDetection script that I use as the entrypoint for a Docker container; it starts up whenever I need it, it's pretty slow. It would be nice if there was an easier way.

daavoo:
This P.R. enables serving MMDetection models with [TorchServe](https://pytorch.org/serve/). A new section `Model Serving` has been added in `useful_tools.md`.

Closes #4661


tools/deployment/serve/Dockerfile
```diff
@@ -0,0 +1,73 @@
+# syntax = docker/dockerfile:experimental
```
ZwwWayne, line 1: Move this dockerfile to the directory [here](https://github.com/open-mmlab/mmdetection/tree/master/docker) may help ease the management of dockers.

ZwwWayne:
Hi @daavoo ,
Overall, this PR LGTM and torchserve will definitely a good feature for mmdet. Let's see whether @hellock has any suggestions.

tools/deployment/serve/Dockerfile
daavoo: Okey! I though about placing it there but didn't know if it was the right place.

ZwwWayne (approved):

hhaAndroid:
Hi, @daavoo Does the writing in Dockerfile require a specific version of docker?  My Docker version is 18.01.0-ce, the following error occurred: 
`Error response from daemon: Dockerfile parse error line 43: Unknown flag: mount`

daavoo:
> Hi, @daavoo Does the writing in Dockerfile require a specific version of docker? My Docker version is 18.01.0-ce, the following error occurred:
> `Error response from daemon: Dockerfile parse error line 43: Unknown flag: mount`

Hi @hhaAndroid , it actually requires 18.09 or higher:
https://docs.docker.com/engine/reference/builder/#buildkit

We can maybe change the syntax to support lower versions?

hhaAndroid:
OK, It's best to be compatible. Thank you

hhaAndroid:
I found some small errors, you can also fix them together：

1. add `mmcv.mkdir_or_exist(output_folder)`
https://github.com/open-mmlab/mmdetection/blob/master/tools/deployment/mmdet2torchserve.py#L44

2. --output_folder -> --output-folder
https://github.com/open-mmlab/mmdetection/blob/master/docs/useful_tools.md#1-convert-model-from-mmdetection-to-torchserve

PR #4813, opened 2021-03-22
Deprecate LoadImage

daavoo:
It's not being used anywhere.

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -53,6 +53,39 @@ def init_detector(config, checkpoint=None, device='cuda:0', cfg_options=None):
     return model
 
 
+class LoadImage(object):
+    """Deprecated.
+
+    A simple pipeline to load image.
+    """
+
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Hi @daavoo ,
Thanks for your kind contribution. It seems that `LoadImage` is indeed not used in this repo. But in case downstream repos use that, shall we add a warning first before totally removing it?  Can we dd a warning in the `__init__` function, and then remove it in the future release (usually after 1-2 months)?

daavoo:
> Hi @daavoo ,
> Thanks for your kind contribution. It seems that `LoadImage` is indeed not used in this repo. But in case downstream repos use that, shall we add a warning first before totally removing it? Can we dd a warning in the `__init__` function, and then remove it in the future release (usually after 1-2 months)?

It seems very reasonable, I replaced the removal with deprecation warnings

ZwwWayne (approved):

PR #4898, opened 2021-04-05
[Fix]: fix mask AP of small/medium/large

ZwwWayne:
This PR delete the key `bbox` in the prediction results when calculating the mask AP of predictions.
When calculating `segm AP`, if key `bbox` is in the prediction, the area of instance will be calculated by the box area rather than the mask area, which will not affect the overall mask AP but will lead to a different mask APs of small/medium/large instances. A comparison is as below

|model|modification|mask AP|AP50|AP75|APs|APm|APl|
|:------:|:------:|:------:|:------:|:------:|:------:|:------:|:------:|
|[Mask R-CNN ms3x](https://github.com/open-mmlab/mmdetection/blob/master/configs/mask_rcnn/mask_rcnn_r50_caffe_fpn_mstrain-poly_3x_coco.py) ([ckpt](http://download.openmmlab.com/mmdetection/v2.0/mask_rcnn/mask_rcnn_r50_caffe_fpn_mstrain-poly_3x_coco/mask_rcnn_r50_caffe_fpn_mstrain-poly_3x_coco_bbox_mAP-0.408__segm_mAP-0.37_20200504_163245-42aa3d00.pth))|before|0.370 |0.584| 0.393| 0.205| 0.402 |0.497|
||after|0.370 |0.584| 0.393|0.182|0.395|0.520|
|[Cascade Mask R-CNN 1x](https://github.com/open-mmlab/mmdetection/blob/master/configs/cascade_rcnn/cascade_mask_rcnn_r50_caffe_fpn_1x_coco.py) ([ckpt](http://download.openmmlab.com/mmdetection/v2.0/cascade_rcnn/cascade_mask_rcnn_r50_caffe_fpn_1x_coco/cascade_mask_rcnn_r50_caffe_fpn_1x_coco_bbox_mAP-0.412__segm_mAP-0.36_20200504_174659-5004b251.pth))|before|0.360 |0.564 |0.387 |0.185 |0.387 |0.494|
||after|0.360 |0.564 |0.387|0.163|0.382|0.529|

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -2,10 +2,10 @@
 
 ## Mask AP Evaluation since MMDetection 2.12.0
 
-Before [PR 4898](https://github.com/open-mmlab/mmdetection/pull/4898) and V2.13.0, the mask AP of small, medium, and large instances is calculated based on the bounding box area rather than the real mask area. This leads to higher APs and APm but lower APl but will not affect the overall mask AP. 
+Before [PR 4898](https://github.com/open-mmlab/mmdetection/pull/4898) and V2.12.0, the mask AP of small, medium, and large instances is calculated based on the bounding box area rather than the real mask area. This leads to higher APs and APm but lower APl but will not affect the overall mask AP.
 [PR 4898](https://github.com/open-mmlab/mmdetection/pull/4898) change it to use mask areas by deleting `bbox` in mask AP calculation.
 The new calculation is consistent with [Detectron2](https://github.com/facebookresearch/detectron2/).
- 
+
 ## Compatibility with MMDetection 1.x
 
 MMDetection 2.0 goes through a big refactoring and addresses many legacy issues. It is not compatible with the 1.x version, i.e., running inference with the same model weights in these two versions will produce different results. Thus, MMDetection 2.0 re-benchmarks all the models and provides their links and logs in the model zoo.
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1,10 +1,18 @@
-# Compatibility with MMDetection 1.x
+# Compatibility of MMDetection 2.x
+
+## Mask AP Evaluation
+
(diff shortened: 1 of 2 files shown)


mmdet/datasets/coco.py
```diff
@@ -435,18 +435,28 @@ def evaluate(self,
...
+                predictions = mmcv.load(result_files[metric])
+                if iou_type == 'segm':
+                    # Refer to https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/coco.py#L331  # noqa
+                    # When evaluating mask AP, if the results contain bbox,
+                    # cocoapi will use the box area instead of the mask area
+                    # for calculating the instance area. Though the overall AP
+                    # is not affected, this leads to different small/medium/large
+                    # mask AP results.
+                    for x in predictions:
+                        x.pop('bbox', None)
```
xvjiarui, line 451:
```suggestion
                        x.pop('bbox')
```

docs/compatibility.md
```diff
@@ -1,10 +1,18 @@
-# Compatibility with MMDetection 1.x
+# Compatibility of MMDetection 2.x
+
+## Mask AP Evaluation
+
+Before [PR 4898](https://github.com/open-mmlab/mmdetection/pull/4898), the mask AP of small, medium, and large instances is calculated based on the bounding box area rather than the real mask area. This leads to higher APs and APm but lower APl but will not affect the overall mask AP.
```
xvjiarui, line 5: Should we also add mmdet version number?

Johnson-Wang:
This is indeed a breaking change. I think we'd better design a better way to inform the users about the change.

ZwwWayne:
> This is indeed a breaking change. I think we'd better design a better way to inform the users about the change.

Any suggestion? For now, I plan to inform to do that through the compatibility documentation and the changelog.

docs/compatibility.md
ZwwWayne: Sure, and this PR will be merged right before the new release.

Johnson-Wang:
How about adding a reminder in the logger?

ZwwWayne:
> How about adding a reminder in the logger?

Agree, will do that.

jshilong (approved):
LGTM

PR #4905, opened 2021-04-06
Fix dynamic clip error in box coder for TensorRT

grimoire:
Hi
This PR use `Tensor.clamp(0, 1)` instead of `torch.where` to fix the conversion error on two-stage detection models such as Faster-RCNN.

### DIFF
--- a/mmdet/core/export/onnx_helper.py
+++ b/mmdet/core/export/onnx_helper.py
@@ -21,19 +21,19 @@ def dynamic_clip_for_onnx(x1, y1, x2, y2, max_shape):
     assert isinstance(
         max_shape,
         torch.Tensor), '`max_shape` should be tensor of (h,w) for onnx'
-    h = max_shape[0].to(x1)
-    w = max_shape[1].to(x1)
-    zero = x1.new_tensor(0)
-    # clip by 0
-    x1 = torch.where(x1 < zero, zero, x1)
-    y1 = torch.where(y1 < zero, zero, y1)
-    x2 = torch.where(x2 < zero, zero, x2)
-    y2 = torch.where(y2 < zero, zero, y2)
-    # clip by h and w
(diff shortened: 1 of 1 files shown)


mmdet/core/export/onnx_helper.py
```diff
@@ -21,19 +21,19 @@ def dynamic_clip_for_onnx(x1, y1, x2, y2, max_shape):
...
-    x1 = torch.where(x1 < zero, zero, x1)
-    y1 = torch.where(y1 < zero, zero, y1)
-    x2 = torch.where(x2 < zero, zero, x2)
-    y2 = torch.where(y2 < zero, zero, y2)
-    # clip by h and w
-    x1 = torch.where(x1 > w, w, x1)
-    y1 = torch.where(y1 > h, h, y1)
-    x2 = torch.where(x2 > w, w, x2)
-    y2 = torch.where(y2 > h, h, y2)
+
```
RunningLeon, line 24: could update docstring in line 10
```
...... we have to use torch.where to workaround.
```

RunningLeon (approved):
LGTM

ZwwWayne (approved):

mmdet/core/export/onnx_helper.py
```diff
@@ -21,19 +21,24 @@ def dynamic_clip_for_onnx(x1, y1, x2, y2, max_shape):
...
+    # scale by 1/max_shape
```
ZwwWayne, line 25: Why we must use mask shape to to the scaling here? If it is necessary, we need comments here.

RunningLeon: The reason is mentioned in the docstring.
```
Since torch.clamp cannot have dynamic `min` and `max`, we scale the
      boxes by 1/max_shape and clamp in the range [0, 1].
```

Issue #4718, opened 2021-03-05
ResNeSt backbone cascade rcnn can not use async_inference_detector, async_simple_test() takes 3 positional arguments but 4 were given


JustinhoCHN:
**Describe the bug**
I was trying to use my ResNeSt backbone cascade rcnn model with ```async_inference_detector```, but it raised an error:
```
Traceback (most recent call last):
  File "tools/batch_inference_async.py", line 77, in <module>
    asyncio.run(main())
  File "/home/huzhihao/anaconda3/envs/open-mmlab/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/home/huzhihao/anaconda3/envs/open-mmlab/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "tools/batch_inference_async.py", line 61, in main
    await async_inference_detector(model, test_img)
  File "/home/huzhihao/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmdet-2.9.0-py3.7.egg/mmdet/apis/inference.py", line 179, in async_inference_detector
    result = await model.aforward_test(rescale=True, **data)
  File "/home/huzhihao/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmdet-2.9.0-py3.7.egg/mmdet/models/detectors/base.py", line 121, in aforward_test
    return await self.async_simple_test(img[0], img_metas[0], **kwargs)
  File "/home/huzhihao/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmdet-2.9.0-py3.7.egg/mmdet/models/detectors/two_stage.py", line 185, in async_simple_test
    x, proposal_list, img_meta, rescale=rescale)
TypeError: async_simple_test() takes 3 positional arguments but 4 were given
```

**BUT !** If I replace the model cfg and weights that ```mmdetection/tests/test_runtime/async_benchmark.py``` supplies, there's no error.

**Reproduction**
codes and models are here: [baiduwangpan](https://pan.baidu.com/s/1rx9i2uik-rMfxOc_weph_A) code: nh11

running command:
```
python tools/batch_inference_async.py workdir/big_model/cascade_rcnn_s101_fpn_syncbn-backbone+head_mstrain-range_1x_coco_fp16.py workdir/big_model/epoch_11.pth --imgs data/random_testset/ --batch_size 8 --device 1
```

mmcv-full: 1.2.7
[...]

PR #4941, opened 2021-04-12
Fix async error and privide aysnc demo

hhaAndroid:
resolved: #4718

### DIFF
--- a/demo/image_demo.py
+++ b/demo/image_demo.py
@@ -15,7 +15,7 @@ def parse_args():
     parser.add_argument(
         '--score-thr', type=float, default=0.3, help='bbox score threshold')
     parser.add_argument(
-        '--async-infer',
+        '--async-test',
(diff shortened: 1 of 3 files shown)


demo/image_demo.py
```diff
@@ -12,8 +14,15 @@ def main():
         '--device', default='cuda:0', help='Device used for inference')
     parser.add_argument(
         '--score-thr', type=float, default=0.3, help='bbox score threshold')
+    parser.add_argument(
+        '--async-infer',
```
ZwwWayne, line 18: --async shoud be OK

hhaAndroid: if use --async, `args.async` will show python keyword conflicts in IDE. Although it will not an error, it is not very good.

ZwwWayne (approved):

PR #4954, opened 2021-04-13
Serve updates

daavoo:
- Remove usage of `DOCKER_BUILDKIT` for compatibility with older versions of docker.
- Minor fixes (https://github.com/open-mmlab/mmdetection/pull/4810#issuecomment-813928693)


hhaAndroid:
Thank you for your contribution！

docs/useful_tools.md
```diff
@@ -184,14 +184,14 @@ In order to serve an `MMDetection` model with [`TorchServe`](https://pytorch.org
...
 ```
 
 ### 2. Build `mmdet-serve` docker image
 
 ```shell
-DOCKER_BUILDKIT=1 docker build -t mmdet-serve:latest docker/serve/
+docker build -t mmdet-serve:latest docker/serve/
 ```
 
 ### 3. Launch `mmdet-serve`
```
hhaAndroid, line 197: (1) I found that there is no `--gpus` parameter in 18.01.0-ce version docker. It will report an error.

(2) Please specify the MODEL_STORE parameter is a folder the same as `--output-folder`, not a file path.

ZwwWayne (approved):

docs/useful_tools.md
daavoo: > (1) I found that there is no `--gpus` parameter in 18.01.0-ce version docker. It will report an error.
@hhaAndroid For running the image with gpu support you would need `nvidia-docker` which requires `docker > 19.03`:

https://github.com/NVIDIA/nvidia-docker/wiki#what-is-the-minimum-supported-docker-version

Maybe we could update the docs for this requirement. Could you check if you can run the image without the `--gpus` arg (theoretically it should run in CPU mode)

hhaAndroid: > > (1) I found that there is no `--gpus` parameter in 18.01.0-ce version docker. It will report an error.
> 
> @hhaAndroid For running the image with gpu support you would need `nvidia-docker` which requires `docker > 19.03`:
> 
> https://github.com/NVIDIA/nvidia-docker/wiki#what-is-the-minimum-supported-docker-version
> 
> Maybe we could update the docs for this requirement. Could you check if you can run the image without the `--gpus` arg (theoretically it should run in CPU mode)

O, I did not install `nvidia-docker`.  Yes, I can run the image in CPU mode without the` --gpus`. You can consider updating the document. Thank you

hhaAndroid: Can you update the document?

daavoo: Done!

PR #4969, opened 2021-04-15
[Feature]: add semantic FPN for panoptic segmentation

ZwwWayne:
This PR continues on PR https://github.com/open-mmlab/mmdetection/pull/953.

It refactors the semantic head and adds semantic FPN for panoptic segmentation. The semantic FPN is used as a decoder in the semantic segmentation head. The `FusedSemanticHead` for HTC is refactored and inherits from the new `SemanticHead`. Thus, it can also be used for panoptic segmentation while providing semantic features in HTC and SCNet.

The BC-breaking of names of convolutional layers in the semantic head is handled by `_load_from_state_dict`. Therefore, the old models can still be successfully loaded for inference and the new models will have different key names in the semantic heads.


mitming:
Hi, @ZwwWayne 

How is this PR going?

Issue #4971, opened 2021-04-15
mmdet cannot train with single class dataset


fcakyon:
Since PR https://github.com/open-mmlab/mmdetection/pull/4508, single class training has been broken.

I create a new dataset class as:

```python
from .builder import DATASETS
from .coco import CocoDataset


@DATASETS.register_module()
class NewCocoDataset(CocoDataset):

    CLASSES = ('uav')
```

Then when i start training with ssd512 config, i get this error:

```bash
The `num_classes` (1) in SSDHead of MMDataParallel does not matches the length of `CLASSES` 3) in NewCocoDataset
  File "mmdetection\mmdet\datasets\utils.py", line 136, in _check_head
    assert module.num_classes == len(dataset.CLASSES), \
```

shinya7y:
Does `CLASSES = ('uav', )` work?
https://github.com/open-mmlab/mmdetection/blob/a616886bf1e8de325e6906b8c76b6a4924ef5520/mmdet/datasets/wider_face.py#L17

fcakyon:
@shinya7y yes that works, maybe we can add a warning that tells to add a semicolon at the end, for the single class case.

ZwwWayne:
> @shinya7y yes that works, maybe we can add a warning that tells to add a semicolon at the end, for the single class case.

I agree with this warning. We may do that in PR #4973.

fcakyon:
@ZwwWayne i updated #4973

PR #4973, opened 2021-04-15
fix training with single class dataset

fcakyon:
fixes https://github.com/open-mmlab/mmdetection/issues/4971

### DIFF
--- a/mmdet/datasets/utils.py
+++ b/mmdet/datasets/utils.py
@@ -133,11 +133,12 @@ class NumClassCheckHook(Hook):
             for name, module in model.named_modules():
                 if hasattr(module, 'num_classes') and not isinstance(
                         module, (RPNHead, VGG, FusedSemanticHead, GARPNHead)):
-                    if type(dataset.CLASSES) is str:
-                        len_dataset_classes = 1
-                    else:
-                        len_dataset_classes = len(dataset.CLASSES)
-                    assert module.num_classes == len_dataset_classes, \
+                    assert type(dataset.CLASSES) is not str, \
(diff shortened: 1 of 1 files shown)


mmdet/datasets/utils.py
```diff
@@ -133,7 +133,11 @@ def _check_head(self, runner):
             for name, module in model.named_modules():
                 if hasattr(module, 'num_classes') and not isinstance(
                         module, (RPNHead, VGG, FusedSemanticHead, GARPNHead)):
-                    assert module.num_classes == len(dataset.CLASSES), \
+                    if type(dataset.CLASSES) is str:
```
ZwwWayne, line 136: dataset.CLASSES must be a tuple of str and should not be str. Otherwise, it will cause errors in other places.

fcakyon: @ZwwWayne i updated the pr with a warning when dataset.CLASSES is a str instead of tuple of str

mmdet/datasets/utils.py
```diff
@@ -133,6 +133,11 @@ def _check_head(self, runner):
             for name, module in model.named_modules():
                 if hasattr(module, 'num_classes') and not isinstance(
                         module, (RPNHead, VGG, FusedSemanticHead, GARPNHead)):
+                    assert type(dataset.CLASSES) is not str, \
```
ZwwWayne, line 136: This part of code is valuable and necessary. Can we move it to another place for a more clear logic and only check it once? For example, move it to line 133 before delving into each modules.

fcakyon: moved outside for loop

ZwwWayne (approved):

PR #4978, opened 2021-04-15
Fix GFL no gt

shinya7y:
https://github.com/open-mmlab/mmdetection/pull/3702 fixed two bugs in the ATSSHead.
This PR and https://github.com/open-mmlab/mmdetection/pull/4631 fix the same bugs in the GFLHead.

### DIFF
--- a/mmdet/models/dense_heads/gfl_head.py
+++ b/mmdet/models/dense_heads/gfl_head.py
@@ -11,6 +11,8 @@ from mmdet.core import (anchor_inside_flags, bbox2distance, bbox_overlaps,
 from ..builder import HEADS, build_loss
 from .anchor_head import AnchorHead
 
+EPS = 1e-12
+
 
 class Integral(nn.Module):
     """A fixed layer for calculating integral result from distribution.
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/gfl_head.py
```diff
@@ -363,6 +365,8 @@ def loss(self,
 
         avg_factor = sum(avg_factor)
         avg_factor = reduce_mean(avg_factor).item()
+        if avg_factor < EPS:
```
ZwwWayne, line 368: Shall we use avg_factor.clamp_(min=1) here?

shinya7y: Should I change ATSSHead in the same way?

ZwwWayne: You can do that if it is OK with you. Thanks~

ZwwWayne: `avg_factor = reduce_mean(avg_factor).clamp_(min=1).item()`  should work.

ZwwWayne (approved):

PR #5015, opened 2021-04-19
Use dataset in `dynamic_rcnn` config

adursun:


ZwwWayne:
Hi @adursun ，
Thanks for your kind contribution. The config path in the configs/dynamic_rcnn/README.md should also be updated with the new file name.

ZwwWayne (approved):

Issue #5009, opened 2021-04-19
Dead lock still exists in vfnet_head.py


zlin3000:
Thanks for your error report and we appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**
 Whiling training vfnet with multiple-GPU and  filter_empty_gt=False, it still possible to meet deadlock.

**Reproduction**

1. What command or script did you run?

bash dist_train.sh vfnet_config.py 8

2. Did you make any modifications on the code or config? Did you understand what you have modified?

Yes

3. What dataset did you use?

voc+custom

**Environment**

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.
2. You may add addition that may be helpful for locating the problem, such as
    - How you installed PyTorch [e.g., pip, conda, source]
    - Other environment variables that may be related (such as `$PATH`, `$LD_LIBRARY_PATH`, `$PYTHONPATH`, etc.)

**Error traceback**
If applicable, paste the error trackback here.

```none
A placeholder for trackback.
```

**Bug fix**
https://github.com/open-mmlab/mmdetection/blob/a2d9047e9a3d70430011895e62764372b3780bc5/mmdet/models/dense_heads/vfnet_head.py#L428

guess should modify this line.

hhaAndroid:
Hi @zlin3000 Thank you for your advice. I will fix it.

PR #5058, opened 2021-04-25
Fix dead lock of vfnet

hhaAndroid:
resolve: #5009

Precision alignment after modification.

### DIFF
--- a/mmdet/models/dense_heads/vfnet_head.py
+++ b/mmdet/models/dense_heads/vfnet_head.py
@@ -411,6 +411,16 @@ class VFNetHead(ATSSHead, FCOSHead):
             bbox_weights_ini.sum()).item()
         bbox_avg_factor_ini = max(iou_targets_ini_avg_per_gpu, 1.0)
 
+        pos_decoded_bbox_preds_refine = \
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/vfnet_head.py
```diff
@@ -411,23 +411,23 @@ def loss(self,
...
         bbox_avg_factor_ini = max(iou_targets_ini_avg_per_gpu, 1.0)
 
+        pos_decoded_bbox_preds_refine = \
+            distance2bbox(pos_points, pos_bbox_preds_refine)
+        iou_targets_rf = bbox_overlaps(
+            pos_decoded_bbox_preds_refine,
+            pos_decoded_target_preds.detach(),
+            is_aligned=True).clamp(min=1e-6)
+        bbox_weights_rf = iou_targets_rf.clone().detach()
+        iou_targets_rf_avg_per_gpu = reduce_mean(bbox_weights_rf.sum()).item()
```
ZwwWayne, line 421: Use `reduce_mean(bbox_weights_rf.sum()).clamp_(min=1).item()` and skip `max()` operation.

ZwwWayne (approved):

PR #5062, opened 2021-04-26
Fix empty bbox error in mask_head of carafe

hhaAndroid:


mmdet/models/roi_heads/test_mixins.py
```diff
@@ -254,6 +254,11 @@ def simple_test_mask(self,
         ori_shapes = tuple(meta['ori_shape'] for meta in img_metas)
         scale_factors = tuple(meta['scale_factor'] for meta in img_metas)
 
+        if all(det_bbox.shape[0] == 0 for det_bbox in det_bboxes):
+            segm_results = [[[] for _ in range(self.mask_head.num_classes)]
```
RunningLeon, line 258: add assertation here at the first line inside if branch:

```
        if torch.onnx.is_in_onnx_export():
            raise RuntimeError('[ONNX Error] Can not record MaskHead '
                               'as it has not been executed this time')
```

RunningLeon (approved):

mmdet/models/roi_heads/test_mixins.py
```diff
@@ -254,6 +254,14 @@ def simple_test_mask(self,
         ori_shapes = tuple(meta['ori_shape'] for meta in img_metas)
         scale_factors = tuple(meta['scale_factor'] for meta in img_metas)
 
+        if all(det_bbox.shape[0] == 0 for det_bbox in det_bboxes):
+            if torch.onnx.is_in_onnx_export():
+                raise RuntimeError('[ONNX Error] Can not record MaskHead '
+                                   'as it has not been executed this time')
+            segm_results = [[[] for _ in range(self.mask_head.num_classes)]
```
ZwwWayne, line 261: line 316 also handles this case, need to simplify the logic.

ZwwWayne (approved):

Issue #5032, opened 2021-04-20
MaskRCNN - IndexError: list index out of range after training 1 epoch


manhph2211:
Hello, I'm having trouble  with this kinda weird error:
```
  File "./mmdetection/mmdet/datasets/coco.py", line 259, in _segm2json
    if isinstance(segms[i]['counts'], bytes):
IndexError: list index out of range
```
Showing up right after I trained 1 epoch :cry: 

Could anyone help me to handle it :love_letter:  Thanks in advance!

hhaAndroid:
I suspect that your data may be incorrectly labeled, you need to check

manhph2211:
Sir, thanks for your reply. I did check the data(only 1 class) but everything seems good :confused:

hhaAndroid:
You can also use [visual scripts](https://github.com/open-mmlab/mmdetection/blob/master/tools/misc/browse_dataset.py) to find problems easily

leemengwei:
+1 here when resume-from model,  at epoch 23

hhaAndroid:
> +1 here when resume-from model, at epoch 23 using dist train

I think this error has nothing to do with the train, it may just be a dataset problem.

manhph2211:
@hhaAndroid I followed the solution in `fix supple_mask` above but the error is still there :sob: :sob:

hhaAndroid:
> @hhaAndroid I followed the solution in `fix supple_mask` above but the error is still there

This problem is more difficult to eliminate, can you debug to see the cause? Thank you.

YiXinChenChen:
+1

hhaAndroid:
> +1

Are you using the coco data set? Can you provide more details？

YiXinChenChen:
> > +1
> 
> Are you using the coco data set? Can you provide more details？

I use the custom dataset but covert it to coco format,  the error appear when evaluation：

2021-04-28 14:33:52,806 - mmdet - INFO - Saving checkpoint at 1 epochs
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 50/50, 6.9 task/s, elapsed: 7s, ETA:     0sTraceback (most recent call last):
  File "tools/train.py", line 190, in <module>
    main()
  File "tools/train.py", line 186, in main
    meta=meta)
  File "/home/chenyixin03/mmdetection/mmdet/apis/train.py", line 171, in train_detector
[...]

PR #5065, opened 2021-04-26
fix supplement_mask

myownskyW7:
Fix the bug when calculating supplement_mask.
Although the risk is low, the last number of a box has chances to be zero.
It may lead to inconsistent numbers of bbox_results and segm_results. And the evaluation process fails at
`File "/mnt/mmdet-dev/mmdetection/mmdet/datasets/coco.py", line 259, in _segm2json`
`if isinstance(segms[i]['counts'], bytes): `
`IndexError: list index out of range`.

### DIFF
--- a/mmdet/models/roi_heads/test_mixins.py
+++ b/mmdet/models/roi_heads/test_mixins.py
@@ -114,8 +114,8 @@ class BBoxTestMixin(object):
         cls_score = cls_score.reshape(batch_size, num_proposals_per_img, -1)
 
         if not torch.onnx.is_in_onnx_export():
-            # remove padding
-            supplement_mask = rois[..., -1] == 0
+            # remove padding, ignore batch_index when calculating mask
+            supplement_mask = rois.abs()[..., 1:].sum(dim=-1) == 0
             cls_score[supplement_mask, :] = 0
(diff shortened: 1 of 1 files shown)


mmdet/models/roi_heads/test_mixins.py
```diff
@@ -114,8 +114,8 @@ def simple_test_bboxes(self,
         cls_score = cls_score.reshape(batch_size, num_proposals_per_img, -1)
 
         if not torch.onnx.is_in_onnx_export():
-            # remove padding
-            supplement_mask = rois[..., -1] == 0
+            # remove padding, ignore batch_index when calculating mask
+            supplement_mask = rois.abs()[..., 1:].sum(dim=-1) == 0
```
hhaAndroid, line 118: Thanks for the contribution. why need `1:`

hhaAndroid:
reslove: #5032

mmdet/models/roi_heads/test_mixins.py
ZwwWayne: the zero dimension is batch idx

ZwwWayne (approved):

PR #4295, opened 2020-12-13
Reimplement AutoAssign

jshilong:
I reimplemented autoassign.
I got a similar accuracy of 40.4 vs 40.5 in origin implementation, and the performance is unstable with  1x setting and may fluctuate by about 0.3 mAP. Results between 40.3 and 40.6 are acceptable.

PR #5092, opened 2021-04-30
fix bug of norm pre

jshilong:
This pr fix a bug of Auto-assign

## Motivation
In the implementation, because in the training phase, the `bbox_pred` and the `target` are normalized by strides, the iou calculation would wrong.
https://github.com/open-mmlab/mmdetection/blob/2f2cb37cdaeac0862710045a066ee1669e286a42/mmdet/models/dense_heads/autoassign_head.py#L361

## BC-breaking (Optional)
None

### DIFF
--- a/configs/autoassign/autoassign_r50_fpn_8x2_1x_coco.py
+++ b/configs/autoassign/autoassign_r50_fpn_8x2_1x_coco.py
@@ -28,7 +28,6 @@ model = dict(
         relu_before_extra_convs=True),
     bbox_head=dict(
         type='AutoAssignHead',
-        norm_on_bbox=True,
         num_classes=80,
         in_channels=256,
         stacked_convs=4,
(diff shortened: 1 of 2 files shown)


mmdet/models/necks/fpn.py
```diff
@@ -93,7 +90,6 @@ def __init__(self,
         self.no_norm_on_lateral = no_norm_on_lateral
         self.fp16_enabled = False
         self.upsample_cfg = upsample_cfg.copy()
-        self.caffe2_xavier_init = caffe2_xavier_init
```
ZwwWayne, line 96: We should keep this option and report a deprecated warning. Otherwise downstream repos like mmdet-projects will fail?

jshilong: I add this argument in `Reimplement AutoAssign #4295`, I am sure this arg only used in autoassign.

ZwwWayne (approved):

mmdet/models/necks/fpn.py
ZwwWayne: OK great

PR #5122, opened 2021-05-07
[Feature] Add MobileNetV2 backbone and inverted residual block

RangiLyu:
### This PR includes these features:
1. Add inverted residual block module which is used in MobileNetV2, V3, and EfficientNet series.
2. Add SE-layer in inverted residual block.
3. Add MobileNetV2 backbone.
4. Add unit tests of these modules.


tests/test_models/test_utils/test_inverted_residual.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+import torch
+from torch.nn.modules import GroupNorm
+from torch.nn.modules.batchnorm import _BatchNorm
+
+from mmdet.models.utils import InvertedResidual, SELayer
+
+
+def is_norm(modules):
```
ZwwWayne, line 9: MMCV should contain such an API.

ZwwWayne (commented):
Overall this PR LGTM.  Since we plan to converge our backbones into mmclassification, we should try to keep the API consistent with [that in mmcls](https://github.com/open-mmlab/mmclassification/blob/master/mmcls/models/backbones/mobilenet_v2.py)  as possible. So we need to check the consistency between these two implementations.

mmdet/models/backbones/mobilenet_v2.py
```diff
@@ -0,0 +1,197 @@
...
+            raise TypeError('pretrained must be a str or None')
+
+        self.widen_factor = widen_factor
+        self.out_indices = out_indices
+        for index in out_indices:
+            if index not in range(0, 8):
+                raise ValueError('the item in out_indices must in '
+                                 f'range(0, 8). But received {index}')
+
+        if frozen_stages not in range(-1, 8):
```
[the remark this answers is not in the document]
RangiLyu: The type of frozen_stages is int. I think there is no need to change.

mmdet/models/backbones/mobilenet_v2.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # Parameters to build layers. 4 parameters are needed to construct a
+    # layer, from left to right: expand_ratio, channel, num_blocks, stride.
+    arch_settings = [[1, 16, 1, 1], [6, 24, 2, 2], [6, 32, 3, 2],
+                     [6, 64, 4, 2], [6, 96, 3, 1], [6, 160, 3, 2],
+                     [6, 320, 1, 1]]
+
+    def __init__(self,
+                 widen_factor=1.,
+                 out_indices=(1, 2, 4, 7),
```
xvjiarui, line 47: Doesn't match the docstring.

mmdet/models/backbones/mobilenet_v2.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+@BACKBONES.register_module()
+class MobileNetV2(BaseModule):
+    """MobileNetV2 backbone.
+
+    Args:
+        widen_factor (float): Width multiplier, multiply number of
+            channels in each layer by this amount. Default: 1.0.
+        out_indices (None or Sequence[int]): Output from which stages.
```
xvjiarui, line 19:
```suggestion
        out_indices (Sequence[int], optional): Output from which stages.
```

xvjiarui (approved):

GUOShuxuan:
Hi, It is a great step to support MobileNet backbone. Is it possible to point out where is the pertained model for it?

ZwwWayne:
> Hi, It is a great step to support MobileNet backbone. Is it possible to point out where is the pertained model for it?

We will add SSD Lite after this PR. At that time the model config will link its corresponding pre-trained model. We are preparing now.

PR #5168, opened 2021-05-16
Supports for exporting DETR to onnx with dynamic shapes and batch inference

v-qjqs:
This PR:
1. Supports for exporting `DETR` to onnx with dynamic shapes and batch inference, achieving exactly the same performance (with no batch inference) between torch and exported onnx model (e.g. 40.1 vs 40.1 for torch and onnxruntime respectively, with batch size 1)
2. Supports batch inference of `DETR` for torch model, achieving 39.9 mAP. The performance gap (e.g. decrease 0.2 mAP) comes from the difference of valid positions for [`masks`](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/dense_heads/detr_head.py#L246): all positions are valid when evaluated with single image, while during batch inference the padded right-bottom positions are invalid.

Though supported, it's *not recommended* to use batch inference in onnxruntime for DETR, because there is huge performance gap between exported onnx model and torch model (e.g. 33.6 vs 39.9 for onnxruntime and torch with a batch size 2; 27.7 mAP for onnxruntime with a batch_size 8). The performance gap (e.g. 6.4 mAP) between torch and onnx *only exists during batch inference*, and the reason comes from the following two aspects:
1. the difference of valid positions for `masks`. (This hase little effect, e.g. 0.2 mAP)
2. the main reason: the [`img_shape`](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/dense_heads/detr_head.py#L674) is not dynamically traceable to onnx, instead the padded [`img_shape_for_onnx`](https://github.com/v-qjqs/mmdetection/blob/onnx_export/mmdet/models/dense_heads/detr_head.py#L801) is used. During batch inference, these is non-negligible effect on the predicted regressions, since the predicted coordinates (cxcywh) is normalized by the `img_shape` (without padding) and should be converted to absolute format by multiplying `img_shape`.


RunningLeon:
@v-qjqs Could you kindly fix CI errors?

mmdet/models/dense_heads/detr_head.py
```diff
@@ -241,18 +241,36 @@ def forward_single(self, x, img_metas):
...
+        if torch.onnx.is_in_onnx_export():
+            # Note `img_shape` is not dynamically traceable to ONNX,
+            # sicne the related augmentation was done with numpy under
+            # CPU. Here `img_shape` and 'batch_input_shape' are discarded,
+            # then `masks` is directly created with zeros (valid tag) and
+            # the same spatial shape as `x`.
+            # The difference between torch and exported ONNX model may be
+            # ignored, since the same performance is achieved (e.g.
+            # 40.1 vs 40.1 for DETR)
+            masks = x.new_zeros(*x.size()[-2:]).unsqueeze(0).repeat(
```
grimoire, line 253: `repeat` might bring an extra cost, can we just create a tensor with shape [B,h,w]?

docs/tutorials/pytorch2onnx.md
```diff
@@ -214,6 +221,7 @@ The table below lists the models that are guaranteed to be exportable to ONNX an
 |    YOLOv3    |    `configs/yolo/yolov3_d53_mstrain-608_273e_coco.py`    |       Y       |        Y        |       |
 | Faster R-CNN |   `configs/faster_rcnn/faster_rcnn_r50_fpn_1x_coco.py`   |       Y       |        Y        |       |
 |  Mask R-CNN  |     `configs/mask_rcnn/mask_rcnn_r50_fpn_1x_coco.py`     |       Y       |        Y        |       |
+|     DETR     |             `configs/detr/detr_r50_8x2_150e_coco.py`     |       Y       |        N        |       |
```
RunningLeon, line 224: Does `DETR` in ONNX support batch inference?

mmdet/models/dense_heads/detr_head.py
ZwwWayne: or considering expand()

RunningLeon:
@v-qjqs Hi,  you may need to refactor the onnx export according to #5205? Thanks a lot.

mmdet/models/detectors/detr.py
```diff
@@ -16,3 +18,29 @@ def __init__(self,
                  init_cfg=None):
         super(DETR, self).__init__(backbone, None, bbox_head, train_cfg,
                                    test_cfg, pretrained, init_cfg)
+
+    # over-write `onnx_export` bacause:
```
RunningLeon, line 22: typo: because

mmdet/models/dense_heads/detr_head.py
```diff
@@ -698,11 +698,121 @@ def simple_test_bboxes(self, feats, img_metas, rescale=False):
...
+        assert len(img_metas) == 1, \
+            'Only support one input image while in exporting to ONNX'
+
+        cls_scores = all_cls_scores_list[-1][-1]
+        bbox_preds = all_bbox_preds_list[-1][-1]
+
+        det_bboxes_list, det_labels_list = [], []
+        batch_size = cls_scores.size(0)
+        batch_size_tensor = torch.arange(batch_size).to(cls_scores.device)
+        for img_id in batch_size_tensor:
```
RunningLeon, line 801: This `for loop` through `batch_size` may be the reason why multi-batch is not supported in DETR. Could we use workaround to support multi-batch for DETR? Or we could add note why DETR in ONNX to not support multi-batch in the doc.

v-qjqs: Hi, how can I use a workaround to support batch inference for ONNX?

RunningLeon: Write a  `self._get_bboxes_single_onnx` for `self._get_bboxes_single` and support batch inference. You could take a chance.

RunningLeon:
@v-qjqs You could fix confits as well.

docs/tutorials/pytorch2onnx.md
v-qjqs: Currently it's not supported and not recommended to use batch inference in ONNX, since huge performance gap between exported onnx model and torch model (e.g. 33.5 vs 39.9 for onnxruntime and torch, with a batch size 2)

mmdet/models/dense_heads/detr_head.py
v-qjqs: I have succeed to dynamically support batch inference in onnxruntime, but it's not recommended to use batch inference for DETR, since my experiments show that there is huge performance gap between exported onnx model and torch model (e.g. 33.5 vs 39.9 for onnxruntime and torch, with a batch size 2)

mmdet/models/dense_heads/detr_head.py
```diff
@@ -795,24 +795,46 @@ def onnx_export(self, all_cls_scores_list, all_bbox_preds_list, img_metas):
...
+            bbox_preds = bbox_preds.view(batch_size, -1, 4)
+            det_labels = det_labels.view(batch_size, -1)
+
+        det_bboxes = bbox_cxcywh_to_xyxy(bbox_preds)
+        # use `img_shape_tensor` for dynamically exporting to ONNX
+        img_shape_tensor = img_shape.flip(0).repeat(2)  # [w,h,w,h]
+        img_shape_tensor = img_shape_tensor.unsqueeze(0).unsqueeze(0).expand(
+            batch_size, det_bboxes.size(1), 4)
+        det_bboxes = det_bboxes * img_shape_tensor
+        det_bboxes[..., 0::2].clamp_(min=0, max=img_shape[1])
```
RunningLeon, line 836: As far as I know, `torch.clamp` would be traced as `clip op` and the `min` and `max` are fixed in ONNX. Maybe you cloud debug here and try with the workaround like https://github.com/open-mmlab/mmdetection/blob/d5dbc0e2ef9bf89e6fc99e163b4c0c9b511846d3/mmdet/core/export/onnx_helper.py#L25

docs/tutorials/pytorch2onnx.md
```diff
@@ -249,6 +249,8 @@ Notes:
 - *All models above are tested with Pytorch==1.6.0 and onnxruntime==1.5.1*, except for CornerNet. For more details about the
 torch version when exporting CornerNet to ONNX, which involves `mmcv::cummax`, please refer to the [Known Issues](https://github.com/open-mmlab/mmcv/blob/master/docs/onnxruntime_op.md#known-issues) in mmcv.
 
+- Though supported, it is *not recommended* to use batch inference in onnxruntime for `DETR`, because there is huge performance gap between ONNX and torch model (e.g. 33.5 vs 39.9 mAP on COCO for onnxruntime and torch respectively, with a batch size 2). The main reason for the gap is that these is non-negligible effect on the predicted regressions during batch inference for ONNX, since the predicted coordinates is normalized by `img_shape` (without padding) and should be converted to absolute format, but `img_shape` is not dynamically traceable thus the padded `img_shape_for_onnx` is used.
```
RunningLeon, line 252: What is the result if you use batch size == 1?

RunningLeon (changes requested):

docs/tutorials/pytorch2onnx.md
v-qjqs: exactly the same as torch model, 40.1mAP

mmdet/models/dense_heads/detr_head.py
```diff
@@ -833,8 +833,11 @@ def onnx_export(self, all_cls_scores_list, all_bbox_preds_list, img_metas):
         img_shape_tensor = img_shape_tensor.unsqueeze(0).unsqueeze(0).expand(
             batch_size, det_bboxes.size(1), 4)
         det_bboxes = det_bboxes * img_shape_tensor
-        det_bboxes[..., 0::2].clamp_(min=0, max=img_shape[1])
-        det_bboxes[..., 1::2].clamp_(min=0, max=img_shape[0])
+        # dynamically clip bboxes
+        x1, y1, x2, y2 = det_bboxes.split((1, 1, 1, 1), dim=-1)
+        from mmdet.core.export import dynamic_clip_for_onnx
+        x1, y1, x2, y2 = dynamic_clip_for_onnx(x1, y1, x2, y2, img_shape)
+        det_bboxes = torch.cat([x1, y1, x2, y2], dim=-1)
```
RunningLeon, line 840: What's the performance after this change? Is it aligned with batchsize==1?

v-qjqs: 40.1 mAP with batch_size 1 and 33.6 mAP with batch_size 2, after the modification.

RunningLeon: Then, we may need to find real causes for the performance drop with batch-size > 1.

v-qjqs: As I mentioned at the beginning, there are two factors affect the performance of torch and onnx model, the second is the main.

RunningLeon: @v-qjqs 
Then, is there a constant shape that images from a batch should be padded into in the original PyTorch inference pipeline?

v-qjqs: Actually the padding for images for constructing a batch tensor is done by mmcv, and the `size_divisor` in DETR is set to 1.

docs/tutorials/pytorch2onnx.md
```diff
@@ -241,6 +249,8 @@ Notes:
...
+
```
RunningLeon, line 253: Maybe we could add info that test with batch is through another config: configs/detr/detr_r50_8x2_150e_coco_batchinference.py

RunningLeon (approved):
LGTM

RunningLeon:
Kindly ping@ZwwWayne, @jshilong

ZwwWayne:
Overall this PR LGTM. Can be merged after resolving the last comment.

PR #5226, opened 2021-05-26
fix py3.8 dict changed during iteration

jshilong:
## Motivation
In python3.8, changing the key of a dictionary during iteration is forbidden.

## BC-breaking (Optional)
None


hhaAndroid (approved):

ZwwWayne (approved):

Jason-cpp:
"RuntimeError: dictionary changed size during iteration" is a common bug. see [dictionary-iterators](https://www.python.org/dev/peps/pep-0234/#dictionary-iterators). It not a version-specific bug of py38.

PR #5227, opened 2021-05-26
Make `cat_ids` be consistent with the order of `CLASSES`

jshilong:
## Motivation
In pycocotools, 
https://github.com/cocodataset/cocoapi/blob/8c9bcc3cf640524c4c20a9c40e89cb6a2f2fa0e9/PythonAPI/pycocotools/coco.py#L173
The `ids`  can be not consistent with the order of `catNms` which is we do not expect,
this can results in a wrong visualization result when you specify a list of classes that is not consistent with the order of inner  `cats`.

## BC-breaking (Optional)
None

### DIFF
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -45,12 +45,7 @@ class CocoDataset(CustomDataset):
         """
 
         self.coco = COCO(ann_file)
-        self.cat_ids = []
-
-        # make cat_ids consistent with the order of class_name
-        for class_name in self.CLASSES:
-            self.cat_ids.append(self.coco.get_cat_ids(cat_names=class_name)[0])
-
+        self.cat_ids = self.coco.get_cat_ids(cat_names=self.CLASSES)
         self.cat2label = {cat_id: i for i, cat_id in enumerate(self.cat_ids)}
         self.img_ids = self.coco.get_img_ids()
         data_infos = []
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -45,7 +45,12 @@ class CocoDataset(CustomDataset):
         """
 
         self.coco = COCO(ann_file)
-        self.cat_ids = self.coco.get_cat_ids(cat_names=self.CLASSES)
+        self.cat_ids = []
+
+        # make cat_ids consistent with the order of class_name
+        for class_name in self.CLASSES:
+            self.cat_ids.append(self.coco.get_cat_ids(cat_names=class_name)[0])
+
         self.cat2label = {cat_id: i for i, cat_id in enumerate(self.cat_ids)}
         self.img_ids = self.coco.get_img_ids()
         data_infos = []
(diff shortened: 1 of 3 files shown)


ZwwWayne (approved):

ZwwWayne:
Need to add a unit test to test whether the behavior is as expected.

jshilong:
Done

mmdet/datasets/coco.py
```diff
@@ -45,7 +45,12 @@ def load_annotations(self, ann_file):
         """
 
         self.coco = COCO(ann_file)
-        self.cat_ids = self.coco.get_cat_ids(cat_names=self.CLASSES)
+        self.cat_ids = []
+
+        # make cat_ids consistent with the order of class_name
+        for class_name in self.CLASSES:
+            self.cat_ids.append(self.coco.get_cat_ids(cat_names=class_name)[0])
```
xvjiarui, line 52: Since we already wrapped coco api in mmdet. Maybe we could add an argument to get some thing like `self.coco.get_cat_ids(sorted=True)`?

xvjiarui (approved):

hhaAndroid (approved):

waterfall414:
Hi @jshilong , I may have got into this problem and can you help me to confirm if it is caused by this bug you fixed?

I have a COCO dataset with a few labels with ids. As examples of the labels, I have 
```json

{ "id": 1, "name": "A-1"}
{ "id": 2, "name": "B-2"}

```

In my experiment config (YOLO-v3), I accidentally changed the order of labels when I was defining my classes list
```python
classes = ('B-2', 'A-1')
```

So when I used show_result_pyplot to plot images with bbox and text labels, I can see the labels for 'A-1' and 'B-2' are wrong consistently evening in training samples.

If you can confirm, then it could be that your fix hasn't incorporated yet. 
My environment is set up by following the installation.md (latest, 2.14.0), which suggests to use pip install mim and then mim install mmdet.

jshilong:
> Hi @jshilong , I may have got into this problem and can you help me to confirm if it is caused by this bug you fixed?
> 
> I have a COCO dataset with a few labels with ids. As examples of the labels, I have
> 
> ```json
> { "id": 1, "name": "A-1"}
> { "id": 2, "name": "B-2"}
> ```
> 
> In my experiment config (YOLO-v3), I accidentally changed the order of labels when I was defining my classes list
> 
> ```python
> classes = ('B-2', 'A-1')
> ```
> 
> So when I used show_result_pyplot to plot images with bbox and text labels, I can see the labels for 'A-1' and 'B-2' are wrong consistently evening in training samples.
> 
> If you can confirm, then it could be that your fix hasn't incorporated yet.
> My environment is set up by following the installation.md (latest, 2.14.0), which suggests to use pip install mim and then mim install mmdet.

Sorry for the late respondence, this pr involves a serious bug, so I revert the master in PR #5246, so this bug is still  in the master, but you can fix it by yourself with #5243 (we did not adopt 5243 due to backward compatible)

PR #5251, opened 2021-05-30
Refactor export One-Stage model to ONNX

jshilong:
## Motivation
The recent ONNX-related development is too quickly, which made the code hard to read, and there is a problem that the return type of the same function such as `_get_bboxes`  in the `DenseHead`.,

## Modification
This PR moves all ONNX related code of One-Stage model to a new function `onnx_export` in the corresponding class.

## BC-breaking (Optional)
None

### DIFF
--- a/mmdet/models/dense_heads/anchor_head.py
+++ b/mmdet/models/dense_heads/anchor_head.py
@@ -559,16 +559,9 @@ class AnchorHead(BaseDenseHead, BBoxTestMixin):
         mlvl_cls_scores = [cls_scores[i].detach() for i in range(num_levels)]
         mlvl_bbox_preds = [bbox_preds[i].detach() for i in range(num_levels)]
 
-        if torch.onnx.is_in_onnx_export():
-            assert len(
-                img_metas
-            ) == 1, 'Only support one input image while in exporting to ONNX'
-            img_shapes = img_metas[0]['img_shape_for_onnx']
-        else:
-            img_shapes = [
-                img_metas[i]['img_shape']
-                for i in range(cls_scores[0].shape[0])
-            ]
+        img_shapes = [
+            img_metas[i]['img_shape'] for i in range(cls_scores[0].shape[0])
+        ]
         scale_factors = [
             img_metas[i]['scale_factor'] for i in range(cls_scores[0].shape[0])
         ]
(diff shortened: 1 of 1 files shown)


RunningLeon:
@jshilong Could fix the CI errors at first.

mmdet/models/dense_heads/base_dense_head.py
```diff
@@ -76,3 +77,185 @@ def simple_test(self, feats, img_metas, rescale=False):
...
+        assert len(cls_scores) == len(bbox_preds)
+        num_levels = len(cls_scores)
+
+        device = cls_scores[0].device
+        featmap_sizes = [cls_scores[i].shape[-2:] for i in range(num_levels)]
+
+        # TODO: @Haian Huang, a unified priori_generator, This func can
+        #  be used in both
+        #  anchor_free_head and anchor_head
+        mlvl_prioris = self.anchor_generator.grid_anchors(
```
RunningLeon, line 140: `mlvl_prioris` to `mlvl_priors`

mmdet/models/dense_heads/base_dense_head.py
```diff
@@ -76,3 +77,185 @@ def simple_test(self, feats, img_metas, rescale=False):
...
+            max_output_boxes_per_class = cfg.nms.get(
+                'max_output_boxes_per_class', 200)
+            iou_threshold = cfg.nms.get('iou_threshold', 0.5)
+            score_threshold = cfg.score_thr
+            nms_pre = cfg.get('deploy_nms_pre', -1)
+            return add_dummy_nms_for_onnx(batch_mlvl_bboxes, batch_mlvl_scores,
+                                          max_output_boxes_per_class,
+                                          iou_threshold, score_threshold,
+                                          nms_pre, cfg.max_per_img)
+        else:
```
RunningLeon, line 249: Have you tested with ` with_nms=False` when exporting to ONNX?

jshilong: Not yet

RunningLeon: we may need to  test whether it could export and how the outputs are organized when `with_nms=False`

mmdet/models/detectors/single_stage.py
```diff
@@ -154,7 +154,10 @@ def onnx_export(self, img, img_metas):
         # `CornerNet` and `CentripetalNet`, which 'pad_shape' is used
         # for inference
         img_metas[0]['pad_shape_for_onnx'] = img_shape
-        # TODO:move all onnx related code in bbox_head to onnx_export function
-        det_bboxes, det_labels = self.bbox_head.get_bboxes(*outs, img_metas)
+        if len(outs) == 2:
+            # add a dummy score_factors
+            outs = list(outs)
```
ZwwWayne, line 159: can we simply use `outs = outs + (None,)`?

v-qjqs:
Hi, I wonder when will this be finished?

jshilong:
> Hi, I wonder when will this be finished?

This PR is blocked now. And it would not be merged to`refactor_dense` instead of `master`

PR #5275, opened 2021-06-02
[Feature] Support non-bbox result show

BIGWangYuDong:
In order to support one-stage instance segmentation(e.g. SOLO, SOLOv2 etc.), it is important to support non-bbox result show. Now this can support SOLO.

Simply using segm_result to estimate bbox and then display the result

Need to improve after mmdet supporting one-stage instance segmentation method.

### DIFF
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -314,7 +314,8 @@ class BaseDetector(BaseModule, metaclass=ABCMeta):
                     for i, bbox in enumerate(bbox_result)
                 ]
             else:  # segm only
-                # estimate center of mask through segm_result
+                # Estimate the center of mask through segm_result
+                # so that the label is in the mask.
                 for scores, segms in \
                         zip(bbox_result, segm_result):
                     assert len(scores) == len(segms)
--- a/mmdet/models/detectors/base.py
+++ b/mmdet/models/detectors/base.py
@@ -275,7 +275,7 @@ class BaseDetector(BaseModule, metaclass=ABCMeta):
 
         Args:
             img (str or Tensor): The image to be displayed.
-            result (Tensor or np.ndarray): The results to draw over `img`
+            result (tuple or np.ndarray): The results to draw over `img`
                 bbox_result or (bbox_result, segm_result).
             score_thr (float, optional): Minimum score of bboxes to be shown.
                 Default: 0.3.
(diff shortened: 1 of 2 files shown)


mmdet/models/detectors/base.py
```diff
@@ -304,13 +304,40 @@ def show_result(self,
...
+                # estimte bbox through segm_result
+                for scores, segms in \
+                        zip(bbox_result, segm_result):
+                    assert len(scores) == len(segms)
+                    if len(segms) == 0:
+                        continue
+                    for score, segm in zip(scores, segms):
+                        bboxes.append(
+                            np.array([
+                                np.where(segm == 1)[1].min(),
```
ZwwWayne, line 323: This does `np.where(segm == 1)` four times, which is inefficient.

mmdet/models/detectors/base.py
```diff
@@ -304,13 +304,40 @@ def show_result(self,
             bbox_result, segm_result = result
             if isinstance(segm_result, tuple):
                 segm_result = segm_result[0]  # ms rcnn
+            if bbox_result[0].shape[1] == 5:  # bbox-segm two stage
```
ZwwWayne, line 307: We can allow to not draw bounding boxes if we only have segm and do not need to manually draw bounding boxes for that.

BIGWangYuDong:
We add a parameter `show_bbox (bool)` to discriminate whether to show the bounding box in the image. the default of `show_bbox` is True. 

For details:
If `show_bbox=True`, the result will show the bounding box, which is support detection(one-stage and two-stage) and two-stage instance segmentation(e.g. mask rcnn).
If `show_bbox=False`, the result will not show the bounding box, which is support one-stage instance segmentation (e.g. SOLO)

mmdet/models/detectors/base.py
hhaAndroid: The description of two stage is wrong, because yolact is not a two-stage.

mmdet/models/detectors/base.py
```diff
@@ -304,13 +307,41 @@ def show_result(self,
...
+                bboxes = np.vstack(bbox_result)
+                labels = [
+                    np.full(bbox.shape[0], i, dtype=np.int32)
+                    for i, bbox in enumerate(bbox_result)
+                ]
+            else:  # segm only
+                # estimate center of mask through segm_result
```
hhaAndroid, line 317: It should specify why the center point is estimated？

mmdet/models/detectors/base.py
hhaAndroid: In extreme cases, it is best to add a unit test.

mmdet/models/detectors/base.py
```diff
@@ -304,13 +307,42 @@ def show_result(self,
...
                 segm_result = segm_result[0]  # ms rcnn
+            if bbox_result[0].shape[1] == 5:  # bbox-segm two stage
+                bboxes = np.vstack(bbox_result)
+                labels = [
+                    np.full(bbox.shape[0], i, dtype=np.int32)
+                    for i, bbox in enumerate(bbox_result)
+                ]
+            else:  # segm only
+                # Estimate the center of mask through segm_result
+                # so that the label is in the mask.
```
hhaAndroid, line 318: so that the label can be displayed in the mask.

mmdet/models/detectors/base.py
```diff
@@ -304,13 +307,42 @@ def show_result(self,
...
+                            np.array([center_x, center_y, 0, 0,
+                                      float(score)]))
+                bboxes = np.vstack(bboxes)
+                labels = [
+                    np.full(len(segm), i, dtype=np.int32)
+                    for i, segm in enumerate(segm_result)
+                ]
+                show_bbox = False
         else:
             bbox_result, segm_result = result, None
```
hhaAndroid, line 340: type?

hhaAndroid (approved):

PR #5280, opened 2021-06-03
[Enhancement]: make deployment tools more friendly to use

RunningLeon:
## Motivation
This PR optimizes `pytorch2onnx.py` and `onnx2tensorrt.py` tools.

## Modification
- [x] Remove some arguments like `--std`, `--mean`, `--dataset` from the tools.
- [x] Use wrappers in `mmdet.core.export.model_wrappers` to do inference in the tools.


docs/tutorials/onnx2tensorrt.md
```diff
@@ -25,45 +25,41 @@
...
 ```
 
 Description of all arguments:
 
+- `config` : The path of a model config file.
 - `model` : The path of an ONNX model file.
 - `--trt-file`: The Path of output TensorRT engine file. If not specified, it will be set to `tmp.trt`.
 - `--input-img` : The path of an input image for tracing and conversion. By default, it will be set to `demo/demo.jpg`.
 - `--shape`: The height and width of model input. If not specified, it will be set to `400 600`.
+- `--min-shape`: The minimum height and width of model input. If not specified, it will be set to the same as `--shape`.
```
ZwwWayne, line 47: I suggest to also clean the `min-shape`, `shape`, and `max-shape`, which can also use the default value in config.

docs/tutorials/onnx2tensorrt.md
```diff
@@ -25,45 +25,41 @@
...
     ${MODEL} \
     --trt-file ${TRT_FILE} \
     --input-img ${INPUT_IMAGE_PATH} \
-    --shape ${IMAGE_SHAPE} \
+    --shape ${INPUT_IMAGE_SHAPE} \
+    --min-shape ${MIN_IMAGE_SHAPE} \
     --max-shape ${MAX_IMAGE_SHAPE} \
-    --mean ${IMAGE_MEAN} \
-    --std ${IMAGE_STD} \
-    --dataset ${DATASET_NAME} \
```
ZwwWayne, line 35: May add a clarification indicating the cleaned arguments and why they are unnecessary.

tools/deployment/pytorch2onnx.py
```diff
@@ -89,79 +89,74 @@ def pytorch2onnx(config_path,
...
-        if test_img is not None:
-            input_config['input_path'] = test_img
+
+        if test_img is None:
+            input_config['input_path'] = input_img
+
+        # prepare input once again
         one_img, one_meta = preprocess_example_input(input_config)
-        tensor_data = [one_img]
+        img_list, img_meta_list = [one_img], [[one_meta]]
```
ZwwWayne, line 124: Does this mean that we only test one image? Can we considering testing at least 100-500 images to ensure the correctness?

RunningLeon: We always test with only one image in this tool. You could evaluate the model with this tool(https://github.com/open-mmlab/mmdetection/blob/master/tools/deployment/test.py) and test with a dataset.

docs/tutorials/onnx2tensorrt.md
RunningLeon: min-shape and max-shape for converting from ONNX to TensorRT which can not be removed.

docs/tutorials/onnx2tensorrt.md
RunningLeon: Now we parse `normalize parameters` such as `std` and `mean` directly from config, thus they could be removed.

mmdet/core/export/model_wrappers.py
```diff
@@ -151,22 +151,30 @@ def forward_test(self, imgs, img_metas, **kwargs):
...
-            )
+            warnings.warn('If input model has custom op from mmcv, \
+                you may have to build mmcv with TensorRT from source.')
 
-        self.output_names = output_names
-        self.model = TRTWraper(engine_file, ['input'], output_names)
+        output_names = ['dets', 'labels']
+        model = TRTWraper(engine_file, ['input'], output_names)
+        with_masks = False
+        if len(model.engine) == 4:
```
RangiLyu, line 168: May add some comments about what this line means?

tools/deployment/onnx2tensorrt.py
```diff
@@ -25,21 +25,22 @@ def onnx2tensorrt(onnx_file,
...
                   show=False,
-                  dataset='coco',
                   workspace_size=1,
                   verbose=False):
     import tensorrt as trt
     onnx_model = onnx.load(onnx_file)
-    input_shape = input_config['input_shape']
     max_shape = input_config['max_shape']
+    min_shape = input_config['min_shape']
+    opt_shape = input_config['opt_shape']
```
RangiLyu, line 34: What does opt stand for?

tools/deployment/onnx2tensorrt.py
```diff
@@ -170,58 +149,92 @@ def parse_args():
...
         type=int,
         default=1,
         help='Max workspace size in GiB')
+
     args = parser.parse_args()
     return args
 
 
+def parse_normalize_cfg(test_pipeline):
+    transforms = test_pipeline[1]['transforms']
```
RangiLyu, line 175: This hardcode may have a bug because not all MultiScaleFlipAug is second in test_pipeline.

tools/deployment/onnx2tensorrt.py
```diff
@@ -170,58 +149,92 @@ def parse_args():
         nargs='+',
         default=[58.395, 57.12, 57.375],
         help='Variance value used for preprocess input data')
```
RangiLyu, line 151: Add some deprecated notice in `help` to avoid users still using these arguments when they run `-h` or `--help`

tools/deployment/onnx2tensorrt.py
```diff
@@ -170,58 +149,92 @@ def parse_args():
...
+
+
 if __name__ == '__main__':
 
     assert is_tensorrt_plugin_loaded(), 'TensorRT plugin should be compiled.'
     args = parse_args()
-
+    warnings.warn(
+        'Arguments like `--to-rgb`, `--mean`, `--std`, `--dataset` are '
+        'deprecated and will be removed in future releases.')
```
RangiLyu, line 188: Add some annotations in the warning message about where these args are moved to.

tools/deployment/pytorch2onnx.py
```diff
@@ -179,6 +174,14 @@ def pytorch2onnx(config_path,
         print('The numerical values are the same between Pytorch and ONNX')
 
 
+def parse_normalize_cfg(test_pipeline):
+    transforms = test_pipeline[1]['transforms']
```
RangiLyu, line 178: Same hardcode issue in onnx2trt.

RangiLyu (approved):
LGTM

ZwwWayne (approved):

PR #5282, opened 2021-06-03
[Feature] Skipping post-processing in Mask R-CNN

SemyonBevzuk:
## Motivation
When exporting Mask R-CNN model to ONNX, it may be useful to skip post-processing of masks.
Then the model will not require an additional `grid_sampler` operation from the MMCV and 
it will be possible to add custom post-processing to the original masks.

## Modification
* `mmdet/models/roi_heads/mask_heads/fcn_mask_head.py`: Added `rescale_mask_to_input_shape` attribute to `FCNMaskHead`. By default, its value is `True`. When `onnx_export` is called, the `rescale_mask_to_input_shape` value is checked and this determines the `_do_paste_mask` function will be called or not.
* `mmdet/models/roi_heads/standard_roi_head.py`: Now `onnx_export` can return a tensor with the shape (N, 28, 28), so I have changed some parameters of the `reshape` function.
* `tests/test_models/test_roi_heads/test_mask_head.py`: Added a test for checking the shape of masks with different `rescale_mask_to_input_shape` values.

## Use cases (Optional)
Example of export without post-processing:
```
pytorch2onnx.py config.py snapshot.pth --output-file config.onnx --cfg-options model.roi_head.mask_head.rescale_mask_to_input_shape=False --dynamic-export
```

### DIFF
--- a/mmdet/models/roi_heads/mask_heads/fcn_mask_head.py
+++ b/mmdet/models/roi_heads/mask_heads/fcn_mask_head.py
@@ -35,8 +35,7 @@ class FCNMaskHead(BaseModule):
                  predictor_cfg=dict(type='Conv'),
                  loss_mask=dict(
                      type='CrossEntropyLoss', use_mask=True, loss_weight=1.0),
-                 init_cfg=None,
-                 rescale_mask_to_input_shape=True):
+                 init_cfg=None):
         assert init_cfg is None, 'To prevent abnormal initialization ' \
                                  'behavior, init_cfg is not allowed to be set'
         super(FCNMaskHead, self).__init__(init_cfg)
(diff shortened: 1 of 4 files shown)


jshilong:
Thanks for your contribution, we would review this PR asap.

mmdet/models/roi_heads/mask_heads/fcn_mask_head.py
```diff
@@ -62,6 +63,7 @@ def __init__(self,
         self.predictor_cfg = predictor_cfg
         self.fp16_enabled = False
         self.loss_mask = build_loss(loss_mask)
+        self.rescale_mask_to_input_shape = rescale_mask_to_input_shape
```
jshilong, line 66: Adding this new arg to `init` of model may be a lack of consideration because it only is used in `onnx_export`,  We may add it to test config

SemyonBevzuk: Done, take a look at the current solution.

guisolsan:
Dear developers,

I am trying to move MaskRCNN from pytorch to onnx but I am getting the same error, I have tried using the information you propose but I still can't get it to work. Any help?

Thanks

PR #5311, opened 2021-06-07
Delete deprecated warning in FPN

AronLin:
## Motivation
Delete deprecated warning in FPN

## BC-Breaking
If you set `add_extra_convs=True` and `extra_convs_on_inputs=False` to add conv layers on the last output feature map of fpn convs instead of setting `add_extra_convs='on_output'` directly, it will not work.


hhaAndroid (approved):

ZwwWayne:
Will be merged just before the next release.

ZwwWayne (approved):

PR #5325, opened 2021-06-09
[Feature] Add dev_script to generate metafile.

RangiLyu:
## Feature
Add stript to generate PWC metafile when gather models.

## Usage
Run `.dev_scripts/gather_models.py` and it will generate a metafile.yml.
You need to fill in the download link in ` Weights:` after uploading the model to the cloud.
Also,  the `Collections` in metafile should be written manually.
And finally, do not forget to write the metafile path into model_zoo.yml.

### DIFF
--- a/.dev_scripts/gather_models.py
+++ b/.dev_scripts/gather_models.py
@@ -4,9 +4,24 @@ import json
 import os.path as osp
 import shutil
 import subprocess
+from collections import OrderedDict
 
 import mmcv
 import torch
+import yaml
+
+
+def ordered_yaml_dump(data, stream=None, Dumper=yaml.SafeDumper, **kwds):
+
+    class OrderedDumper(Dumper):
+        pass
+
+    def _dict_representer(dumper, data):
+        return dumper.represent_mapping(
+            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, data.items())
+
+    OrderedDumper.add_representer(OrderedDict, _dict_representer)
+    return yaml.dump(data, stream, OrderedDumper, **kwds)
 
 
 def process_checkpoint(in_file, out_file):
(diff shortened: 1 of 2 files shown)


.dev_scripts/gather_models.py
```diff
@@ -50,6 +68,61 @@ def get_final_results(log_json_path, epoch, results_lut):
...
+        if 'segm_mAP ' in model['results']:
+            metric = round(model['results']['segm_mAP'] * 100, 1)
+            results.append(
+                OrderedDict(
+                    Task='Instance Segmentation',
+                    Dataset=dataset_name,
+                    Metrics={'mask AP': metric}))
+
+        pwc_model_info['Results'] = results
+        pwc_model_info['Weights'] = 'Need to fill in after upload'
```
hhaAndroid, line 121: The weight should be filled in automatically.

.dev_scripts/gather_models.py
```diff
@@ -160,6 +233,10 @@ def main():
     print(f'Totally gathered {len(publish_model_infos)} models')
     mmcv.dump(models, osp.join(models_out, 'model_info.json'))
 
+    pwc_infos = convert_model_info_to_pwc(model_infos)
+    with open(osp.join(models_out, 'metafile.yml'), 'w') as f:
```
hhaAndroid, line 237: Each model should have a separate metafile, so it is correct ?

.dev_scripts/gather_models.py
```diff
@@ -28,7 +43,10 @@ def process_checkpoint(in_file, out_file):
 
 def get_final_epoch(config):
     cfg = mmcv.Config.fromfile('./configs/' + config)
-    return cfg.runner.max_epochs
+    epoch = cfg.runner.max_epochs
+    if cfg.data.train.type == 'RepeatDataset':
+        epoch *= cfg.data.train.times
+    return epoch
```
hhaAndroid, line 49: This code cannot be modified because `epoch_{}.pth` will make an error.

ZwwWayne (approved):

PR #5330, opened 2021-06-10
[Enhancement] Add main to pytorch2onnx

SemyonBevzuk:
## Motivation
This RP adds an argument list as an argument to the `parse_args` function, which makes it easier to use additional parsers.
The creation of an additional `main` function simplifies the use of wrappers over `pytorch2onnx`, because you can reuse argument handling.

## Modification
 `tools/deployment/pytorch2onnx.py`:
- Added an argument with a list of arguments to the `parse_args` function.
- Moved arguments handling with `pytorch2onnx` call to `main` function.

### DIFF
--- a/tools/deployment/pytorch2onnx.py
+++ b/tools/deployment/pytorch2onnx.py
@@ -179,7 +179,7 @@ def pytorch2onnx(config_path,
         print('The numerical values are the same between Pytorch and ONNX')
 
 
-def parse_args():
+def parse_args(args_list=None):
     parser = argparse.ArgumentParser(
         description='Convert MMDetection models to ONNX')
     parser.add_argument('config', help='test config file path')
@@ -235,13 +235,11 @@ def parse_args():
         '--dynamic-export',
         action='store_true',
         help='Whether to export onnx with dynamic axis.')
-    args = parser.parse_args()
+    args = parser.parse_args(args=args_list)
     return args
 
 
-if __name__ == '__main__':
-    args = parse_args()
-
+def main(args):
     assert args.opset_version == 11, 'MMDet only support opset 11 now'
 
     if not args.input_img:
@@ -276,3 +274,7 @@ if __name__ == '__main__':
         do_simplify=args.simplify,
         cfg_options=args.cfg_options,
         dynamic_export=args.dynamic_export)
+
+
+if __name__ == '__main__':
+    main(parse_args())


RunningLeon:
> ## Motivation
> This RP adds an argument list as an argument to the `parse_args` function, which makes it easier to use additional parsers.
> The creation of an additional `main` function simplifies the use of wrappers over `pytorch2onnx`, because you can reuse argument handling.
> 
> ## Modification
> `tools/deployment/pytorch2onnx.py`:
> 
> * Added an argument with a list of arguments to the `parse_args` function.
> * Moved arguments handling with `pytorch2onnx` call to `main` function.
@SemyonBevzuk Hi, thanks for the contribution. Could you give an example of "makes it easier to use additional parsers."?

SemyonBevzuk:
> > ## Motivation
> > This RP adds an argument list as an argument to the `parse_args` function, which makes it easier to use additional parsers.
> > The creation of an additional `main` function simplifies the use of wrappers over `pytorch2onnx`, because you can reuse argument handling.
> > ## Modification
> > `tools/deployment/pytorch2onnx.py`:
> > 
> > * Added an argument with a list of arguments to the `parse_args` function.
> > * Moved arguments handling with `pytorch2onnx` call to `main` function.
> 
> @SemyonBevzuk Hi, thanks for the contribution. Could you give an example of "makes it easier to use additional parsers."?

Hi. 
This can be useful in the following case:
Let's say I'm writing a script in which I want to reuse functionality from `pytorch2onnx.py`, but I need additional arguments on the command line.
On the current `master` branch, I cannot use `parse_args`, because it always processes `_sys.argv`.
Adding the `args_list` argument to `parse_args` will allow you to preprocess additional custom command line arguments.
Example:
```python
def parse_args():
    wrapper_args, other_args_list = parse_args_wrapper()
    from pytorch2onnx import parse_args as parse_args_pytorch2onnx
    pytorch2onnx_args = parse_args_pytorch2onnx(other_args_list)
    return wrapper_args, pytorch2onnx_args
```

RunningLeon:
@SemyonBevzuk Thanks for the example. This is a minor change and you could keep this change to your local branch. Based on the feedbacks, few users would be beneficial from this change. Thus, we may not need to merge this PR right now. If there is a strong need for this change in the future, we could discuss it again. Finally, thanks again.

RunningLeon:
@SemyonBevzuk 
Could we close this PR for now?
Thanks for the contribution and the understanding.

PR #5338, opened 2021-06-11
fix the import path of msda

jshilong:
## Motivation
In https://github.com/open-mmlab/mmcv/pull/978, `MultiScaleDeformableAttention` has been moved to `mmcv/ops/multi_scale_deform_attn.py`. We have to fix the import path of  `MultiScaleDeformableAttention`.

## BC-breaking
None

### DIFF
--- a/mmdet/models/utils/transformer.py
+++ b/mmdet/models/utils/transformer.py
@@ -1,4 +1,5 @@
 import math
+import warnings
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

noahc1510:
https://github.com/open-mmlab/mmdetection/blob/master/docs/get_started.md
This doc should be update that mmdet cannot work with mmcv>=1.3.6 because of this change.

PR #5349, opened 2021-06-14
Refactor anchor_generator and point_generator

jshilong:
## Motivation
As an essential step of unifying the `get_bboxes` of `anchor_head` and `anchor_free_head`, we need to design a more unified interface for the generation of anchors and points.

## Modification 
1. Rename the ``ANCHOR_GENERATORS`` to  ``PROIRS_GENERATORS``.
2. Add  `MlvlPointGenerator` which follow the logic of `AnchorGenerator`.
3. Add `spares_priors` to all `AnchorGenerator` and `PointGenerator` 
4. Rename `grid_anchors` and `grid_points` to `grid_priors`.
5. Add a more unified function as `single_level_grid_priors` to `AnchorGenerator` which follow the original implementation as `single_level_grid_anchors`, but we remove the `base_anchors` and `strides` from dataflow and add `level_idx` to be consistent with `MlvlPointGenerator`

## BC-Breaking
None.
I retain all the original interfaces and add deprecated warnings.

### DIFF
--- a/mmdet/core/anchor/anchor_generator.py
+++ b/mmdet/core/anchor/anchor_generator.py
@@ -268,7 +268,7 @@ class AnchorGenerator:
                       prior_idx,
                       featmap_size,
                       level_idx,
-                      dtype,
+                      dtype=torch.float32,
                       device='cuda'):
         """Generate sparse anchors according to the ``prior_idx``.
 
@@ -278,7 +278,8 @@ class AnchorGenerator:
             featmap_size (tuple[int]): feature map size arrange as (h, w).
             level_idx (int): The level index of corresponding feature
                 map.
-            dtype (obj:`torch.dtype`): Date type of points.
+            dtype (obj:`torch.dtype`): Date type of points.Defaults to
+                ``torch.float32``.
             device (obj:`torch.device`): The device where the points is
                 located.
         Returns:
--- a/mmdet/core/anchor/anchor_generator.py
+++ b/mmdet/core/anchor/anchor_generator.py
@@ -212,7 +212,7 @@ class AnchorGenerator:
         Args:
             featmap_sizes (list[tuple]): List of feature map sizes in
                 multiple feature levels.
-            device (str): Device where the anchors will be put on.
+            device (str): The device where the anchors will be put on.
 
         Return:
             list[torch.Tensor]: Anchors in multiple feature levels. \
@@ -267,8 +267,8 @@ class AnchorGenerator:
             base_anchors (torch.Tensor): The base anchors of a feature grid.
             featmap_size (tuple[int]): Size of the feature maps.
             stride (tuple[int], optional): Stride of the feature map in order
-                (w, h). Defaults to (16, 16).
-            device (str, optional): Device the tensor will be put on.
+                (h, w). Defaults to (16, 16).
+            device (str, optional): The device the tensor will be put on.
                 Defaults to 'cuda'.
 
         Returns:
@@ -276,8 +276,9 @@ class AnchorGenerator:
         """
         # keep as Tensor, so that we can covert to ONNX correctly
         feat_h, feat_w = featmap_size
-        shift_x = torch.arange(0, feat_w, device=device) * stride[0]
-        shift_y = torch.arange(0, feat_h, device=device) * stride[1]
+        stride_w, stride_h = stride
+        shift_x = torch.arange(0, feat_w, device=device) * stride_w
+        shift_y = torch.arange(0, feat_h, device=device) * stride_h
 
         shift_xx, shift_yy = self._meshgrid(shift_x, shift_y)
         shifts = torch.stack([shift_xx, shift_yy, shift_xx, shift_yy], dim=-1)
@@ -299,7 +300,7 @@ class AnchorGenerator:
             featmap_sizes (list(tuple)): List of feature map sizes in
                 multiple feature levels.
             pad_shape (tuple): The padded shape of the image.
-            device (str): Device where the anchors will be put on.
+            device (str): The device where the anchors will be put on.
 
         Return:
             list(torch.Tensor): Valid flags of anchors in multiple levels.
@@ -327,10 +328,11 @@ class AnchorGenerator:
         """Generate the valid flags of anchor in a single feature map.
 
         Args:
-            featmap_size (tuple[int]): The size of feature maps.
+            featmap_size (tuple[int]): The size of feature maps, arrange
+                as (h, w).
             valid_size (tuple[int]): The valid size of the feature maps.
             num_base_anchors (int): The number of base anchors.
-            device (str, optional): Device where the flags will be put on.
+            device (str, optional): The device where the flags will be put on.
                 Defaults to 'cuda'.
 
         Returns:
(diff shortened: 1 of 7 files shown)


mmdet/core/anchor/anchor_generator.py
```diff
@@ -1,12 +1,14 @@
+import warnings
+
 import mmcv
 import numpy as np
 import torch
 from torch.nn.modules.utils import _pair
 
-from .builder import ANCHOR_GENERATORS
+from .builder import PROIRS_GENERATORS
```
ZwwWayne, line 8: PROIRS -> PRIORS

mmdet/core/anchor/anchor_generator.py
```diff
@@ -230,6 +232,29 @@ def grid_anchors(self, featmap_sizes, device='cuda'):
...
         return multi_level_anchors
 
+    def grid_anchors(self, featmap_size, stride=16, device='cuda'):
+
+        warnings.warn('``grid_anchors`` would be deprecated soon. Please use '
+                      '``grid_priors``')
+        return self.grid_priors(
+            self, featmap_size=featmap_size, stride=stride, device=device)
+
+    def sparse_priors(self, level_idx, featmap_size, dtype, device, topk_inds):
```
ZwwWayne, line 242: dtype and device should be the last two arguments with default values.
Suggest the argument order as topkinds, featmap_size, level_idx, dtype, device based on their importance.

mmdet/core/anchor/anchor_generator.py
```diff
@@ -230,6 +232,29 @@ def grid_anchors(self, featmap_sizes, device='cuda'):
...
+        # recover the grid index in feature map from topk_inds
+        # only generate anchors on the filtered grids to reduce latency
+
+        height, width = featmap_size
+        num_anchors = self.num_base_anchors[level_idx]
+        anchor_id = topk_inds % num_anchors
+        x = (topk_inds // num_anchors) % width
+        y = (topk_inds // width // num_anchors) % height
+        priors = torch.stack(
+            [x, y, x, y], 1).to(dtype) * self.anchor_generator.strides[
```
ZwwWayne, line 252: self.anchor_generator -> self?

mmdet/core/anchor/anchor_generator.py
```diff
@@ -222,39 +224,61 @@ def grid_anchors(self, featmap_sizes, device='cuda'):
...
                 Defaults to 'cuda'.
 
         Returns:
             torch.Tensor: Anchors in the overall feature maps.
         """
         # keep as Tensor, so that we can covert to ONNX correctly
         feat_h, feat_w = featmap_size
-        shift_x = torch.arange(0, feat_w, device=device) * stride[0]
-        shift_y = torch.arange(0, feat_h, device=device) * stride[1]
+        stride_w, stride_h = stride
```
hhaAndroid, line 279: Code and docs are inconsistent.

mmdet/core/anchor/point_generator.py
```diff
@@ -35,3 +37,193 @@ def valid_flags(self, featmap_size, valid_size, device='cuda'):
...
+                                 stride=None,
+                                 device='cuda',
+                                 with_stride=False):
+        """Generate grid Points of a single level.
+
+        Note:
+            This function is usually called by method ``self.grid_priors``.
+
+        Args:
+            featmap_size (tuple[int]): Size of the feature maps, arrange as
```
hhaAndroid, line 110: featmap_size and stride. Do they need a unified order ?

mmdet/core/anchor/point_generator.py
```diff
@@ -35,3 +37,193 @@ def valid_flags(self, featmap_size, valid_size, device='cuda'):
...
+        return all_points
+
+    def valid_flags(self, featmap_sizes, pad_shape, device='cuda'):
+        """Generate valid flags of points of multiple feature levels.
+
+        Args:
+            featmap_sizes (list(tuple)): List of feature map sizes in
+                multiple feature levels, each size arrange as
+                as (h, w).
+            pad_shape (tuple): The padded shape of the image.
```
hhaAndroid, line 151: It’s better to explain the hw order

mmdet/core/anchor/point_generator.py
jshilong: This would cause BC-breaking,  The reason for `featmap_size` being (H, W) is it using the last two value of the feature map's shape (N, C, H, W), however in other cases, `image_meta` for example, all arrange as (W, H ).

hhaAndroid (approved):

hhaAndroid (approved):

jshilong:
I benchmark the performance of `reppoints` and test all checkpoints, all the results look good to me.

mmdet/core/anchor/point_generator.py
```diff
@@ -35,3 +37,198 @@ def valid_flags(self, featmap_size, valid_size, device='cuda'):
         valid_xx, valid_yy = self._meshgrid(valid_x, valid_y)
         valid = valid_xx & valid_yy
         return valid
+
+
+@PRIOR_GENERATORS.register_module()
+class MlvlPointGenerator:
+
```
ZwwWayne, line 44: This class also need a docstring, also explain what does mlvl means

jshilong: Actually, it should be `PointGenerator`, but it has been used by 
https://github.com/open-mmlab/mmdetection/blob/0e09f00a5bf76169386cadbf4c41bd0ff7f6208d/mmdet/core/anchor/point_generator.py#L7
So I add the prefix `Mlvl` which means `multi-level`, I would explain it in docs.

mmdet/core/anchor/point_generator.py
```diff
@@ -35,3 +37,199 @@ def valid_flags(self, featmap_size, valid_size, device='cuda'):
         valid_xx, valid_yy = self._meshgrid(valid_x, valid_y)
         valid = valid_xx & valid_yy
         return valid
+
+
+@PRIOR_GENERATORS.register_module()
+class MlvlPointGenerator:
+    """Standard points generator for multi-level(Mlvl) feature maps in 2D
```
ZwwWayne, line 44: multi-level(Mlvl) -> multi-level (Mlvl)

mmdet/core/anchor/point_generator.py
```diff
@@ -35,3 +37,199 @@ def valid_flags(self, featmap_size, valid_size, device='cuda'):
...
+    points-based detectors.
```
ZwwWayne, line 45: Suggest to merge PointGenerator and MlvlPointGenerator. 
From[here](https://github.com/open-mmlab/mmdetection/blob/670ecc25460ae583ccabb0b4f0182a284e17d0cc/mmdet/models/dense_heads/reppoints_head.py#L95) it seems RepPoints essentially use multi-level PointGenerator. So at least we can replace that.
To prevent BC-breaking, we can add an option named `return_list=False` by default.

jshilong: The original `PointGenerator` actually is a `single_level_points_genertator`, the original `grid_point` in `PointGenerator`  actually is `single_level_grid_priors` of `MlvlPointGenerator`, It is hard to merge them because the original implementation does not follow the design of `AnchorGenerator`.
I believe the best way is to replace `PointGenerator`  with `MlvlPointGenerator`, but I am afraid of BC-breaking.

mmdet/core/anchor/anchor_generator.py
```diff
@@ -87,7 +89,7 @@ def __init__(self,
 
         # calculate scales of anchors
         assert ((octave_base_scale is not None
-                and scales_per_octave is not None) ^ (scales is not None)), \
+                 and scales_per_octave is not None) ^ (scales is not None)), \
```
ZwwWayne, line 92: Should we also change num_base_anchors to num_base_priors and add a similar interface in PointGenerator?

ZwwWayne (approved):

Issue #5356, opened 2021-06-15
gcnet x101-32x4d checkpoint is wrong


greathope:
Please see https://github.com/open-mmlab/mmdetection/issues/4654#issue-813190257.
The backbone corresponding to this weight is resnet50. When I changed it, I can load it, but the ap is very poor in coco validation, below 0.1.
please check it, thanks.

hhaAndroid:
Hi @greathope You are right, there is indeed a problem with this config.

PR #5360, opened 2021-06-15
fix gcnet x101 config error

hhaAndroid:
## Motivation
The configuration of `cascade_mask_rcnn_x101_32x4d_fpn_syncbn-backbone_dconv_c3_xxx.py` is incorrect, backbone should not be resne50. 

Resolve: https://github.com/open-mmlab/mmdetection/issues/5356


ZwwWayne (approved):

ZwwWayne:
Need to check whether the checkpoint is correct.

hhaAndroid:
> Need to check whether the checkpoint is correct.



> Need to check whether the checkpoint is correct.

I checked that the config and weights are the same, which means that the config and weights are wrong. It needs to be retrained.

ZwwWayne:
> > Need to check whether the checkpoint is correct.
> 
> > Need to check whether the checkpoint is correct.
> 
> I checked that the config and weights are the same, which means that the config and weights are wrong. It needs to be retrained.

OK，then we need to re-benchmark the model in this PR.

PR #5363, opened 2021-06-15
[Fix] Not use deprecated pretrained kwarg when it is None

RangiLyu:
Currently, there is still some code setting `backbone.pretrained = pretrained` even when it is not used. This will cause unable to use mmcls backbones since there is no `pretrained` kwarg in mmcls.
This PR fixes this problem.

### DIFF
--- a/mmdet/models/detectors/rpn.py
+++ b/mmdet/models/detectors/rpn.py
@@ -1,3 +1,5 @@
+import warnings
+
 import mmcv
 import torch
 from mmcv.image import tensor2imgs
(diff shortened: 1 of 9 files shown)


ZwwWayne (approved):

mmdet/models/detectors/rpn.py
```diff
@@ -20,7 +22,10 @@ def __init__(self,
                  pretrained=None,
                  init_cfg=None):
         super(RPN, self).__init__(init_cfg)
-        backbone.pretrained = pretrained
+        if pretrained:
+            warnings.warn('DeprecationWarning: pretrained is a deprecated, '
```
ZwwWayne, line 26: a deprecated -> deprecated

innerlee:
Gentle reminder: for bug fixes, it is advised to add tests.

PR #5374, opened 2021-06-16
Unified FCOS and  VFNet bbox coder

hhaAndroid:
## Motivation
In order to facilitate the unified post-processing process and optimize the code, it is necessary to consider the unified FCOS and VFNet bbox coder

## Modification
1  Add `DistancePointBBoxCoder` in `mmdet/core/bbox/coder`
2  Replace distance2bbox in FCOS and VFNet with bbox_coder in get_bbox_single.

VFNet still has problems.


jshilong (commented):
It is better to benchmark the performance of training and testing.
And it is better to merge it to `refactor_dense` to avoid BC-breaking.

ZwwWayne:
Overall this PR LGTM. May fully apply box coder in these two heads, e.g., in get_targets etc.

jshilong (approved):

hhaAndroid:
> Overall this PR LGTM. May fully apply box coder in these two heads, e.g., in get_targets etc.

I have unified the FCOS Head, but VFNet has problems because it inherits ATSS and FCOS and there will be problems. The best solution is to refactor the VFNet Head code In the future. The current inheritance logic makes the code very difficult to understand.

PR #5403, opened 2021-06-21
[Feature] Add support for mask diagonal flip in TTA

yeliudev:
Add support for mask diagonal flip in `mmdet/core/post_processing/merge_augs.py` for test-time augmentation.


ZwwWayne:
Hi @yeliudev ,
Thanks for the PR, would you like to also write a unit test for that?

yeliudev:
> Hi @yeliudev ,
> Thanks for the PR, would you like to also write a unit test for that?

Thanks for your review. I've added the unit tests but I'm not sure whether they are in the right place.

hhaAndroid (approved):

PR #5431, opened 2021-06-24
[Fix] fix ssd_vgg backbone bug

BIGWangYuDong:
Fix SSD-VGG backbone bug,
`elif pretrained is None` need follow `if instance(Pretrained, str)`


RangiLyu (approved):
Thanks for fixing the misposition of code. Need to merge asap.

PR #5479, opened 2021-06-29
Update analyze_logs.py

Y-M-Y:
_Thank you very much_ for bringing great help to my research. I encountered a small problem in the process of using it, and I have already proposed the bug and my solution in _**issue 5396**_. I was told that I could create a PR to fix that, so I give it a try by creating this PR.

**Describe the bug**
When using the `plot_curve` function `tools/analysis_tools/analyze_logs.py` to plot the curve of loss-iteration, loss values with a large number of iterations will be misplaced in front of the loss with a small number of iterations. Therefore, some details of the loss-iteration curve are disorderly.
Figure_1

There could be some problems with the setting of abscissa(iters). I checked the `plot_curve` fucntion in `tools/analysis_tools/analyze_logs.py` and found that `num_iters_per_epoch` could not return the right value. It returns the iter of `mode:val` instead of the last/biggest number of `mode:train` in log.json file.

**Reproduction**
Using `python tools/analysis_tools/analyze_logs.py plot_curve log.json --keys loss_cls --legend loss_cls` to plot curve.
Here's my log and log.json file.
[logandlogjason20210619_165226.zip](https://github.com/open-mmlab/mmdetection/files/6681733/logandlogjason20210619_165226.zip)

**Bug fix**
I tried a method that can solve my problem, but it may not work for all related problems.

ln `line 68` of `tools/analysis_tools/analyze_logs.py`, I changed `num_iters_per_epoch = log_dict[epochs[0]]['iter'][-1]` to `num_iters_per_epoch = log_dict[epochs[0]]['iter'][-2]`. After the change, `num_iters_per_epoch` could return the last value of `mode:train` instead of the value of `mode:val`.
Figure_2

The solution only works when there is only one value of `mode:val`.

Thanks again.

**PS**
I do **NOT** check the potential influence on downstream projects, please pay attention.

### DIFF
--- a/tools/analysis_tools/analyze_logs.py
+++ b/tools/analysis_tools/analyze_logs.py
@@ -65,7 +65,7 @@ def plot_curve(log_dicts, args):
             else:
                 xs = []
                 ys = []
-                num_iters_per_epoch = log_dict[epochs[0]]['iter'][-1]
+                num_iters_per_epoch = log_dict[epochs[0]]['iter'][-2]
                 for epoch in epochs:
                     iters = log_dict[epoch]['iter']
                     if log_dict[epoch]['mode'][-1] == 'val':


OpenMMLab-Assistant-004:
Hi @Y-M-Y ！First of all, we want to express our gratitude for your significant PR in the mmdetection project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

If you have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）

PR #5515, opened 2021-07-02
[Refactor] refactor losses function, support key `reduction_override`

BIGWangYuDong:
Firstly, I used `weight_reduce_loss` to rewrite the forward function and, added `reduction` key in `__init__`, the reduction defaults to mean. Meanwhile, I refactored the GHMC, GHMR forward function in `ghm_loss.py`, which make the loss function can support the key `reduction_override `.

Then I supplemented the losses test unit, to ensure that the losses of the same task can be replaced.

I have confirmed the results of GHMC and GHMR are the same as before the modification.

### DIFF
--- a/mmdet/models/losses/ghm_loss.py
+++ b/mmdet/models/losses/ghm_loss.py
@@ -3,6 +3,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 
 from ..builder import LOSSES
+from .utils import weight_reduce_loss
 
 
 def _expand_onehot_labels(labels, label_weights, label_channels):
@@ -30,9 +31,16 @@ class GHMC(nn.Module):
         momentum (float): The parameter for moving average.
         use_sigmoid (bool): Can only be true for BCE based loss now.
         loss_weight (float): The weight of the total GHM-C loss.
+        reduction (str): Options are "none", "mean" and "sum".
+            Defaults to "sum"
     """
 
-    def __init__(self, bins=10, momentum=0, use_sigmoid=True, loss_weight=1.0):
+    def __init__(self,
+                 bins=10,
+                 momentum=0,
+                 use_sigmoid=True,
+                 loss_weight=1.0,
+                 reduction='mean'):
         super(GHMC, self).__init__()
         self.bins = bins
         self.momentum = momentum
@@ -46,8 +54,14 @@ class GHMC(nn.Module):
         if not self.use_sigmoid:
             raise NotImplementedError
         self.loss_weight = loss_weight
+        self.reduction = reduction
 
-    def forward(self, pred, target, label_weight, *args, **kwargs):
+    def forward(self,
+                pred,
+                target,
+                label_weight,
+                reduction_override=None,
+                **kwargs):
         """Calculate the GHM-C loss.
 
         Args:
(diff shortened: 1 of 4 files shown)


ZwwWayne:
We can consider add a unit test that uses Faster R-CNN as a base config, then change to use different regression loss in the config and run forward. This could ensure compatibility in the interfaces of each loss.

tests/test_models/test_loss_compatibility.py
```diff
@@ -0,0 +1,193 @@
...
+    }
+    img_metas = [img_matas]
+    cfg_model = ConfigDict(cfg_model)
+
+    # verify bbox loss function compatibility
+    for loss_bbox in loss_bboxes:
+        cfg_model.roi_head.bbox_head.loss_bbox = loss_bbox
+        model = build_detector(cfg_model)
+        loss = model.forward_train(imgs, img_metas, gt_bboxes, gt_labels)
+        assert isinstance(loss['loss_cls'], torch.Tensor)
```
ZwwWayne, line 184: we usually test whether loss is greater than 0

tests/test_models/test_loss_compatibility.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def test_loss_compatibility():
+    """Test loss_cls and loss_bbox compatibility.
+
+    Using Faster R-CNN as a sample, modifying the loss function in the config
+    file to verify the compatibility of Loss APIS
+    """
+    # Faster R-CNN config dict
+    cfg_model = dict(
```
ZwwWayne, line 15: Can we directly use _base_/models/faster_rcnn_r50_fpn.py?

BIGWangYuDong:
We split the `test_loss_compatibility` unit into `test_cls_loss_compatibility` and `test_box_loss_compatibility` 

We first use `_demo_mm_inputs` from test_forward.py to load the config file and build the detector. Then, we use `_get_detector_cfg` to generate a  non-empty truth batch and feed it into the `detector.forward_train`. It is important to set roi_head.bbox_head.reg_decoded_bbox be True when using IoULoss.

We rewrite the test unit to support pytest and change the way of assertion. In the latest version, we assert whether loss is greater than 0.

tests/test_models/test_loss_compatibility.py
```diff
@@ -0,0 +1,78 @@
+import pytest
+from test_forward import _demo_mm_inputs, _get_detector_cfg
```
ZwwWayne, line 2: may put this into the function to avoid modifing setup_cfg

mmdet/models/losses/ghm_loss.py
```diff
@@ -3,6 +3,7 @@
 import torch.nn.functional as F
 
 from ..builder import LOSSES
+from .utils import weight_reduce_loss
```
ZwwWayne, line 6: need to rebenchmark a retinanet-GHM

BIGWangYuDong:
Being re-benchmark retinanet-GHM on the slurm, will get the result tomorrow

BIGWangYuDong:
Being re-benchmark retinanet-GHM on the slurm, will get the result tomorrow

BIGWangYuDong:
@ZwwWayne 
got 37.4 or 37.1 mAP on R-50-FPN, which is higher than the original 37.0 mAP.
image

ZwwWayne (approved):

PR #5540, opened 2021-07-06
[Fix] Use `warnings.warn` instead of `logger.warning`

AronLin:
Logger.warning() will report a warning message every time during inference.
We should fix it.

### DIFF
--- a/mmdet/models/roi_heads/point_rend_roi_head.py
+++ b/mmdet/models/roi_heads/point_rend_roi_head.py
@@ -1,6 +1,6 @@
 # Modified from https://github.com/facebookresearch/detectron2/tree/master/projects/PointRend  # noqa
-import logging
 import os
+import warnings
 
 import numpy as np
 import torch
@@ -11,8 +11,6 @@ from mmdet.core import bbox2roi, bbox_mapping, merge_aug_masks
 from .. import builder
 from ..builder import HEADS
 from .standard_roi_head import StandardRoIHead
-
-logger = logging.getLogger(__name__)
 
 
 @HEADS.register_module()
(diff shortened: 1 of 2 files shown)


mmdet/models/dense_heads/corner_head.py
```diff
@@ -791,8 +793,8 @@ def _bboxes_nms(self, bboxes, labels, cfg):
             return bboxes, labels
 
         if 'nms_cfg' in cfg:
-            warning.warn('nms_cfg in test_cfg will be deprecated. '
-                         'Please rename it as nms')
+            logger.warning('nms_cfg in test_cfg will be deprecated. '
```
jshilong, line 796: logger.warning can not print only once

mmdet/models/dense_heads/ga_rpn_head.py
```diff
@@ -83,7 +85,7 @@ def _get_bboxes_single(self,
 
         # deprecate arguments warning
         if 'nms' not in cfg or 'max_num' in cfg or 'nms_thr' in cfg:
-            warnings.warn(
+            logger.warning(
```
jshilong, line 88: Every forward will print it

jshilong:
logger.warning can not print message only once, you should check all your changed warning

mmdet/core/anchor/anchor_generator.py
```diff
@@ -320,8 +323,10 @@ def grid_anchors(self, featmap_sizes, device='cuda'):
                 are the sizes of the corresponding feature level, \
                 num_base_anchors is the number of anchors for that level.
         """
-        warnings.warn('``grid_anchors`` would be deprecated soon. '
-                      'Please use ``grid_priors`` ')
+        if rank == 0:
+            # warnings.warn only print once
```
ZwwWayne, line 327: See https://github.com/open-mmlab/mmdetection/blob/c6515a7b84f8adb248adb789fd04358d0093b88d/mmdet/apis/inference.py#L46

jshilong (approved):
LGTM

ZwwWayne (approved):

PR #5544, opened 2021-07-06
Support DropBlock with Retinanet

hhaAndroid:
add DropBlock to improve model performance.


mmdet/models/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .losses import *  # noqa: F401,F403
 from .necks import *  # noqa: F401,F403
 from .roi_heads import *  # noqa: F401,F403
+from .plugin import *  # noqa: F401,F403
```
ZwwWayne, line 11: plugin -> plugins

ZwwWayne (approved):

PR #5554, opened 2021-07-07
[WIP]Base Classes for Single-Stage Instance Segmentation

jshilong:
## Motivation
We need some general-enough BaseClasses for One-Stage Instance Segmentation, Such as `SingleStageInstanceSegmentor` compare to One-Stage detector and `BaseMaskHead` compare to `BaseDenseHead`

## Modification
This PR adds `base_mask_head.py` and `single_stage_instance_seg.py`  and modify the existed method Yolact and SOLO to be compatible with base classes,
This base class will be used in the following methods  to prove its versatility
- [ ] SOLO
- [x] YOLACT
- [ ] SOLOv2
- [ ] CondInst

## BC-breaking (Optional)
For Yolact


mmdet/core/post_processing/matrix_nms.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Point NMS
+    Args:
+        heat (Tensor): shape (b, n, h, w)
+        kernel (int): kernel size, must be 2
+    Returns:
+        Tensor: cate_pred, tensors of shape (b, n, h, w)
+    """
+    hmax = nn.functional.max_pool2d(
+        heat, (kernel, kernel), stride=1, padding=1)
```
RangiLyu, line 76: This function can be more general. Since the logic is the same with `get_local_maximum` in CornerNet and CenterNet, it can be used in these algorithms.

jshilong: This PR only confirm the base class is general enough for four Algorithm, then @BIGWangYuDong would refactor the SOLO on his own PR, @BIGWangYuDong Please note this comment

BIGWangYuDong: OK

mmdet/models/losses/dice_loss.py
```diff
@@ -0,0 +1,55 @@
...
+        Args:
+            use_sigmoid (bool, optional): Whether to the prediction is
+                used for sigmoid or softmax. Defaults to True.
+            loss_weight (float, optional): Weight of loss. Defaults to 1.0.
+        """
+        super(DiceLoss, self).__init__()
+        self.use_sigmoid = use_sigmoid
+        self.loss_weight = loss_weight
+
+    def forward(self, pred, target):
```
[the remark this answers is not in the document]
jshilong: we have supported it before, The code is under refactoring

mmdet/models/detectors/single_stage_instance_seg.py
```diff
@@ -0,0 +1,189 @@
...
+import numpy as np
+import torch
+
+from mmdet.core import bbox2result
+from ..builder import DETECTORS, build_backbone, build_head, build_neck
+from .base import BaseDetector
+
+
+@DETECTORS.register_module()
+class SingleStageInstanceSegmentor(BaseDetector):
```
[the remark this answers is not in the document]
jshilong: Sometimes we do not need bbox_head sometimes

Issue #5037, opened 2021-04-21
a bug in test_mixins.py


ToumaKazusa3:
Thanks for your error report and we appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**
When I run the simple_test_bboxes function in the D:\anaconda3\envs\open-mmlab\Lib\site-packages\mmdet\models\roi_heads\test_mixins.py file,  If there is no proposal in this image (like [tensor([], device='cuda :0', size=(0, 4))]), the function will have the following error:
RuntimeError: cannot reshape tensor of 0 elements into shape [1, 0, -1] because the unspecified dimension size -1 can be any value and is ambiguous

**Reproduction**

1. What command or script did you run?
I run a script in mmtracking project, which is built upon MMDetection. 
```none
./tools/slurm_test.sh mediaf test-1 configs/mot/tracktor/tracktor_faster-rcnn_r50_fpn_4e_mot17-public-half.py 8 --eval track
```
**Environment**
sys.platform: linux
Python: 3.7.6 (default, Jan  8 2020, 19:59:22) [GCC 7.3.0]
CUDA available: True
GPU 0,1,2,3,4,5,6,7: GeForce GTX 1080
CUDA_HOME: /mnt/lustre/share/polaris/dep/cuda-9.0-cudnn7.6.5
NVCC: Cuda compilation tools, release 9.0, V9.0.176
GCC: gcc (GCC) 5.4.0
PyTorch: 1.5.0
PyTorch compiling details: PyTorch built with:
  - GCC 5.4
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.1 Product Build 20200208 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v0.21.1 (Git Hash 912ce228837d1ce28e1a61806118835de03f5751)
  - OpenMP 201307 (a.k.a. OpenMP 4.0)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 9.0
[...]

PR #5559, opened 2021-07-07
Fix two-stage runtime error when empty proposal

hhaAndroid:
## Motivation
Fix two-stage runtime error when empty proposal

Resolve: https://github.com/open-mmlab/mmdetection/issues/5037

## Modification
All two-stage models deal with the case of empty proposal specially.

### DIFF
--- a/mmdet/models/roi_heads/cascade_roi_head.py
+++ b/mmdet/models/roi_heads/cascade_roi_head.py
@@ -335,14 +335,14 @@ class CascadeRoIHead(BaseRoIHead, BBoxTestMixin, MaskTestMixin):
                         self.bbox_head[i].loss_cls.get_activation(s)
                         for s in cls_score
                     ]
-                refine_roi_list = []
+                refine_rois_list = []
                 for j in range(num_imgs):
                     if rois[j].shape[0] > 0:
                         bbox_label = cls_score[j][:, :-1].argmax(dim=1)
-                        refine_roi = self.bbox_head[i].regress_by_class(
+                        refined_rois = self.bbox_head[i].regress_by_class(
                             rois[j], bbox_label[j], bbox_pred[j], img_metas[j])
-                        refine_roi_list.append(refine_roi)
-                rois = torch.cat(refine_roi_list)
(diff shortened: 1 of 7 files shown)


jshilong (commented):
Add unitests for empty `rois` can be better

hhaAndroid:
> Add unitests for empty `rois` can be better

done

ZwwWayne (approved):

mmdet/models/roi_heads/cascade_roi_head.py
```diff
@@ -317,13 +335,14 @@ def simple_test(self, x, proposal_list, img_metas, rescale=False):
...
-                                                       img_metas[j])
-                    for j in range(num_imgs)
-                ])
+                refine_roi_list = []
+                for j in range(num_imgs):
+                    if rois[j].shape[0] > 0:
+                        bbox_label = cls_score[j][:, :-1].argmax(dim=1)
+                        refine_roi = self.bbox_head[i].regress_by_class(
+                            rois[j], bbox_label[j], bbox_pred[j], img_metas[j])
+                        refine_roi_list.append(refine_roi)
```
ZwwWayne, line 344: 1. It seems that rois.shape==0 is not handled, will that have influence?
2. refine_roi -> refined_rois

mmdet/models/roi_heads/cascade_roi_head.py
```diff
@@ -317,13 +335,14 @@ def simple_test(self, x, proposal_list, img_metas, rescale=False):
...
-                                                       img_metas[j])
-                    for j in range(num_imgs)
-                ])
+                refine_rois_list = []
+                for j in range(num_imgs):
+                    if rois[j].shape[0] > 0:
+                        bbox_label = cls_score[j][:, :-1].argmax(dim=1)
+                        refined_rois = self.bbox_head[i].regress_by_class(
+                            rois[j], bbox_label[j], bbox_pred[j], img_metas[j])
+                        refine_rois_list.append(refined_rois)
```
ZwwWayne, line 344: refined_roi_list

PR #5562, opened 2021-07-07
Fix valid_mask logic error of rpn_head

hhaAndroid:
## Motivation

There is an error in the rpn filtering minimum bbox logic.

## Modification

1. Fix rpn and cascade rpn head errors of filtering minimum bbox logic
2. Modify ga rpn head to unify the writing of filtering minimum bbox logic


jshilong (approved):
LGTM

PR #5566, opened 2021-07-08
fix all_ious shape mismatch bug

fangwudi:
all_iours use list instead of array format  because of mismatch shape between different images


jshilong:
Will it cause any error？

jshilong:
Any update?

PR #4912, opened 2021-04-07
Post process

jshilong:

PR #5592, opened 2021-07-12
The `multiclass_nms` should return the global indices.

zeakey:
## Motivation
This PR improves the codes for multi_class bbox NMS by returning the global indice.

Currently [multiclass_nms](https://github.com/open-mmlab/mmdetection/blob/b9dedfdc8b660804bb31ae031a251e86a4905ade/mmdet/core/post_processing/bbox_nms.py#L7-L94) returns the local indices after batched_nms:

<https://github.com/open-mmlab/mmdetection/blob/b9dedfdc8b660804bb31ae031a251e86a4905ade/mmdet/core/post_processing/bbox_nms.py#L85-L92>

The returned indices is useless outside the function unless it's a global indice w.r.t the inputs.

## Modification
The modification is quite slight: change <https://github.com/open-mmlab/mmdetection/blob/b9dedfdc8b660804bb31ae031a251e86a4905ade/mmdet/core/post_processing/bbox_nms.py#L92>

to `return dets, labels[keep], inds[keep]`

### DIFF
--- a/mmdet/core/post_processing/bbox_nms.py
+++ b/mmdet/core/post_processing/bbox_nms.py
@@ -89,7 +89,7 @@ def multiclass_nms(multi_bboxes,
         keep = keep[:max_num]
 
     if return_inds:
-        return dets, labels[keep], keep
+        return dets, labels[keep], inds[keep]
     else:
         return dets, labels[keep]
 


jshilong:
Originally I fixed it in pr #4912.However #4912 is blocked, I think we can merge this pr first

jshilong (approved):
LGTM

zeakey:
@jshilong Thx for the quick response. The only concern is that there may be other codes just used the 'local indices' but I think this case is very rare.

jshilong:
Ping @ZwwWayne

jshilong (approved):

jshilong:
> @jshilong Thx for the quick response. The only concern is that there may be other codes just used the 'local indices' but I think this case is very rare.

@ZwwWayne We should notify this modification to MMtracking and MMAction and MMpose

ZwwWayne (approved):

PR #5603, opened 2021-07-13
[Fix] Fix some backbone init from pretrained checkpoint error in config file

BIGWangYuDong:
Due to moving the `model.pretrained` into `model.backbone.init_cfg`, this is a large-scale changes in config file. We add a test unit to check whether the backbone is successfully loaded the pretrained checkpoint. Now, we support check `ResNet, ResNext, VGG, Darknet` .etc.

We traversed all potential config files under the `config` directory. we ignore the folder  `'_base_', 'common'` because it does not contain the model. Meanwhile, we ignore the folder `legacy_1.x` due to it does not belong to mmdet 2.X. 

First, using `mmcv._load_checkpoint` to load the checkpoint without loading models.
Then, using `build_detector` to build models, and using `model.init_weights()` to initialize the parameters.
Finally, assert weights and bias of each layer loaded from pretrained checkpoint are equal to the weights and bias of the original checkpoint. For the convenience of comparison, we sum up the weights and biases of each loaded layer separately.

After traversing the config file, we fix some backbone init from pretrained checkpoint error in the config file.

### DIFF
--- a/.dev_scripts/test_init_backbone.py
+++ b/.dev_scripts/test_init_backbone.py
@@ -131,20 +131,10 @@ def _check_backbone(config, print_cfg=True):
             test_cfg=cfg.get('test_cfg'))
         model.init_weights()
 
-        checkpoint_layers = dict()
-        backbone_layers = dict()
-        for name, value in state_dict.items():
-            split_name = name.split('.')
-            if split_name[-1] == 'weight':
-                checkpoint_layers[name] = value.sum()
+        checkpoint_layers = state_dict.keys()
         for name, value in model.backbone.state_dict().items():
-            split_name = name.split('.')
-            if split_name[-1] == 'weight':
-                backbone_layers[name] = value.sum()
-
-        check_layers = list(set(backbone_layers) & set(checkpoint_layers))
-        for layer_name in check_layers:
-            assert backbone_layers[layer_name] == checkpoint_layers[layer_name]
+            if name in checkpoint_layers:
+                assert value.equal(state_dict[name])
 
         if print_cfg:
             print('-' * 10 + 'Successfully load checkpoint' + '-' * 10 +
--- a/.dev_scripts/test_init_backbone.py
+++ b/.dev_scripts/test_init_backbone.py
@@ -112,100 +112,32 @@ def _check_backbone(config, print_cfg=True):
         init_flag = False
     if init_flag:
         checkpoint = _load_checkpoint(init_cfg.checkpoint)
+        if 'state_dict' in checkpoint:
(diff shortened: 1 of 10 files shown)


.dev_scripts/test_init_backbone.py
```diff
@@ -0,0 +1,280 @@
...
+def _test_load_pretrained():
+    """We traversed all potential config files under the `config` file. If you
+    need to print details or debug code, you can use this function.
+
+    Returns
+        check_cfg_names (list[str]): Config files that backbone initialized
+            from pretrained checkpoint might be problematic. Need to recheck
+            the config file. The output including the config files that the
+            backbone.init_cfg is None
+    # >>> from test_init_backbone import _test_load_pretrained
```
jshilong, line 242: Please follow the format of  [`Example`](https://github.com/open-mmlab/mmdetection/blob/5f6134796ed07f443d1bf389f57d3cee40bbd044/mmdet/core/anchor/anchor_generator.py#L41) or just remove it

.dev_scripts/test_init_backbone.py
```diff
@@ -0,0 +1,280 @@
...
+            train_cfg=cfg.get('train_cfg'),
+            test_cfg=cfg.get('test_cfg'))
+        model.init_weights()
+
+        layers = []
+        names = []
+        checkpoint_names = []
+        layer_indexes = []
+        darknet_indexs = []
+        for name in model.backbone.state_dict():
```
jshilong, line 126: We can simplify the logic by directly comparinging the `state_dict` and `checkpoitn`

.dev_scripts/test_init_backbone.py
```diff
@@ -0,0 +1,190 @@
...
+        if 'state_dict' in checkpoint:
+            state_dict = checkpoint['state_dict']
+        else:
+            state_dict = checkpoint
+
+        model = build_detector(
+            cfg.model,
+            train_cfg=cfg.get('train_cfg'),
+            test_cfg=cfg.get('test_cfg'))
+        model.init_weights()
```
jshilong, line 132: How about simplifying the code as
```
check_layers = set(state_dict.keys()) & set(model.backbone.state_dict().keys())
for key in check_layers:
    assert (state_dict[key] == model.backbone.state_dict()[key]).all()
```

jshilong: Because in your implementation, `bias` is not checked.

ZwwWayne:
ping @jshilong

ZwwWayne (approved):

PR #5625, opened 2021-07-16
fix metric of voc dataset

jshilong:
## Motivation
In the official [metric implementation](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCdevkit_18-May-2011.tar) of VOC dataset, it adopts the  Legacy coordinate system in mmdet 1.x, which means h, w should be computed as `x2 - x1 + 1` and `y2 - y1 + 1`.

## Modification
I add argument `Legacy_coordinate` to `eval_map`， `tpfp_default` and `tpfp_imagenet`，add `extra_length` to `bbox_overlaps` to control whether to use legacy coordinate system when caculate the width and length.

## BC-breaking 
None

### DIFF
--- a/mmdet/core/evaluation/bbox_overlaps.py
+++ b/mmdet/core/evaluation/bbox_overlaps.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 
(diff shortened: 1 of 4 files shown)


ZwwWayne:
Need to also fix compatibility doc

PR #5626, opened 2021-07-16
[Fix] Fix unittest for Numclasshook when it is not used.

AronLin:
When `Numclasshook` is not used in the config, the unit test for this config still runs `_check_numclasscheckhook`.

I fix it by adding checking.

### DIFF
--- a/tests/test_runtime/test_config.py
+++ b/tests/test_runtime/test_config.py
@@ -83,7 +83,6 @@ def test_config_build_detector():
     for config_fname in config_names:
         config_fpath = join(config_dpath, config_fname)
         config_mod = Config.fromfile(config_fpath)
-        config_mod.model
         print(f'Building detector, config_fpath = {config_fpath}')
 
         # Remove pretrained keys to allow for testing in an offline environment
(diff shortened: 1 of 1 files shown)


tests/test_runtime/test_config.py
```diff
@@ -93,7 +92,16 @@ def test_config_build_detector():
         detector = build_detector(config_mod.model)
         assert detector is not None
 
-        _check_numclasscheckhook(detector, config_mod)
+        # Check whether NumClassCheckHook is used.
+        assert config_mod.custom_hooks is None or \
```
jshilong, line 96: config_mod.get("custom_hooks", None)

jshilong (approved):
LGTM

tests/test_runtime/test_config.py
ZwwWayne: config_mod.get('custom_hooks', [])

tests/test_runtime/test_config.py
```diff
@@ -93,7 +92,16 @@ def test_config_build_detector():
         detector = build_detector(config_mod.model)
         assert detector is not None
 
-        _check_numclasscheckhook(detector, config_mod)
+        # Check whether NumClassCheckHook is used.
+        custom_hooks = config_mod.get('custom_hooks', [])
+        assert custom_hooks is None or isinstance(custom_hooks, list)
```
ZwwWayne, line 97: unnecessary assertion for None

AronLin: `custom_hooks` is possibly set to `None`.

PR #5640, opened 2021-07-19
add annotations for instaboost

likyoo:
## Motivation

add annotations for instaboost, referencing https://github.com/GothicAi/InstaBoost-pypi#instaboostconfig

```
Args:
        action_candidate (tuple): action candidates. "normal", "horizontal", \
            "vertical", "skip" are supported.
        action_prob (tuple): corresponding action probabilities. Should be the \
            same length as action_candidate.
        scale (tuple): (min scale, max scale)
        dx (int): the maximum x-axis shift will be (instance width) / dx
        dy (int): the maximum y-axis shift will be (instance height) / dy
        theta (tuple): (min rotation degree, max rotation degree)
        color_prob (float): probability of images for color augmentation
        heatmap_flag (bool): whether to use heatmap guided
        aug_ratio (float): probability of applying this transformation
```

### DIFF
--- a/mmdet/datasets/pipelines/instaboost.py
+++ b/mmdet/datasets/pipelines/instaboost.py
@@ -10,6 +10,19 @@ class InstaBoost:
     <https://arxiv.org/abs/1908.07801>`_.
 
     Refer to https://github.com/GothicAi/Instaboost for implementation details.
+
+    Args:
+        action_candidate (tuple): action candidates. "normal", "horizontal", \
+            "vertical", "skip" are supported.
+        action_prob (tuple): corresponding action probabilities. Should be the \
+            same length as action_candidate.
+        scale (tuple): (min scale, max scale)
+        dx (int): the maximum x-axis shift will be (instance width) / dx
+        dy (int): the maximum y-axis shift will be (instance height) / dy
+        theta (tuple): (min rotation degree, max rotation degree)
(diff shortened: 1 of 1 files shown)


mmdet/datasets/pipelines/instaboost.py
```diff
@@ -10,6 +10,19 @@ class InstaBoost:
...
+            "vertical", "skip" are supported.
+        action_prob (tuple): corresponding action probabilities. Should be the \
+            same length as action_candidate.
+        scale (tuple): (min scale, max scale)
+        dx (int): the maximum x-axis shift will be (instance width) / dx
+        dy (int): the maximum y-axis shift will be (instance height) / dy
+        theta (tuple): (min rotation degree, max rotation degree)
+        color_prob (float): probability of images for color augmentation
+        heatmap_flag (bool): whether to use heatmap guided
+        aug_ratio (float): probability of applying this transformation
```
hhaAndroid, line 25: Thank you for your contribution. Can you add the default parameter description and fix lint？ e.g aug_ratio (float): probability of applying this transformation. Default 0.5.

likyoo: ok !

ZwwWayne:
Other default values should also be added.

hhaAndroid (approved):

PR #5646, opened 2021-07-19
[Enhance] Add ignore_index to CrossEntropyLoss

AronLin:
The `CrossEntropyLoss` has no interface `ignore_index`, which makes it inconvenient to complete the segmentation task.

### DIFF
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -48,23 +48,22 @@ def cross_entropy(pred,
     return loss
 
 
-def _expand_onehot_labels(labels, label_weights, target_shape, ignore_index):
+def _expand_onehot_labels(labels, label_weights, label_channels, ignore_index):
     """Expand onehot labels to match the size of prediction."""
-    bin_labels = labels.new_zeros(target_shape)
+    bin_labels = labels.new_full((labels.size(0), label_channels), 0)
     valid_mask = (labels >= 0) & (labels != ignore_index)
-    inds = torch.nonzero(valid_mask, as_tuple=True)
+    inds = torch.nonzero(
+        valid_mask & (labels < label_channels), as_tuple=False)
 
-    if inds[0].numel() > 0:
-        if labels.dim() == 3:
-            bin_labels[inds[0], labels[valid_mask], inds[1], inds[2]] = 1
-        else:
-            bin_labels[inds[0], labels[valid_mask]] = 1
+    if inds.numel() > 0:
+        bin_labels[inds, labels[inds]] = 1
 
-    valid_mask = valid_mask.unsqueeze(1).expand(target_shape).float()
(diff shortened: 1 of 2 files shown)


AronLin:
I will check the training and inference processes of HTC.

AronLin:
kindly ping @jshilong.

AronLin:
The training and inference processes of Mask R-CNN are ok.

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -23,12 +24,20 @@ def cross_entropy(pred,
         avg_factor (int, optional): Average factor that is used to average
             the loss. Defaults to None.
         class_weight (list[float], optional): The weight for each class.
+        ignore_index (int | None): The label index to be ignored.
+            If None, it will be set to default value. Default: -100.
```
jshilong, line 28: I am confused about this default value

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -23,12 +24,20 @@ def cross_entropy(pred,
         avg_factor (int, optional): Average factor that is used to average
             the loss. Defaults to None.
         class_weight (list[float], optional): The weight for each class.
+        ignore_index (int | None): The label index to be ignored.
```
jshilong, line 27: `ignore_index` may be confusing, no keyword in the name to indicate it  is only related  to `label`

jshilong: Is it right the type `int? Does it only support ignore one class?

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -60,7 +74,8 @@ def binary_cross_entropy(pred,
                          weight=None,
                          reduction='mean',
                          avg_factor=None,
-                         class_weight=None):
+                         class_weight=None,
+                         ignore_index=255):
```
jshilong, line 78: 255 is so confusing

mmdet/models/losses/cross_entropy_loss.py
AronLin: ignore_index is a key of F.cross_entropy_loss(). It only supports ignore one class. Usually, this number does not belong to the classes of a model.

mmdet/models/losses/cross_entropy_loss.py
AronLin: -100 is the default value of ignore_index in F.cross_entropy_loss.

mmdet/models/losses/cross_entropy_loss.py
AronLin: In this function, we do not use ignore_index in the F.cross_entropy_loss.
So I refer to the setting of mmsegmentation, setting it to 255 by default.

ZwwWayne (approved):

PR #5705, opened 2021-07-28
Complete Support of YOLOX

ZwwWayne:


ZwwWayne:
Close this PR because it is outdated, see https://github.com/open-mmlab/mmdetection/pull/5828

PR #5732, opened 2021-07-30
[Feature] Add function upsample_like and its unit test

AronLin:
## Motivation

The preparation work for Panoptic Segmentation, used in semantic heads.
This function upsamples the logits to the shape of the target.
The input logits can be 3D or 4D. The target can be a Tensor or a `np.ndarray` with the shape (..., target_h, target_w).

## BC-breaking (Optional)

No.


mmdet/models/utils/misc.py
```diff
@@ -0,0 +1,42 @@
+from torch.nn import functional as F
+
+
+def upsample_like(logits, target, mode='bilinear', align_corners=False):
+    """Up-sample the logits to the shape of target.
```
ZwwWayne, line 5: Upsample

AronLin: Done.

ZwwWayne (approved):

PR #5746, opened 2021-08-01
[Feature] Add ConvUpsample module and its unit test

AronLin:
Convupsample performs 2x upsampling after Conv.

There are several `ConvModule` layers. In the first few layers, upsampling will be applied after each layer of convolution. The number of upsampling must be no more than the number of ConvModule layers.

`num_layer` means the number of convolutional layers.  
`num_upsample` means the number of upsampling, it must be no more than `num_layers`.

### DIFF
--- a/mmdet/models/utils/conv_upsample.py
+++ b/mmdet/models/utils/conv_upsample.py
@@ -6,8 +6,9 @@ from mmcv.runner import BaseModule, ModuleList
 class ConvUpsample(BaseModule):
     """ConvUpsample performs 2x upsampling after Conv.
 
-    This module performs 2x upsampling after Conv. Upsampling will be applied
-    after the first few layers of convolution.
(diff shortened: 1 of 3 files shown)


mmdet/models/utils/conv_upsample.py
```diff
@@ -0,0 +1,65 @@
+import torch.nn.functional as F
+from mmcv.cnn import ConvModule
+from mmcv.runner import BaseModule, ModuleList
+
+
+class ConvUpsample(BaseModule):
+    """ConvUpsample performs 2x upsampling after Conv.
+
+    This module performs 2x upsampling after Conv. Upsampling will be applied
```
ZwwWayne, line 9: The description is inaccurate. Because it performs more than 2x upsamples if num_upsample > 1

ZwwWayne (approved):

PR #5776, opened 2021-08-03
[bug] fix bug of classwise weight in bce loss

Ezra-Yu:

## Motivation

 fix bug of classwise_weight in bce loss, and add unit test

## Modification

[this line](https://github.com/open-mmlab/mmdetection/blob/90e0ec78a705f26d4553503a433a5fdbc73757f2/mmdet/models/losses/cross_entropy_loss.py#L107) has a bug. the loss only scale in the location where the label is 1. The class_weight has no effect on the location where the label is 0.

my exp:
image

### DIFF
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -104,7 +104,7 @@ def binary_cross_entropy(pred,
     if weight is not None:
         weight = weight.float()
     loss = F.binary_cross_entropy_with_logits(
-        pred, label.float(), pos_weight=class_weight, reduction='none')
+        pred, label.float(), weight=class_weight, reduction='none')
     # do the reduction for the weighted loss
     loss = weight_reduce_loss(
         loss, weight, reduction=reduction, avg_factor=avg_factor)
--- a/tests/test_metrics/test_losses.py
+++ b/tests/test_metrics/test_losses.py
@@ -30,6 +30,39 @@ def test_ce_loss():
     loss_cls = build_loss(loss_cls_cfg)
     assert torch.allclose(loss_cls(fake_pred, fake_label), torch.tensor(200.))
 
+    # test bce_loss
+    cls_score = torch.Tensor([[-200, 100], [500, -1000], [300, -300]])
+    label = torch.Tensor([[1, 0], [0, 1], [1, 0]])
+    weight = torch.Tensor([0.6, 0.4, 0.5])
+    class_weight = torch.tensor([0.1, 0.9])  # class 0: 0.1, class 1: 0.9
+
+    # test bce_loss without class weight
+    loss_cfg = dict(
+        type='CrossEntropyLoss',
+        use_sigmoid=True,
+        reduction='mean',
+        loss_weight=1.0)
+    loss = build_loss(loss_cfg)
+    assert torch.allclose(loss(cls_score, label), torch.tensor(300.))
+    # test ce_loss with weight
+    assert torch.allclose(
+        loss(cls_score, label, weight=weight), torch.tensor(130.))
+
+    # test bce_loss with class weight
+    loss_cfg = dict(
+        type='CrossEntropyLoss',
+        use_sigmoid=True,
+        reduction='mean',
+        loss_weight=1.0,
+        class_weight=class_weight)
+    loss = build_loss(loss_cfg)
+    assert torch.allclose(loss(cls_score, label),
+                          torch.tensor(176.667)), loss(cls_score, label)
+    # test bce_loss with weight
+    assert torch.allclose(
+        loss(cls_score, label, weight=weight), torch.tensor(74.333)), loss(
+            cls_score, label, weight=weight)
+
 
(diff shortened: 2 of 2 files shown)


mzr1996:
I think here we need to @xvjiarui, refers to https://github.com/ElectronicElephant/YOLOv3-mmdetection/commit/f526be06c56d0af212adad506975e8b5ff5bfe17

xvjiarui (approved):
LGTM
Just to make sure there is no BC breaking.

mzr1996:
> LGTM
> Just to make sure there is no BC breaking.

Does Yolov3 need to use `pos_weight`? I notice it's modified in `YOLOv3-mmdetection`

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -104,7 +104,7 @@ def binary_cross_entropy(pred,
     if weight is not None:
         weight = weight.float()
     loss = F.binary_cross_entropy_with_logits(
-        pred, label.float(), pos_weight=class_weight, reduction='none')
+        pred, label.float(), weight=class_weight, reduction='none')
```
AronLin, line 107: Please take a look at the docstring of `F.binary_cross_entropy_with_logits`.
`weight` should be a tensor that matches the input tensor shape. Not the class-aware weight.
`pos_weight` should be a vector with a length equal to the number of classes.

tests/test_metrics/test_losses.py
```diff
@@ -30,6 +30,37 @@ def test_ce_loss():
     loss_cls = build_loss(loss_cls_cfg)
     assert torch.allclose(loss_cls(fake_pred, fake_label), torch.tensor(200.))
 
+    # test bce_loss
+    cls_score = torch.Tensor([[-200, 100], [500, -1000], [300, -300]])
```
AronLin, line 34: `bce_loss` now only supports the input tensor with shape (n, 1).

Ezra-Yu: [yolov3 cls head](https://github.com/open-mmlab/mmdetection/blob/5c7c74d77dfba3663431fe00b6872dfff8bcce1d/configs/yolo/yolov3_d53_mstrain-608_273e_coco.py#L29) uses bce loss, it's [pred input tensor](https://github.com/open-mmlab/mmdetection/blob/5c7c74d77dfba3663431fe00b6872dfff8bcce1d/mmdet/models/dense_heads/yolo_head.py#L494) has  shpae (batchsize, number_booxes, number_class)

mmdet/models/losses/cross_entropy_loss.py
Ezra-Yu: it will broadcasts to (....., C)
Of course, the premise is to convert pred tensor into shape of (..., C). the yolov3 in mmdet has done this transform

Ezra-Yu: pos_weight is a vector with a length equal to the number of classes， but it is different with the classwise weight. Using pos_weight will pos items and neg items loss unblanced.

AronLin:
pred = torch.Tensor([[2, 4],
                         [-2, 1], 
                         [-1, 3]])  # (N, C)
    label = torch.Tensor([[0, 1], 
                         [1, 0], 
                         [1, 1]])  # (N, C)
    class_weight = torch.tensor([0.1, 0.9])  # (C, )

    # loss without class_weight
    loss_cls_cfg = dict(
        type='CrossEntropyLoss',
        use_sigmoid=True,
        loss_weight=1.0,
        reduction='none'
    )
    loss_cls = build_loss(loss_cls_cfg)
    origin_loss = loss_cls(pred, label)
    weight_between_classes = loss_cls(pred, label, weight=class_weight)

    # loss with class_weight
    loss_cls_cfg = dict(
        type='CrossEntropyLoss',
        use_sigmoid=True,
        loss_weight=1.0,
        class_weight = torch.tensor([0.1, 0.9]),
        reduction='none'
    )
    loss_cls = build_loss(loss_cls_cfg)
    weight_between_samples = loss_cls(pred, label)

    # output
    origin_loss: tensor([[2.1269, 0.0181],
                         [2.1269, 1.3133],
                         [1.3133, 0.0486]])
    weight_between_classes: tensor([[0.2127, 0.0163],
                                    [0.2127, 1.1819],
                                    [0.1313, 0.0437]])
    weight_between_samples: tensor([[2.1269, 0.0163],
                                    [0.2127, 1.3133],
                                    [0.1313, 0.0437]])

After discussion, we think it is not a bug and we do not need to change the interface of `class_weight`, here are the reasons:
(1). In the original design, `class_weight` is used to adjust the weight between the positive and negative samples of each class. So the `class_weight` is applied to positive samples. 
(2). If we want to adjust the importance between classes, we can use the interface `'weight'` in the `forward` function.
(3). If we approve this PR, it will be hard to implement the function in (1). But the function in (2) is easy to implement now.

Please check whether the above items are in reason, I will close this PR if you agree with me.
Kindly ping @ZwwWayne @xvjiarui @mzr1996 @Ezra-Yu .

PR #5577, opened 2021-07-09
[Feature] Panoptic Two-stage base Segmentor

AronLin:
# Motivation
Support panoptic segmentation base model and PanopticFPN.

# Modification
- Add PanopticTwoStageSegmentor which inherits TwoStageDetector and add PanopticFPN method.
- Add seg_heads and panoptic_fusion_heads.
- Add a config of PanopticFPN.
- Fix evaluation bug in CocoPanopticDataset.

# Desgin
1. As well as the components in TwoStageDetector, Panoptic Segmentor has extra semantic_head and panoptic_fusion_head.
2. Semantic head requires the features from Backbone and Neck, outputs logits with shape (N, C, H, W). C is the number of classes, which need to be specified during the registry.
3. Panoptic Fusion Head is for the post-process of panoptic segmentation. It might be based on the heuristic method or an additional module. `num_things_classes and num_stuff_classes` are the necessary arguments.

# BC-Breaking
This PR has no BC-Breaking. But we will change the position of some components such as `fused_semantic_head`. This will be done with a new PR.

# Result
Results for panoptic_fpn_r50_1x_coco:

|              |    PQ    |     SQ    |    RQ   |
| ---------  |  --------- |  --------  |   ------   |
| ALL       |  40.22  |   77.62  |   49.32 |
|THINGS |  47.60  |   80.79  |   57.35 |
|STUFF   |  29.07  |  72.83   |  37.19  |

PR #5788, opened 2021-08-04
[Enhance] Rename the function `upsample_like` to `interpolate_as`

AronLin:
Actually, this function can not only upsample the input Tensor but also downsample the input.
So we should rename it to reflect its real function.

### DIFF
--- a/mmdet/models/utils/misc.py
+++ b/mmdet/models/utils/misc.py
@@ -1,29 +1,29 @@
 from torch.nn import functional as F
 
 
-def upsample_like(source, target, mode='bilinear', align_corners=False):
-    """Upsample the source to the shape of the target.
+def interpolate_as(source, target, mode='bilinear', align_corners=False):
+    """Down/up samples the source to the shape of the target.
 
-    Upsample the source to the shape of target. The input must be a Tensor,
-    but the target can be a Tensor or a np.ndarray with the shape
+    Down/up samples the source to the shape of target. The input must be a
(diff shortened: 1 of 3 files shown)


mmdet/models/utils/misc.py
```diff
@@ -1,29 +1,29 @@
 from torch.nn import functional as F
 
 
-def upsample_like(source, target, mode='bilinear', align_corners=False):
-    """Upsample the source to the shape of the target.
+def interpolate_as(source, target, mode='bilinear', align_corners=False):
+    """Down/up samples the source to the shape of the target.
```
ZwwWayne, line 5: Interpolate the source

ZwwWayne (approved):

mmdet/models/utils/misc.py
```diff
@@ -1,29 +1,29 @@
 from torch.nn import functional as F
 
 
-def upsample_like(source, target, mode='bilinear', align_corners=False):
-    """Upsample the source to the shape of the target.
+def interpolate_as(source, target, mode='bilinear', align_corners=False):
```
jshilong, line 4:
```
Interpolate the `source` to the shape of the `target`.

The `source` must be a Tensor, but the `target` can be a Tensor or a np.ndarray with the shape
(..., target_h, target_w).
```

jshilong (approved):
Please resolve my comments first

mmdet/models/utils/misc.py
AronLin: ok, I will resolve it in #5577.

PR #5848, opened 2021-08-10
Fix some error of YOLOX

hhaAndroid:


ZwwWayne (approved):

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2193,13 +2191,13 @@ class MixUp:
            Default to (0.5, 1.5).
         flip_ratio (float): Horizontal flip ratio of mixup image.
            Default to 0.5.
-        pad_value (int): Pad value. Default to 114.
+        pad_val (int): Pad value. Default to 114.
```
RangiLyu, line 2194: `Default to 114.` -> `Default: 114. ` Should keep the same format.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2193,13 +2191,13 @@ class MixUp:
            Default to (0.5, 1.5).
```
RangiLyu, line 2191: `Default to` -> `Default: ` Should keep the same format.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -1964,19 +1964,17 @@ class Mosaic:
            image. Default to (640, 640).
         center_ratio_range (Sequence[float]): Center ratio range of mosaic
            output. Default to (0.5, 1.5).
-        pad_value (int): Pad value. Default to 114.
+        pad_val (int): Pad value. Default to 114.
```
RangiLyu, line 1967: `Default to` -> `Default: ` Should keep the same format.

RangiLyu (approved):

PR #5851, opened 2021-08-10
[Feature]: Support export onnx models without post process.

RangiLyu:
# Motivation

Many users ask for supporting export raw models without post-process. This is useful when they want to convert the onnx to other formats such as caffe and ncnn, and implement post-process by themselves.

## Usage

When run `tools/deployment/pytorch2onnx.py`, add `--skip-postprocess`. For example:

```
python tools/deployment/pytorch2onnx.py configs/yolo/yolov3_mobilenetv2_mstrain-416_300e_coco.py \
checkpoints/yolov3_mobilenetv2_mstrain-416_300e_coco_20210718_010823-f68a07b3.pth --output-file \
yolov3_mobilenetv2_no_post.onnx --shape 416 416 --skip-postprocess
```

## Notice

Exporting models without post-process only works on some one-stage models.

### DIFF
--- a/docs/tutorials/pytorch2onnx.md
+++ b/docs/tutorials/pytorch2onnx.md
@@ -59,7 +59,7 @@ python tools/deployment/pytorch2onnx.py \
 - `--verify`: Determines whether to verify the correctness of an exported model. If not specified, it will be set to `False`.
 - `--simplify`: Determines whether to simplify the exported ONNX model. If not specified, it will be set to `False`.
 - `--cfg-options`: Override some settings in the used config file, the key-value pair in `xxx=yyy` format will be merged into config file.
-- `--skip-postprocess`: Determines whether export model without post process. If not specified, it will be set to `False`. Notice: This is an Experimental option. Only work for some single stage models. Users need to implement the post-process by themselves. We do not guarantee the correctness of the exported model.
+- `--skip-postprocess`: Determines whether export model without post process. If not specified, it will be set to `False`. Notice: This is an experimental option. Only work for some single stage models. Users need to implement the post-process by themselves. We do not guarantee the correctness of the exported model.
 
 Example:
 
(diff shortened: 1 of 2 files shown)


nihui:
The great feature comes !

RunningLeon:
@RangiLyu Hi, how do you validate the exported model since it has more than ten outputs and it is only a part of the whole model.

RunningLeon:
> > @RangiLyu Hi, how do you validate the exported model since it has more than ten outputs and it is only a part of the whole model.
> 
> We do not validate the model. This is an optional function and we are not responsible for the model correctness. Maybe we can add some disclaimers in code and readme.

Ok, could remind users  in the doc:
https://github.com/open-mmlab/mmdetection/blob/master/docs/tutorials/pytorch2onnx.md#how-to-convert-models-from-pytorch-to-onnx

docs/tutorials/pytorch2onnx.md
```diff
@@ -59,6 +59,7 @@ python tools/deployment/pytorch2onnx.py \
 - `--verify`: Determines whether to verify the correctness of an exported model. If not specified, it will be set to `False`.
 - `--simplify`: Determines whether to simplify the exported ONNX model. If not specified, it will be set to `False`.
 - `--cfg-options`: Override some settings in the used config file, the key-value pair in `xxx=yyy` format will be merged into config file.
+- `--skip-postprocess`: Determines whether export model without post process. If not specified, it will be set to `False`. Notice: This is an Experimental option. Only work for some single stage models. Users need to implement the post-process by themselves. We do not guarantee the correctness of the exported model.
```
RunningLeon, line 62: Experimental -> experimental

RunningLeon (approved):
LGTM

RangiLyu:
Kindly ping @ZwwWayne

PR #5853, opened 2021-08-10
fix the error of polt map when the logs in training time.

jiangwenj02:
## Motivation

fix the error of polt map when the logs in training time.

## Modification

In training time it appears error "raise ValueError(f"x and y must have same first dimension, but " when you plot the curve about mAP. It is because the epoch is greater than the length of ys, while the length of ys is equal to epoch -1.
Then we add a code "xs = xs[:len(ys)]" that can solve it and does not affect ploting curve in finished training time.


hhaAndroid:
@jiangwenj02 Thank you very much for your contribution. To make it easier for others to understand, can you post an example of the log when the error occurred?

PR #5866, opened 2021-08-12
update digit_version

ZwwWayne:
Following https://github.com/open-mmlab/mmcv/pull/1185 to update digit_version in mmdet


tests/test_utils/test_digit_version.py
```diff
@@ -0,0 +1,20 @@
+from mmdet import digit_version
```
ZwwWayne, line 1: test_digital_version -> test_digit_version

PR #5881, opened 2021-08-13
[Enhance] Refactor the loss function in HTC and SCNet

AronLin:
Now we support `ignore_index` in `CrossEntropyLoss`.

We should refactor the loss function in HTC and SCNet to use our own CrossEntropyLoss instead of PyTorch.

### DIFF
--- a/mmdet/models/roi_heads/mask_heads/fused_semantic_head.py
+++ b/mmdet/models/roi_heads/mask_heads/fused_semantic_head.py
@@ -3,7 +3,7 @@ import torch.nn.functional as F
 from mmcv.cnn import ConvModule
 from mmcv.runner import BaseModule, auto_fp16, force_fp32
 
-from mmdet.models.builder import HEADS
+from mmdet.models.builder import HEADS, build_loss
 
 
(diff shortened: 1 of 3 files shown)


mmdet/models/roi_heads/mask_heads/fused_semantic_head.py
```diff
@@ -30,10 +30,12 @@ def __init__(self,
                  in_channels=256,
                  conv_out_channels=256,
                  num_classes=183,
-                 ignore_label=255,
```
ZwwWayne, line 33: should not delete them immediately. We should keep them and add deprecated warning for this two argument for three month at least

hhaAndroid (commented):
OK

AronLin:
The training results for htc_r50_fpn_1x_coco:

Backbone  |  box AP | mask AP | 
|:---------:|:-------:|:-------:|
| origin |     42.3 |  37.4 |
| now   |     42.1 |  37.5 |

ZwwWayne (approved):

PR #5882, opened 2021-08-13
[Fix] Fix the log error when `IterBasedRunner` is used

haotian-liu:
## Motivation

If the priority of `EvalHook` is higher than `IterTimerHook`, it will cause `KeyError: 'data_time'` (https://github.com/open-mmlab/mmsegmentation/issues/758, https://github.com/open-mmlab/mmcv/issues/1261).
Since the time key will add to the output of `log_buffer` after `IterTimeHook`, the `TextLoggerHook` will print the time and `data_time` at the same time.

This PR is based on https://github.com/open-mmlab/mmsegmentation/pull/766.

This PR will be useful for models that uses `IterBasedRunner`, e.g., yolof.

## Modification

Set the priority of `EvalHook` to `LOW`.

## BC-breaking (Optional)

No

## Use cases (Optional)

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -147,7 +147,10 @@ def train_detector(model,
         eval_cfg = cfg.get('evaluation', {})
         eval_cfg['by_epoch'] = cfg.runner['type'] != 'IterBasedRunner'
         eval_hook = DistEvalHook if distributed else EvalHook
-        runner.register_hook(eval_hook(val_dataloader, **eval_cfg))
+        # In this PR (https://github.com/open-mmlab/mmcv/pull/1193), the
+        # priority of IterTimerHook has been modified from 'NORMAL' to 'LOW'.
+        runner.register_hook(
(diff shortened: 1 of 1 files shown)


mmdet/apis/train.py
```diff
@@ -147,7 +147,10 @@ def train_detector(model,
         eval_cfg = cfg.get('evaluation', {})
         eval_cfg['by_epoch'] = cfg.runner['type'] != 'IterBasedRunner'
         eval_hook = DistEvalHook if distributed else EvalHook
-        runner.register_hook(eval_hook(val_dataloader, **eval_cfg))
+        # In this PR (https://github.com/open-mmlab/mmcv/pull/1193), the
+        # priority of IterTimerHook has been modified from 'NORMAL' to 'LOW'.
```
hhaAndroid, line 151: We should specify why we need to modify.

hhaAndroid (approved):

PR #5925, opened 2021-08-20
fix dynamic shape bug

www516717402:
Cant modify dataset param when dataloader multiprocessing running.

### DIFF
--- a/mmdet/core/hook/sync_random_size_hook.py
+++ b/mmdet/core/hook/sync_random_size_hook.py
@@ -16,7 +16,9 @@ class SyncRandomSizeHook(Hook):
             by 32, and then change the dataset output image size.
             Default: (14, 26).
         img_scale (tuple[int]): Size of input image. Default: (640, 640).
-        interval (int): The interval of change image size. Default: 10.
+        interval (int): The interval of change image size. Default: 1.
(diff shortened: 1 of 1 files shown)


RangiLyu:
Great thanks for helping us find out this bug! We will fix the random size bug and re-benchmark the model as soon as possible!

hhaAndroid:
Please refer to https://github.com/open-mmlab/mmdetection/pull/5932

jshilong:
We can not modify the dataset in the subprocesses in the loop, the modification would work when iter(dataloader)

WXWorkCapture_16282604726933(2)

mmdet/core/hook/sync_random_size_hook.py
```diff
@@ -33,9 +35,9 @@ def __init__(self,
         self.interval = interval
         self.device = device
 
-    def after_train_iter(self, runner):
+    def after_train_epoch(self, runner):
```
hhaAndroid, line 38: Please add some notes: its behavior is different from YOLOX's official implementation, the official is to change the size every fixed iteration interval, and we write it to change the size every fixed epoch. And the reason for the modification is that under the current design, the dataloader multi-process cannot modify the size in real-time.

PR #5932, opened 2021-08-23
[WIP] Update yolox

hhaAndroid:
## Motivation

Based on the latest open-source code, we synchronize and benchmark the code. The changes are as follows:

-  No longer synchronously change the image shape between multiple GPUs but directly use the Resize function. The logic is simpler
-  Delete SyncRandomSizeHook
-  Fix known bugs caused by multiple processes
-  Optimize unnecessary `all_reduce_dict` code
- Compatible with the previous version


mmdet/datasets/dataset_wrappers.py
```diff
@@ -358,10 +358,10 @@ def __getitem__(self, idx):
                 ]
                 results['mix_results'] = mix_results
 
-            if self._dynamic_scale is not None:
-                # Used for subsequent pipeline to automatically change
-                # the output image size. E.g MixUp, Resize.
-                results['scale'] = self._dynamic_scale
+            # if self._dynamic_scale is not None:
```
jshilong, line 361: Should we remove these comments lines?

jshilong (changes requested):

mmdet/datasets/dataset_wrappers.py
hhaAndroid: Yes. Once we have decided on the plan, we can delete it.

mmdet/core/hook/yolox_mode_switch_hook.py
```diff
@@ -15,14 +15,20 @@ class YOLOXModeSwitchHook(Hook):
             training to close the data augmentation and switch to L1 loss.
             Default: 15.
        skip_type_keys (list[str], optional): Sequence of type string to be
-            skip pipeline. Default: ('Mosaic', 'RandomAffine', 'MixUp')
+            skip pipeline. Default: ('Mosaic', 'RandomAffine', 'MixUp').
+       is_multiprocess (bool): Determine whether it is a multi-process.
```
RangiLyu, line 19: It needs to be noted that this variable is related to the multi-process of dataloader. Because dist training is also a kind of multi-process, but when the workers of dataloader can also be set to 0.

mmdet/core/hook/yolox_mode_switch_hook.py
```diff
@@ -32,7 +38,15 @@ def before_train_epoch(self, runner):
         if is_module_wrapper(model):
             model = model.module
         if (epoch + 1) == runner.max_epochs - self.num_last_epochs:
-            runner.logger.info('No mosaic and mixup aug now!')
-            train_loader.dataset.update_skip_type_keys(self.skip_type_keys)
             runner.logger.info('Add additional L1 loss now!')
             model.bbox_head.use_l1 = True
+
+        if self.is_multiprocess:
+            # Due to the multi-process feature, closing the augmentation
```
RangiLyu, line 45: Need to explain in detail why need to do this in comments, or add relevant links.

mmdet/core/utils/dist_utils.py
```diff
@@ -94,20 +94,19 @@ def _get_global_gloo_group():
         return dist.group.WORLD
 
 
-def all_reduce_dict(py_dict, op='sum', group=None, to_float=True):
+def all_reduce_dict(py_dict, op='sum', to_float=True):
```
RangiLyu, line 97: This will cause BC breaking if anyone used this function.

Issue #5917, opened 2021-08-19
Zero proposals for bbox head during training


Hussni-V:
Thanks for your error report and we appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**
When I am training tridentnet on my custom dataset. During the first iteration, the model produces 0 proposals for the box head.
I will be given this error:
`File "/home/jovyan/work/vdetection/mmdetection/mmdet/models/roi_heads/bbox_heads/bbox_head.py", line 113, in forward`
`x = x.view(x.size(0), -1)`
`RuntimeError: cannot reshape tensor of 0 elements into shape [0, -1] because the unspecified dimension size -1 can be any value and is ambiguous`

When I try to print the shape of x it is
`x.shape torch.Size([0, 2048, 1, 1])`

However, this error goes away if I change the backbone depth & init_cfg checkpoint of tridentnet or if I use another dataset.
So I know this is an edge case.

**Reproduction**

1. What command or script did you run?

```none
A placeholder for the command.
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?
3. What dataset did you use?

**Environment**

```
sys.platform: linux
Python: 3.8.8 (default, Feb 24 2021, 21:46:12) [GCC 7.3.0]
CUDA available: True
GPU 0,1: NVIDIA Tesla P100-SXM2-16GB
CUDA_HOME: /usr/local/cuda
NVCC: Build cuda_11.3.r11.3/compiler.29745058_0
GCC: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
PyTorch: 1.9.0+cu111
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.1.2 (Git Hash 98be7e8afa711dc9b66c8ff3504129cb82013cdb)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.1
[...]

PR #5941, opened 2021-08-24
Fix empty proposal errors in the training of some two-stage models

hhaAndroid:
## Motivation

Fix empty proposal errors in the training of some two-stage models.

Resolve: https://github.com/open-mmlab/mmdetection/issues/5917


mmdet/models/roi_heads/scnet_roi_head.py
```diff
@@ -102,7 +102,9 @@ def _bbox_forward(self,
         bbox_head = self.bbox_head[stage]
         bbox_feats = bbox_roi_extractor(
             x[:len(bbox_roi_extractor.featmap_strides)], rois)
-        if self.with_semantic and semantic_feat is not None:
+        # bbox_feats.shape[0] > 0 is mean the number of proposal is not 0.
```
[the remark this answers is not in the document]
jshilong: This would raise an error in DDP because there are parameters not used in forward

jshilong:
Please make sure all parameters should be used in forwarding process

hhaAndroid:
@jshilong Please help deal with the problem of the sparse-rcnn empty proposal.

ZwwWayne:
> @jshilong Please help deal with the problem of the sparse-rcnn empty proposal.

sparse rcnn should not have such a problem because it always has a constant number of proposals？

mmdet/models/roi_heads/bbox_heads/bbox_head.py
```diff
@@ -108,8 +108,12 @@ def custom_accuracy(self):
     @auto_fp16()
     def forward(self, x):
         if self.with_avg_pool:
-            x = self.avg_pool(x)
-        x = x.view(x.size(0), -1)
+            if x.shape[0] > 0:
+                x = self.avg_pool(x)
```
ZwwWayne, line 112: 1. May indicate why we need to do that, e.g., avg_pool does not support empty tensor 
2. may use x.numel() rather than x.shape[0] for consistency

mmdet/models/roi_heads/cascade_roi_head.py
```diff
@@ -271,6 +271,11 @@ def forward_train(self,
                     if self.bbox_head[i].custom_activation:
                         cls_score = self.bbox_head[i].loss_cls.get_activation(
                             cls_score)
+
+                    # Empty proposal.
+                    if cls_score.shape[0] == 0:
+                        break
```
ZwwWayne, line 277: Simply break here might cause different graph during training?

mmdet/models/utils/brick_wrappers.py
```diff
@@ -0,0 +1,50 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+from mmcv.cnn.bricks.wrappers import NewEmptyTensorOp, obsolete_torch_version
+
+if torch.__version__ == 'parrots':
+    TORCH_VERSION = torch.__version__
+else:
+    # torch.__version__ could be 1.3.1+cu92, we only need the first two
+    # for comparison
+    TORCH_VERSION = tuple(int(x) for x in torch.__version__.split('.')[:2])
```
jshilong, line 11: You can use 
https://github.com/open-mmlab/mmdetection/blob/b1f97c1b6d2b3c735d8c5690952264c598c2f206/mmdet/__init__.py#L7

mmdet/models/roi_heads/bbox_heads/bbox_head.py
```diff
@@ -108,8 +108,12 @@ def custom_accuracy(self):
...
+                x = x.view(x.size(0), -1)
+            else:
+                # Empty proposal
+                x = torch.mean(x, dim=(-1, -2))
```
jshilong, line 116: Why here does not be tested

mmdet/models/roi_heads/cascade_roi_head.py
hhaAndroid: This is the refining stage, no gradient is needed. It should not cause deadlock.

mmdet/models/utils/brick_wrappers.py
hhaAndroid: We just follow https://github.com/open-mmlab/mmcv/blob/master/mmcv/cnn/bricks/wrappers.py#L16.  If digit_version is adopted, I need special treatment for parrots.

mmdet/models/roi_heads/bbox_heads/bbox_head.py
hhaAndroid: Because `self.with_avg_pool` is only set to True for certain algorithms, unit tests are not added.

ZwwWayne (approved):

PR #5943, opened 2021-08-24
Allow XML dataset generalization

guigarfr:
## Motivation

Current CustomXML dataset expects a PASCAL VOC-like database (JPEGImages and Annotations directories). There are some datasets that might name this directories differently (images and annotations) or even have the annotations in the same path than the images.

## Modification

Instead of hardcoding the directories for images and annotations, will use a default for a Pascal VOC-like database, but user could redefine this directories via configuration file.

## BC-breaking (Optional)

It's compatible, the default is exactly as before.

## Use cases (Optional)

If a dataset has a structure different than PASCAL VOC directories, users can modify them in the config via `img_subdir` and `ann_subdir` keys.

Example:
```config
data = dict(
    samples_per_gpu=2,
    workers_per_gpu=2,
    train=dict(
        type='XMLDataset',
        classes=[....],
        ann_file=data_root + 'train.txt',
        ann_subdir='images',
        img_prefix=data_root,
        img_subdir='annotations',
        pipeline=train_pipeline),
)
```


mmdet/datasets/xml_style.py
```diff
@@ -18,11 +18,19 @@ class XMLDataset(CustomDataset):
         min_size (int | float, optional): The minimum size of bounding
             boxes in the images. If the size of a bounding box is less than
             ``min_size``, it would be add to ignored field.
+        img_subdir (str): Subdir where images are stored. Default is JPEGImages.
```
ZwwWayne, line 21: 1. can we use `img_prefix` rather than `img_subdir` to do what we want through setting proper values to `img_prefix`?
2. if we can do 1, considering rename ann_subdir -> ann_prefix

guigarfr: We could... but then, img_prefix will be behaving differently than before...

A reimplementation for this class could be done. but my purpose for this PR was just removing the hardcoded parts, so it could be configured, as i had to do this for using it with another dataset.

Let me think about it...

jshilong (approved):
LGTM

RangiLyu (approved):
LGTM

mmdet/datasets/xml_style.py
guigarfr: @ZwwWayne if you think it's useful we can have a look at a little refactor on that part for later on.

BIGWangYuDong (approved):
LGTM

guigarfr:
Should something else be changed @ZwwWayne ?

ZwwWayne (approved):

PR #5969, opened 2021-08-30
[Enhance]: Check dtype in transform unit tests.

RangiLyu:

## Motivation

Add data type check in transform unit test.

## Modification

* Move some common functions into utils.py and add data type check.
* Assert dtype of box is float32.

## BC-breaking (Optional)

None.

### DIFF
--- a/tests/test_data/test_pipelines/test_transform/test_transform.py
+++ b/tests/test_data/test_pipelines/test_transform/test_transform.py
@@ -205,6 +205,7 @@ def test_random_crop():
     h, w, _ = img.shape
     gt_bboxes = create_random_bboxes(8, w, h)
     gt_bboxes_ignore = create_random_bboxes(2, w, h)
+    results['gt_labels'] = np.ones(gt_bboxes.shape[0], dtype=np.int64)
     results['gt_bboxes'] = gt_bboxes
     results['gt_bboxes_ignore'] = gt_bboxes_ignore
     transform = dict(type='RandomCrop', crop_size=(h - 20, w - 20))
@@ -213,6 +214,9 @@ def test_random_crop():
     assert results['img'].shape[:2] == (h - 20, w - 20)
     # All bboxes should be reserved after crop
     assert results['img_shape'][:2] == (h - 20, w - 20)
+    assert results['gt_labels'].shape[0] == results['gt_bboxes'].shape[0]
+    assert results['gt_labels'].dtype == np.int64
+    assert results['gt_bboxes'].dtype == np.float32
     assert results['gt_bboxes'].shape[0] == 8
     assert results['gt_bboxes_ignore'].shape[0] == 2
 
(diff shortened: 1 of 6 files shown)


tests/test_data/test_pipelines/test_transform/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    for key in keys:
+        if isinstance(results[key], (BitmapMasks, PolygonMasks)):
+            assert np.equal(results[key].to_ndarray(),
+                            pipeline_results[key].to_ndarray()).all()
+        else:
+            assert np.equal(results[key], pipeline_results[key]).all()
+            assert results[key].dtype == pipeline_results[key].dtype
+
+
+def check_result_same(results, pipeline_results):
```
AronLin, line 17: Please add docstring for this function.
The function name cannot accurately make people understand its function.

tests/test_data/test_pipelines/test_transform/utils.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import numpy as np
+
+from mmdet.core.mask import BitmapMasks, PolygonMasks
+
+
+def _check_fields(results, pipeline_results, keys):
```
AronLin, line 7: Can you have a better name for this function？

tests/test_data/test_pipelines/test_transform/test_transform.py
```diff
@@ -10,6 +10,7 @@
 
 from mmdet.core.evaluation.bbox_overlaps import bbox_overlaps
 from mmdet.datasets.builder import PIPELINES
+from .utils import create_random_bboxes
```
hhaAndroid, line 13: If relative import is adopted, is it impossible to run unit tests in the current folder?

tests/test_data/test_pipelines/test_transform/test_transform.py
```diff
@@ -227,6 +221,8 @@ def area(bboxes):
 
     assert (area(results['gt_bboxes']) <= area(gt_bboxes)).all()
     assert (area(results['gt_bboxes_ignore']) <= area(gt_bboxes_ignore)).all()
+    assert results['gt_bboxes'].dtype == np.float32
```
hhaAndroid, line 224: gt_labels should also consider asserting.

tests/test_data/test_pipelines/test_transform/test_transform.py
RangiLyu: Import is okay because there is a __init__.py here.
However, some config and image paths use relative paths. So it still needs to run pytest at the root folder of the project.

tests/test_data/test_pipelines/test_transform/test_transform.py
RangiLyu: Did not create dummy labels here.

RangiLyu: I notice that some `gt_labels`'s type is tensor but actually we use np int64 array. I'll change them in this PR.

tests/test_data/test_pipelines/test_transform/test_transform.py
hhaAndroid: That's a bit troublesome, especially for users who want to use IDE directly. In this case, it is acceptable.

hhaAndroid (approved):

ZwwWayne (approved):

Issue #5966, opened 2021-08-29
Yolox training error when using L1 Loss


maxrumi:
Hi, I ran into some problems when training Yolox.
Yolox used extra L1 Loss for the last 15 epochs. I used a customized voc format data set with only one category. 100 epochs were set in the configuration file, and there was no problem with the previous 85 epochs.  But when training for the last 15 epoch, something went wrong. I checked the source code and the asset statement caused the training to go wrong: assert pred.size() == target.size() and target.numel() > 0 .
Here is the error traceback:
![error](https://user-images.githubusercontent.com/17705768/131244847-f21b2716-b302-4638-b2ca-9353fd9fc14c.PNG)
**And this is my yolox-s setting:**

model = dict(
    type='YOLOX',
    backbone=dict(type='CSPDarknet', deepen_factor=0.33, widen_factor=0.5),
    neck=dict(
        type='YOLOXPAFPN',
        in_channels=[128, 256, 512],
        out_channels=128,
        num_csp_blocks=1),
    bbox_head=dict(
        type='YOLOXHead', num_classes=1, in_channels=128, feat_channels=128),
    train_cfg=dict(assigner=dict(type='SimOTAAssigner', center_radius=2.5)),
    test_cfg=dict(score_thr=0.01, nms=dict(type='nms', iou_threshold=0.65)))

img_scale = (640, 640)
dataset_type = 'CustomVOCDataset'
img_norm_cfg = dict(
    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
train_pipeline = [
    dict(type='Mosaic', img_scale=img_scale, pad_val=114.0),
    dict(
        type='RandomAffine',
        scaling_ratio_range=(0.1, 2),
        border=(-img_scale[0] // 2, -img_scale[1] // 2)),
    dict(
        type='MixUp',
        img_scale=img_scale,
        ratio_range=(0.8, 1.6),
        pad_val=114.0),
    dict(
        type='PhotoMetricDistortion',
        brightness_delta=32,
        contrast_range=(0.5, 1.5),
        saturation_range=(0.5, 1.5),
        hue_delta=18),
    dict(type='RandomFlip', flip_ratio=0.5),
    dict(type='Resize', keep_ratio=True),
    dict(type='Pad', pad_to_square=True, pad_val=114.0),
    dict(type='Normalize', **img_norm_cfg),
    dict(type='DefaultFormatBundle'),
[...]

PR #5976, opened 2021-08-30
Fix regression loss does not support empty tensor problem.

hhaAndroid:
## Motivation

Fix YOLOX l1loss assert error when the number of positive samples is 0.

resolve: https://github.com/open-mmlab/mmdetection/issues/5966

## Modification
When the positive sample is 0, the l1loss function returns 0.


hhaAndroid:
Note: In fact, all L1Loss series have this problem, should it be processed directly in the loss function instead of in the YOLOX algorithm?

ZwwWayne:
> Note: In fact, all L1Loss series have this problem, should it be processed directly in the loss function instead of in the YOLOX algorithm?

Such a fix might be better.

RangiLyu (approved):
LGTM

ZwwWayne:
> > Note: In fact, all L1Loss series have this problem, should it be processed directly in the loss function instead of in the YOLOX algorithm?
> 
> Such a fix might be better.

Still suggest fixing it in L1Loss rather than in the head.

RangiLyu (approved):

mmdet/models/losses/iou_loss.py
```diff
@@ -88,6 +91,7 @@ def bounded_iou_loss(pred, target, beta=0.2, eps=1e-3):
                             (target_w + eps))
     loss_dh = 1 - torch.min(target_h / (pred_h + eps), pred_h /
                             (target_h + eps))
+    # view(..., -1) does not work for empty tensor
     loss_comb = torch.stack([loss_dx, loss_dy, loss_dw, loss_dh],
                             dim=-1).view(loss_dx.size(0), -1)
```
ZwwWayne, line 96: We can use flatten rather than view to avoid that

ZwwWayne (approved):

Issue #5981, opened 2021-08-31
YOLOX support FP16


RaymondByc:
**Describe the feature**

when I add the line

> " fp16 = dict(loss_scale=512.) " 

to the config yolox, I encounter a problem, which is shown as follows:

![截图](https://user-images.githubusercontent.com/11907000/131435152-b08f4298-3380-49c7-9fcb-95e4c6408c26.PNG)

Does the YOLOX support FP16?

hhaAndroid:
@RaymondByc Thank you for your feedback, we will fix it as soon as possible.

PR #5983, opened 2021-08-31
YOLOX supports fp16 training

hhaAndroid:
## Motivation

YOLOX supports fp16 training

Resolve: https://github.com/open-mmlab/mmdetection/issues/5981

### DIFF
--- a/mmdet/models/dense_heads/yolox_head.py
+++ b/mmdet/models/dense_heads/yolox_head.py
@@ -1,3 +1,4 @@
+# Copyright (c) OpenMMLab. All rights reserved.
 import math
 
 import torch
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/yolox_head.py
```diff
@@ -1,4 +1,3 @@
-# Copyright (c) OpenMMLab. All rights reserved.
```
RangiLyu, line 1: Why delete this?

hhaAndroid: Oh, Maybe I forgot to sync.

ZwwWayne (approved):

ZwwWayne:
suggest to leave comments in the pr message to indicate 1 whether it can run the training, 2 whether the performance is verified as expected or we leave the benchmark issue in a future pr

hhaAndroid:
> suggest to leave comments in the pr message to indicate 1 whether it can run the training, 2 whether the performance is verified as expected or we leave the benchmark issue in a future pr

ok

ZwwWayne (approved):

PR #6003, opened 2021-09-01
Onestage onnx

jshilong:
## Motivation
The recent ONNX-related development is too quickly, which made the code hard to read, and there is a problem that the return type of the same function such as `_get_bboxes`  in the `DenseHead`.,

## Modification
This PR moves all ONNX related code of the One-Stage model to a new function `onnx_export` in the corresponding class.

## BC-breaking (Optional)
None

PR #6042, opened 2021-09-06
fix yolox to onnx

www516717402:
fix yolox to onnx

### DIFF
--- a/mmdet/models/dense_heads/yolox_head.py
+++ b/mmdet/models/dense_heads/yolox_head.py
@@ -278,7 +278,8 @@ class YOLOXHead(BaseDenseHead, BBoxTestMixin):
             flatten_bboxes[..., :4] /= flatten_bboxes.new_tensor(
                 scale_factors).unsqueeze(1)
 
-        dets = [], labels = []
+        dets = []
+        labels = []
         for img_id in range(len(img_metas)):
             cls_scores = flatten_cls_scores[img_id]
             score_factor = flatten_objectness[img_id]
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Lint failed, could you first fix the format issue?

mmdet/models/dense_heads/yolox_head.py
```diff
@@ -278,16 +278,19 @@ def get_bboxes(self,
             flatten_bboxes[..., :4] /= flatten_bboxes.new_tensor(
                 scale_factors).unsqueeze(1)
 
-        result_list = []
+        dets = [], labels = []
```
sergevkim, line 281: You need to separate this line into two lines without a comma

www516717402: rewrite already.

mmdet/models/dense_heads/yolox_head.py
```diff
@@ -278,16 +278,20 @@ def get_bboxes(self,
             flatten_bboxes[..., :4] /= flatten_bboxes.new_tensor(
                 scale_factors).unsqueeze(1)
 
-        result_list = []
+        dets = []
+        labels = []
         for img_id in range(len(img_metas)):
```
RunningLeon, line 283: Does the exported ONNX of YOLOX support batch inference and dynamic shape features?

ZwwWayne:
Suggest merging the master branch so that you can fully use the CI.

RangiLyu:
We are going to move the onnx export logic out of the get_bboxes function. Would you like to follow the refectoring in #6003 and add an onnx_export function in yolox?

mmdet/models/dense_heads/yolox_head.py
RunningLeon: any feedback here?

mathmanu:
YOLOX is one of the best light weight models for embedded inference, considering the complexity and accuracy.

Now that https://github.com/open-mmlab/mmdetection/pull/6003 is merged, can support for onnx export of YOLOX be added.

PR #6079, opened 2021-09-10
Implementation of YOLACT forward_dummy to enable get_flops

### LINKED ISSUE #5891: Issue about tools\analysis_tools\get_flops.py
When I used the tools/analysis_tools/get_flops.py to compute the Flops and Parameters of YOLACT , I got an error "NotImplementedError".

  File "tools/analysis_tools/get_flops.py", line 81, in <module>
    main()
  File "tools/analysis_tools/get_flops.py", line 71, in main
    flops, params = get_model_complexity_info(model, input_shape)
  File "/home/scsc01/anaconda3/envs/mmlab/lib/python3.8/site-packages/mmcv/cnn/utils/flops_counter.py", line 104, in get_model_complexity_info
    _ = flops_model(batch)
  File "/home/scsc01/anaconda3/envs/mmlab/lib/python3.8/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/home/scsc01/lxx/mmlab/mmdet/models/detectors/yolact.py", line 32, in forward_dummy
    raise NotImplementedError
NotImplementedError
Does the get_flops.py not support the complexity calculation of YOLACT? thank you~~

collinzrj:
## Motivation
To enable the `get_flops` feature on YOLACT, the `forward_dummy` needs to be implemented

## Modification
The `forward_dummy` function is implemented.

Closes #5891

### DIFF
--- a/mmdet/models/detectors/yolact.py
+++ b/mmdet/models/detectors/yolact.py
@@ -30,15 +30,18 @@ class YOLACT(SingleStageDetector):
 
         See `mmdetection/tools/analysis_tools/get_flops.py`
         """
-        img_metas = [{'img_shape': (300, 300), 'ori_shape': (300, 300), 'scale_factor': (1, 1)} for _ in range(len(img))]
+        ## img_metas do not influence the flops calculation
+        dummy_img_metas = [
+            dict(img_shape=(300, 300), ori_shape=(300, 300), scale_factor=(1, 1)) for _ in range(len(img))
+        ]
         rescale = False
         feat = self.extract_feat(img)
         det_bboxes, det_labels, det_coeffs = self.bbox_head.simple_test(
-            feat, img_metas, rescale=rescale)
+            feat, dummy_img_metas, rescale=rescale)
         # det_bboxes will be empty for randomly generated images, generate non-empty bboxes so mask_head can be executed
-        det_bboxes = torch.rand((1, 2, 5)).cuda()
-        det_labels = torch.tensor([[0, 0]]).cuda()
-        det_coeffs = torch.rand((1, 2, 32)).cuda()
+        det_bboxes = torch.rand((1, 1, 5)).cuda()
+        det_labels = torch.tensor([[0]]).cuda()
+        det_coeffs = torch.rand((1, 1, 1)).cuda()
         bbox_results = [
             bbox2result(det_bbox, det_label, self.bbox_head.num_classes)
             for det_bbox, det_label in zip(det_bboxes, det_labels)
@@ -49,7 +52,7 @@ class YOLACT(SingleStageDetector):
             det_bboxes,
             det_labels,
             det_coeffs,
-            img_metas,
+            dummy_img_metas,
             rescale=rescale)
 
         return list(zip(bbox_results, segm_results))
--- a/mmdet/models/detectors/yolact.py
+++ b/mmdet/models/detectors/yolact.py
@@ -30,7 +30,30 @@ class YOLACT(SingleStageDetector):
 
         See `mmdetection/tools/analysis_tools/get_flops.py`
         """
-        raise NotImplementedError
+        img_metas = [{'img_shape': (300, 300), 'ori_shape': (300, 300), 'scale_factor': (1, 1)} for _ in range(len(img))]
+        rescale = False
+        feat = self.extract_feat(img)
+        det_bboxes, det_labels, det_coeffs = self.bbox_head.simple_test(
+            feat, img_metas, rescale=rescale)
+        # det_bboxes will be empty for randomly generated images, generate non-empty bboxes so mask_head can be executed
+        det_bboxes = torch.rand((1, 2, 5)).cuda()
+        det_labels = torch.tensor([[0, 0]]).cuda()
+        det_coeffs = torch.rand((1, 2, 32)).cuda()
+        bbox_results = [
+            bbox2result(det_bbox, det_label, self.bbox_head.num_classes)
+            for det_bbox, det_label in zip(det_bboxes, det_labels)
+        ]
+
+        segm_results = self.mask_head.simple_test(
+            feat,
+            det_bboxes,
+            det_labels,
+            det_coeffs,
+            img_metas,
+            rescale=rescale)
+
+        return list(zip(bbox_results, segm_results))
+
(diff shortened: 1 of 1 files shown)


collinzrj:
@ZwwWayne

mmdet/models/detectors/yolact.py
```diff
@@ -30,7 +30,36 @@ def forward_dummy(self, img):
 
         See `mmdetection/tools/analysis_tools/get_flops.py`
         """
-        raise NotImplementedError
+        # img_metas do not influence the flops calculation
+        dummy_img_metas = [
+            dict(
+                img_shape=(300, 300),
```
ZwwWayne, line 36: It might not be a good idea to hardcode the shape of images in this place. Can we use img.shape?

mmdet/models/detectors/yolact.py
```diff
@@ -25,12 +25,40 @@ def __init__(self,
...
 
-    def forward_dummy(self, img):
+    def forward_dummy(self, imgs):
         """Used for computing network flops.
 
         See `mmdetection/tools/analysis_tools/get_flops.py`
         """
-        raise NotImplementedError
+        # img_metas do not influence the flops calculation
+        dummy_img_metas = [
```
hhaAndroid, line 34: We only need to consider network forward instead of simple_test. You can refer to two-stage:  https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/detectors/two_stage.py#L72

collinzrj: I'll fix that. Is there any requirements for the outputs? Do they need to be real results of the forward? Can I simply generate the outputs?

hhaAndroid: No need for real output, as long as flop can be calculated reasonably.

collinzrj: fixed

mmdet/models/detectors/yolact.py
```diff
@@ -25,12 +25,24 @@ def __init__(self,
         self.segm_head = build_head(segm_head)
         self.mask_head = build_head(mask_head)
 
-    def forward_dummy(self, img):
+    def forward_dummy(self, imgs):
```
hhaAndroid, line 28: imgs -> img

mmdet/models/detectors/yolact.py
```diff
@@ -25,12 +25,24 @@ def __init__(self,
...
-        raise NotImplementedError
+        # img_metas do not influence the flops calculation
+        dummy_img_metas = [
+            dict(
+                img_shape=img.shape, ori_shape=img.shape, scale_factor=(1, 1))
+            for img in imgs
+        ]
+        feat = self.extract_feat(imgs)
+        bbox_outs = self.bbox_head(feat)
+        bboxes = torch.rand((1, 1, 4)).to(imgs.device)
```
hhaAndroid, line 41: This number cannot be set to 1 will cause inaccurate calculations, and should be the same as the actual output number.

collinzrj: It seems that `bboxes` passed to `forward` of `mask_head` is used for cropping the results, so it does not influence the calculation of flops. I have experimented with different shape of `bboxes` and the results are all the same. Or should the cropping of the results also be counted as flops? 
Actually the part of `mask_head`'s forward really contains the forwarding of a network is
```
prototypes = self.protonet(x)
```
Can I create a `forward_dummy` for the `mask_head` that just calls the `protonet`? In this way we do not need to generate `img_meta`, `bboxes` and `det_coeffs`. The problem may be what we output from `forward_dummy` is different from that of `simple_test`, should the outputs of `forward_dummy` represent anything?

collinzrj: @hhaAndroid

hhaAndroid: @collinzrj Just follow `forward_dummy` not including the post-processing part.

collinzrj:
The change does not influence `test_yolo_normal`, is there anything wrong with the CI?

mmdet/models/dense_heads/yolact_head.py
```diff
@@ -663,6 +663,9 @@ def _init_layers(self):
             protonets = protonets[:-1]
         return nn.Sequential(*protonets)
 
+    def forward_dummy(self, x):
+        self.protonet(x)
+
```
RangiLyu, line 668: return prototypes here

tests/test_models/test_forward.py
```diff
@@ -519,6 +519,7 @@ def test_yolact_forward():
         gt_masks=gt_masks,
         return_loss=True)
     assert isinstance(losses, dict)
+    detector.forward_dummy(imgs)
```
RangiLyu, line 522: Add a comment to clarify what this line is testing.
Also, add some assert to test the output.

mmdet/models/detectors/yolact.py
```diff
@@ -30,7 +30,10 @@ def forward_dummy(self, img):
 
         See `mmdetection/tools/analysis_tools/get_flops.py`
         """
-        raise NotImplementedError
+        feat = self.extract_feat(img)
+        bbox_outs = self.bbox_head(feat)
+        proposals = self.mask_head.forward_dummy(feat[0])
```
RangiLyu, line 35: rename `proposals` to `prototypes`

tests/test_models/test_forward.py
collinzrj: I am following example of testing `forward_dummy` in `test_forward.py` `test_sparse_rcnn_forward` L83
```
assert isinstance(losses, dict)
loss, _ = detector._parse_losses(losses)
assert float(loss.item()) > 0
detector.forward_dummy(imgs)
```
What's the requirements of outputs of `forward_dummy`?

collinzrj:
My local lint passed, is there any updates on the docformatter check?

RangiLyu:
> My local lint passed, is there any updates on the docformatter check?

My local lint passed too. Really wired. @hhaAndroid @ZwwWayne any ideas?

collinzrj:
lint passed

collinzrj:
The test failed for something not related to this change

PR #6125, opened 2021-09-17
[Fix] Support seg_pad_val in Pad

AronLin:
In the current codebase, setting the padding value for maps in `seg_fileds` is not supported, and the padding value of the semantic segmentation map is 0 by default. 

In HTC and other tasks that use semantic segmentation branches, the wrong boundary value will be pad.

So, I support `seg_pad_val` in `Pad` to fix this bug and unify the padding value of panoptic segmentation.


AronLin:
I will train a model for HTC and PanopticFPN to ensure that the performance will not drop.

docs/conventions.md
```diff
@@ -72,7 +72,4 @@ If you have customized `RoIHead`, you can refer to the above method to deal with
 
 In MMDetection, we have supported COCO Panoptic dataset. We clarify a few conventions about the implementation of `CocoPanopticDataset` here.
 
-1. In the `gt_semantic_seg`, different from the default setting of MMDetection, the label `0` stands for `VOID` label and the category labels start from `1`.
```
ZwwWayne, line 75: These contents should be kept and indicate their version. So that users using versions before this PR will not be confused.

mmdet/models/seg_heads/base_semantic_head.py
```diff
@@ -21,14 +21,15 @@ def __init__(self,
...
         super(BaseSemanticHead, self).__init__(init_cfg)
         self.loss_seg = build_loss(loss_seg)
         self.num_classes = num_classes
 
     @force_fp32(apply_to=('seg_preds', ))
-    def loss(self, seg_preds, gt_semantic_seg, label_bias=1):
+    def loss(self, seg_preds, gt_semantic_seg):
         """Get the loss of semantic head.
 
         Args:
```
ZwwWayne, line 34: suggest to indicate the value range & meaning in docstring

mmdet/datasets/pipelines/transforms.py
```diff
@@ -585,10 +585,12 @@ def __init__(self,
                  size=None,
                  size_divisor=None,
                  pad_to_square=False,
-                 pad_val=0):
+                 pad_val=0,
```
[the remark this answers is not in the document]
AronLin: use `img`, `masks`, and `seg` to correspond to the function name.

docs/conventions.md
```diff
@@ -72,7 +72,6 @@ If you have customized `RoIHead`, you can refer to the above method to deal with
...
-   It will be fixed after we support setting the padding value of `gt_segmantic_seg`.
-2. In the evaluation, the panoptic result is a map with the same shape as the original image.
-   Each value in the result map has the format of `instance_id * INSTANCE_OFFSET + category_id`.
+1. For mmdet<=2.16.0, the range of foreground and background labels in the `gt_semantic_seg` are different from the default setting of MMDetection. The label `0` stands for `VOID` label and the category labels start from `1`.
```
ZwwWayne, line 75: Since mmdet=2.17.0, the category labels of semantic segmentation start from `0` and label `255` stands for `VOID` for consistency with labels of bounding boxes. To achieve that, the `Pad` pipeline supports setting the padding value for `seg`.

ZwwWayne (approved):

ZwwWayne (approved):

PR #6144, opened 2021-09-22
fix dynamic_shape bug of SyncRandomSizeHook

hhaAndroid:
## Motivation

Fixed the dynamic shape bug of SyncRandomSizeHook following to https://github.com/open-mmlab/mmdetection/pull/5925. When dataloader is enabled for multiple processes, SyncRandomSizeHook cannot changing the size at a fixed iteration interval.

## Modification
Modified to change the size of every fixed epoch interval.

### DIFF
--- a/mmdet/core/hook/sync_random_size_hook.py
+++ b/mmdet/core/hook/sync_random_size_hook.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import random
+import warnings
 
 import torch
 from mmcv.runner import get_dist_info
@@ -17,7 +18,7 @@ class SyncRandomSizeHook(Hook):
             by 32, and then change the dataset output image size.
             Default: (14, 26).
         img_scale (tuple[int]): Size of input image. Default: (640, 640).
-        interval (int): The interval of change image size. Default: 10.
+        interval (int): The epoch interval of change image size. Default: 1.
         device (torch.device | str): device for returned tensors.
             Default: 'cuda'.
     """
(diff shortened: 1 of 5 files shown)


mmdet/core/hook/sync_random_size_hook.py
```diff
@@ -17,26 +18,33 @@ class SyncRandomSizeHook(Hook):
...
             Default: 'cuda'.
     """
 
     def __init__(self,
                  ratio_range=(14, 26),
                  img_scale=(640, 640),
-                 interval=10,
+                 interval=1,
                  device='cuda'):
+        warnings.warn('DeprecationWarning: SyncRandomSizeHook is deprecated. '
```
RangiLyu, line 31: This docstring needs to be polished.

mmdet/core/hook/sync_random_size_hook.py
```diff
@@ -17,26 +18,33 @@ class SyncRandomSizeHook(Hook):
...
+                      'Please use Resize class to achieve similar functions. '
```
RangiLyu, line 32: Resize class -> Resize pipeline. Also need to indicate the specific replacement method. Such as giving an example pipeline:```dict(
        type='Resize',
        img_scale=[(448, 448), (832, 832)],
        multiscale_mode='range',
        keep_ratio=True),```

mmdet/core/hook/sync_random_size_hook.py
```diff
@@ -10,33 +11,47 @@
 @HOOKS.register_module()
 class SyncRandomSizeHook(Hook):
     """Change and synchronize the random image size across ranks, currently
```
RangiLyu, line 13: No longer used. "currently used in YOLOX" should be deleted.

ZwwWayne (approved):

RangiLyu (approved):

PR #6190, opened 2021-09-28
[Fix] Fix a lot of typos

AronLin:
Fix a lot of typos.


mmdet/core/post_processing/matrix_nms.py
```diff
@@ -37,7 +37,7 @@ def mask_matrix_nms(masks,
             - scores (Tensor): Updated scores, has shape (n,).
             - labels (Tensor): Remained labels, has shape (n,).
             - masks (Tensor): Remained masks, has shape (n, w, h).
-            - keep_inds (Tensor): The indexs number of
+            - keep_inds (Tensor): The indices number of
```
BIGWangYuDong, line 40: indexes or indices?

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2022,8 +2022,8 @@ def get_indexes(self, dataset):
             list: indexes.
         """
 
-        indexs = [random.randint(0, len(dataset)) for _ in range(3)]
-        return indexs
+        indices = [random.randint(0, len(dataset)) for _ in range(3)]
```
BIGWangYuDong, line 2025: indexes or indices? seems using indexes

mmdet/models/dense_heads/sabl_retina_head.py
```diff
@@ -600,11 +600,11 @@ def get_bboxes_single(self,
                 bbox_cls_pred.contiguous(),
                 bbox_reg_pred.contiguous()
             ]
-            bboxes, confids = self.bbox_coder.decode(
+            bboxes, confidences = self.bbox_coder.decode(
                 anchors.contiguous(), bbox_preds, max_shape=img_shape)
             mlvl_bboxes.append(bboxes)
             mlvl_scores.append(scores)
-            mlvl_confids.append(confids)
+            mlvl_confids.append(confidences)
```
BIGWangYuDong, line 607: do we also need to change `mlvl_confid` -> `mlvl_confidences`>

mmdet/core/post_processing/matrix_nms.py
AronLin: Indices and indexes are all right.
`indexes` is often used in North American countries, and other regions often use `indices`.

mmdet/datasets/pipelines/transforms.py
AronLin: `indexes` will be better here, because the function name is `get_indexes`. I will fix it.

mmdet/models/dense_heads/sabl_retina_head.py
AronLin: yes

mmdet/core/post_processing/matrix_nms.py
BIGWangYuDong: It is suggested to use indexes due to we used `indexes` below. Or we need to unify

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2022,8 +2022,8 @@ def get_indexes(self, dataset):
             list: indexes.
         """
 
-        indexs = [random.randint(0, len(dataset)) for _ in range(3)]
-        return indexs
+        indexes = [random.randint(0, len(dataset)) for _ in range(3)]
```
BIGWangYuDong, line 2025: We need to unify whether to use `indexes` or `indices`

PR #6268, opened 2021-10-13
Refactor dense_head and speedup

hhaAndroid:
# Motivation

- Unify the post-processing flow of the one-stage object detection algorithm as much as possible
- Speed up most one-stage object detection algorithm testing process


jshilong:
@ZwwWayne  idxes or idxs?
seems we use idxs for a quite long time.

PR #4750, opened 2021-03-11
[Refactor]: Unified parameter initialization

hhaAndroid:
## Motivation
Refactoring model initialization methods based MMCV. For more information, please refer to https://github.com/open-mmlab/mmcv/pull/780

## Dependency

- [x] https://github.com/open-mmlab/mmcv/pull/780
- [x]  https://github.com/open-mmlab/mmcv/pull/884
- [x] https://github.com/open-mmlab/mmcv/pull/893
- [x]  https://github.com/open-mmlab/mmcv/pull/899
- [x]  https://github.com/open-mmlab/mmcv/pull/902
- [x] https://github.com/open-mmlab/mmcv/pull/918
- [x] https://github.com/open-mmlab/mmcv/pull/930
- [x] https://github.com/open-mmlab/mmcv/pull/959
- [x] https://github.com/open-mmlab/mmcv/pull/971

## Models

Models that have been refactored as follow:

- [x] ATSS
- [x] CARAFE 
- [x] Cascade R-CNN
- [x] Faster R-CNN
- [x] DCN
- [x] Double Head
- [x] Dynamic R-CNN
- [x] Empirical Attention
- [x] Foveabox
- [x] FreeAnchor
- [x]  FASF
- [x] GCNet
- [x] GFL
- [x] GN
- [x] GROIE 
- [x] Guided Anchoring
- [x] HRNet
- [x] HTC
- [x] MS R-CNN
- [x] NAS FPN
- [x] PAA
- [x] PAFPN
- [x] PISA
- [x] RepPoints
- [x] ResneSt
- [x] RetinaNet
- [x] GHM
- [x] RPN
- [x] SABL
- [x] SCNet
- [x] TridentNet
- [x] Sparse R-CNN
- [x] VFNet
- [x] YOLOV3
- [x] FPG
- [x] cascade_rpn
- [x] centripetalnet
- [x] cornernet
- [x] detectors
- [x] detr
- [x] nas_fcos
- [x] gn+ws (40.2/36.2) vs (40.6/36.6)
- [x] grid_rcnn
- [x] libra_rcnn
- [x] point_rend
- [x] regnet
- [x] res2net
- [x] ssd
- [x] yolact

All models have been checked and passed！

PR #6279, opened 2021-10-14
[BUG]fix init of convfc head

### LINKED ISSUE #6238: [BUG]bug in ConvFCBBoxHead's init cfg.
https://github.com/open-mmlab/mmdetection/blob/c88509cb9a73d6bd1edcba64eb924d3cf3cfe85d/mmdet/models/roi_heads/bbox_heads/convfc_bbox_head.py#L103

This line will override initializers for fc_cls and fc_reg because they are also nn.Linear. 
Or is it what's intended? But I see the old way to initialize fc_cls and fc_reg is using Normal.

jshilong:
# Motivation
https://github.com/open-mmlab/mmdetection/blob/c88509cb9a73d6bd1edcba64eb924d3cf3cfe85d/mmdet/models/roi_heads/bbox_heads/convfc_bbox_head.py#L103

This line will override initializers for fc_cls and fc_reg because they are also `nn.Linear`. 
which is not consistent with the old way to initialize `fc_cls` and `fc_reg` before #4750.

Closes #6238 

# BC breaking
None

### DIFF
--- a/mmdet/models/roi_heads/bbox_heads/convfc_bbox_head.py
+++ b/mmdet/models/roi_heads/bbox_heads/convfc_bbox_head.py
@@ -97,10 +97,16 @@ class ConvFCBBoxHead(BBoxHead):
                 out_features=out_dim_reg)
 
         if init_cfg is None:
+            # when init_cfg is None,
+            # It has been set to
+            # [[dict(type='Normal', std=0.01, override=dict(name='fc_cls'))],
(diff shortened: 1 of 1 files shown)


jshilong:
> I only see comments added, does this PR fix anything?

I remove `layer='Linear',` from init_cfg which would overwide all linear layers when doing the initilization

PR #6317, opened 2021-10-19
[Fix]: Remove sampling hardcode.

### LINKED ISSUE #5717: [Bug]: Can not using custom loss when determine sampling by loss type.
This hardcode has caused some bugs when using custom loss.

https://github.com/open-mmlab/mmdetection/blob/2295154d0413a5d9df26579c56ef067be30e48b9/mmdet/models/dense_heads/anchor_head.py#L65-L68

RangiLyu:

## Motivation

Remove hardcode of determining sampling by loss type. 

Resolve #5717 [Bug]: Can not using custom loss when determine sampling by loss type.

## Modification

Sampling or not is determined by `sampler` in `train_cfg` now. 

If there is no `sampler` or setting `PseudoSampler` in train_cfg, sampling will be False, otherwise, be True.

And to avoid BC-breaking, when `sampler` is set in `train_cfg`, we still keep the loss type logic and raise a warning. These codes will be removed after several versions.

## BC-breaking (Optional)

Yes. sampling will not be determined by loss type in the future.

Need to check backward compatibility.

## Use cases (Optional)

## Checklist

### DIFF
--- a/docs/tutorials/customize_losses.md
+++ b/docs/tutorials/customize_losses.md
@@ -8,17 +8,38 @@ This tutorial first elaborate the computation pipeline of losses, then give some
 
 Given the input prediction and target, as well as the weights, a loss function maps the input tensor to the final loss scalar. The mapping can be divided into four steps:
 
-1. Get **element-wise** or **sample-wise** loss by the loss kernel function.
+1. Set the sampling method to sample positive and negative samples.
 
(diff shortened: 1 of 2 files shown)


ZwwWayne:
May also update docs https://mmdetection.readthedocs.io/en/latest/tutorials/customize_losses.html.

PR #6344, opened 2021-10-22
[Feature]: Support plot confusion matrix.

RangiLyu:

## Motivation

Many users ask for supporting confusion matrix calculation. This PR implemented this feature.

## Modification

Add a script to calculate and draw confusion matrix.

First, run tools/test.py to save the .pkl detection results.

Then, run

```shell
python tools/analysis_tools/confusion_matrix.py ${CONFIG}  ${DETECTION_RESULTS}  ${SAVE_DIR} --show
```

will get this plot

confution_matrix_example

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

- [ ] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/tools/analysis_tools/confusion_matrix.py
+++ b/tools/analysis_tools/confusion_matrix.py
@@ -29,7 +29,7 @@ def parse_args():
     parser.add_argument(
         '--score-thr',
         type=float,
-        default=0,
+        default=0.3,
         help='score threshold to filter detection bboxes')
     parser.add_argument(
         '--tp-iou-thr',
@@ -89,10 +89,7 @@ def calculate_confusion_matrix(dataset,
         analyze_per_img_dets(confusion_matrix, gt_bboxes, labels, res_bboxes,
                              score_thr, tp_iou_thr, nms_iou_thr)
         prog_bar.update()
-    per_label_sums = confusion_matrix.sum(axis=1)[:, np.newaxis]
-    normalized_matrix = \
-        confusion_matrix.astype('float') / per_label_sums * 100
-    return normalized_matrix
+    return confusion_matrix
 
 
 def analyze_per_img_dets(confusion_matrix,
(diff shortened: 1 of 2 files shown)


tools/analysis_tools/confusion_matrix.py
```diff
@@ -0,0 +1,250 @@
...
+        'in xxx=yyy format will be merged into config file. If the value to '
+        'be overwritten is a list, it should be like key="[a,b]" or key=a,b '
+        'It also allows nested list/tuple values, e.g. key="[(a,b),(c,d)]" '
+        'Note that the quotation marks are necessary and that no white space '
+        'is allowed.')
+    args = parser.parse_args()
+    return args
+
+
+def calculate_confusion_matrix(dataset,
```
ZwwWayne, line 59: I think we can put those functions in the core.utils.conf_matrix, so that users can also draw the confusion matrix through jupyter notebook

tools/analysis_tools/confusion_matrix.py
```diff
@@ -0,0 +1,250 @@
...
+        analyze_per_img_dets(confusion_matrix, gt_bboxes, labels, res_bboxes,
+                             score_thr, tp_iou_thr, nms_iou_thr)
+        prog_bar.update()
+    per_label_sums = confusion_matrix.sum(axis=1)[:, np.newaxis]
+    normalized_matrix = \
+        confusion_matrix.astype('float') / per_label_sums * 100
+    return normalized_matrix
+
+
+def analyze_per_img_dets(confusion_matrix,
```
ZwwWayne, line 98: Can we reuse the TP/FP calculation implemented in mean_ap?

tools/analysis_tools/confusion_matrix.py
RangiLyu: We can support this in the future.

tools/analysis_tools/confusion_matrix.py
RangiLyu: The function in mean_ap only returns the information of TP and FP, but in the confusion matrix, FN and category info are also needed. This implementation can do this in one loop.

RangiLyu:
@hhaAndroid please review this PR

ZwwWayne (approved):

Issue #6339, opened 2021-10-21
Consider set a random seed when seed is not set


HarborYuan:
Hi,

Currently, when seed is not set in the command line, the training process will not do anything with seed. [This](https://github.com/open-mmlab/mmdetection/blob/9874180a015beee874c904357fe4ed4fab4b46a4/tools/train.py#L151) may cause potential error in some cases (when different threads have different seeds). In fact, I have met a very misleading error caused by this.

For example, I guess [the design of group distributed sampler](https://github.com/open-mmlab/mmdetection/blob/9874180a015beee874c904357fe4ed4fab4b46a4/mmdet/datasets/samplers/group_sampler.py#L88) takes this into consideration carefully. We can consider setting a random seed as the generalization of this.

I can raise a PR if you agree this is necessary.

Thanks.

PR #6354, opened 2021-10-24
Random Seed Issue

HarborYuan:

## Motivation

#6339 

## Modification

#6339 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -3,6 +3,7 @@ import argparse
 import copy
 import os
 import os.path as osp
+import random
 import time
 import warnings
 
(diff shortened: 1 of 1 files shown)


HarborYuan:
@hhaAndroid @RangiLyu

RangiLyu:
Thanks for your contribution! We are now assessing the impact of this change.
Here are some brief suggestions:

- When the seed is not set in the args, it will randomly generate a seed and broadcast to all ranks. How about integrating these codes into the `set_random_seed` function in `mmdet/apis/train.py`?
- When `args.seed` is `None`, the `seed` in `cfg` will set to `None`. The hyperparam `cfg.seed` is used in some other places such as the data loader sampler. I think the `args.seed` should also be set to the `random_num` in this PR.

HarborYuan:
Thanks for your reponse.

You can make changes on the PR based on your understanding. By the way, I think there is another thing needs to be considered:
The current codes only support gpu, do we need to support multi process cpu tasks?

RangiLyu:
> You can make changes on the PR based on your understanding. 

Okay, we will take over this PR and do some modifications.

> By the way, I think there is another thing needs to be considered: The current codes only support gpu, do we need to 
> support multi process cpu tasks?

CPU DDP or DP training support is not in our plan.

hhaAndroid:
@HarborYuan @RangiLyu I will follow this PR to continue to modify.

PR #6399, opened 2021-10-28
mmdet: r3_cnn original code

hachreak:
Signed-off-by: Leonardo Rossi <leonardo.rossi@unipr.it>

## Motivation

Original code for the paper: Recursively Refined R-CNN: Instance Segmentation with Self-RoI Rebalancing

## Modification

R3RoIHead is a more generic version of HybridTaskCascadeRoIHead, in the sense that permits to specify which bbox/mask head use at each step.

## BC-breaking (Optional)

It should not break anything.


ZwwWayne:
Hi @hachreak ,
Thanks for your kind contribution. Would you like to fix the lint issue first? You can click the buttons in this section to see the details.
image

hachreak:
> Would you like to fix the lint issue first?

Hi @ZwwWayne , thanks for your comment!
I rebased and I hope to have fixed the issues now (I run `pre-commit run --all-files`). :smile:

mmdet/models/roi_heads/cascade_roi_head.py
```diff
@@ -58,13 +58,11 @@ def init_bbox_head(self, bbox_roi_extractor, bbox_head):
         """
         self.bbox_roi_extractor = ModuleList()
         self.bbox_head = ModuleList()
+        count = len(bbox_head)
```
hhaAndroid, line 61: The name is not good, maybe you can consider `num_bbox_heads`, or other

mmdet/models/roi_heads/mask_heads/r3_mask_head.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+from mmcv.cnn import ConvModule
+
+from mmdet.models.builder import HEADS
+from .fcn_mask_head import FCNMaskHead
+
+
+@HEADS.register_module()
+class R3MaskHead(FCNMaskHead):
```
hhaAndroid, line 10: Missing unit tests.

mmdet/models/roi_heads/mask_heads/r3_mask_head.py
```diff
@@ -0,0 +1,42 @@
...
+                self.conv_out_channels,
+                1,
+                conv_cfg=self.conv_cfg,
+                norm_cfg=self.norm_cfg)
+
+    def forward(self, x, res_feat=None, return_logits=True, return_feat=True):
+        if res_feat is None:
+            res_feat = torch.zeros_like(x)
+
+        res_feat = self.conv_res(res_feat)
```
hhaAndroid, line 27: Raise an error if self.with_conv_res=False.

hhaAndroid:
@hachreak Please add the unit test, because the code coverage drops more.

mmdet/models/roi_heads/cascade_roi_head.py
hachreak: done! :smile:

mmdet/models/roi_heads/mask_heads/r3_mask_head.py
hachreak: done :smile: 
I hope it's enough.

mmdet/models/roi_heads/mask_heads/r3_mask_head.py
hachreak: fixed.

hhaAndroid:
@hachreak Please confirm whether the Cascade Mask RCNN has BC breaking. At the same time, you can simply run an experiment to see if there are any problems. If you need other help, you can contact me.

hachreak:
Hi @hhaAndroid 
I think it should not break anything for Cascade R-CNN and neither for HTC.
But I did't consider Cascade Mask RCNN.
Where is the source code for that model to check?

Actually, to run experiments for me now could be a problem because I don't have enough gpus to start some training. :disappointed:

hachreak:
Hi @hhaAndroid any news for the PR? :smile: 
How could I help to speed up the integration?

ZwwWayne:
1. Conflicts should be resolved.
2. htc_roi_head should not be deleted.
3. cascade_roi_head.py should not be modified. We suggest restricting the modification inside the r3_roi_head.py You can override these functions in cascade_roi_head.py

hachreak:
Hi @ZwwWayne 
thanks to answering :smile: 
 1. of course! I'll do it
 2. actually I'm not deleting but generalize it, because R3 RoI Head is a generalization of HTC RoI Head, where I can select the BBox Head to use each time. I could implement a R3 RoI Head separately but it means doubling 90% of the code. Do you think I should reimplement anyway everything? Maybe I can find another way to do it. let me think about it :smile: 
 3. I needed to make small changes (but without affecting the Cascade behaviour) to permit me to implement the loop. I mean, I needed to be able to select the BBox Head I want during each step of the for loop (the for loop which implement the cascade, and in my case the loop).
Basically, for Cascade it is mandatory to select the next one, in my case I configure it.

PR #6405, opened 2021-10-29
Always map location to cpu when load checkpoint

st9007a:

## Motivation

In the multi-GPU environment, `init_detector()` may have potential memory leak or crash with a device error.

Assume we have a model whose weights was saved in `cuda:0` during training. And we want to load this model on a 2-GPU machine with the following code:

```python
from mmdet import apis

model = apis.init_detector(config, checkpoint, device='cuda:1')
```

See line 42, 43 in `mmdet/apis/inference.py`. `init_detector()` won't remap the location. So, the checkpoint will be loaded to `cuda:0` (The device where checkpoint was saved in during training). 

```python
41     if checkpoint is not None:
42         map_loc = 'cpu' if device == 'cpu' else None
43         checkpoint = load_checkpoint(model, checkpoint, map_location=map_loc)
```

And then, `init_detector()` transfer the weights to the device we specific (`cuda:1`).

```python
52     model.to(device)
```

After `init_detector()` done, we get a model located on `cuda:1`. But `cuda:0` still keep a copy of its weights. So, the gpu memory leak. 

In another case, if we have a model saved in `cuda:1` and we want to load it on a single-GPU machine, this may raise a runtime error because `cuda:1` doesn't exist.

I also traced the code of `init_segmentor()` in mmsegmentation. It always remaps checkpoints to cpu. So, this problem won't happen in mmsegmentation. I think modifying the behavior of `init_detector()` to make it work as `init_segmentor()` can resolve this issue.

## Modification

Always set `map_location` to cpu in `init_detector()` when initialize a model with a checkpoint.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -39,8 +39,7 @@ def init_detector(config, checkpoint=None, device='cuda:0', cfg_options=None):
     config.model.train_cfg = None
     model = build_detector(config.model, test_cfg=config.get('test_cfg'))
     if checkpoint is not None:
-        map_loc = 'cpu' if device == 'cpu' else None
-        checkpoint = load_checkpoint(model, checkpoint, map_location=map_loc)
+        checkpoint = load_checkpoint(model, checkpoint, map_location='cpu')
         if 'CLASSES' in checkpoint.get('meta', {}):
             model.CLASSES = checkpoint['meta']['CLASSES']
         else:


ZwwWayne (approved):

Ezra-Yu (commented):
(Ignore this review)

PR #6457, opened 2021-11-08
[Enchance] Set a random seed when the user does not set a seed.

hhaAndroid:
## Motivation

We follow https://github.com/open-mmlab/mmdetection/pull/6354  to continue to modify.

- When the seed is not set in the args, it will randomly generate a seed and broadcast to all ranks.
- cfg.seed=seed

Reslove: https://github.com/open-mmlab/mmdetection/issues/6339

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -4,16 +4,49 @@ import warnings
 
 import numpy as np
 import torch
+import torch.distributed as dist
 from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
 from mmcv.runner import (HOOKS, DistSamplerSeedHook, EpochBasedRunner,
                          Fp16OptimizerHook, OptimizerHook, build_optimizer,
-                         build_runner)
+                         build_runner, get_dist_info)
 from mmcv.utils import build_from_cfg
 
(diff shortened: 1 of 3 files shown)


HarborYuan:
Thanks for your work @hhaAndroid 

When device were cpu, the broadcast would not work. Maybe we need to consider this situation.

hhaAndroid:
> Thanks for your work @hhaAndroid
> 
> When device were cpu, the broadcast would not work. Maybe we need to consider this situation.
> 
> e.g. add an assertation word_size == 1 when device is with cpu.

If the process group is gloo, the CPU can also be broadcast.

mmdet/apis/train.py
```diff
@@ -16,6 +17,38 @@
...
+        device (str): The device where the seed will be put on.
+            Default to 'cuda'.
+
+    Returns:
+        int: Seed to be used.
+    """
+    if seed is not None:
+        return seed
+
+    # When the seed is not set, unknown behavior may occur.
```
RangiLyu, line 37:
```suggestion
    # Make sure all ranks share the same random seed to prevent some potential bugs.
```

ZwwWayne (approved):

RangiLyu (approved):
LGTM

Issue #6495, opened 2021-11-13
Distributed training will hang if log_vars has different length among GPUs


fingertap:
Similar to the [issue in mmseg](https://github.com/open-mmlab/mmsegmentation/issues/1030). Here I just quote:

> In the `_parse_log` function of the `mmseg.segmentors.base.BaseSegmentor`, it attempts to synchronize the loss values among all GPUs. What happened is that in this loop ([line 194](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/models/segmentors/base.py#L194)):
>
> ```python
> for loss_name, loss_value in log_vars.items():
>    # reduce loss when distributed training
>    if dist.is_available() and dist.is_initialized():
>        loss_value = loss_value.data.clone()
>        dist.all_reduce(loss_value.div_(dist.get_world_size()))
>    log_vars[loss_name] = loss_value.item()
>```
>
>One GPU A does not have a `"roi_acc"` as `loss_name` (suppose it is the last key in `log_vars`). Then this GPU A thinks it has done all its work, and jump out of the loop. Other GPUs with the last `"roi_acc"` will try to call `torch.distributed.all_reduce`, which infinitely waits for the reply from GPU A.

This bug is hard to debug, as it just gets blocked without error messages.

RangiLyu:
Thanks for finding this bug! Would you like to open a PR for MMDetection just like the one in mmseg https://github.com/open-mmlab/mmsegmentation/pull/1035 ?

fingertap:
Of course. Also, do you think I should add support for skipping such loss terms by setting them to `None`? An example can be found [here](https://github.com/open-mmlab/mmsegmentation/pull/1035#:~:text=%40staticmethod%0Adef%20_parse_losses,return%20loss%2C%20log_vars).

RangiLyu:
> Of course. Also, do you think I should add support for skipping such loss terms by setting them to `None`? An example can be found [here](https://github.com/open-mmlab/mmsegmentation/pull/1035#:~:text=%40staticmethod%0Adef%20_parse_losses,return%20loss%2C%20log_vars).

This feature is worth discussing. @hhaAndroid @ZwwWayne Please take a lot.

hhaAndroid:
[...]

PR #6501, opened 2021-11-15
[Fix] Avoid infinite GPU waiting in dist training

fingertap:
See #6495 for details. Here I just add an assertion before reducing the log_vars to ensure every GPU has the same length of log_vars to prevent infinite waiting.


mmdet/models/detectors/base.py
```diff
@@ -198,6 +198,13 @@ def _parse_losses(self, losses):
         loss = sum(_value for _key, _value in log_vars.items()
                    if 'loss' in _key)
 
+        # If the loss_vars has different length, GPUs will wait infinitely
+        if dist.is_available() and dist.is_initialized():
```
hhaAndroid, line 202: You can directly use `reduce_mean` in mmdet/core/utils/dist_utils.py

mmdet/models/detectors/base.py
```diff
@@ -198,6 +198,13 @@ def _parse_losses(self, losses):
...
+            log_var_length = torch.tensor(len(log_vars), device=loss.device)
+            dist.all_reduce(log_var_length)
+            assert log_var_length == len(log_vars) * dist.get_world_size(), \
```
hhaAndroid, line 205: Once an error occurs, print out all the keys in each process to facilitate troubleshooting.

fingertap: does mmdet provide any thread-safe print utilities?

mmdet/models/detectors/base.py
fingertap: Both ways are good

mmdet/models/detectors/base.py
hhaAndroid: No need to think about it, only need to be printed after asset, and keys and rank need to be printed at the same time.

mmdet/models/detectors/base.py
hhaAndroid: It's better to be able to write uniformly.

fingertap: The communication in the [following lines](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/detectors/base.py#L206) are also through `torch.dist`. So I just use a similar way to do this. If you want to use the utilities from mmdet, I suggest a more complete style check.

mmdet/models/detectors/base.py
fingertap: Also, how can other GPUs know that this GPU raises error?

mmdet/models/detectors/base.py
hhaAndroid: Okay

mmdet/models/detectors/base.py
fingertap: It seems that if we want to inform all GPUs about the assertion error, we have to do one more communication among GPUs. If this overhead is Okay, I will add it to this PR.

hhaAndroid: There is no need to distinguish which GPU is wrong, just print all the keys, the user can check the output to determine which key is wrong

`assert Flase, f'{...get_rank(), log_val.keys(), len(log_val.keys())...}'`

fingertap:
Before I could merge, the upstream master need to fix the BC bug as in mmseg.

mmdet/models/detectors/base.py
fingertap: Yes, you are right. At least two GPUs will raise exception, so users can compare the error messages.

hhaAndroid: Yes

RangiLyu:
Hi, we have fixed the compatibility with MMCV in v2.18.1. Now the CI can work normally. Please pull the latest master and re-run the unit tests.

fingertap:
> Hi, we have fixed the compatibility with MMCV in v2.18.1. Now the CI can work normally. Please pull the latest master and re-run the unit tests.

Merged.

RangiLyu:
Lint failed. Please pip install pre-commit and then run  `pre-commit run --all-files` to fix.

RangiLyu (approved):

mmdet/models/detectors/base.py
```diff
@@ -198,6 +198,16 @@ def _parse_losses(self, losses):
         loss = sum(_value for _key, _value in log_vars.items()
                    if 'loss' in _key)
 
+        # If the loss_vars has different length, GPUs will wait infinitely
```
ZwwWayne, line 201: suggest to use rank, word_size = get_dist_info(), then use word_size to decide whether to synchronize

mmdet/models/detectors/base.py
```diff
@@ -198,6 +198,16 @@ def _parse_losses(self, losses):
...
+        if dist.is_available() and dist.is_initialized():
+            log_var_length = torch.tensor(len(log_vars), device=loss.device)
+            dist.all_reduce(log_var_length)
```
ZwwWayne, line 204: You may consider reduce_mean in mmdet, and use assert log_var_length == len(log_vars)

mmdet/models/detectors/base.py
fingertap: No need to broadcast the error. At least two GPUs will have their len(log_vars) different from the mean. So users can compare the error logs to determine the missing loss terms.

hhaAndroid (approved):

ZwwWayne (approved):

OpenMMLab-Coodinator:
Hi Han！First of all, we want to express our gratitude for your significant PR in the MMDet project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

fingertap:
Thanks for your invitation.

PR #6516, opened 2021-11-16
update get_indexes in mixup

zhaoxin111:
## Motivation

When the parameter dataset in MultiImageMixDataset is ConcatDataset/RepeatDataset/ClassBalancedDataset and mixup is added in the pipeline, the get_indexes in mixup will crash since there are no get_ann_info in ConcatDataset/RepeatDataset/ClassBalancedDataset. 

## Modification

change gt_bboxes_i = dataset.get_ann_info(index)['bboxes']
to gt_bboxes_i = dataset[index]['gt_bboxes']

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #6352, opened 2021-10-23
[Enhance] Add 'get_ann_info' to dataset_wrappers.

zhaoxin111:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

add get_ann_info in RepeatDataset and ConcatDataset to support MultiImageMixDataset

## Modification

add get_ann_info in RepeatDataset and ConcatDataset

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #6526, opened 2021-11-17
Add 'get_ann_info' to dataset_wrappers

zhaoxin111:
## Motivation

When the parameter dataset in MultiImageMixDataset is ConcatDataset/RepeatDataset/ClassBalancedDataset and mixup is added in the pipeline, the get_indexes in mixup will crash since there are no get_ann_info in ConcatDataset/RepeatDataset/ClassBalancedDataset.

I submitted a PR before, but it was closed by me. I added a unit test on the basis of the previous PR
Related PR：
https://github.com/open-mmlab/mmdetection/pull/6352
https://github.com/open-mmlab/mmdetection/pull/6516
## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/datasets/dataset_wrappers.py
+++ b/mmdet/datasets/dataset_wrappers.py
@@ -71,8 +71,6 @@ class ConcatDataset(_ConcatDataset):
     def get_ann_info(self, idx):
         """Get annotation of concatenated dataset by index.
 
-        This is needed by MixUp.
-
         Args:
             idx (int): Index of data.
 
--- a/mmdet/datasets/dataset_wrappers.py
+++ b/mmdet/datasets/dataset_wrappers.py
@@ -48,7 +48,7 @@ class ConcatDataset(_ConcatDataset):
 
     def get_cat_ids(self, idx):
         """Get category ids of concatenated dataset by index.
-
+        
         Args:
             idx (int): Index of data.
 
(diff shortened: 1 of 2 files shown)


RangiLyu:
Lint failed. Please install pre-commit and then run pre-commit run --all-files to fix.

zhaoxin111:
> Lint failed. Please install pre-commit and then run pre-commit run --all-files to fix.

Hi, I tried pre_commit, but the pyflakes version is always wrong. I tried to install the 2.2.0 version, but this is still wrong. Do you have any suggestions?
image

mmdet/datasets/dataset_wrappers.py
```diff
@@ -68,6 +68,30 @@ def get_cat_ids(self, idx):
...
+        if idx < 0:
+            if -idx > len(self):
+                raise ValueError(
+                    'absolute value of index should not exceed dataset length')
+            idx = len(self) + idx
+        dataset_idx = bisect.bisect_right(self.cumulative_sizes, idx)
+        if dataset_idx == 0:
+            sample_idx = idx
+        else:
+            sample_idx = idx - self.cumulative_sizes[dataset_idx - 1]
```
ZwwWayne, line 92: line 83-92 can be encapsulated into a function like get_sample_idx and we can use it in many places.

zhaoxin111: This function is only used in two places. Where should this function be placed?

mmdet/datasets/dataset_wrappers.py
```diff
@@ -68,6 +68,30 @@ def get_cat_ids(self, idx):
             sample_idx = idx - self.cumulative_sizes[dataset_idx - 1]
         return self.datasets[dataset_idx].get_cat_ids(sample_idx)
 
+    def get_ann_info(self, idx):
+        """Get annotation of concatenated dataset by index.
+
+        This is needed by MixUp.
```
RangiLyu, line 74: This line is not necessary.

RangiLyu (approved):

RangiLyu:
ping @hhaAndroid @ZwwWayne

hhaAndroid (approved):

ZwwWayne (approved):

PR #6527, opened 2021-11-17
[Fix] add --format-options arg for test to allow --eval and --format_only concurrently

XenonPDash:
## Motivation

Currently, in `tools/test.py`, `--eval` and `--format_only` can not be used at the same time.

The cause is that these two modes (actually `dataset.format_results()` func and `dataset.evaluate()` func) both use `--eval_options` as their additional kwargs, while the args they allowed are not the same.

There is also an assertion to ensure that `--eval` and `--format_only` are not both given.

In my case, I need to obtain the evalution results and a '.json' output file at one run which requires `--eval` and `--format_only` concurrently. Otherwise I have to run test twice.

## Modification

1. A new arg `--format_options` is added to specify kwargs for `--format_only`. `--eval-options` is now only for `--eval`.
2. The assertion to ensure `--eval` and `--format_only` are not both given is removed.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -85,6 +85,12 @@ def parse_args():
         help='custom options for evaluation, the key-value pair in xxx=yyy '
         'format will be kwargs for dataset.evaluate() function')
     parser.add_argument(
+        '--format-options',
+        nargs='+',
+        action=DictAction,
+        help='custom options for format_only, the key-value pair in xxx=yyy '
+        'format will be kwargs for dataset.format_results() function')
+    parser.add_argument(
         '--launcher',
         choices=['none', 'pytorch', 'slurm', 'mpi'],
         default='none',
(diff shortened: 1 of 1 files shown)


AronLin:
Actually, the `format_result` process is included in the evaluation process and the `.json` file will be saved in the temp directory. The evidence is here: https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/coco.py#L416.

If you want to get the `.json` file, just set `--options jsonfile_prefix=the_directory_you_want_to_save_the_file`.
So I will close this PR if you have no further comments in 3 days.

XenonPDash:
Thanks for your reply!

Unfortunately this does not work in my case cause the dataset in my project directly inherits from `CustomDataset` rather then `CocoDataset` so `jsonfile_prefix` is not a valid arg in `evaluate()`. 

But this is obviously out of scope of mmdet, so for now I will restrict these changes in my project and close the PR.

DominicccLUO:
In MMDet 1.1, it's `--format_only`, yet in MMDet 2.12, it's `--format-only`.

OpenMMLab-Assistant005:
Hi @XenonPDash ！We are grateful for your efforts in helping improve this open-source project during your personal time. 

PR #6621, opened 2021-11-29
[Fix]Add an example of combining swin and one-stage models

jshilong:
## Motivation

When combining swin with one-stage models, [a norm_layer will be applied to each stage in `out_indices`](https://github.com/open-mmlab/mmdetection/blob/a7a16afbf2a4bdb4d023094da73d325cb864838b/mmdet/models/backbones/swin.py#L757), however, we usually
set `out_indices` as [0,1,2,3] and `set start_level` as 1 in FPN for all one-stage models, the norm_layer of the first stage would not be used in forward and DDP would raise an error in this case.

## Modification
I add an example for combining swin with one-stage model RetinaNet


configs/swin/retinanet_swin-t-p4-w7_fpn_1x_coco.py
```diff
@@ -0,0 +1,30 @@
+_base_ = [
+    '../_base_/models/retinanet_r50_fpn.py',
+    '../_base_/datasets/coco_detection.py',
+    '../_base_/schedules/schedule_1x.py', '../_base_/default_runtime.py'
+]
+# optimizer
```
hhaAndroid, line 6: The position of the comment is wrong.

RangiLyu:
I think we should also add the notice in the documents.

jshilong:
> I think we should also add the notice in the documents.

Would you mind giving a suggestion where the notice should be added?

jshilong:
> > I think we should also add the notice in the documents.
> 
> Would you mind giving a suggestion about the place where notice should be added?

I add a notice to the readme of Swin Transformer.

jshilong:
Ping @ZwwWayne

ZwwWayne (approved):

PR #6625, opened 2021-11-29
[Fix]fix init when densehead contains dcn

jshilong:
## Motivation

When there is DCN in `DenseHead`, the zero initialization for `conv_offset` would be overwide.

## Modification

Add specific process for `conv_offset` of DCN and DCNv2 in `BaseDenseHead`


mmdet/models/dense_heads/base_dense_head.py
```diff
@@ -14,6 +14,15 @@ class BaseDenseHead(BaseModule, metaclass=ABCMeta):
     def __init__(self, init_cfg=None):
         super(BaseDenseHead, self).__init__(init_cfg)
 
+    def init_weights(self):
+        super(BaseDenseHead, self).init_weights()
+        # avoid init_cfg overwrite the initialization of `conv_offset`
+        for m in self.modules():
+            # DeformConv2dPack, ModulatedDeformConv2dPack
+            if hasattr(m, 'conv_offset'):
```
hhaAndroid, line 22: DCN should not be able to use `init_cfg`.

jshilong: conv_offset is a Conv2D

jshilong: You can reproduce this bug by running NASFCOS

mmdet/models/dense_heads/base_dense_head.py
```diff
@@ -14,6 +14,15 @@ class BaseDenseHead(BaseModule, metaclass=ABCMeta):
...
+                m.conv_offset.weight.data.zero_()
+                m.conv_offset.bias.data.zero_()
```
RangiLyu, line 24: Does `conv_offset` always have bias? Use `constant_init` in mmcv would be better.

hhaAndroid (approved):
LGTM

RangiLyu (approved):
LGTM

ZwwWayne (approved):

PR #6730, opened 2021-12-08
Support `bbox_clip_border` for the augmentations of YOLOX

GT9505:

## Motivation

Support `bbox_clip_border` argument for the data augmentations of YOLOX.

## Modification

Add `bbox_clip_border` argument for `Mosaic`, `MixUp` and `RandomAffine` augmentations. The argument allow the augmentations not to clip the bboxes based on the image border, just like [`Resize`](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/pipelines/transforms.py#L57).

## BC-breaking (Optional)

No BC-breaking

## Use cases (Optional)

Just set `bbox_clip_border=False` in these augmentations to not clip the bboxes.

## Checklist

### DIFF
--- a/mmdet/core/bbox/transforms.py
+++ b/mmdet/core/bbox/transforms.py
@@ -3,6 +3,24 @@ import numpy as np
 import torch
 
 
+def remove_outside_bboxes(bboxes, img_h, img_w):
+    """Remove bboxes that are out of the image boundary.
+
+    Args:
+        bboxes (Tensor): Shape (N, 4).
+        img_h (int): Image height.
+        img_w (int): Image width.
+
+    Returns:
+        Tensor: Index of the remaining bboxes.
+    """
+    inside_inds = bboxes[:, 0] < img_w
+    inside_inds = inside_inds & (bboxes[:, 2] > 0)
+    inside_inds = inside_inds & (bboxes[:, 1] < img_h)
+    inside_inds = inside_inds & (bboxes[:, 3] > 0)
+    return inside_inds
+
+
 def bbox_flip(bboxes, img_shape, direction='horizontal'):
     """Flip bboxes horizontally or vertically.
 
--- a/mmdet/core/bbox/transforms.py
+++ b/mmdet/core/bbox/transforms.py
@@ -4,7 +4,7 @@ import torch
 
 
 def remove_outside_bboxes(bboxes, img_h, img_w):
-    """Remove bboxes that are out of the image boundary.
+    """Remove bboxes that are completely out of the image boundary.
 
     Args:
         bboxes (Tensor): Shape (N, 4).
@@ -14,10 +14,8 @@ def remove_outside_bboxes(bboxes, img_h, img_w):
     Returns:
         Tensor: Index of the remaining bboxes.
     """
-    inside_inds = bboxes[:, 0] < img_w
-    inside_inds = inside_inds & (bboxes[:, 2] > 0)
-    inside_inds = inside_inds & (bboxes[:, 1] < img_h)
-    inside_inds = inside_inds & (bboxes[:, 3] > 0)
+    inside_inds = (bboxes[:, 0] < img_w) & (bboxes[:, 2] > 0) \
+        & (bboxes[:, 1] < img_h) & (bboxes[:, 3] > 0)
     return inside_inds
 
 
--- a/mmdet/core/bbox/transforms.py
+++ b/mmdet/core/bbox/transforms.py
@@ -3,8 +3,8 @@ import numpy as np
 import torch
 
 
-def remove_outside_bboxes(bboxes, img_h, img_w):
-    """Remove bboxes that are completely out of the image boundary.
+def find_inside_bboxes(bboxes, img_h, img_w):
+    """Find bboxes as long as a part of bboxes is inside the image.
 
     Args:
         bboxes (Tensor): Shape (N, 4).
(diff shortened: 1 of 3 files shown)


mmdet/core/bbox/transforms.py
```diff
@@ -3,6 +3,24 @@
...
+        img_w (int): Image width.
+
+    Returns:
+        Tensor: Index of the remaining bboxes.
+    """
+    inside_inds = bboxes[:, 0] < img_w
+    inside_inds = inside_inds & (bboxes[:, 2] > 0)
+    inside_inds = inside_inds & (bboxes[:, 1] < img_h)
+    inside_inds = inside_inds & (bboxes[:, 3] > 0)
+    return inside_inds
```
hhaAndroid, line 21: Can be combined into one condition, it's will more concise.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2099,16 +2103,25 @@ def _mosaic_transform(self, results):
...
+                                                 2 * self.img_scale[1])
+                mosaic_bboxes[:, 1::2] = np.clip(mosaic_bboxes[:, 1::2], 0,
+                                                 2 * self.img_scale[0])
+
             if not self.skip_filter:
                 mosaic_bboxes, mosaic_labels = \
                     self._filter_box_candidates(mosaic_bboxes, mosaic_labels)
 
+        # remove outside bboxes
+        inside_inds = remove_outside_bboxes(mosaic_bboxes,
```
hhaAndroid, line 2119: The original logic of this operation does not exist, and it is unclear whether it will affect performance.

GT9505: In the original logic, all `bboxes` are clipped, which means that all `bboxes` are inside the image. 
Therefore, there is no difference after `remove_outside_bboxes`.

GT9505: The `remove_outside_bboxes` only remove `bboxes` that are completely out of the image boundary.
The `bbox` will be kept as long as a part of it is inside the image.

hhaAndroid: The clip may have data like [0,0,2,2], but I think it’s okay to filter it out.

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2099,16 +2103,25 @@ def _mosaic_transform(self, results):
...
         if len(mosaic_labels) > 0:
             mosaic_bboxes = np.concatenate(mosaic_bboxes, 0)
-            mosaic_bboxes[:, 0::2] = np.clip(mosaic_bboxes[:, 0::2], 0,
-                                             2 * self.img_scale[1])
-            mosaic_bboxes[:, 1::2] = np.clip(mosaic_bboxes[:, 1::2], 0,
-                                             2 * self.img_scale[0])
             mosaic_labels = np.concatenate(mosaic_labels, 0)
 
+            if self.bbox_clip_border:
+                mosaic_bboxes[:, 0::2] = np.clip(mosaic_bboxes[:, 0::2], 0,
```
hhaAndroid, line 2109: Can you add some notes to explain when not bbox_clip_border？

hhaAndroid (approved):

mmdet/core/bbox/transforms.py
```diff
@@ -3,6 +3,22 @@
 import torch
 
 
+def remove_outside_bboxes(bboxes, img_h, img_w):
```
ZwwWayne, line 6: This function does not really remove the outside bboxes, suggest using find_inside_bboxes

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2242,6 +2261,8 @@ class MixUp:
         max_aspect_ratio (float): Aspect ratio of width and height
             threshold to filter bboxes. If max(h/w, w/h) larger than this
             value, the box will be removed. Default: 20.
+        bbox_clip_border (bool, optional): Whether clip the objects outside
```
ZwwWayne, line 2264: Whether to clip

mmdet/core/bbox/transforms.py
GT9505: `find_inside_bboxes` sounds like the `bboxes` are completely inside the image?

GT9505: The function allows a part of `bboxes` is inside the image, meanwhile the other parts are out of boundary.

GT9505: Do we have a better name for the function?

ZwwWayne:
Suggest adding unit tests to test the feature. Furthermore, may indicate the purpose of bbox_clip_border in docstrings rather than in the code comments.

GT9505:
> Suggest adding unit tests to test the feature. Furthermore, may indicate the purpose of bbox_clip_border in docstrings rather than in the code comments.

@ZwwWayne I have tried and found that it's hard to create a sample that meets not clipping the border of bbox for the augmentations in YOLOX. The reason is that there is randomness in these augmentations, and we cann't guarantee the output bbox to be out of boundary no matter the input bbox is/isn't out of boundary.

Therefore, it is hard to add unit tests to test the feature. Considering the simpleness of the feature, I think it is unnecessary.

ZwwWayne (approved):

PR #6732, opened 2021-12-08
Support keeping image ratio in the multi-scale training of YOLOX

GT9505:

## Motivation

Support keeping image ratio in the multi-scale training of YOLOX

## Modification

Modify the `_random_resize()` function in the YOLOX detector.

## BC-breaking (Optional)

No BC-breaking.

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/detectors/yolox.py
+++ b/mmdet/models/detectors/yolox.py
@@ -119,9 +119,10 @@ class YOLOX(SingleStageDetector):
 
         if self.rank == 0:
             size = random.randint(*self._random_size_range)
-            size = (self._size_multiplier * size, self._size_multiplier * size)
-            tensor[0] = size[0]
-            tensor[1] = size[1]
+            scale_factor = float(
+                self._default_input_size[1]) / self._default_input_size[0]
+            size = (self._size_multiplier * size,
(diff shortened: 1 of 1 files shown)


hhaAndroid (approved):

mmdet/models/detectors/yolox.py
```diff
@@ -119,7 +119,10 @@ def _random_resize(self):
 
         if self.rank == 0:
             size = random.randint(*self._random_size_range)
-            size = (self._size_multiplier * size, self._size_multiplier * size)
+            scale_factor = float(
```
ZwwWayne, line 122: scale_factor has other meanings in mmdet, should use aspect_ratio

ZwwWayne:
Do we need to run a model to ensure performance? I think originally the image ratio is not kept during training but not it is kept. @hhaAndroid may run an experiment to check the performance

GT9505:
> Do we need to run a model to ensure performance? I think originally the image ratio is not kept during training but not it is kept. @hhaAndroid may run an experiment to check the performance

In the orginal YOLOX, the height and width of image are always the same. Therefore, I think it's unnecessary to rerun an experiments.

PR #6734, opened 2021-12-08
Update video_demo.py

GuangjiHuang:
1) wait-time is the wrong name of the variable;

2) the cv.waitKey's parameter delay must be the interger intead of the float datatype.

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/demo/video_demo.py
+++ b/demo/video_demo.py
@@ -19,8 +19,8 @@ def parse_args():
     parser.add_argument('--out', type=str, help='Output video file')
     parser.add_argument('--show', action='store_true', help='Show video')
     parser.add_argument(
-        '--wait-time',
-        type=float,
+        '--wait_time',
+        type=int,
         default=1,
         help='The interval of show (s), 0 is block')
     args = parser.parse_args()


GuangjiHuang:
.

hhaAndroid (approved):

demo/video_demo.py
```diff
@@ -19,8 +19,8 @@ def parse_args():
     parser.add_argument('--out', type=str, help='Output video file')
     parser.add_argument('--show', action='store_true', help='Show video')
     parser.add_argument(
-        '--wait-time',
```
ZwwWayne, line 22: MMDetection use '-' to separate words in command arguments, like resume-from, wait-time to ease the usages. So the change in line 22 should be omitted.

GuangjiHuang: Oh, absolutely, you are right! But, there is a situation that when I see the code "args.resume_from", I don't know what the argument is and then I want to jump to the parser.add_argument(...) for more information about the args.reume_from, so in the vim normal mode, I will key "mm*ggn" to jump to the destination code line. If the using the _ to separate the words, I can easily jump to line of the code parser.add_argument, but I can't do that if using the - to separate words.

ZwwWayne: I see... The original purpose is to ease the typing in the command line, which we think might be more commonly needed. For your case with VIM, you may directly search for 'wait' rather than 'wait_time'?
For now, we also do not wish break the compatibility of arguments actually

demo/video_demo.py
```diff
@@ -19,8 +19,8 @@ def parse_args():
...
-        type=float,
+        '--wait_time',
```
RangiLyu, line 22: All args in mmdet use `-` instead of `_`. It will be converted automatically when parsing args.
```suggestion
        '--wait-time',
```

RangiLyu (changes requested):

PR #6773, opened 2021-12-13
[docs] update docs/conf.py and docs_zh-CN/conf.py to use shared items

Czm369:
## Motivation

The pytorch theme has been updated, so we need to directly configure conf.py to use the sharing menu in the theme

## Modification

Update docs/conf.py and docs_zh-CN/conf.py to use shared items.


hhaAndroid:
@Czm369 Please fix the conflict.

PR #6781, opened 2021-12-14
[Enhance] Support select gpu-ids in non-distribute testing time

BIGWangYuDong:
Add a parameter `gpu_ids` in `test.py`, which can select gpu id in testing time.

Meanwhile, in non-distribute testing time, setting multi gpu ids will occur an error, because MMDataParallel only supports single GPU training. In this PR, if the user set multi gpu ids, it will select the first id to run the code.


hhaAndroid:
@BIGWangYuDong Please complete PR conversation information.

tools/test.py
```diff
@@ -33,6 +33,18 @@ def parse_args():
         action='store_true',
         help='Whether to fuse conv and bn, this will slightly increase'
         'the inference speed')
+    group_gpus = parser.add_mutually_exclusive_group()
```
hhaAndroid, line 36: Please specify the command used in the corresponding document.

BIGWangYuDong:
```python
if args.gpu_ids is not None:
    cfg.gpu_ids = args.gpu_ids
else:
    cfg.gpu_ids = range(1) if args.gpus is None else range(args.gpus)
```
We used gpus or gpu_id here

tools/test.py
```diff
@@ -155,9 +161,20 @@ def main():
...
+        cfg.gpu_ids = args.gpu_ids
+    else:
+        cfg.gpu_ids = range(1)
+
     # init distributed env first, since logger depends on the dist info.
     if args.launcher == 'none':
         distributed = False
+        if len(cfg.gpu_ids) > 1:
+            warnings.warn(
+                'Only supports single GPU in non-distribute testing time.'
```
jshilong, line 174: a warning message is confusing，please polish the description

BIGWangYuDong: change to f'We treat {cfg.gpu_ids} as gpu-ids, and reset to {cfg.gpu_ids[0:1]} as gpu-ids to avoid potential error in non-distribute testing time.

hhaAndroid (approved):

ZwwWayne (approved):

PR #6837, opened 2021-12-20
[Fix] Support setting pad_val in batch collating

AronLin:
Now, the padding value is set to 0 by default, but in `gt_semantic_seg` it needs to be 255.

So I modified the `DefaultFormatBundle` and added `padding_value` in the `DataContainer`.
In batch collating, the `DataContainer` will fill the data with the `padding_value`.

### DIFF
--- a/mmdet/datasets/pipelines/formating.py
+++ b/mmdet/datasets/pipelines/formating.py
@@ -191,10 +191,15 @@ class DefaultFormatBundle:
     Args:
         img_to_float (bool): Whether to force the image to be converted to
             float type. Default: True.
+        pad_val (dict): A dict for padding value when collcate in parallel,
+            the default value is `dict(img=0, masks=0, seg=255)`.
     """
 
-    def __init__(self, img_to_float=True):
+    def __init__(self,
+                 img_to_float=True,
+                 pad_val=dict(img=0, masks=0, seg=255)):
         self.img_to_float = img_to_float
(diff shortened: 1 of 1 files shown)


AronLin:
I will run an `HTC` model and a `Panoptic FPN` model to ensure the performance stability.

mmdet/datasets/pipelines/formating.py
```diff
@@ -191,10 +191,15 @@ class DefaultFormatBundle:
...
         img_to_float (bool): Whether to force the image to be converted to
             float type. Default: True.
+        pad_val (dict): A dict for padding value when collcate in parallel,
+            the default value is `dict(img=0, masks=0, seg=255)`.
     """
 
-    def __init__(self, img_to_float=True):
+    def __init__(self,
+                 img_to_float=True,
+                 pad_val=dict(img=0, masks=0, seg=255)):
```
hhaAndroid, line 200: Please add unit test

AronLin: We have not filled the data with pad_val in DefaultFormatBundle. So it is hard to add the unit test here.

AronLin:
PanopticFPN: 40.1PQ.
HTC: bbox mAP:42.2, segm mAP:37.4.
They are normal.

hhaAndroid (approved):

PR #6955, opened 2022-01-06
Fix typo in formatting

nijkah:
## Motivation

"formating" in mmdet.datasets.pipelines is not an appropriate word.
**formatting** is correct.

## Modification

So I changed its name and added a deprecated warning as in [mmseg](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/datasets/pipelines/formating.py)


nijkah:
I didnt' forked the latest one. I'll request again after fix.

PR #7016, opened 2022-01-17
[Feature] Support CPU training

AronLin:
We support CPU training.

The model is default put on cuda device, we modify this setting so that if there are no cuda devices, the model will be put on cpu.
This PR is based on open-mmlab/mmcv#1621, we should merge this PR after 1621.

Now we can use the CPU to train/debug our model and test our model with batch size >=2. Before running the program we need to export CUDA_VISIBLE_DEVICES=-1  to disable GPU visibility.


tools/train.py
```diff
@@ -109,20 +104,11 @@ def main():
     if args.resume_from is not None:
         cfg.resume_from = args.resume_from
     cfg.auto_resume = args.auto_resume
-    if args.gpu_ids is not None:
-        cfg.gpu_ids = args.gpu_ids
-    else:
-        cfg.gpu_ids = range(1) if args.gpus is None else range(args.gpus)
+    cfg.gpu_ids = [args.gpu_ids]
```
BIGWangYuDong, line 107: `gpu_id`,

tools/test.py
```diff
@@ -161,20 +161,11 @@ def main():
             for ds_cfg in cfg.data.test:
                 ds_cfg.pipeline = replace_ImageToTensor(ds_cfg.pipeline)
 
-    if args.gpu_ids is not None:
-        cfg.gpu_ids = args.gpu_ids
-    else:
-        cfg.gpu_ids = range(1)
+    cfg.gpu_ids = [args.gpu_id]
```
BIGWangYuDong, line 164: Do we need to change `cfg.gpu_ids` -> `cfg.gpu_id`?

tools/test.py
```diff
@@ -161,20 +161,11 @@ def main():
...
 
     # init distributed env first, since logger depends on the dist info.
     if args.launcher == 'none':
         distributed = False
-        if len(cfg.gpu_ids) > 1:
```
BIGWangYuDong, line 172: maybe we should keep this warning, because if we set gpu_ids to [0, 1], we may still meet the error during non-distribute training or testing.

tools/test.py
AronLin: No, it is a list when distributed.

tools/train.py
AronLin: yes

AronLin:
The bug about non-distributed multi-GPU training is removed to #7019

docs/en/1_exist_data_model.md
```diff
@@ -521,6 +540,20 @@ Difference between `resume-from` and `load-from`:
...
+
+```shell
+export CUDA_VISIBLE_DEVICES=-1
+```
+
+And then run the script [above](#training-on-a-single-GPU).
+
+**Note**:
+
+We do not recommend users to use CPU for training because it is too slow. We support this feature to allow users to debug on machines without GPU, which is more convenient.
```
ZwwWayne, line 555: without GPU for convenience.

docs/zh_cn/1_exist_data_model.md
```diff
@@ -495,7 +527,22 @@ evaluation = dict(interval=12)
...
+
+```shell
+export CUDA_VISIBLE_DEVICES=-1
+```
+
+之后运行单 GPU 训练脚本即可。
+
+**注意**：
+
+我们不推荐用户使用 CPU 进行训练，这太过缓慢。我们支持这个功能是为了调试方便，它允许用户在没有 GPU 的机器上进行调试。
```
ZwwWayne, line 542: 我们支持这个功能是为了方便用户在没有 GPU 的机器上进行调试。

docs/en/1_exist_data_model.md
```diff
@@ -521,6 +540,20 @@ Difference between `resume-from` and `load-from`:
...
+
+```shell
+export CUDA_VISIBLE_DEVICES=-1
+```
+
+And then run the script [above](#training-on-a-single-GPU).
+
+**Note**:
+
+We do not recommend users to use CPU for training because it is too slow. We support this feature to allow users to debug on machines without GPU for convenience.
```
gaotongxiao, line 555: Should have used the "note" wrapper
``````
```{note}
```
``````

EasonQYS:
[question] If I want to use mmdet/apis/train.py , how to understand "device_ids=cfg.gpu_ids"?
What should be equel to cfg.gpu_ids?
As I know,  cfg.gpu_ids should be a non-negtive sequence like [0,] or [0,1,]. but not accept [-1,] or -1.

PR #7019, opened 2022-01-17
[Fix] Fix bug in non-distributed multi-gpu training/testing

AronLin:
Since MMDP does not support non-distributed multi-GPU training, --gpus in train.py lost its role, so it is removed, --gpu-ids is changed to --gpu-id, because only one GPU can be specified for non-distribution training and testing.
If the number of GPUs is more than 1, there will be an assertion error in MMDP because MMDP does not support the situation with more than one GPU.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -35,15 +35,10 @@ def parse_args():
         help='whether not to evaluate the checkpoint during training')
     group_gpus = parser.add_mutually_exclusive_group()
     group_gpus.add_argument(
-        '--gpus',
+        '--gpu-id',
         type=int,
-        help='number of gpus to use '
-        '(only applicable to non-distributed training)')
-    group_gpus.add_argument(
-        '--gpu-ids',
-        type=int,
-        nargs='+',
-        help='ids of gpus to use '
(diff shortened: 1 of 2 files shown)


BIGWangYuDong (approved):
LGTM

ZwwWayne (approved):

tools/train.py
```diff
@@ -35,15 +35,10 @@ def parse_args():
         help='whether not to evaluate the checkpoint during training')
     group_gpus = parser.add_mutually_exclusive_group()
     group_gpus.add_argument(
-        '--gpus',
+        '--gpu-id',
         type=int,
-        help='number of gpus to use '
```
ZwwWayne, line 40: We should add deprecation warning of --gpu-ids and --gpus to indicate users that they are deprecated and useless. We should not delete them immediately.

AronLin: ok

ZwwWayne (approved):

gaotongxiao (commented):
Never mind... It should have been posted to mmocr :)

PR #6744, opened 2021-12-09
Add fix on the bbox loss_weight of the PAA head

gkagkos:
## Motivation

I was having a look on the PAA head and I accidentally saw that the loss weight used on the bbox, belongs on the cls. Which I think is wrong. 

## Modification

I changed [this](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/dense_heads/paa_head.py#L247-L252) to use the weight of cls.

PR #7032, opened 2022-01-19
[Fix] Keep the same weights before reassign in the PAA head

zimoqingfeng:
## Motivation

Worried about "Fix wrong bbox loss_weight of the PAA head (#6744)" in MMDetection V2.20.0.

According to the PAA's source code, it seems that the authors calculated the cls_loss and reg_loss with the same weight (always equal to 1.0) before reassignment, and calculated the final loss with different loss_weight

https://github.com/kkhoot/PAA/blob/master/paa_core/modeling/rpn/paa/loss.py#L306
https://github.com/kkhoot/PAA/blob/master/paa_core/modeling/rpn/paa/loss.py#L356

## Modification

Before reassignment, always set loss_weight as 1.0 (maybe it will lead to some confusion for setting all losses as self.loss_cls.loss_weight)

### DIFF
--- a/mmdet/models/dense_heads/paa_head.py
+++ b/mmdet/models/dense_heads/paa_head.py
@@ -241,14 +241,14 @@ class PAAHead(ATSSHead):
             pos_scores,
             pos_label,
             pos_label_weight,
-            avg_factor=self.loss_cls.loss_weight,
+            avg_factor=1.0,
             reduction_override='none')
 
         loss_bbox = self.loss_bbox(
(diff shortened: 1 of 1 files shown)


zimoqingfeng:
@hhaAndroid can you please check the PR ?

hhaAndroid (approved):

zimoqingfeng:
It looks like it's blocked at **codecov**. Is there anything I can do to help?  @hhaAndroid @ZwwWayne

PR #7036, opened 2022-01-19
[Enhance] Add setup multi-processing both in train and test.

RangiLyu:
## Motivation

Add setup multi-processing both in train and test.
Add unit tests.

### DIFF
--- a/mmdet/utils/setup_env.py
+++ b/mmdet/utils/setup_env.py
@@ -1,10 +1,10 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import multiprocessing as mp
 import os
 import platform
 import warnings
 
 import cv2
+import torch.multiprocessing as mp
 
 
 def setup_multi_processes(cfg):
--- a/mmdet/utils/setup_env.py
+++ b/mmdet/utils/setup_env.py
@@ -1,0 +1,47 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import multiprocessing as mp
+import os
+import platform
+import warnings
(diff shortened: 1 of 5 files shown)


mmdet/utils/setup_env.py
```diff
@@ -0,0 +1,47 @@
...
+        mp.set_start_method(mp_start_method, force=True)
+
+    # disable opencv multithreading to avoid system being overloaded
+    opencv_num_threads = cfg.get('opencv_num_threads', 0)
+    cv2.setNumThreads(opencv_num_threads)
+
+    # setup OMP threads
+    # This code is referred from https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py  # noqa
+    if 'OMP_NUM_THREADS' not in os.environ and cfg.data.workers_per_gpu > 1:
+        omp_num_threads = 1
```
ZwwWayne, line 31: can we also add an variable in cfg named omp_num_threads and mkl_num_threads? Because in some other repos like MMOCR, setting it as 1 slows down the training speed.

hhaAndroid (approved):

mmdet/utils/setup_env.py
RangiLyu: These two are environment variables. It would be better to set them in command-line instead of the config.

ZwwWayne: Ideally, we should print this information out in the log. We can merge this PR for now and update log in new version.

ZwwWayne (approved):

PR #7074, opened 2022-01-25
[Fix] Fix bbox clamp in pt1.10

RangiLyu:
## Motivation
FCOS and AutoAssign cannot run with PyTorch 1.10. 
Error message:
```
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [2, 4, 20, 20]], which is output 0 of ReluBackward0, is at version 1; expected version 0 instead.
```

After using  `with torch.autograd.set_detect_anomaly(True):` to debug and find out that it was caused by the ReLU in bbox branch.

## Modification

Replace `F.relu(bbox_pred)` with `bbox_pred.clamp(min=0)`

### DIFF
--- a/mmdet/models/dense_heads/autoassign_head.py
+++ b/mmdet/models/dense_heads/autoassign_head.py
@@ -197,7 +197,7 @@ class AutoAssignHead(FCOSHead):
         # scale the bbox_pred of different level
         # float to avoid overflow when enabling FP16
         bbox_pred = scale(bbox_pred).float()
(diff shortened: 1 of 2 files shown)


ZwwWayne (approved):

ZwwWayne:
Several things to be done after merging this PR:
1. Should add comments above the line of code to indicate the reason for the change.
2. CI of PyTorch 1.10.

PR #6716, opened 2021-12-07
[Enhance] Upgrage visualization to custom colors of different classes.

jbwang1997:
## Motivation

To custom colors for objects with different classes.
Figure_1

## Modification

The main changes are in [image.py](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/visualization/image.py)
I also add a new argument named --show-colors in test.py to custom colors in the command line.

PR #7085, opened 2022-01-27
[Fix] Relax the requirement of PALETTE in dataset wrappers

gaotongxiao:
## Motivation

The recent new feature https://github.com/open-mmlab/mmdetection/pull/6716 introduced a new assumption on the `PALETTE` attribute in all dataset wrappers, which breaks external references to these wrappers without such an attribute in their datasets. This PR turns this requirement optional.

## Modification

Dataset wrappers take `PALETTE` from their dataset classes only if available, or None will be used to fill out the attribute otherwise.

## BC-breaking (Optional)

None

### DIFF
--- a/mmdet/datasets/dataset_wrappers.py
+++ b/mmdet/datasets/dataset_wrappers.py
@@ -169,7 +169,7 @@ class RepeatDataset:
         self.dataset = dataset
         self.times = times
         self.CLASSES = dataset.CLASSES
-        self.PALETTE = dataset.PALETTE
+        self.PALETTE = dataset.PALETTE if hasattr(dataset, 'PALETTE') else None
         if hasattr(self.dataset, 'flag'):
             self.flag = np.tile(self.dataset.flag, times)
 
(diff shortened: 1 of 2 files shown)


RangiLyu (approved):

mmdet/datasets/dataset_wrappers.py
```diff
@@ -30,7 +30,8 @@ class ConcatDataset(_ConcatDataset):
     def __init__(self, datasets, separate_eval=True):
         super(ConcatDataset, self).__init__(datasets)
         self.CLASSES = datasets[0].CLASSES
-        self.PALETTE = datasets[0].PALETTE
+        self.PALETTE = datasets[0].PALETTE if hasattr(datasets[0],
+                                                      'PALETTE') else None
```
shinya7y, line 34: `self.PALETTE = getattr(datasets[0], 'PALETTE', None)` work?

hhaAndroid (approved):

ZwwWayne (approved):

PR #7090, opened 2022-01-28
Fix None grad problem during training TOOD by adding SigmoidGeometricMean

Johnson-Wang:

## Motivation

The training of TOOD often encounters None gradient during backpropagation, which would further cause None tensors in the next training step. Some issues in the original repo (https://github.com/fcjian/TOOD/issues/11) might be also due to this error. The problem is caused by the naive implementation of sigmoid geometric mean function ``cls_score = (cls_logits.sigmoid() * cls_prob.sigmoid()).sqrt()``. This output might be 0 if  ``cls_logits`` or ``cls_prob`` is a low negative value, which causes either inf grad of none grad during backpropagation. 

## Modification

A reimplementation of ``SigmoidGeometricMean`` class as an inheritance of ``torch.autograd.Function`` is proposed. The backward function is derived analytically and would avoid and inf or none grad during bp. 
- This modification has little influence on the final results (42.3 mAP after modification vs. 42.4 mAP as reported).
- This modification enables users to train TOOD without ATSS warmup, yet with some performance drop (~41.8 mAP)

## Checklist

- [x] This PR does not involve any function interface change.
- [x] Docstring has been added.

### DIFF
--- a/mmdet/models/utils/misc.py
+++ b/mmdet/models/utils/misc.py
@@ -29,6 +29,9 @@ class SigmoidGeometricMean(Function):
         return grad_x, grad_y
 
 
+sigmoid_geometric_mean = SigmoidGeometricMean.apply
+
+
 def interpolate_as(source, target, mode='bilinear', align_corners=False):
     """Interpolate the `source` to the shape of the `target`.
 
--- a/mmdet/models/utils/misc.py
+++ b/mmdet/models/utils/misc.py
@@ -1,5 +1,32 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from torch.autograd import Function
 from torch.nn import functional as F
+
+
+class SigmoidGeometricMean(Function):
+    """Forward and backward function of geometric mean of two sigmoid
+    functions.
+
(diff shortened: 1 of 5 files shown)


ZwwWayne (approved):

mmdet/models/utils/misc.py
```diff
@@ -1,7 +1,34 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from torch.autograd import Function
 from torch.nn import functional as F
 
 
+class SigmoidGeometricMean(Function):
```
ZwwWayne, line 6: How about we implement an interface named sigmoid_geometric_mean = SigmoidGeometricMean.apply here so that in tood_head we can simply use sigmoid_geometric_mean(xxx)?

jshilong: We have to use `SigmoidGeometricMean.apply` in the high PyTorch version such as  Pytorch 1.10

jshilong (approved):
LGTM

RangiLyu (approved):

ZwwWayne (approved):

Issue #7105, opened 2022-01-30
Error Occurs when running train.py with the "image" field of json file (val2017.json) set empty


HiiiXinyiii:
**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**

+ It's not a critical bug but more like an improvement. 
+ The bug occurs when the "image" field of the json file named val2017.json is empty. So when validating after each train epoch, the index subscript will be out of range. 
+ It's advisable to assert this potential error when an empty dataloader is set or before it's used. Or the alternative plan is not to skip validating if the dataloader is empty.

**Reproduction**

1. The json file(modified from val2017.json) that I used. 

```json
    "licenses": [
        ...
    ],
    "images": [],
    "annotations": [
    {
        ...
    }
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?

- I didn't modified the code and the config. I just extract a smaller dataset from the original coco dataset, including modifying the json file and extracting the pictures corresponding to the json file.

3. What dataset did you use?

+ COCO dataset which is modifed by me as said above

**Environment**

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.

```
sys.platform: linux
Python: 3.9.7 (default, Sep 16 2021, 13:09:58) [GCC 7.5.0]
CUDA available: True
GPU 0: NVIDIA GeForce GTX 1660 Ti
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 10.2, V10.2.89
GCC: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
PyTorch: 1.9.0+cu102
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.1.2 (Git Hash 98be7e8afa711dc9b66c8ff3504129cb82013cdb)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 10.2
[...]

PR #7123, opened 2022-02-07
check error when the "image" field of json file of dataset is empty, which may lead to an empty dataloader

Xinyi-Adv:
#7105

### DIFF
--- a/mmdet/core/evaluation/eval_hooks.py
+++ b/mmdet/core/evaluation/eval_hooks.py
@@ -53,6 +53,7 @@ class EvalHook(BaseEvalHook):
             return
 
         from mmdet.apis import single_gpu_test
+        assert len(self.dataloader) != 0, "empty dataloader, which may be due to the \"image\" field of json file"
         results = single_gpu_test(runner.model, self.dataloader, show=False)
         runner.log_buffer.output['eval_iter_num'] = len(self.dataloader)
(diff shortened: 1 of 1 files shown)


mmdet/core/evaluation/eval_hooks.py
```diff
@@ -53,6 +53,8 @@ def _do_evaluate(self, runner):
             return
 
         from mmdet.apis import single_gpu_test
+        # assert when dataloader is empty, in case that Error Occurs when running train.py with the "image" field of json file (val2017.json) set empty
```
RangiLyu, line 56: Not all task uses json as an annotation file. Try to make the error message and comments be more specific.

mmdet/core/evaluation/eval_hooks.py
```diff
@@ -53,6 +53,8 @@ def _do_evaluate(self, runner):
...
+        assert len(self.dataloader) != 0, "empty dataloader, which may be due to the \"image\" field of json file"
```
RangiLyu, line 57: Adding the assertion directly after building dataloader instead of here would be better.

Xinyi-Adv: ohhh Thank you. I have posted a new PR refering to your advice #7129

hhaAndroid:
Duplicate. I will close it and open it myself if necessary

PR #7129, opened 2022-02-08
check error when the "image" field of json file of dataset is empty, which may lead to an empty dataloader

HiiiXinyiii:
Check error when the "image" field of json file of dataset is empty, which may lead to an empty dataloader.
It could be an empty train dataloader or an empty validation dataloader.
This PR is related to the previous PR  #7123  and the previous issue  #7105

### DIFF
--- a/mmdet/datasets/custom.py
+++ b/mmdet/datasets/custom.py
@@ -100,6 +100,10 @@ class CustomDataset(Dataset):
                 'might cause errors if the path is not a local path. '
                 'Please use MMCV>= 1.3.16 if you meet errors.')
             self.data_infos = self.load_annotations(self.ann_file)
+        # assert an error when the data info is empty
+        assert len(self.data_infos) != 0, \
+            'empty data_infos, ' \
+            'which may be due to the \"image\" field of annotation json file'
 
         if self.proposal_file is not None:
             if hasattr(self.file_client, 'get_local_path'):
--- a/mmdet/datasets/custom.py
+++ b/mmdet/datasets/custom.py
@@ -103,7 +103,7 @@ class CustomDataset(Dataset):
         # assert an error when the data info is empty
         assert len(self.data_infos) != 0, \
             'empty data_infos, ' \
-            'which may be due to the \"image\" field of annotation json file'
+            'Please check your dataset files'
 
         if self.proposal_file is not None:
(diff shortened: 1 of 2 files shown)


RangiLyu:
Please install pre-commit and then run `pre-commit run --all-files` to fix the lint.

RangiLyu:
To solve the review comments, you can directly commit to your branch instead of opening another PR.

Xinyi-Adv:
Thank you. I have corrected it according to your advice.

hhaAndroid:
@Xinyi-Adv Thanks a lot for your contribution, I think assert is best written in `mmdet/datasets/custom.py`.

hhaAndroid:
@Xinyi-Adv do you have any progress？

Xinyi-Adv:
> @Xinyi-Adv do you have any progress？

Yeah, I have updated my repo branch

mmdet/datasets/custom.py
```diff
@@ -100,6 +100,10 @@ def __init__(self,
                 'might cause errors if the path is not a local path. '
                 'Please use MMCV>= 1.3.16 if you meet errors.')
             self.data_infos = self.load_annotations(self.ann_file)
+        # assert an error when the data info is empty
+        assert len(self.data_infos) != 0, \
+            'empty data_infos, ' \
+            'which may be due to the \"image\" field of annotation json file'
```
chhluo, line 106: datasets like voc didn't use json file to save annotation. It's beter to replace it with `'Please check your dataset files.'`

Xinyi-Adv: Thank you for your advice. I have updated it as you suggested.

chhluo: Adding this assert will cause some problem, unit test will fail https://github.com/open-mmlab/mmdetection/blob/7a9bc498d5cc972171ec4f7332afcd70bb50e60e/tests/test_data/test_datasets/test_custom_dataset.py#L21 .In this unit test, we use a empty dataset to test costum classes overriding. One possible solution: use a dataset which contains a sample instead of a empty dataset in unit test, the dataset can be put in `./test/data/`.

hhaAndroid:
@Xinyi-Adv You may have to update the dev branch code because the ci did not pass. Once fixed we will merge.

PR #5868, opened 2021-08-12
[Docs] Remove opencv-python-headless dependency by albumentations

BIGWangYuDong:
Follow https://github.com/open-mmlab/mmpose/pull/833,

By default, installing albumentations by pip has opencv-python-headless dependency even if opencv-python has already been installed. This could cause different opencv versions installed in the environment and lead to unexpected errors.

This PR updates the installation of albumentations to prevent using a wheel distribution of it with opencv-python-headless dependency. See its [docs](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies) for details.

**Notice, ** now we can't use `pip install -v -e .`  normally. Need to resolve the error

PR #7143, opened 2022-02-11
[Fix] update albumentations installed way

BIGWangYuDong:
Continue to follow https://github.com/open-mmlab/mmdetection/pull/5868.

By default, installing `albumentations` by pip has `opencv-python-headless` dependency even if `opencv-python` has already been installed. This could cause different opencv versions installed in the environment and lead to unexpected errors. https://github.com/open-mmlab/mmdetection/pull/5868 has fixed this potential issue, but albumentations has updated their installing way, which changes `pip install albumentations --no-binary imgaug,albumentations` to `pip install -U albumentations --no-binary qudida,albumentations`. See its [docs](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies) for details.

If simply use `pip install albumentations>=0.3.2 --no-binary qudida,albumentations` may cause some error or installed opencv-python-headless simultaneously. so it is suggested to use `pip install -r requirements/albu.txt` or `pip install -U albumentations --no-binary qudida,albumentations` to install albu.

### DIFF
--- a/docs/en/get_started.md
+++ b/docs/en/get_started.md
@@ -134,7 +134,7 @@ Or you can still install MMDetection manually:
     # for LVIS dataset
     pip install git+https://github.com/lvis-dataset/lvis-api.git
     # for albumentations
-    pip install albumentations>=0.3.2 --no-binary imgaug,albumentations
+    pip install -r requirements/albu.txt
     ```
 
 **Note:**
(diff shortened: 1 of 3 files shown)


ZwwWayne (approved):

docs/en/get_started.md
```diff
@@ -134,7 +134,7 @@ Or you can still install MMDetection manually:
     # for LVIS dataset
     pip install git+https://github.com/lvis-dataset/lvis-api.git
     # for albumentations
-    pip install albumentations>=0.3.2 --no-binary imgaug,albumentations
+    pip install -r requirements/albu.txt
```
AronLin, line 137: Why not just change the command directly but use a file?

AronLin (approved):

ZwwWayne (approved):

PR #7150, opened 2022-02-13
Fix get_indexes method of pipelines

nijkah:
## Motivation

get_indexes method in the pipelines has a potential problem if it draws len(dataset).

https://github.com/open-mmlab/mmdetection/blob/98949809b7179fab9391663ee5a4ab5978425f90/mmdet/datasets/dataset_wrappers.py#L408

## Modification
Fix it to len(dataset) -1.

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -2032,7 +2032,7 @@ class Mosaic:
             list: indexes.
         """
 
-        indexes = [random.randint(0, len(dataset)) for _ in range(3)]
+        indexes = [random.randint(0, len(dataset) - 1) for _ in range(3)]
         return indexes
 
     def _mosaic_transform(self, results):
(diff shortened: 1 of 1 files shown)


dvansa:
According to the [numpy docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html) the high value of randint() is excluded.  ie. randint(x,y) already will draw values in [x,y-1]

nijkah:
> According to the [numpy docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html) the high value of randint() is excluded. ie. randint(x,y) already will draw values in [x,y-1]

It uses random.randint from python native package.
https://docs.python.org/3/library/random.html
And it includes a <= x <= b.

dvansa:
@nijkah 
[Looking at the imports](https://github.com/open-mmlab/mmdetection/blob/b99641b6a43b78495dc1358a3f2ea40911f53ecf/mmdet/datasets/pipelines/transforms.py#L10) it should be using the numpy one. 

> from numpy import random

In any case, looks more like an import related issue. Also, don't know if there is a specific criteria to be using `np.random.randint` vs `random.randint` or it's just missing standarization.

nijkah:
@dvansa Oh what a foolish mistake. Thank you for your comments. I'll close it.

dvansa:
@nijkah. Not foolish at all. Using `np.random.randint` and `random.randint` interchangeably where both refer to the same np version is confusing. Thank you for spotting that.

PR #7194, opened 2022-02-18
Add alias LoadImageFromNdarray to LoadImageFromWebcam

### LINKED ISSUE #7171: [Feature Request] Add LoadImageFromNdarray  alias for LoadImageFromWebcam
**Describe the feature**

**Motivation**
LoadImageFromWebcam is not easy to understand how it works.
Actually it transform ndarray to default mmdetection data format.

So I think it is good idea to add alias as LoadImageFromNdarray.
implementation will be like just below
```
# end of the mmdet.datasets.pipelines.py
LoadImageFromNdarray = LoadImageFromWebcam
```

If it seems good idea, I will create the PR.

nijkah:
## Motivation
Closes #7171 
`LoadImageFromWebcam` is just a pipeline adding meta-information from results['img'] which is np.ndarray.
But its name is not easy to understand how it works.

## Modification

Added alias `LoadImageFromNdarray` to `LoadImageFromWebcam`.

## Considerations
Should we deprecate `LoadImageFromWebcam`?


ZwwWayne:
Invite @gaotongxiao to take a look

PR #7241, opened 2022-02-24
update MaskFormer readme and docs

chhluo:
update readme and docs in maskformer

PR #7242, opened 2022-02-24
Add deprecation message for deploy tools

chhluo:
Related PR https://github.com/open-mmlab/mmclassification/pull/697

Motivation
Since [MMDeploy](https://github.com/open-mmlab/mmdeploy) has been released, the deployment tools scattered in each OpenMMLab codebase will be deprecated in future.

Modification
Add deprecation message in deployment tools.
Used package [colorama](https://github.com/tartley/colorama#colored-output)[](https://user-images.githubusercontent.com/28671653/154436568-9b36c248-812a-402f-bef8-5631ed8abd7a.png) to print cross-platform colored text.

BC-breaking (Optional)
None

Use cases (Optional)
And the message would look like it in Ubuntu18.04 terminal:
image

### DIFF
--- a/tools/deployment/test.py
+++ b/tools/deployment/test.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import argparse
+import warnings
 
 import mmcv
 from mmcv import Config, DictAction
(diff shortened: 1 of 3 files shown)


ZwwWayne (approved):

tools/deployment/test.py
```diff
@@ -140,4 +141,16 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    # main()
```
RunningLeon, line 144: this function should not be commented.

chhluo: forgot to uncomment it.......

chhluo: fixed in this pr https://github.com/open-mmlab/mmdetection/pull/7241

PR #7252, opened 2022-02-25
[Fix] Fix Open Images testunit to avoid error in Windows CI

BIGWangYuDong:
1. `csv.writerow` and `csv.writerows` may create blank line in `Windows` system. To avoid this potential error, add `newline=''` while open the file. 
   
   `with open(file_path, 'w') as f` -> `with open(file_path, 'w', newline='') as f`

3. Windows system path is A\b\c, which is different from Linux or macOS. To avoid potential error while splitting the filename, change `.split('/')` to `osp.split()`


mmdet/datasets/openimages.py
```diff
@@ -286,7 +286,7 @@ def get_img_shape(self, metas):
         """Set images original shape into data_infos."""
         assert len(metas) == len(self)
         for i in range(len(metas)):
-            file_name = metas[i].data['filename'].split('/')[-1]
+            file_name = metas[i].data['ori_filename'].split('/')[-1]
```
ZwwWayne, line 289: We should use osp.split? On windows the path is A\b\c

BIGWangYuDong:
Done

tests/test_data/test_datasets/test_openimages_dataset.py
```diff
@@ -14,7 +14,7 @@ def _create_ids_error_oid_csv(
     fake_csv_file,
 ):
     label_description = ['/m/000002', 'Football']
-    with open(label_file, 'w') as f:
+    with open(label_file, 'w', newline='') as f:
```
ZwwWayne, line 17: We should leave comments here indicating it is necessary for running on Windows.

hhaAndroid, line 17: # `newline=''` is used to avoid index error of out of bounds
    # in Windows system

PR #7315, opened 2022-03-03
[Enhance] add cpu_num in cocopanoptic for pq computing

### LINKED ISSUE #7309: Limit the number of process in Panoptic Evaluation
**Describe the feature**
Allow to set maximum number of multi-processing in panoptic evaluation

**Motivation**
In the current evaluation https://github.com/open-mmlab/mmdetection/blob/2733aa961ae98f635c4336e9d88a6f1428b958f9/mmdet/datasets/api_wrappers/panoptic_evaluation.py#L198, panoptic evaluation API count the number of CPUs and launch the same number of multiple processes. This will cost too much memory in some cases. We should allow to set a maximum number of multi-processing in the interface to avoid OOM.

**Related resources**
If there is an official code release or third-party implementations, please also provide the information here, which would be very helpful.

**Additional context**
If you would like to implement the feature and create a PR, please leave a comment here and that would be much appreciated.

chhluo:
resolve #7309 
1. add arg `cpu_num` in  pq_compute_multi_core
2. add arg `cpu_num` and docstring for args in cocopanoptic

### DIFF
--- a/mmdet/datasets/api_wrappers/panoptic_evaluation.py
+++ b/mmdet/datasets/api_wrappers/panoptic_evaluation.py
@@ -171,7 +171,7 @@ def pq_compute_multi_core(matched_annotations_list,
                           pred_folder,
                           categories,
                           file_client=None,
-                          cpu_num=-1):
+                          nproc=32):
     """Evaluate the metrics of Panoptic Segmentation with multithreading.
 
     Same as the function with the same name in `panopticapi`.
@@ -185,8 +185,8 @@ def pq_compute_multi_core(matched_annotations_list,
         categories (str): The categories of the dataset.
         file_client (object): The file client of the dataset. If None,
             the backend will be set to `disk`.
-        cpu_num (int): Number of cpus for panoptic quality computing.
-            Default to -1, which means use all cpus.
+        nproc (int): Number of processes for panoptic quality computing.
+            Default to 32.
     """
     if PQStat is None:
         raise RuntimeError(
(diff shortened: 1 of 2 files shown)


mmdet/datasets/api_wrappers/panoptic_evaluation.py
```diff
@@ -195,7 +198,11 @@ def pq_compute_multi_core(matched_annotations_list,
         file_client_args = dict(backend='disk')
         file_client = mmcv.FileClient(**file_client_args)
 
-    cpu_num = multiprocessing.cpu_count()
+    if cpu_num == -1:
```
ZwwWayne, line 201: 1. cpu_num -> nproc, indicating number of multi-processing. By default, it can be 32.
2. we should set cpu_num = min(multiprocessing.cpu_count(), nproc), so that it can the number of process is suitable for the current number of CPU and avoid exceeding the memory limitation.

mmdet/datasets/coco_panoptic.py
```diff
@@ -140,6 +140,28 @@ class CocoPanopticDataset(CocoDataset):
...
+        proposal_file (str, optional): Path to proposal file. Defaults to None.
+        test_mode (bool, optional): If set True, annotation will not be loaded.
+        filter_empty_gt (bool, optional): If set true, images without bounding
+            boxes of the dataset's classes will be filtered out. This option
+            only works when `test_mode=False`, i.e., we never filter images
+            during tests.
+        file_client_args (dict): Arguments to instantiate a FileClient.
+            See :class:`mmcv.fileio.FileClient` for details.
+            Defaults to ``dict(backend='disk')``.
+        nproc (int): Number of processes for panoptic quality computing.
```
ZwwWayne, line 163: nproc should be the arguments of evaluate rather than __init__.

ZwwWayne (approved):

Issue #7284, opened 2022-02-28
YOLOX: num_last_epochs might not be the exact number what it's supposed to be


HJoonKwon:
Hello. I might have found a small issue, please look into it :) 

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**
While tuning the parameter ```python num_last_epochs``` in ```shell yolox_s_8x8_300e_coco.py```, I found that ```python num_last_epochs ``` is not the exact number what it meant to be? 

Here is the simple example. 
Let's say 
```python 
max_epochs=20
num_last_epochs=5
```
Then ```shell yolox_mode_switch_hook.py``` comes in before training epoch as a hook to turn off augmentation 
at the 14th epoch because the line number 35
```python
if (epoch + 1) == runner.max_epochs - self.num_last_epochs:
```
is satisfied. 
Then during 6 epochs (from 14th to 19th epoch) instead of 5 epochs (which is num_last_epochs), training proceeds without augmentation.
Shouldn't line 35 be like below to ensure num_last_epochs is what it meant to be?
```python
if epoch == runner.max_epochs - self.num_last_epochs:
```
Thank you!

**Bug fix**
If you have already identified the reason, you can provide the information here. If you are willing to create a PR to fix it, please also leave a comment here and that would be much appreciated!

HJoonKwon:
If it is clearly an issue then I can make a PR for you!

PR #7322, opened 2022-03-04
[Fix] Fix num_last_epochs logic for the hooks in YOLOX training

HJoonKwon:
## Motivation

Based on #7284 , I ran a few more experiments with ``` num_last_epochs```. 
I could still see that the ``` before_train_epoch``` hooks didn't come in at the proper epoch. 
They came in one epoch earlier than I expected them to, so I made a PR to fix it. 

On the other hand, the learning rate schedule, which is the learning rate is fixed during the ``` num_last_epochs```, works fine. 
This means that the ``` number of epochs``` for fixed learning rate and the ``` number of epochs``` for non-mixup-mosaic augmentation didn't match as well. 

## Modification

I modified two files, ``` sync_norm_hook.py``` and  ``` yolox_mode_switch_hook.py``` that implement ``` before_train_epoch``` hook using ``` num_last_epochs```. 
In both of the files, the if logic was modified from
```python
        if (epoch + 1) == runner.max_epochs - self.num_last_epochs:
```
to
```python
        if epoch == runner.max_epochs - self.num_last_epochs:
```

## BC-breaking (Optional)

Yes

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/core/hook/sync_norm_hook.py
+++ b/mmdet/core/hook/sync_norm_hook.py
@@ -33,7 +33,7 @@ class SyncNormHook(Hook):
 
     def before_train_epoch(self, runner):
         epoch = runner.epoch
-        if (epoch + 1) == runner.max_epochs - self.num_last_epochs:
+        if epoch == runner.max_epochs - self.num_last_epochs:
             # Synchronize norm every epoch.
             self.interval = 1
 
--- a/mmdet/core/hook/yolox_mode_switch_hook.py
+++ b/mmdet/core/hook/yolox_mode_switch_hook.py
@@ -32,7 +32,7 @@ class YOLOXModeSwitchHook(Hook):
         model = runner.model
         if is_module_wrapper(model):
             model = model.module
-        if (epoch + 1) == runner.max_epochs - self.num_last_epochs:
+        if epoch == runner.max_epochs - self.num_last_epochs:
             runner.logger.info('No mosaic and mixup aug now!')
             # The dataset pipeline cannot be updated when persistent_workers
             # is True, so we need to force the dataloader's multi-process


shinya7y:
The official YOLOX code also adds one more epoch. https://github.com/Megvii-BaseDetection/YOLOX/pull/1152 proposes a fix for the code.
Since the authors have not commented so far, we cannot determine whether it is a bug or an intentional specification.

If it is intentional,
`num_last_epochs=15` -> `num_last_epochs=16` in `yolox_mode_switch_hook.py`, `sync_norm_hook.py`

shinya7y:
This PR makes a BC-breaking change.
If we want to reproduce the settings of the published trained models, we need to modify `yolox_s_8x8_300e_coco.py`.
For users that use their own configs, it is worth considering raising warnings.

HJoonKwon:
@shinya7y The author in the [paper](https://arxiv.org/abs/2107.08430) says 

> We adopt the MixUp and Mosaic implementation in our model and close it for the last 15 epochs

So I think it's a bug but let's wait for the answer. 

What a coincidence that person made a PR 1 day after I published an issue here :)

hhaAndroid:
@HJoonKwon Sorry for the late reply. Although the understanding is a bit wrong, in fact, this is not a bug, this is completely according to the official way of writing. If num_last_epochs=15, it means switching from 300-15

HJoonKwon:
@hhaAndroid Still confused.. Switching from 300-15(in the meta data perspective) means that training closes the augmentation from 285th epoch to 300th epoch. And that is 16 epochs, not 15 epochs. There might be some confusion in the '[writing](https://github.com/open-mmlab/mmdetection/pull/7322#issuecomment-1059692767)' in the paper. But, sure we may wait! :)

nemonameless:
> 

Have you tried it? Are there any differences in the results？

voldemortX:
While the official code still haven't updated. I would suggest that the `+1` be removed from `num_last_epochs` , and setting it to `16` in the configs with a comment that leads to these discussions.

This could be extra confusing as mentioned in #8560, when people expect to tune off this thing through `num_last_epochs=0`.

PR #7360, opened 2022-03-10
[Feature] Add diff seeds to diff ranks.

hhaAndroid:

## Motivation
This demand comes from [mmselfsup](https://github.com/open-mmlab/mmselfsup/pull/228/). In the training of [MAE](https://arxiv.org/pdf/2111.06377.pd), random occlusion operations on the input image blocks are required. In order to ensure that each rank has different randomness, it is necessary to set different seeds for different ranks. 

There is no way to make different models run with different seeds in the previous implementation. So, it is necessary to set different seeds for different ranks at the beginning of the program and synchronize them in the sampler to ensure that the seeds are the same.

MMDetection does not have this demand for the time being, but in order to unify the various opennmmlab repo, we have also made corresponding modifications.

## BC-breaking (Optional)
No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -265,7 +265,7 @@ jobs:
       - name: Build and install
         run: pip install -e .
       - name: Run unittests
-        run: coverage run --branch --source mmdet -m pytest tests -sv
+        run: coverage run --branch --source mmdet -m pytest tests
       - name: Generate coverage report
         run: |
           coverage xml
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Results and models are available in the [model zoo](docs/en/model_zoo.md).
       <td>
         <ul>
           <li><a href="configs/panoptic_fpn">Panoptic FPN (CVPR'2019)</a></li>
-          <li><a href="configs/maskformer">MaskFormer (NeurIPS'2019)</a></li>
+          <li><a href="configs/maskformer">MaskFormer (NeurIPS'2021)</a></li>
         </ul>
       </td>
       <td>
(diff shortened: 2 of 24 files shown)


mmdet/core/utils/dist_utils.py
```diff
@@ -151,3 +152,28 @@ def all_reduce_dict(py_dict, op='sum', group=None, to_float=True):
     if isinstance(py_dict, OrderedDict):
         out_dict = OrderedDict(out_dict)
     return out_dict
+
+
+def shared_random_seed(seed=None, device='cuda'):
```
ZwwWayne, line 157: shared_random_seed -> sync_random_seed

tools/train.py
```diff
@@ -168,7 +170,12 @@ def main():
     logger.info(f'Config:\n{cfg.pretty_text}')
 
     # set random seeds
-    seed = init_random_seed(args.seed)
+    # make sure each rank has a different
+    if args.seed is None:
+        seed = np.random.randint(2**31)
+    else:
+        seed = args.seed
+    seed = seed + dist.get_rank()
```
ZwwWayne, line 178: make it an option and keep the default behavior.

RangiLyu, line 178: I think we still need an option to enable all ranks to use the same seed.

tools/train.py
```diff
@@ -52,6 +53,10 @@ def parse_args():
         help='id of gpu to use '
         '(only applicable to non-distributed training)')
     parser.add_argument('--seed', type=int, default=None, help='random seed')
+    parser.add_argument(
+        '--diff_seed',
```
RangiLyu, line 57:
```suggestion
        '--diff-seed',
```

mmdet/datasets/samplers/infinite_sampler.py
```diff
@@ -48,7 +50,7 @@ def __init__(self,
         self.world_size = world_size
         self.dataset = dataset
         self.batch_size = batch_size
-        self.seed = seed if seed is not None else 0
+        self.seed = sync_random_seed(seed) if seed is not None else 0
```
RangiLyu, line 53:
```suggestion
        self.seed = sync_random_seed(seed)
```

Setting the seed defaults to 0 will reduce randomness after resume. Better to generate a random number as the default seed.

jshilong:
Please check whether this pr would break down the initialization of the dataset in mmfewshot

tools/train.py
```diff
@@ -169,6 +174,7 @@ def main():
 
     # set random seeds
     seed = init_random_seed(args.seed)
```
RangiLyu, line 176: Can this be replaced with sync_random_seed?

jshilong: May `init_random_seed` not be equal to`sync_random_seed`?

mmdet/datasets/samplers/infinite_sampler.py
hhaAndroid: Currently the default value of seed is 0, so if the user does not set it, it will be 0.

hhaAndroid:
> Please check whether this pr would break down the initialization of the dataset in mmfewshot

should not, without  bc breaking.

tools/train.py
hhaAndroid: The logic is not the same, and bc breaking is not expected.

mmdet/datasets/samplers/infinite_sampler.py
hhaAndroid: At present, the program will definitely pass the seed

RangiLyu (approved):

PR #7371, opened 2022-03-11
[Enchance] Add a probability parameter to Mosaic transform

RangeKing:
## Motivation

Add a probability parameter to the mosaic transform

## Modification

Add
```python
if random.uniform(0, 1) > self.prob:
            return results
```
in \_\_call__ of the mosaic transform.

## BC-breaking (Optional)

> Does the modification introduce changes that break the backward-compatibility of the downstream repos?

Not BC-breaking

## Use cases (Optional)

In config file,
```python
train_pipeline = [
    dict(type='Mosaic', img_scale=img_scale, pad_val=114, prob=0.5),
...
```

## Checklist

- [x] 1. Pre-commit or other linting tools are used to fix the potential lint issues.
- [ ] 2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [x] 3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
- [ ] 4. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -1994,7 +1994,8 @@ class Mosaic:
             is True, the filter rule will not be applied, and the
             `min_bbox_size` is invalid. Default to True.
         pad_val (int): Pad value. Default to 114.
-        prob (float): probability of applying this transformation.
+        prob (float): Probability of applying this transformation.
+            Default to 1.0.
     """
 
     def __init__(self,
--- a/mmdet/datasets/pipelines/transforms.py
+++ b/mmdet/datasets/pipelines/transforms.py
@@ -11,6 +11,7 @@ from numpy import random
 
 from mmdet.core import PolygonMasks, find_inside_bboxes
 from mmdet.core.evaluation.bbox_overlaps import bbox_overlaps
+from mmdet.utils import log_img_scale
 from ..builder import PIPELINES
 
 try:
@@ -1979,9 +1980,10 @@ class Mosaic:
 
     Args:
         img_scale (Sequence[int]): Image size after mosaic pipeline of single
-           image. Default to (640, 640).
+            image. The shape order should be (height, width).
+            Default to (640, 640).
         center_ratio_range (Sequence[float]): Center ratio range of mosaic
-           output. Default to (0.5, 1.5).
+            output. Default to (0.5, 1.5).
         min_bbox_size (int | float): The minimum pixel for filtering
             invalid bboxes after the mosaic pipeline. Default to 0.
         bbox_clip_border (bool, optional): Whether to clip the objects outside
@@ -2002,6 +2004,7 @@ class Mosaic:
                  skip_filter=True,
                  pad_val=114):
         assert isinstance(img_scale, tuple)
+        log_img_scale(img_scale, skip_square=True)
         self.img_scale = img_scale
         self.center_ratio_range = center_ratio_range
         self.min_bbox_size = min_bbox_size
@@ -2232,7 +2235,7 @@ class MixUp:
                 |             pad              |
                 +------------------------------+
 
-     The mixup transform steps are as follows::
+     The mixup transform steps are as follows:
 
         1. Another random image is picked by dataset and embedded in
            the top left patch(after padding and resizing)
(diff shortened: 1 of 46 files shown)


mmdet/datasets/pipelines/transforms.py
```diff
@@ -2022,6 +2026,9 @@ def __call__(self, results):
             dict: Result dict with mosaic transformed.
         """
 
+        if random.uniform(0, 1) > self.prob:
```
hhaAndroid, line 2029: Please add unit test to prob.

hhaAndroid:
@RangeKing Hello. do you have any progress?

mmdet/datasets/pipelines/transforms.py
RangeKing: > Please add unit test to prob.

I add an assertion test for invalid `prob` in `tests/test_data/test_pipelines/test_transform/test_transform.py`.
Does the unit test need to set different probabilities like `test_img_augment.py`?
https://github.com/open-mmlab/mmdetection/blob/e045607843f439f064791506257e9b4fdbf268c0/tests/test_data/test_pipelines/test_transform/test_img_augment.py#L138-L142

hhaAndroid:
@RangeKing Please fix the conflicting files

hhaAndroid:
@RangeKing Please fix the lint

mmdet/datasets/pipelines/transforms.py
```diff
@@ -2002,15 +2004,20 @@ def __init__(self,
                  min_bbox_size=0,
                  bbox_clip_border=True,
                  skip_filter=True,
-                 pad_val=114):
+                 pad_val=114,
+                 prob=0.5):
```
hhaAndroid, line 2008:
```suggestion
                 prob=1.):
```

mmdet/datasets/pipelines/transforms.py
hhaAndroid: Yes

RangeKing:
> @RangeKing Please fix the lint

@hhaAndroid I saw that the [lint check](https://github.com/open-mmlab/mmdetection/runs/5600909421?check_suite_focus=true#logs) was successful. But some build checks were not successful. I don't know much about the build error, for example, [build_cuda101 (3.7, 1.7.0+cu101)](https://github.com/open-mmlab/mmdetection/runs/5600910041?check_suite_focus=true#logs). I don't know how to fix it. Any suggestions, please.

mmdet/datasets/pipelines/transforms.py
RangeKing: image

ZwwWayne gave this suggestion that 0.5 might be better.

RangeKing: I also think the default value set to 1 has the least impact.

hhaAndroid: Change it to 1 because there cannot be bc breaking

hhaAndroid:
> 

This is because you modified the logic to cause the unit test assert error.

tests/test_data/test_pipelines/test_transform/test_transform.py
```diff
@@ -913,7 +918,7 @@ def test_mosaic():
     results['gt_labels'] = np.ones(gt_bboxes.shape[0], dtype=np.int64)
     results['gt_bboxes'] = gt_bboxes
     results['gt_bboxes_ignore'] = gt_bboxes_ignore
-    transform = dict(type='Mosaic', img_scale=(10, 12))
+    transform = dict(type='Mosaic', img_scale=(10, 12), prob=1.)
```
hhaAndroid, line 921: No changes required

hhaAndroid (approved):

ZwwWayne (approved):

feixiang7701:
@RangeKing Hi. RandomAffine is often used with mosaic, if set the prob=0.5, the border of RandomAffine need to change together?
-----------------------------------------------------------------------------------------------------------------------------------------
border (tuple[int]): Distance from height and width sides of input image to adjust output shape. Only used in mosaic dataset.
            Default: (0, 0).
-----------------------------------------------------------------------------------------------------------------------------------------

feixiang7701:
def __call__(self, results):
        img = results['img']
        height = img.shape[0] + self.border[0] * 2 
        width = img.shape[1] + self.border[1] * 2 #the height and width should be obtained from pipeline

feixiang7701:
def __call__(self, results):
        img = results['img']
        height = img.shape[0] + self.border[0] * 2 
        width = img.shape[1] + self.border[1] * 2 #the height and width should be obtained from pipeline?

PR #7376, opened 2022-03-11

PeterH0323:
## Motivation

## Modification

- Add `scale_lr` before build_optimizer in function `train_detector`.
- Add `default_gpu_number = 8` and `default_samples_per_gpu = 2` in `configs/_base_/default_runtime.py`.
- Add `mmdet_official_special_gpu_number` or `mmdet_official_special_samples_per_gpu` in the config file which is not using the default setting.

image

- print `original LR` and `scaled LR` to the terminal log when the training start.

4027845ee62c82a23fb1439edc6f7fc

## Checklist

## Need Help:
I don't konw hwo to do the scale LR when using distributed mode. 

pls review, thx 😁   @jbwang1997

### DIFF
--- a/mmdet/apis/train.py
+++ b/mmdet/apis/train.py
@@ -69,13 +69,25 @@ def set_random_seed(seed, deterministic=False):
 
 
 def scale_lr(cfg, logger):
-    """Automatically scaling LR according to GPU number.
+    """Automatically scaling LR according to GPU number and sample per GPU.
 
     Args:
         cfg (config): training config.
         logger (logging.Logger): logger.
     """
-    if len(cfg.gpu_ids) == cfg.default_gpu_number:
+
+    if "mmdet_official_special_samples_per_gpu" in cfg:
+        original_samples_per_gpu = cfg.mmdet_official_special_samples_per_gpu
+    else:
+        original_samples_per_gpu = cfg.default_samples_per_gpu
+
+    if "mmdet_official_special_gpu_number" in cfg:
+        original_gpu_number = cfg.mmdet_official_special_gpu_number
+    else:
+        original_gpu_number = cfg.default_gpu_number
+
+    if len(cfg.gpu_ids) == original_gpu_number and \
+            cfg.data.samples_per_gpu == original_samples_per_gpu:
         logger.info(f'You are using {len(cfg.gpu_ids)} GPU(s) ,'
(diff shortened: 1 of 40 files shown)


mmdet/apis/train.py
```diff
@@ -68,6 +68,54 @@ def set_random_seed(seed, deterministic=False):
...
+
+    if "mmdet_official_special_gpu_number" in cfg:
+        original_gpu_number = cfg.mmdet_official_special_gpu_number
+    else:
+        original_gpu_number = cfg.default_gpu_number
+
+    if "mmdet_official_special_samples_per_gpu" in cfg:
+        original_samples_per_gpu = cfg.mmdet_official_special_samples_per_gpu
+    else:
+        original_samples_per_gpu = cfg.default_samples_per_gpu
```
jbwang1997, line 88: It's maybe better to let users control the number of samples by themselves. The samples_per_gpu and learning rate are all can modified in config, but the number of GPU is modified in the command.

PeterH0323: User can control the number of samples by themselves by changing `data.samples_per_gpu` in the config file like before. This default_samples_per_gpu means the pretrain model default sample per gpu = 2 or mmdet_official_special_samples_per_gpu = 1 or 5 or 8 and so on

PeterH0323: User can control the number of samples by themselves by changing `data.samples_per_gpu` in the config file like before. This default_samples_per_gpu means the pretrain model default sample per gpu = 2 or mmdet_official_special_samples_per_gpu = 1 or 5 or 8 and so on

jbwang1997: Yes. My opinion is that we don't set default sample_per_gpu for users and users need to match the lr with their sample_per_gpu by themselves. We only scale the lr according to the number of GPU.

jbwang1997: It's easy to change the sample_per_gpu and lr for users in config. So, we left it to users. But for the GPU number, it can be changed in the command line. Thus, we need to automatically change it in codes.

PeterH0323:
Switch to #7377

mmdet/apis/train.py
PeterH0323: Oh, sorry i miss the point. This reason why i want to add the default sample per gpu is that the user may get OOM. I opened another PR #7377, that is based on this.

Issue #3079, opened 2020-06-19
does autoscale_lr take affect?


iamRyanChia:
https://github.com/open-mmlab/mmdetection/blob/bf5c7185e8153af1d9aadd79e472d889594198a4/tools/train.py#L92

when running with `./tools/dist_train.sh config 4 --autoscale_lr`, the lr is not scaled coorectly, unless i specify the gpu-ids or gpus like `./tools/dist_train.sh config 4 --autoscale_lr --gpus 4`, it is unnatural.

hellock:
Thanks for reporting! This argument is risky for distributed training and we decide to remove the misleading arg in #3080.

FreyWang:
I have changed it into:
`cfg.optimizer['lr'] = cfg.optimizer['lr'] * cfg.data['samples_per_gpu'] * torch.cuda.device_count() / 16`

hellock:
> I have changed it into:
> `cfg.optimizer['lr'] = cfg.optimizer['lr'] * cfg.data['samples_per_gpu'] * torch.cuda.device_count() / 16`

Sometimes users only use part of GPUs on the server, e.g., 4 out of 8. `torch.cuda.device_count()` is not safe.

FreyWang:
> > I have changed it into:
> > `cfg.optimizer['lr'] = cfg.optimizer['lr'] * cfg.data['samples_per_gpu'] * torch.cuda.device_count() / 16`
> 
> Sometimes users only use part of GPUs on the server, e.g., 4 out of 8. `torch.cuda.device_count()` is not safe.

yes it is true, cause I always use `CUDA_VISIBLE_DEVICES` to assign gpu 😁

hellock:
That's good. However the usage cases are various thus we just remove this argument to prevent bugs and always make users specify the learning rate.

PR #7377, opened 2022-03-12

PeterH0323:
## Motivation

## Modification

- Add `scale_lr` before `build_optimizer` in function `train_detector`.
- Add `default_batch_size = 16` ,which means `(8 GPUs) x (2 samples per GPU) or (4 GPUs) x (4 samples per GPU)`,  in `configs/_base_/default_runtime.py`.

image

- Add `mmdet_official_batch_size` in the config file which is not using the default setting.

image

- print `original LR` and `scaled LR` to the terminal log after scaled the LR.

image

- Using the flag `--auto-scale-lr` to enable this featrure

```python
python tools/train.py configs/balloon/mask_rcnn_r50_caffe_fpn_mstrain-poly_1x_balloon.py --auto-scale-lr
```

- Add Automatically scaling LR in the doc

## Checklist

- [ ] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.


jbwang1997:
Have you solved the problem of distributed mode? The number of GPUs can get from cfg.gpu_ids.

PeterH0323:
> Have you solved the problem of distributed mode? The number of GPUs can get from cfg.gpu_ids.

I can get the GPU number when using locat machine by `cfg.gpu_ids`, but in distributed mode, I saw MMDistributedDataParallel using `[torch.cuda.current_device()]` while MMDataParallel using `cfg.gpu_ids`.

So I wondar if it is the way to get gpu_ids by using `[torch.cuda.current_device()]` in distributed mode or I still cant use `cfg.gpu_ids` in distributed mode?

jbwang1997:
In train.py, the cfg.gpu_ids has been modified according to distributed model at [here](https://github.com/open-mmlab/mmdetection/blob/master/tools/train.py#L137-L144). `torch.cuda.current_device()` only return the cuda index for one process.

PeterH0323:
> In train.py, the cfg.gpu_ids has been modified according to distributed model at [here](https://github.com/open-mmlab/mmdetection/blob/master/tools/train.py#L137-L144). `torch.cuda.current_device()` only return the cuda index for one process.

Wow, Thx a lot ! I had commit new code about this.

jbwang1997:
Continuing with the previous OOM question, I still consider it can be solved by users. But directly setting default samples_per_gpu may cause some confusion to users.

PeterH0323:
> Continuing with the previous OOM question, I still consider it can be solved by users. But directly setting default samples_per_gpu may cause some confusion to users.

OK, Let me elaborate on my ideas.

Learning rate depend on `GPU number` and `sample per GPU`. In the config file, the sample per GPU is set by `data.samples_per_gpu`, if the user have a correct of GPU type, this will be okay. 
But, if the machine of user is not good enough and cause the OOM, then the user will change the sample per GPU and this will confusion the scaling LR.
So according to this, i add the `default sample per gpu` at the bottom of the config file and let the user to modefide the data.samples_per_gpu

I show you a picture to illustrate:

image

shinya7y:
Automatic LR scaling has been removed to prevent bugs. https://github.com/open-mmlab/mmdetection/issues/3079
Even if `cfg.gpu_ids` is correct now, careful design will be needed to restore the feature.

PeterH0323:
> Automatic LR scaling has been removed to prevent bugs. #3079 Even if `cfg.gpu_ids` is correct now, careful design will be needed to restore the feature.

Thanks for reminding 👍

jbwang1997:
@PeterH0323 Sorry for the late reply.  After consideration, I agree with your idea. Both the number of GPUs and samples_per_gpu should be related to lr scaling.

And I also consider the flag in config is too much. Could we only set `default_batch_size` in default_runtime.py and reload in subconfig. In `train_detector`, we compare the current batch size with `default_batch_size` to get the scaling ratio.

Also, this function should be disabled if not specified. Because we need to avoid some potential bc-breaking in downstream codebases and keep the action of the new vision consistent with the previous vision. We can add an argument like `--auto-scale-lr` in train.py and `auto_scale_lr = True` in config to enable this function.

PeterH0323:
> @PeterH0323 Sorry for the late reply. After consideration, I agree with your idea. Both the number of GPUs and samples_per_gpu should be related to lr scaling.
> 
> And I also consider the flag in config is too many. Could we only set `default_batch_size` in default_runtime.py and reload it in subconfig when the default batch size is not 16. In `train_detector`, we compare the current batch size with `default_batch_size` to get the scaling ratio.
> 
> Also, this function should be disabled if not specified. Because we need to avoid some potential bc-breaking in downstream codebases and keep the action of the new vision consistent with the previous vision. We can add an argument like `--auto-scale-lr` in train.py and `auto_scale_lr = True` in config to enable this function.

@jbwang1997  According to your review suggestion, I updated my description of this PR. Also I commited new codes and docs at the meantime. Pls review, thx 😄

configs/_base_/datasets/cityscapes_detection.py
```diff
@@ -54,3 +54,6 @@
         img_prefix=data_root + 'leftImg8bit/test/',
         pipeline=test_pipeline))
 evaluation = dict(interval=1, metric='bbox')
+
+# NOTE: This is for automatically scaling LR, USER CAN'T CHANGE THIS VALUE
+mmdet_official_special_batch_size = 8  # (8 GPUs) x (1 samples per GPU)
```
jbwang1997, line 59: We can set the batch size in specific models.

jbwang1997: the configs in datasets only contain the dataset arguments.

PeterH0323: I didn't add this `mmdet_official_special_batch_size ` in all the config, just in the config with `data.samples_per_gpu`, is that OK ?

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
```diff
@@ -103,3 +103,6 @@
     warmup_ratio=1.0 / 3,
     step=[190])
 runner = dict(type='EpochBasedRunner', max_epochs=210)
+
+# NOTE: This is for automatically scaling LR, USER CAN'T CHANGE THIS VALUE
+mmdet_official_special_batch_size = 96  # (16 GPUs) x (6 samples per GPU)
```
jbwang1997, line 108: We can both use `default_batch_size` for unity.

configs/_base_/datasets/cityscapes_detection.py
PeterH0323: Oh, do you mean i need to skip this directory: `configs/_base_/datasets` ?

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
PeterH0323: OK, I will use `default_batch_size` in each config file instead of `mmdet_official_special_batch_size`

configs/_base_/datasets/cityscapes_detection.py
jbwang1997: You can set `default_batch_size` in model configs like [faster_rcnn_r50_fpn_1x_cityscapes.py](https://github.com/open-mmlab/mmdetection/blob/11f3ca2ba61b01e84e425bca2b8c6109e525ae67/configs/cityscapes/faster_rcnn_r50_fpn_1x_cityscapes.py).

jbwang1997: You can also refer to the metafile.yml in each folder to find the model not training with 16 batch size.

mmdet/apis/train.py
```diff
@@ -68,6 +68,49 @@ def set_random_seed(seed, deterministic=False):
...
+    """
+
+    if cfg.enable_auto_scale_lr is False:
+        logger.info(f'You disabled automatically scaling LR.')
+        return
+
+    if "mmdet_official_special_batch_size" in cfg:
+        original_batch_size = cfg.mmdet_official_special_batch_size
+    else:
+        original_batch_size = cfg.default_batch_size
```
jbwang1997: `enable_auto_scale_lr` and `default_batch_size` don't definitely exists in cfg. We need to use `get`

mmdet/apis/train.py
```diff
@@ -68,6 +68,49 @@ def set_random_seed(seed, deterministic=False):
...
+        # scale LR according to paper [linear scaling rule](https://arxiv.org/abs/1706.02677)
+        scaled_lr = (batch_size / original_batch_size) * original_lr
+        cfg.optimizer.update({"lr": scaled_lr})
+
+        logger.info(f'You are using {gpu_number} GPU(s) '
+                    f'and {cfg.data.samples_per_gpu} samples per gpu '
+                    f'means the batch is {batch_size}, '
+                    f'while mmdet original batch size of this model '
+                    f'is {original_batch_size}, '
+                    f'automatically scaling LR from {original_lr} to {scaled_lr}')
```
jbwang1997, line 109: When `batch_size == original_batch_size` the scaling ratio is 1. We can merge the if branch.

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
PeterH0323: Done

configs/_base_/datasets/cityscapes_detection.py
PeterH0323: Done. I remove the `default_batch_size` out of dataset configs

PeterH0323: When I read the metafile.yml, I have little confuse.

Take yolact as example, In README.md, is said only using single GPU,
image

But in the metafile.yml, It show `8x GPUs`
image

and some of it shows that batch size is 8 and another is 64, I totally confuse...
image

PeterH0323:
When I read the metafile.yml, I have little confuse.

Take yolact as example, In README.md, is said only using single GPU,
image

But in the metafile.yml, It show `8x GPUs`
image

and some of it shows that batch size is 8 and another is 64, I totally confuse...
image


What's more , some metafile. yml is like that 
image

PeterH0323:
And the most thing is that, I use the `Collections.Training Resources` in metafile.yml as the default gpu number , pls help me clear this. I think I'm stuck in this. I don't know which Training Resources show I use for the default gpu number

PeterH0323:
I saw the doc [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets) said `The default learning rate in config files is for 8 GPUs and 2 img/gpu (batch size = 8\*2 = 16).` Or show I just set the `default_batch_size` and that will be okay? 😆

PeterH0323:
I had commited codes and docs with only set `default_batch_size = 16` in `default_running.py`

PeterH0323:
> I saw the doc [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets) said `The default learning rate in config files is for 8 GPUs and 2 img/gpu (batch size = 8\*2 = 16).` That is to say, I just need to only set the `default_batch_size = 16` in the `default_runtime.py` and that will be okay? 😆

I had commited codes and docs with only set `default_batch_size = 16` in `default_running.py`

jbwang1997:
@PeterH0323 Awesome work! The batch size of existing models is complicated, so it's better to let users make decisions. The code looks good to me. There are only some small points need to fix. I have left some comments below.

configs/_base_/default_runtime.py
```diff
@@ -19,3 +19,8 @@
 opencv_num_threads = 0
 # set multi-process start method as `fork` to speed up the training
 mp_start_method = 'fork'
+
+# In order to support automatically scaling LR. Default batch size
+# (8 GPUs) x (2 samples per GPU) or (4 GPUs) x (4 samples per GPU)
+default_batch_size = 16
+enable_auto_scale_lr = False
```
jbwang1997, line 26: We can set `default_batch_size = None` and let users decide the batch size of their models.

docs/en/1_exist_data_model.md
```diff
@@ -499,9 +499,6 @@ pipelines = [
 MMDetection also provides out-of-the-box tools for training detection models.
 This section will show how to train _predefined_ models (under [configs](https://github.com/open-mmlab/mmdetection/tree/master/configs)) on standard datasets i.e. COCO.
 
-**Important**: The default learning rate in config files is for 8 GPUs and 2 img/gpu (batch size = 8\*2 = 16).
-According to the [linear scaling rule](https://arxiv.org/abs/1706.02677), you need to set the learning rate proportional to the batch size if you use different GPUs or images per GPU, e.g., `lr=0.01` for 4 GPUs \* 2 imgs/gpu and `lr=0.08` for 16 GPUs \* 4 imgs/gpu.
-
 ### Prepare datasets
```
jbwang1997, line 502: We can save this line because this function is normally disabled.

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,22 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
...
+```shell
+python tools/train.py \
+    ${CONFIG_FILE} \
+    --auto-scale-lr \
+    [optional arguments]
+```
+
+If you enabled this feature, the learning rate will be automatically scaled according to the number of GPUs of the machine and the batch size of training. See [linear scaling rule](https://arxiv.org/abs/1706.02677) for details. For example, If there are 4 GPUs and 2 pictures on each GPU, 'lr = 0.01', then if there are 16 GPUs and 4 pictures on each GPU, it will automatically scale to 'lr = 0.08'. 
+
+If you don't want to use it, the default learning rate in config files is based on batch size of 16, you need to calculate the learning rate manually according to the practical training batch size and change the learning rate in specific config file.
```
jbwang1997, line 524: We need to mention the function of `default_batch_size` and `enable_auto_scale_lr` in config.

jbwang1997:
And I found the lint seems failed. You can follow https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md to set up pre-commit hooks to fix the lint error.

configs/_base_/default_runtime.py
PeterH0323: According to the mmdet original doc [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets), It seems that they already modefided the LR to batch size 16. Cuz if there are multi version of batch size of different model , the user will be confused.

What's more if the user need to change the LR according to the metafile.yml, the doc will show them how to do, but I can't find any doc about the metafile.yml, so I think this default batch size of mmdet is all 16.

In conclusion, I think we can set it to 16.

PeterH0323: Cuz if it set to `None`, I think is half-auto not full auto. Pls consider it.

docs/en/1_exist_data_model.md
PeterH0323: I add some detail when it is disabled on line `524` : 

```text
If you don't want to use it, the default learning rate in config files is based on batch size of 16, you need to calculate the learning rate manually according to the practical training batch size and change the learning rate in specific config file.
```

docs/en/1_exist_data_model.md
PeterH0323: Done

PeterH0323:
> And I found the lint seems failed. You can follow https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md to set up pre-commit hooks to fix the lint error.

Done 😄

jbwang1997:
Hi @PeterH0323. The learning rates of some models are not based on 16 batch size. We still need to set `default_batch_size` in those models.

For the special batch size, you can take both readme and metafile.yml into consideration. Some information may be missing or incorrect in metafile.yml, but we have described the special batch size in the readme.

jbwang1997:
If you cannot decide the batch size of some models, you can list them here and we will check their batch size.

PeterH0323:
@jbwang1997  For my point of view, I think we can change the angle to reconsider this.

If I am a user, I want to use mmdet to train my own model, I will do the follow setps:
1. Read the doc about LR on [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets). 
2. Check my own mechine about the gpu number and calculate the LR according to the doc. And may be at the same time the user will modify the `sample_per_gpu` to `2` if it isn't. 
3. Run the training
4. If I my machine is good enough, then I won't get the OOM. But If the user get OOM then they will change the `sample_per_gpu` then calculate the LR again. 
5. Run the training again.

I think those steps are the user all gonna do. **So here lead to a serious problem**, the default batch size not match the LR from the begining, If I am lucky then I will have a correct LR but it depends on the luck. That is to say user can't re-implement the pre-trained model according to the config from scratch, or the user will have the wrong LR when they train a model base on pre-trained model.

Thus, I think this is a best opportunity for mmdet to make a list to show the default batch size of all the `.py` configs or make the metafile.yml correctly. This will help the user know the default batch size of each model and help them with the calculate LR. As a developer, I can't 100% sure I set it correct, so I think this must be done by mmdet.

**I just want to make mmdet better. No offense.**

Pls consider that, and I think it is better let the team leader knows about this.

configs/_base_/default_runtime.py
ZwwWayne: We can set it as 16, and users can modify their config if they need to change it.

docs/en/1_exist_data_model.md
```diff
@@ -499,9 +499,6 @@ pipelines = [
 MMDetection also provides out-of-the-box tools for training detection models.
 This section will show how to train _predefined_ models (under [configs](https://github.com/open-mmlab/mmdetection/tree/master/configs)) on standard datasets i.e. COCO.
 
-**Important**: The default learning rate in config files is for 8 GPUs and 2 img/gpu (batch size = 8\*2 = 16).
```
ZwwWayne, line 502: We do not need to clean the doc. We should update it and tell users we already support auto-scaling LR.

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+    if enable_auto_scale_lr is False:
+        logger.info(f'You disabled automatically scaling LR.')
+        return
+
+    # default batch size of pre-train model
+    original_batch_size = cfg.get('default_batch_size', None)
+    assert original_batch_size is not None
+
+    # calculate the batch size
+    gpu_number = len(cfg.gpu_ids)
```
ZwwWayne, line 88: We can use get_world_size() to get the full world size, which should be safe in multi-machine DDP.

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+        scaled_lr = (batch_size / original_batch_size) * original_lr
+        cfg.optimizer.update({"lr": scaled_lr})
+
+        logger.info(f'While default batch size of this model '
+                    f'is {original_batch_size}, '
+                    f'automatically scaling LR '
+                    f'from {original_lr} to {scaled_lr}')
+
+    else:
+        logger.info(f'Which match the '
```
ZwwWayne, line 111: We should remove f here because there is no variable.

mmdet/apis/train.py
PeterH0323: In file `tools\train.py`, when distributed enable, the `cfg.gpu_ids` will get info from `dist.get_world_size()`. And those steps process before function `scale_lr`. So it's safe to get the gpu number using `len(cfg.gpu_ids)` here.

image

image

mmdet/apis/train.py
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done , I added it under the `Learning rate automatically scale`. Also I added instructions on how to distinguish whether it is the default batch size and how to modify the default batch size.

jbwang1997:
> > Hi @PeterH0323. The learning rates of some models are not based on 16 batch size. We still need to set `default_batch_size` in those models.
> > For the special batch size, you can take both readme and metafile.yml into consideration. Some information may be missing or incorrect in metafile.yml, but we have described the special batch size in the readme.
> 
> @jbwang1997 For my point of view, I think we need to change the angle to reconsider this.
> 
> If I am a user, I want to use mmdet to train my own model, I will do the follow setps:
> 
> 1. Read the doc about LR on [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets).
> 2. Check my own mechine about the gpu number and calculate the LR according to the doc, **which default batch size is `16`**. And may be at the same time the user will modify the `sample_per_gpu` to `2` in the specific config file if it isn't `2` by default,.
> 3. Run the training
> 4. If my machine is good enough, then I won't get the OOM. But If the user get OOM then they will change the `sample_per_gpu` then calculate the LR again.
> 5. Run the training again.
> 
> I think those steps are the user all gonna do. **So here lead to a serious problem**, the default batch size not match the LR from the begining, If I am lucky then I will have a correct LR but it depends on the luck. That is to say user can't re-implement the pre-trained model according to the config from scratch, or the user will have the wrong LR when they train a model base on pre-trained model.
> 
> Thus, I think this is a best opportunity for mmdet to make a list to show the default batch size of all the `.py` configs or make the metafile.yml correctly. This will help the user know the default batch size of each model and help them with the calculate LR. As a developer, I can't 100% sure I set it correct, so I think this must be done by mmdet.
> 
> **I just want to make mmdet better. No offense.**
> 
> Pls consider that, and I think it is better let the team leader knows about this.

Great suggestions. We will continue to improve the related documents.
I consider we can also set `default_batch_size` in configs whose learning rate is not based on 16 batch size, like the earlier vision of this pr. This will avoid potential bugs of auto-scale-lr and will also remind users to pay attention to batch size change.

PeterH0323:
> > > Hi @PeterH0323. The learning rates of some models are not based on 16 batch size. We still need to set `default_batch_size` in those models.
> > > For the special batch size, you can take both readme and metafile.yml into consideration. Some information may be missing or incorrect in metafile.yml, but we have described the special batch size in the readme.
> > 
> > 
> > @jbwang1997 For my point of view, I think we need to change the angle to reconsider this.
> > If I am a user, I want to use mmdet to train my own model, I will do the follow setps:
> > 
> > 1. Read the doc about LR on [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets).
> > 2. Check my own mechine about the gpu number and calculate the LR according to the doc, **which default batch size is `16`**. And may be at the same time the user will modify the `sample_per_gpu` to `2` in the specific config file if it isn't `2` by default,.
> > 3. Run the training
> > 4. If my machine is good enough, then I won't get the OOM. But If the user get OOM then they will change the `sample_per_gpu` then calculate the LR again.
> > 5. Run the training again.
> > 
> > I think those steps are the user all gonna do. **So here lead to a serious problem**, the default batch size not match the LR from the begining, If I am lucky then I will have a correct LR but it depends on the luck. That is to say user can't re-implement the pre-trained model according to the config from scratch, or the user will have the wrong LR when they train a model base on pre-trained model.
> > Thus, I think this is a best opportunity for mmdet to make a list to show the default batch size of all the `.py` configs or make the metafile.yml correctly. This will help the user know the default batch size of each model and help them with the calculate LR. As a developer, I can't 100% sure I set it correct, so I think this must be done by mmdet.
> > **I just want to make mmdet better. No offense.**
> > Pls consider that, and I think it is better let the team leader knows about this.
> 
> Great suggestions. We will continue to improve the related documents. I consider we can also set `default_batch_size` in configs whose learning rate is not based on 16 batch size, like the earlier vision of this pr. This will avoid potential bugs of auto-scale-lr and will also remind users to pay attention to batch size change.

I think I can use the config file name to calculate the default batch size, and I will add a unittest to test it. This will make the changes minimize and good for the review

jbwang1997:
> > > > Hi @PeterH0323. The learning rates of some models are not based on 16 batch size. We still need to set `default_batch_size` in those models.
> > > > For the special batch size, you can take both readme and metafile.yml into consideration. Some information may be missing or incorrect in metafile.yml, but we have described the special batch size in the readme.
> > > 
> > > 
> > > @jbwang1997 For my point of view, I think we need to change the angle to reconsider this.
> > > If I am a user, I want to use mmdet to train my own model, I will do the follow setps:
> > > 
> > > 1. Read the doc about LR on [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets).
> > > 2. Check my own mechine about the gpu number and calculate the LR according to the doc, **which default batch size is `16`**. And may be at the same time the user will modify the `sample_per_gpu` to `2` in the specific config file if it isn't `2` by default,.
> > > 3. Run the training
> > > 4. If my machine is good enough, then I won't get the OOM. But If the user get OOM then they will change the `sample_per_gpu` then calculate the LR again.
> > > 5. Run the training again.
> > > 
> > > I think those steps are the user all gonna do. **So here lead to a serious problem**, the default batch size not match the LR from the begining, If I am lucky then I will have a correct LR but it depends on the luck. That is to say user can't re-implement the pre-trained model according to the config from scratch, or the user will have the wrong LR when they train a model base on pre-trained model.
> > > Thus, I think this is a best opportunity for mmdet to make a list to show the default batch size of all the `.py` configs or make the metafile.yml correctly. This will help the user know the default batch size of each model and help them with the calculate LR. As a developer, I can't 100% sure I set it correct, so I think this must be done by mmdet.
> > > **I just want to make mmdet better. No offense.**
> > > Pls consider that, and I think it is better let the team leader knows about this.
> > 
> > 
> > Great suggestions. We will continue to improve the related documents. I consider we can also set `default_batch_size` in configs whose learning rate is not based on 16 batch size, like the earlier vision of this pr. This will avoid potential bugs of auto-scale-lr and will also remind users to pay attention to batch size change.
> 
> I think I can use the config file name to calculate the default batch size automatically, and I will add a unittest to test it. This will make the changes minimize and good for the review

The users may don't follow the config name rule. It may break the work of this function.

PeterH0323:
> > > > > Hi @PeterH0323. The learning rates of some models are not based on 16 batch size. We still need to set `default_batch_size` in those models.
> > > > > For the special batch size, you can take both readme and metafile.yml into consideration. Some information may be missing or incorrect in metafile.yml, but we have described the special batch size in the readme.
> > > > 
> > > > 
> > > > @jbwang1997 For my point of view, I think we need to change the angle to reconsider this.
> > > > If I am a user, I want to use mmdet to train my own model, I will do the follow setps:
> > > > 
> > > > 1. Read the doc about LR on [train-predefined-models-on-standard-datasets](https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md#train-predefined-models-on-standard-datasets).
> > > > 2. Check my own mechine about the gpu number and calculate the LR according to the doc, **which default batch size is `16`**. And may be at the same time the user will modify the `sample_per_gpu` to `2` in the specific config file if it isn't `2` by default,.
> > > > 3. Run the training
> > > > 4. If my machine is good enough, then I won't get the OOM. But If the user get OOM then they will change the `sample_per_gpu` then calculate the LR again.
> > > > 5. Run the training again.
> > > > 
> > > > I think those steps are the user all gonna do. **So here lead to a serious problem**, the default batch size not match the LR from the begining, If I am lucky then I will have a correct LR but it depends on the luck. That is to say user can't re-implement the pre-trained model according to the config from scratch, or the user will have the wrong LR when they train a model base on pre-trained model.
> > > > Thus, I think this is a best opportunity for mmdet to make a list to show the default batch size of all the `.py` configs or make the metafile.yml correctly. This will help the user know the default batch size of each model and help them with the calculate LR. As a developer, I can't 100% sure I set it correct, so I think this must be done by mmdet.
> > > > **I just want to make mmdet better. No offense.**
> > > > Pls consider that, and I think it is better let the team leader knows about this.
> > > 
> > > 
> > > Great suggestions. We will continue to improve the related documents. I consider we can also set `default_batch_size` in configs whose learning rate is not based on 16 batch size, like the earlier vision of this pr. This will avoid potential bugs of auto-scale-lr and will also remind users to pay attention to batch size change.
> > 
> > 
> > I think I can use the config file name to calculate the default batch size automatically, and I will add a unittest to test it. This will make the changes minimize and good for the review
> 
> The users may don't follow the config name rule. It may break the work of this function.

We can tell the user that we use the config name to calculate the new LR, and tell them must follow the rule if they want to copy one of them. Or if they need to change the name of the config file then need to add the `default_batch_size` at the buttom of the new config file. How about this ?

ZwwWayne:
This PR resolves https://github.com/open-mmlab/mmdetection/projects/7#card-79129764

jbwang1997:
> We can tell the user that we use the config name to calculate the new LR, and tell them must follow the rule if they want to copy one of them. Or if they need to change the name of the config file then need to add the default_batch_size at the buttom of the new config file. How about this ?

It's hard to regular the actions of users. I think we can add `default_batch_size` in config which is more obvious and straight.

PeterH0323:
> > We can tell the user that we use the config name to calculate the new LR, and tell them must follow the rule if they want to copy one of them. Or if they need to change the name of the config file then need to add the default_batch_size at the buttom of the new config file. How about this ?
> 
> It's hard to regular the actions of users. I think we can add `default_batch_size` in config which is more obvious and straight.

Done. Also fix the lint in the commit. pls review 😄  @jbwang1997

jbwang1997:
It looks good to me. Let's see if the CI passes now.

PeterH0323:
> It looks good to me. Let's see if the CI passes now.

I use the pre-commit to check the code and I fixed another lint problem, I think this time the ci test will pass 😄

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+    if enable_auto_scale_lr is False:
+        logger.info('You disabled automatically scaling LR.')
+        return
+
+    # default batch size of pre-train model
+    original_batch_size = cfg.get('default_batch_size', None)
+    assert original_batch_size is not None
+
+    # calculate the batch size
+    gpu_number = len(cfg.gpu_ids)
```
shinya7y, line 88: `num_gpus` may be better to improve consistency with the code below.
https://github.com/open-mmlab/mmdetection/blob/70f6d9cfade4a2f0b198e4f64776521d181b28be/mmdet/apis/train.py#L104

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+    batch_size = gpu_number * cfg.data.samples_per_gpu
+
+    logger.info(f'You are using {gpu_number} GPU(s) '
+                f'and {cfg.data.samples_per_gpu} samples per gpu '
+                f'means the batch is {batch_size}.')
```
shinya7y, line 93:
```python
f'and {cfg.data.samples_per_gpu} samples per GPU, '
f'which means the total batch size is {batch_size}.')
```
or simply
```python
f'and {cfg.data.samples_per_gpu} samples per GPU. '
f'Total batch size is {batch_size}.')
```

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+
+    if batch_size != original_batch_size:
+        # Get original LR from config
+        original_lr = cfg.optimizer.get('lr', 0)
+        assert original_lr != 0
+
+        # scale LR according to paper
+        # [linear scaling rule](https://arxiv.org/abs/1706.02677)
```
shinya7y, line 101:
```python
        # scale LR with linear scaling rule https://arxiv.org/abs/1706.02677
```
will be sufficient.

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+        scaled_lr = (batch_size / original_batch_size) * original_lr
+        cfg.optimizer.update({'lr': scaled_lr})
+
+        logger.info('While default batch size of this model '
```
shinya7y, line 105: 'Since' rather than 'While'?

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+                    f'is {original_batch_size}, '
+                    'automatically scaling LR '
+                    f'from {original_lr} to {scaled_lr}')
+
+    else:
+        logger.info('Which match the '
```
shinya7y, line 111: Users cannot know what 'Which' means.

docs/zh_cn/1_exist_data_model.md
```diff
@@ -487,17 +487,35 @@ pipelines = [
...
 
 **注意**：
 目前，`configs/cityscapes` 文件夹下的配置文件都是使用 COCO 预训练权值进行初始化的。如果网络连接不可用或者速度很慢，你可以提前下载现存的模型。否则可能在训练的开始会有错误发生。
 
+### 学习率自动缩放
+
+**注意**：在配置文件中的学习率是在 8 块 GPU，每块 GPU 有 2 张图像（批大小为 8*2=16）的情况下设置的。其已经设置在`config/_base_/default_runtime.py` 中的 `default_batch_size`。当配置文件的批次大小为`16`时，学习率自动缩放基于该值。同时，为了不影响其他基于 mmdet 的 codebase，启用自动缩放标志 `enable_auto_scale_lr` 默认设置为 `False`。
+
+如果要启用此功能，需在命令添加参数 `--auto-scale-lr`。并且在启动命令之前，请检查下即将使用的配置文件的名称，因为配置名称指示默认的批处理大小。
+在默认情况下，批次大小是 `8 x 2 = 16`，例如：`faster_rcnn_r50_caffe_fpn_90k_coco.py` 或者 `pisa_faster_rcnn_x101_32x4d_fpn_1x_coco.py`；若不是默认批次，你可以看到像 `_NxN_` 字样的，例如：`faster_rcnn_x101_32x4d_fpn_2x_coco.py` 的批次大小是 `32 x 4 = 128`, 或者 `scnet_x101_64x4d_fpn_8x1_20e_coco.py` 的批次大小是 `8 x 1 = 8`。
```
shinya7y, line 502: `faster_rcnn_x101_32x4d_fpn_2x_coco.py` is wrong as an example and inconsistent with `docs/en/1_exist_data_model.md`.

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,28 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
...
 
+### Learning rate automatically scale
+
+**Important**: The default learning rate in config files is for 8 GPUs and 2 sample per gpu (batch size = 8 * 2 = 16). And it had been set to `default_batch_size` in `config/_base_/default_runtime.py`. Learning rate automatically scale base on this value when the batch size is `16`. Meanwhile, in order not to affect other codebase which based on mmdet, the flag `enable_auto_scale_lr` is set to `False` by default.
+
+If you want to enable this feature, you need to add argument `--auto-scale-lr`. And you need to check the config name which you want to use before you process the command, because the config name indicates the default batch size.
+By default, it is `8 x 2 = 16 batch size`, like `faster_rcnn_r50_caffe_fpn_90k_coco.py` or `pisa_faster_rcnn_x101_32x4d_fpn_1x_coco.py`. If it is other, you will see like `_NxN_` at the end of the config name, like `cornernet_hourglass104_mstest_32x3_210e_coco.py` which batch size is `32 x 3 = 128`, or `scnet_x101_64x4d_fpn_8x1_20e_coco.py` which batch size is `8 x 1 = 8`.
```
shinya7y, line 515: `32 x 3 = 96`

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done

docs/zh_cn/1_exist_data_model.md
PeterH0323: Done

mmdet/apis/train.py
shinya7y: Markdown is invalid (as far as I know).
This comment can be written in one line.

jbwang1997:
Kindly ping @hhaAndroid to have a review.

mmdet/apis/train.py
shinya7y: mmdet basically follows pep8. https://github.com/open-mmlab/mmdetection/blob/master/setup.cfg
`Limit all lines to a maximum of 79 characters.` https://peps.python.org/pep-0008/#maximum-line-length

PeterH0323:
@hhaAndroid  Pls have a review if you have time. Thx 😄

jbwang1997:
Hi @PeterH0323. I find the configs in `configs/cityscapes` are also needed to set `default_batch_size`.

PeterH0323:
> Hi @PeterH0323. I find the configs in `configs/cityscapes` are also needed to set `default_batch_size`.

Oh, Should I change the config file name, add `8x1` in name?

PeterH0323:
> Hi @PeterH0323. I find the configs in `configs/cityscapes` are also needed to set `default_batch_size`.

Done, I have updated the code.

jbwang1997:
The config name is relating to metafile.yml. Only setting `default_batch_size` in configs is okay.

PeterH0323:
> `scratch`

PeterH0323:
> The config name is relating to metafile.yml. Only setting `default_batch_size` in configs is okay.

Okay, pls check the config files of `scratch` , I don't know if it needed to add default batch size  😄

jbwang1997:
Seems the batch size is still 16 in `scratch`, referring to the paper.

PeterH0323:
> Seems the batch size is still 16 in `scratch`, referring to the paper.

Okay, thx

PeterH0323:
I rebased my branch to the newest dev branch which commit is 8603765 by hhaAndroid last Saturday. But I don't know if I do it right, pls check 😄  @jbwang1997

jbwang1997:
It seems to still have unrelated changes. Do you update your branch to the newest dev branch?

PeterH0323:
> It seems to still have unrelated changes. Do you update your branch to the newest dev branch?

`unrelated changes` means there were some commits not belong mine ?

jbwang1997:
> > It seems to still have unrelated changes. Do you update your branch to the newest dev branch?
> 
> `unrelated changes` means there were some commits not belong mine ?

Yes.

jbwang1997:
You can try `git rebase --onto` to rebase this pr.

PeterH0323:
> You can try `git rebase --onto` to rebase this pr.

Okay, I will try later today, Thx 😄

PeterH0323:
Cuz I met some rebase problem that I can't fix, I open another PR #7482, we can continue conversation on that. Close this one.

PR #7433, opened 2022-03-17
[Enhancement] support file client in OpenImages dataset

BIGWangYuDong:


mmdet/datasets/openimages.py
```diff
@@ -786,8 +825,8 @@ def get_relation_matrix(self, hierarchy_file):
             relationship between the parent class and the child class,
             of shape (class_num, class_num).
         """
-
-        assert hierarchy_file.endswith('np')
+        if self.assert_suffix:
+            assert hierarchy_file.endswith('np')
```
ZwwWayne, line 829: This might not be compatible with ceph, consider convert it to pkl format.

mmdet/datasets/openimages.py
```diff
@@ -786,8 +825,8 @@ def get_relation_matrix(self, hierarchy_file):
             relationship between the parent class and the child class,
             of shape (class_num, class_num).
         """
-
-        assert hierarchy_file.endswith('np')
+        if self.assert_suffix:
```
ZwwWayne, line 828: We do not need to use this option. The files are generated by us, we should always check them.

zytx121: That means we should remove all the `self.assert_suffix`?

ZwwWayne: Yes, we should check it by default.

mmdet/datasets/openimages.py
BIGWangYuDong: this is the official format, have checked. This can work

mmdet/datasets/openimages.py
```diff
@@ -66,18 +71,41 @@ def __init__(self,
...
+            class_names = self.get_classes_from_csv(label_file)
+        super(OpenImagesDataset, self).__init__(
+            ann_file=ann_file, file_client_args=file_client_args, **kwargs)
         self.CLASSES = class_names
         self.image_level_ann_file = image_level_ann_file
         self.load_image_level_labels = load_image_level_labels
         if get_supercategory is True:
-            assert hierarchy_file is not None
-            self.class_label_tree = self.get_relation_matrix(hierarchy_file)
+            assert hierarchy_file is not None and \
```
zytx121, line 98: It seems that the test failed because when the default value of `get_supercategory` is true, `hierarchy_file` must be passed in.

BIGWangYuDong:
Kindly ping @zytx121 have a look

zytx121 (approved):
LGTM

ZwwWayne (approved):

Issue #6908, opened 2021-12-28
RuntimeError: CUDA out of memory. Tried to allocate 258.96 GiB (GPU 0; 15.75 GiB total capacity; 2.26 GiB already allocated; 11.11 GiB free; 172.34 MiB cached)


sanmulab:
I use the SOLO framework to train the LVIS datasets. After the model training is completed, the evaluation is performed, but the memory is greatly exceeded. And I tested MaskRcnn in the same environment without any problems.
error：RuntimeError: CUDA out of memory. Tried to allocate 258.96 GiB (GPU 0; 15.75 GiB total capacity; 2.26 GiB already allocated; 11.11 GiB free; 172.34 MiB cached)

Env:  mmdet=2.18.1    mmcv-full=1.3.17  pytorch=1.3.1

config：

dataset_type = 'LVISV05Dataset'
data_root = 'data/lvis_v05/'
img_norm_cfg = dict(
    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
train_pipeline = [
    dict(type='LoadImageFromFile'),
    dict(type='LoadAnnotations', with_bbox=True, with_mask=True),
    dict(
                type='Resize',
                img_scale=[(1333, 640), (1333, 672), (1333, 704), (1333, 736),
                           (1333, 768), (1333, 800)],
                multiscale_mode='value',
                keep_ratio=True),
    dict(type='RandomFlip', flip_ratio=0.5),
    dict(
        type='Normalize',
        mean=[123.675, 116.28, 103.53],
        std=[58.395, 57.12, 57.375],
        to_rgb=True),
    dict(type='Pad', size_divisor=32),
    dict(type='DefaultFormatBundle'),
    dict(type='Collect', keys=['img', 'gt_bboxes', 'gt_labels', 'gt_masks'])
]
test_pipeline = [
    dict(type='LoadImageFromFile'),
    dict(
        type='MultiScaleFlipAug',
        img_scale=(1333, 800),
        flip=False,
        transforms=[
            dict(type='Resize', keep_ratio=True),
            dict(type='RandomFlip'),
            dict(
                type='Normalize',
                mean=[123.675, 116.28, 103.53],
                std=[58.395, 57.12, 57.375],
                to_rgb=True),
            dict(type='Pad', size_divisor=32),
            dict(type='ImageToTensor', keys=['img']),
            dict(type='Collect', keys=['img'])
        ])
]
data = dict(
[...]

PR #7434, opened 2022-03-17
[Feature] Add AvoidOOM to avoid OOM

BIGWangYuDong:
First, trying to change `torch.mm` to `torch.einsum` to avoid OOM:
before change, mAP: 0.331
```python
inter_matrix = torch.mm(flatten_masks, flatten_masks.transpose(1, 0))
```
after change, mAP: 0.331
```python
inter_matrix = torch.einsum('ik, kj -> ij', flatten_masks,  flatten_masks.transpose(1, 0))
```

But found it cannot save GPU memory.

|            |torch.mm | torch.einsum |
|:--------:|:--------:|:-------:|
|  process time (s)  |    4.69856       |     4.5713     |
| GPU memory  (MiB)  |    807     |      807     |

To avoid OOM, we add a class, which can try to convert inputs to FP16 and CPU if got a PyTorch's CUDA Out of Memory error.
It will do the following steps:
1. first retry after calling `torch.cuda.empty_cache()`.
2. If that still fails, it will then retry by converting inputs to FP16.
3. If that still fails try to convert inputs to CPUs.  In this case, it expects the function to dispatch to the CPU implementation.

TODO:
- [x] Add docs in FAQ

Close: https://github.com/open-mmlab/mmdetection/issues/6908


mmdet/utils/memory.py
```diff
@@ -0,0 +1,172 @@
...
+            self.logger.info(f'Attempting to copy inputs of {str(func)} '
+                             f'to fp16 due to CUDA OOM')
+            if self.out_fp32:
+                if self.output_type is not None:
+                    dtype = self.output_type
+                else:
+                    dtype = torch.float
+                with _ignore_torch_cuda_oom():
+                    self.logger.info(f'Trying to convert output to {dtype}')
+                    return func(*fp16_args, **fp16_kwargs).to(dtype=dtype)
```
jbwang1997, line 133: The `func` may have multiple outputs.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,172 @@
...
+                self.logger.info(f'Cannot convert output to {dtype} due to '
+                                 'CUDA OOM.')
+            with _ignore_torch_cuda_oom():
+                # try to convert outputs to fp16
+                self.logger.info('Trying to convert outputs to fp16')
+                return func(*fp16_args, **fp16_kwargs)
+            warnings.warn('Cannot convert outputs to fp16 due to CUDA OOM')
+            self.logger.info('Cannot convert output to fp16 due to '
+                             'CUDA OOM.')
```
jbwang1997, line 142: Repeat implement `func` with fp16, we can merge this branch into above `if self.out_fp32` branch.

jbwang1997:
If we can only set `keep_dtype` to make users decide whether the outputs are still the same as the input data type or are converted to the avoid OOM data type (fp16, cpu).

mmdet/utils/memory.py
```diff
@@ -0,0 +1,172 @@
...
+            stateless.
+    """
+
+    def __init__(self,
+                 logger=None,
+                 out_fp32=True,
+                 return_gpu=True,
+                 output_type=None,
+                 convert_cpu=True):
+        self.logger = logger
```
ZwwWayne, line 59: we do not need to pass logger. We could simply use `logger = get_root_logger()`

mmdet/utils/memory.py
```diff
@@ -0,0 +1,172 @@
...
+        self.return_gpu = return_gpu
+        if output_type is not None:
```
ZwwWayne, line 61: I think this is determined by the input type rather than the config. We should ensure that tensors have the same type when they are passed in and processed out.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    if isinstance(inputs, torch.Tensor):
+        # we need to ensure that the type of inputs to be casted are the same
+        # as the argument `src_type`.
+        return inputs.to(dst_device) \
+            if inputs.device == src_device else inputs
+    elif isinstance(inputs, str):
+        return inputs
+    elif isinstance(inputs, np.ndarray):
+        return inputs
```
jbwang1997, line 34: Can be merged into the `else` branch.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,215 @@
...
+        dst_device (torch.device): Destination device.
+
+    Returns:
+        The same device with inputs, but all contained Tensors have been cast.
+    """
+    if isinstance(inputs, torch.Tensor):
+        # we need to ensure that the type of inputs to be casted are the same
+        # as the argument `src_type`.
+        return inputs.to(dst_device) \
+            if inputs.device == src_device else inputs
```
jbwang1997, line 30: Why do we need to determine the source device?

mmdet/utils/memory.py
```diff
@@ -0,0 +1,215 @@
...
+            self.logger.info(f'Attempting to copy inputs of {str(func)} '
+                             f'to fp16 due to CUDA OOM')
+
+            if self.keep_type:
+                # get input tensor type, the output type will same as
+                # the first parameter type.
+                if len(args) > 0:
+                    dtype = args[0].dtype
+                elif len(kwargs) > 0:
+                    dtype = list(kwargs.values())[0].dtype
```
jbwang1997, line 160: The first item of args and kwargs may not a tensor.

BIGWangYuDong: Yes, you are right. But I did not have any idea how to avoid this potential error

mmdet/utils/memory.py
BIGWangYuDong: This logic is simply copied from mmcv.

mmdet/utils/memory.py
jbwang1997: If we can iterate the args + kwargs.values(), find the first tensor and get its dtyoe and device.

mmdet/utils/memory.py
jbwang1997: I consider it's unnecessary here to determine the source device. And, if we can make `cast_tensor_device` more general and can cast both tensor's device and type. So that, we can replace `maybe_to_cpu` and `maybe_to_fp16` with it.

mmdet/utils/memory.py
BIGWangYuDong: Ok

mmdet/utils/memory.py
```diff
@@ -0,0 +1,215 @@
...
+            # therefore print a notice.
+            if self.convert_cpu:
+                self.logger.info(f'Attempting to copy inputs of {str(func)} '
+                                 f'to CPU due to CUDA OOM')
+                cpu_args = (maybe_to_cpu(x) for x in args)
+                cpu_kwargs = {k: maybe_to_cpu(v) for k, v in kwargs.items()}
+                if self.return_gpu:
+                    # try to convert outputs to GPU
+                    if len(args) > 0:
+                        device = args[0].device
```
chhluo, line 191: How about adding output_dtype, output_device in args or kwargs?

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+
+                warnings.warn('Cannot convert output to GPU due to CUDA OOM, '
+                              'the output is now on CPU, which might cause '
+                              'errors if the output need to interact with GPU '
+                              'data in subsequent operations')
+                self.logger.info('Cannot convert output to GPU due to '
+                                 'CUDA OOM, the output is on CPU now.')
+
+                return func(*cpu_args, **cpu_kwargs)
+            else:
```
chhluo, line 213: The dtype of output tensor fp16 when  `keep_type=False, return_fp16=False, convert_cpu=False`, how to handle with this case?

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+            1. first retry after calling `torch.cuda.empty_cache()`.
+            2. If that still fails, it will then retry by trying to
+              convert inputs to FP16.
+            3. If that still fails trying to convert inputs to CPUs.
+              In this case, it expects the function to dispatch to
+              CPU implementation.
+
+        Args:
+            func: a stateless callable that takes tensor-like objects
+                as arguments
```
chhluo, line 130: add '.'

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+            type will be fp16. Default: True.
+        convert_cpu (bool): Whether to convert outputs to CPU if get an OOM
+            error. This will slows down the code significantly.
+            Default: True.
+        return_gpu (bool): Whether convert outputs back to GPU, which will
+            used in when `convert_cpu` is True. Default: True.
+        test (bool): Whether skip torch.cuda.empty_cache(), only used in
+            test unit. Default: False
+        return_fp16 (bool): Whether return torch.half, only used in test unit.
+            Default: True
```
chhluo, line 86: add '.'

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+        keep_type (bool): Whether to ensure that tensors have the same type
+            when they are passed in and processed out. If False, the output
+            type will be fp16. Default: True.
+        convert_cpu (bool): Whether to convert outputs to CPU if get an OOM
+            error. This will slows down the code significantly.
+            Default: True.
+        return_gpu (bool): Whether convert outputs back to GPU, which will
+            used in when `convert_cpu` is True. Default: True.
+        test (bool): Whether skip torch.cuda.empty_cache(), only used in
+            test unit. Default: False
```
chhluo, line 84: add '.'

mmdet/utils/memory.py
BIGWangYuDong: It will re-try to use fp16 as input by using `func(*fp16_args, **fp16_kwargs)`.

chhluo: Shall we add this description for this case (The dtype of output tensor is fp16 when keep_type=False, return_fp16=False, convert_cpu=False)?

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+        if 'CUDA out of memory. ' in str(e):
+            pass
+        else:
+            raise
+
+
+class AvoidOOM(object):
+    """Try to convert inputs to FP16 and CPU if got a PyTorch's CUDA Out of
+    Memory error.
+
```
ZwwWayne, line 73: Describe the processing logic in the docstring to help users better understand its usage.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+        self.logger = get_root_logger()
+        self.keep_type = keep_type
+        self.convert_cpu = convert_cpu
+        self.return_gpu = return_gpu
+        self.test = test
+        self.return_fp16 = return_fp16
+
+    def retry_if_cuda_oom(self, func):
+        """Makes a function retry itself after encountering pytorch's CUDA OOM
+        error. It will do the following steps:
```
ZwwWayne, line 119: check whether the API doc is as expected.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,217 @@
...
+                             f'to fp16 due to CUDA OOM')
+
+            if self.keep_type:
+                # get input tensor type, the output type will same as
+                # the first parameter type.
+                with _ignore_torch_cuda_oom():
+                    self.logger.info(f'Trying to convert output to {dtype}')
+                    output = func(*fp16_args, **fp16_kwargs)
+                    output = convert_tensor_type(
+                        output, src_type=torch.half, dst_type=dtype)
```
ZwwWayne, line 173: we should have function in mmcv to convert tensors to fp16.

docs/en/faq.md
```diff
@@ -79,6 +79,16 @@ We list some common troubles faced by many users and their corresponding solutio
...
     3. Try mixed precision training using following the examples in `config/fp16`. The `loss_scale` might need further tuning for different models.
+    4. Try to use `AvoidOOM` to avoid GPU out of memory. It will first retry after calling `torch.cuda.empty_cache()`. If it still fails, it will then retry by trying to convert the dtype of inputs to FP16. If it still fails, it will try to copy inputs from GPUs to CPUs to continue computing. Here is an example to use `AvoidOOM`:
+
+    ```python
+    from mmdet.utils import AvoidOOM
+
+    AvoidOOM = AvoidOOM()
+    # GPU OOM error
+    # outputs = some_torch_function(input1, input2)
+    output = AvoidOOM.retry_if_cuda_oom(some_torch_function)(input1, input2)
```
ZwwWayne, line 90: can it work as a decorator? If we can, we should also show use cases.

mmdet/utils/memory.py
BIGWangYuDong: yes,  the function is `mmcv.runner.cast_tensor_type`. But we also need to convert tensors from GPU to CPU or from CPU to GPU. MMCV does not have a related function. So we write a new function that can both covert device and dtype. 
I did not use `mmcv.runner.cast_tensor_type` because the new function is more intuitive for users

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+        3. If that still fails trying to convert inputs to CPUs.
+          In this case, it expects the function to dispatch to
+          CPU implementation.
+
+    Args:
+        keep_type (bool): Whether to ensure that tensors have the same type
+            when they are passed in and processed out. If False, the output
+            type will be fp16. Default: True.
+        convert_cpu (bool): Whether to convert outputs to CPU if get an OOM
+            error. This will slows down the code significantly.
```
chhluo, line 89: `slows` -> `slow`

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+        >>> # To use as a decorator
+        >>> # from mmdet.utils import AvoidCUDAOOM
+        >>> @AvoidCUDAOOM.retry_if_cuda_oom
+        >>> def function(*args, **kwargs):
+        >>>     return None
+    ```
+
+    Note:
+        1. The output may be on CPU even if inputs are on GPU. Processing
+            on CPU will slows down the code significantly.
```
chhluo, line 112: `slows` -> `slow`

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+
+    Args:
+        inputs: Inputs that to be casted.
+        src_type (torch.dtype | torch.device): Source type.
+        dst_type (torch.dtype | torch.device): Destination type.
+
+    Returns:
+        The same type with inputs, but all contained Tensors have been cast.
+    """
+    assert dst_type is not None
```
chhluo, line 23: add unit test for `convert_tensor_type`

docs/en/faq.md
```diff
@@ -79,6 +79,25 @@ We list some common troubles faced by many users and their corresponding solutio
...
+    ```python
+    # To use as a function
+    from mmdet.utils import AvoidOOM
+
+    AvoidCUDAOOM = AvoidOOM()
+    # GPU OOM error
+    # outputs = some_torch_function(input1, input2)
+    output = AvoidCUDAOOM.retry_if_cuda_oom(some_torch_function)(input1, input2)
+
+    # To use as a decorator
```
ZwwWayne, line 93: use two codeblock to show these two examples with explanation.

docs/en/faq.md
```diff
@@ -79,6 +79,25 @@ We list some common troubles faced by many users and their corresponding solutio
     1. There are some scenarios when there are large amount of ground truth boxes, which may cause OOM during target assignment. You can set `gpu_assign_thr=N` in the config of assigner thus the assigner will calculate box overlaps through CPU when there are more than N GT boxes.
     2. Set `with_cp=True` in the backbone. This uses the sublinear strategy in PyTorch to reduce GPU memory cost in the backbone.
     3. Try mixed precision training using following the examples in `config/fp16`. The `loss_scale` might need further tuning for different models.
+    4. Try to use `AvoidOOM` to avoid GPU out of memory. It will first retry after calling `torch.cuda.empty_cache()`. If it still fails, it will then retry by trying to convert the dtype of inputs to FP16. If it still fails, it will try to copy inputs from GPUs to CPUs to continue computing. Here is an example to use `AvoidOOM`:
```
ZwwWayne, line 82: it will then retry by converting the type of inputs to FP16 format.

ZwwWayne, line 82: Tre avoidoom in you code to make the code continue to run when GPU memory runs out.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+    if isinstance(inputs, torch.Tensor):
+        if isinstance(dst_type, torch.device):
+            # convert Tensor to dst_device
```
ZwwWayne, line 26: check whether this could support instancedata

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+    Args:
+        keep_type (bool): Whether to ensure that tensors have the same type
+            when they are passed in and processed out. If False, the output
+            type will be fp16. Default: True.
+        convert_cpu (bool): Whether to convert outputs to CPU if get an OOM
+            error. This will slow down the code significantly.
+            Default: True.
+        return_gpu (bool): Whether convert outputs back to GPU, which will
+            used in when `convert_cpu` is True. Default: True.
+        test (bool): Whether skip torch.cuda.empty_cache(), only used in
```
ZwwWayne, line 93: if these interface is.only used for UT, we should avoid adding them in the code. We could use mock to make sure some API is called.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,233 @@
...
+
+        1. First retry after calling `torch.cuda.empty_cache()`.
+        2. If that still fails, it will then retry by trying to
+            convert inputs to FP16.
+        3. If that still fails trying to convert inputs to CPUs.
+          In this case, it expects the function to dispatch to
+          CPU implementation.
+
+    Args:
+        keep_type (bool): Whether to ensure that tensors have the same type
```
ZwwWayne, line 85: Is there any case when the type and device do not need to be kept? If not, this module should keep the input output in the same type and device.

mmdet/utils/memory.py
BIGWangYuDong: We can support InstanceData by adding some codes.

mmdet/utils/memory.py
BIGWangYuDong: We should keep this interface because we cannot skip the following codes if we did not meet OOM error.
```python
 with _ignore_torch_cuda_oom():
                    return func(*args, **kwargs)

                # Clear cache and retry
                torch.cuda.empty_cache()
                with _ignore_torch_cuda_oom():
                    return func(*args, **kwargs)
```
In UT, we should avoid large operations in UT

BIGWangYuDong:
Update the logic in AvoidOOM, which defaults to return source type and device without any interface. This makes the codes look simpler.

docs/en/faq.md
```diff
@@ -150,9 +150,35 @@ We list some common troubles faced by many users and their corresponding solutio
...
+
+     AvoidCUDAOOM = AvoidOOM()
+     # GPU OOM error
+     # outputs = some_torch_function(input1, input2)
+     output = AvoidCUDAOOM.retry_if_cuda_oom(some_torch_function)(input1, input2)
+     ```
+
+     ```python
+     # To use as a decorator
+     from mmdet.utils import AvoidCUDAOOM
```
ZwwWayne, line 172: should be AvoidOOM?

docs/en/faq.md
```diff
@@ -150,9 +150,35 @@ We list some common troubles faced by many users and their corresponding solutio
...
+     ```python
+     # To use as a function
+     from mmdet.utils import AvoidOOM
+
+     AvoidCUDAOOM = AvoidOOM()
+     # GPU OOM error
+     # outputs = some_torch_function(input1, input2)
+     output = AvoidCUDAOOM.retry_if_cuda_oom(some_torch_function)(input1, input2)
+     ```
+
```
ZwwWayne, line 169: May also add some description in between the two usages.

mmdet/utils/memory.py
```diff
@@ -0,0 +1,214 @@
...
+from collections import abc
+from contextlib import contextmanager
+from functools import wraps
+
+import torch
+
+from mmdet.utils import get_root_logger
+
+
+def convert_tensor_type(inputs, src_type=None, dst_type=None):
```
ZwwWayne, line 12: convert_tensor_type -> cast_tensor_type

mmdet/utils/memory.py
```diff
@@ -0,0 +1,214 @@
...
+    """Recursively convert Tensor in inputs from src_type to dst_type.
```
ZwwWayne, line 13: src_type -> `src_typ`

mmdet/utils/memory.py
```diff
@@ -0,0 +1,214 @@
...
+        yield
+    except RuntimeError as e:
+        # NOTE: the string may change?
+        if 'CUDA out of memory. ' in str(e):
+            pass
+        else:
+            raise
+
+
+class AvoidOOM(object):
```
ZwwWayne, line 79: AvoidOOM(object) -> AvoidOOM

mmdet/utils/memory.py
```diff
@@ -0,0 +1,214 @@
...
+
+        1. First retry after calling `torch.cuda.empty_cache()`.
+        2. If that still fails, it will then retry by converting inputs
+          to FP16.
+        3. If that still fails trying to convert inputs to CPUs.
+          In this case, it expects the function to dispatch to
+          CPU implementation.
+
+    Args:
+        convert_cpu (bool): Whether to convert outputs to CPU if get an OOM
```
ZwwWayne, line 91: convert_cpu -> to_cpu

docs/en/faq.md
ZwwWayne: I think we should simply use AvoidCUDAOOM here, which squeeze the code into two lines:
```
from mmdet.utils import AvoidCUDAOOM

output = AvoidCUDAOOM.retry_if_cuda_oom(some_torch_function)(input1, input2)
```

ZwwWayne (approved):

PR #7449, opened 2022-03-17
[Fix] Fix reduction=mean in CELoss.

### LINKED ISSUE #7367: Inconsistency with loss function of PyTorch when using ingore_index arg
Hello，I'm a user of mmdet3d, as this toolbox import a lot of funcs from mmdet, which is the reason why I report here.

For example the commly used module  `CrossEntropyLoss` with `ignore_index`

Here is the definition from PyTorch

```
        ignore_index (int, optional): Specifies a target value that is ignored
            and does not contribute to the input gradient. When :attr:`size_average` is
            ``True``, the loss is averaged over non-ignored targets.
```

And the following is from MMDetection

```
        ignore_index (int | None): The label index to be ignored.
            If None, it will be set to default value. Default: -100.
```

This is the [code segment](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/cross_entropy_loss.py) from mmdet 

```python
def cross_entropy(pred,
                  label,
                  weight=None,
                  reduction='mean',
                  avg_factor=None,
                  class_weight=None,
                  ignore_index=-100):
    """Calculate the CrossEntropy loss.

    Args:
        pred (torch.Tensor): The prediction with shape (N, C), C is the number
            of classes.
        label (torch.Tensor): The learning label of the prediction.
        weight (torch.Tensor, optional): Sample-wise loss weight.
        reduction (str, optional): The method used to reduce the loss.
        avg_factor (int, optional): Average factor that is used to average
            the loss. Defaults to None.
        class_weight (list[float], optional): The weight for each class.
        ignore_index (int | None): The label index to be ignored.
            If None, it will be set to default value. Default: -100.

    Returns:
        torch.Tensor: The calculated loss
    """
    # The default value of ignore_index is the same as F.cross_entropy
    ignore_index = -100 if ignore_index is None else ignore_index
    # element-wise losses
    loss = F.cross_entropy(
        pred,
        label,
        weight=class_weight,
        reduction='none',
        ignore_index=ignore_index)

    # apply weights and do the reduction
    if weight is not None:
        weight = weight.float()
    loss = weight_reduce_loss(
        loss, weight=weight, reduction=reduction, avg_factor=avg_factor)

    return loss
```

The key is that, the results are inconsistent. If someone uses mmdets `CrossEntropyLoss` by default and give a param of `ignore_index` such as 255, torch version will ignores the ignored elements, while mmdet version will first caclulate element-wise loss, then apply `mean` reduction for all elements by default. 

In my scenario of point cloud segmentation, the ground truth is a voxelized tensor with most of the labels ignored. By torch version, `mean` reduction will be performed over only no-ingored voxels while mmdet version over all elements.

It troubles me for days  because of this tiny, sort of, 'bugs'?

Pls figure it out and keep consistency with common funcs of torch

Bests

Singal

RangiLyu:

## Motivation

resolve #7367 

## Modification

Fix avg_factor in CELoss with ignore index.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -41,6 +41,7 @@ def cross_entropy(pred,
         reduction='none',
         ignore_index=ignore_index)
 
+    # loss is averaged over non-ignored targets
     if reduction == 'mean' and avg_factor is None:
         avg_factor = max(1, (label != ignore_index).sum())
 
@@ -104,9 +105,10 @@ def binary_cross_entropy(pred,
     """
     # The default value of ignore_index is the same as F.cross_entropy
     ignore_index = -100 if ignore_index is None else ignore_index
+
+    # loss is averaged over non-ignored targets
     if reduction == 'mean' and avg_factor is None:
-        avg_factor = max(1,
-                         label.numel() - (label == ignore_index).sum().item())
+        avg_factor = max(1, ((label >= 0) & (label != ignore_index)).sum())
 
     if pred.dim() != label.dim():
         label, weight = _expand_onehot_labels(label, weight, pred.size(-1),
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -1,4 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import warnings
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -13,7 +15,8 @@ def cross_entropy(pred,
                   reduction='mean',
                   avg_factor=None,
                   class_weight=None,
-                  ignore_index=-100):
+                  ignore_index=-100,
+                  avg_non_ignore=False):
     """Calculate the CrossEntropy loss.
 
     Args:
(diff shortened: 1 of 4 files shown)


mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -41,6 +41,9 @@ def cross_entropy(pred,
         reduction='none',
         ignore_index=ignore_index)
 
+    if reduction == 'mean' and avg_factor is None:
```
ZwwWayne, line 44: Add comments to explain the reason. It is not straightforward to understand that it aims to calculate the real mean value of the loss.

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -41,6 +41,10 @@ def cross_entropy(pred,
         reduction='none',
         ignore_index=ignore_index)
 
+    # loss is averaged over non-ignored targets
+    if reduction == 'mean' and avg_factor is None:
+        avg_factor = max(1, (label != ignore_index).sum())
```
hhaAndroid, line 46: Why not `avg_factor = max(1, ((label >= 0) & (label != ignore_index)).sum())`

mmdet/models/losses/cross_entropy_loss.py
hhaAndroid: Post-related issues.

ZwwWayne (approved):

ZwwWayne (approved):

PR #6938, opened 2022-01-01
[Feature] add Mask2former to mmdet

chhluo:
This PR Mask2former is based on maskformer's PR(#6607 ).The components of maskformer in mask2former are same with the origin PR(#6607 ).

## Motivation

Add a greater panoptic segmentation model to mmdet.

## Modification

- Components in maskformer(not list here).
- Multi-scale-Deformable-Attention Pixeldecoder.
- mask2former head and detector.
- two configs for r50 and swin-tiny.
- instance segmentation evaluation in CocoPanopticDataset.
```
evaluation result of mask2former_r50_lsj_8x2_50e_coco
{
'PQ': 51.85113493826705, 'SQ': 83.07220152800042, 'RQ': 61.59141241640433, 'PQ_th': 57.712003700037954, 'SQ_th': 84.06726498653896, 'RQ_th': 68.07799535626707, 'PQ_st': 43.00454058087707, 'SQ_st': 81.57021894907417, 'RQ_st': 51.80034382793225, 
'PQ_copypaste': '51.9 83.1 61.6 57.7 84.1 68.1 43.0 81.6 51.8', 
'bbox_mAP': 0.443, 'bbox_mAP_50': 0.637, 'bbox_mAP_75': 0.477, 'bbox_mAP_s': 0.264, 'bbox_mAP_m': 0.475, 'bbox_mAP_l': 0.632, 
'bbox_mAP_copypaste': '0.443 0.637 0.477 0.264 0.475 0.632', 
'segm_mAP': 0.417, 'segm_mAP_50': 0.64, 'segm_mAP_75': 0.449, 'segm_mAP_s': 0.223, 'segm_mAP_m': 0.454, 'segm_mAP_l': 0.638, 
'segm_mAP_copypaste': '0.417 0.640 0.449 0.223 0.454 0.638'
}
```
- processing for instance segmengtation results in test.py.
- More general version of match cost: DiceCost, MaskFocalLossCost, and the new match cost: CrossEntropyLossCost.
- bce loss for input and pred which have same shape.

PR #7472, opened 2022-03-20
[Enhance] Update docstring of cross entropy loss

chhluo:
## Motivation

This is part of Mask2Former( PR #6938 )

## Modification

1. update docstring in `binary_cross_entropy`. When `pred`  have shape (N, 1), it will be  expanded to onehot, while for `pred` with shape (N, ), it will not be expanded to onehot.

### DIFF
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -81,7 +81,10 @@ def binary_cross_entropy(pred,
     """Calculate the binary CrossEntropy loss.
 
     Args:
-        pred (torch.Tensor): The prediction with shape (N, 1).
+        pred (torch.Tensor): The prediction with shape (N, 1) or (N, ).
+            When pred is of shape (N, 1) it will be expanded to one-hot
+            format, while for pred of shape (N, ) it will not be expanded
+            to one-hot format.
         label (torch.Tensor): The learning label of the prediction.
         weight (torch.Tensor, optional): Sample-wise loss weight.
         reduction (str, optional): The method used to reduce the loss.
--- a/mmdet/models/losses/cross_entropy_loss.py
+++ b/mmdet/models/losses/cross_entropy_loss.py
@@ -82,10 +82,11 @@ def binary_cross_entropy(pred,
 
     Args:
         pred (torch.Tensor): The prediction with shape (N, 1) or (N, ).
-            When pred is of shape (N, 1) it will be expanded to one-hot
-            format, while for pred of shape (N, ) it will not be expanded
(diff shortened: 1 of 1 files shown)


mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -171,6 +181,7 @@ def __init__(self,
                  use_sigmoid=False,
                  use_mask=False,
                  reduction='mean',
+                 bce_expand_one_hot=True,
```
AronLin, line 184: I have a question.  Wich is better to put this arg in `_init_` or in `forward`.
Can @BIGWangYuDong @jbwang1997 give some advice?
If put in `_init_`, it is convenient to switch the loss in the config file directly, but the disadvantage is that ce loss and mask ce loss do not have this parameter. I don't know if it is suitable or not to put an arg that only belongs to a part of `CrossEntropyLoss` in `_init_`.

jbwang1997: The original bce loss can directly input onehot labels. When the `pred` and `label` have the same `dim`, the program will skip the `_expand_onehot_labels` function. So I think it's unnecessary to set `bce_expand_one_hot`.

AronLin: > The original bce loss can directly input onehot labels. When the `pred` and `label` have the same `dim`, the program will skip the `_expand_onehot_labels` function. So I think it's unnecessary to set `bce_expand_one_hot`.

It seems that the labels are not one-hot in maskformer.

jbwang1997: Yes, the label can also be float if it doesn't pass the `bce_expand_one_hot` function.

chhluo: yes, `bce_expand_one_hot` is unnecessary. The docstring is misleading. I have updated it.

jbwang1997 (approved):
LGTM

mmdet/models/losses/cross_entropy_loss.py
```diff
@@ -81,7 +81,10 @@ def binary_cross_entropy(pred,
     """Calculate the binary CrossEntropy loss.
 
     Args:
-        pred (torch.Tensor): The prediction with shape (N, 1).
+        pred (torch.Tensor): The prediction with shape (N, 1) or (N, ).
+            When pred is of shape (N, 1) it will be expanded to one-hot
```
ZwwWayne, line 85: what does `it will be expanded to one-hot format` mean? Does this mean pred's shape will be changed from what to what?

Issue #7397, opened 2022-03-15
Support automatically scaling LR according to GPU number


ZwwWayne:
## Description

The default learning rate in mmdetection is based on 8 GPUs. We hope to support automatically scaling LR according to GPU number.

## Steps

- Implement the feature
- Add unit tests to test your code

## Reviewers
@jbwang1997

PR #7482, opened 2022-03-21

PeterH0323:
This PR previous conversation link to #7377 

## Motivation

## Modification

- Add `auto_scale_lr` before `build_optimizer` in function `train_detector`.

image

- Add a config dict named `auto_scale_lr` to contain all the setting of auto scale lr in `default_runtime.py`:

image

- Add specific setting at the bottom of the config file which not using default setting:

image

- print `original LR` and `scaled LR` to the terminal log after scaled the LR.

image

- Using the flag `--auto-scale-lr` to enable this featrure

```python
python tools/train.py configs/balloon/mask_rcnn_r50_caffe_fpn_mstrain-poly_1x_balloon.py --auto-scale-lr
```

- Add Automatically scaling LR in the doc

image

## Checklist

- [ ] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.


PeterH0323:
This PR resolves https://github.com/open-mmlab/mmdetection/projects/7#card-79129764

PeterH0323:
This PR link to issue #7397

configs/_base_/default_runtime.py
```diff
@@ -19,3 +19,8 @@
 opencv_num_threads = 0
 # set multi-process start method as `fork` to speed up the training
 mp_start_method = 'fork'
+
+# In order to support automatically scaling LR. Default batch size
```
ZwwWayne, line 23: The default batch size of configs in MMDet. This key allows to scale LR automatically.

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
```diff
@@ -103,3 +103,6 @@
     warmup_ratio=1.0 / 3,
     step=[190])
 runner = dict(type='EpochBasedRunner', max_epochs=210)
+
+# NOTE: This is for automatically scaling LR, USER CAN'T CHANGE THIS VALUE
```
ZwwWayne, line 107: User should not change this value.

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,28 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
 Currently, the config files under `configs/cityscapes` use COCO pretrained weights to initialize.
 You could download the existing models in advance if the network connection is unavailable or slow. Otherwise, it would cause errors at the beginning of training.
 
+### Learning rate automatically scale
+
+**Important**: The default learning rate in config files is for 8 GPUs and 2 sample per gpu (batch size = 8 * 2 = 16). And it had been set to `default_batch_size` in `config/_base_/default_runtime.py`. Learning rate automatically scale base on this value when the batch size is `16`. Meanwhile, in order not to affect other codebase which based on mmdet, the flag `enable_auto_scale_lr` is set to `False` by default.
```
ZwwWayne, line 512: Learning rate will be automatically scaled

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,28 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
...
+
+If you want to enable this feature, you need to add argument `--auto-scale-lr`. And you need to check the config name which you want to use before you process the command, because the config name indicates the default batch size.
+By default, it is `8 x 2 = 16 batch size`, like `faster_rcnn_r50_caffe_fpn_90k_coco.py` or `pisa_faster_rcnn_x101_32x4d_fpn_1x_coco.py`. If it is other, you will see like `_NxN_` at the end of the config name, like `cornernet_hourglass104_mstest_32x3_210e_coco.py` which batch size is `32 x 3 = 96`, or `scnet_x101_64x4d_fpn_8x1_20e_coco.py` which batch size is `8 x 1 = 8`.
```
ZwwWayne, line 515: If it is other, -> In other cases, you will see the name have `nxm` in dictating

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,28 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
...
+
+**Please remember to check the bottom of the specific config file you want to use, it will have `default_batch_size` if the batch size is not `16` and please do not modify it.**
```
ZwwWayne, line 517: and please do not modify it.** -> and please do not modify it if you want to automatically scale the LR.**

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+def scale_lr(cfg, logger):
+    """Automatically scaling LR according to GPU number and sample per GPU.
+
+    Args:
+        cfg (config): training config.
+        logger (logging.Logger): logger.
+    """
+    enable_auto_scale_lr = cfg.get('enable_auto_scale_lr', False)
+    if enable_auto_scale_lr is False:
+        logger.info('You disabled automatically scaling LR.')
```
ZwwWayne, line 80: You disabled automatically scaling LR -> Automatic scaling of learning rate (LR) has been disabled.

mmdet/apis/train.py
```diff
@@ -68,6 +68,51 @@ def set_random_seed(seed, deterministic=False):
...
+        assert original_lr != 0
+
+        # scale LR with
+        # [linear scaling rule](https://arxiv.org/abs/1706.02677)
+        scaled_lr = (batch_size / default_batch_size) * original_lr
+        cfg.optimizer.update({'lr': scaled_lr})
+
+        logger.info('Since default batch size of this model '
+                    f'is {default_batch_size}, '
+                    'automatically scaling LR '
```
ZwwWayne, line 107: LR has been automatically scaled from {} to {}

mmdet/apis/train.py
```diff
@@ -122,9 +167,11 @@ def train_detector(model,
     else:
         model = MMDataParallel(model, device_ids=cfg.gpu_ids)
 
-    # build runner
+    # build optimizer
+    scale_lr(cfg, logger)
```
ZwwWayne, line 171: autoscale_lr

configs/_base_/default_runtime.py
PeterH0323: Done

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

configs/_base_/default_runtime.py
PeterH0323: Done

configs/centripetalnet/centripetalnet_hourglass104_mstest_16x6_210e_coco.py
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

docs/en/1_exist_data_model.md
PeterH0323: Done

tools/train.py
```diff
@@ -106,6 +110,9 @@ def main():
     if args.cfg_options is not None:
         cfg.merge_from_dict(args.cfg_options)
 
+    if args.auto_scale_lr:
```
hhaAndroid, line 113: I feel that the configuration usage and args usage are inconsistent. It is best to unify all of them as `auto_scale_lr`

mmdet/apis/train.py
```diff
@@ -68,6 +68,50 @@ def set_random_seed(seed, deterministic=False):
...
+    """
+    enable_auto_scale_lr = cfg.get('enable_auto_scale_lr', False)
+    if enable_auto_scale_lr is False:
+        logger.info('Automatic scaling of learning rate (LR)'
+                    ' has been disabled.')
+        return
+
+    # default batch size of pre-train model
+    default_batch_size = cfg.get('default_batch_size', None)
+    assert default_batch_size is not None
```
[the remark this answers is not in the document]
PeterH0323: Done. I modefied this assert to:
```python
    if default_batch_size is None:
        logger.warning('Can not find default_batch_size in config file, '
                       'please check if it existed. '
                       'Disable automatic scaling of learning rate.')
        return
```

tools/train.py
PeterH0323: I saw configuration like `--cfg-options` and `diff-seed` were using `-` and use `_` which were `args.cfg_options` and `args.diff_seed`. So I used `--auto-scale-lr` and `args.auto_scale_lr`. Do I still need to change `--auto-scale-lr` to `--auto_scale_lr` ? 😄

PeterH0323: Or do you mean the `enable_auto_scale_lr` should unify to `auto_scale_lr` ?

ZwwWayne: > Or do you mean the `enable_auto_scale_lr` should unify to `auto_scale_lr` ?

I think we can try that. It is clear enough.

mmdet/apis/train.py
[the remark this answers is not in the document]
PeterH0323: I saw all the config files inherit `default_runtime.py`, so I added `default_batch_size=16` in it. And also I added `default_batch_size` in some other config files which default batch size size is not 16.

Should I needed to add `Please update all the config file to mmdet 2.23` in this `logger.warning` ?

tools/train.py
PeterH0323: Done

mmdet/apis/train.py
```diff
@@ -68,6 +68,54 @@ def set_random_seed(seed, deterministic=False):
...
+    # default batch size of pre-train model
+    default_batch_size = cfg.get('default_batch_size', None)
+    if default_batch_size is None:
+        logger.warning('Can not find default_batch_size in config file, '
+                       'please check if it existed. '
+                       'Disable automatic scaling of learning rate.')
+        return
+
+    # calculate the batch size
+    num_gpus = len(cfg.gpu_ids)
```
ZwwWayne, line 93: should use world_size

ZwwWayne: we should do autoscale_lr after distributed env is initialized.

PeterH0323: In file `tools\train.py`, when distributed enable, the `cfg.gpu_ids` will get info from `dist.get_world_size()`. And those steps process before function `scale_lr`. So it's safe to get the gpu number using `len(cfg.gpu_ids)` here.

image

image

jbwang1997: Hi @PeterH0323. After discussion, we consider using `word_size` is safer than directly using `args.gpu_ids`.

PeterH0323: Done

mmdet/apis/train.py
```diff
@@ -68,6 +68,61 @@ def set_random_seed(seed, deterministic=False):
...
+        scaled_lr = (batch_size / default_batch_size) * original_lr
+        cfg.optimizer.update({'lr': scaled_lr})
+
+        logger.info('LR has been automatically scaled '
+                    f'from {original_lr} to {scaled_lr}')
+
+    else:
+        logger.info('The batch size match the '
+                    f'default batch size: {default_batch_size}, '
+                    f'won\'t scaling the LR ({cfg.optimizer.get("lr")}).')
```
ZwwWayne, line 123: won\'t -> will not

ZwwWayne:
This PR is almost done after resolving the remaining comments. Given its importance, we are inviting developers in other OpenMMLab repos to take a look and leave comments.

mmdet/apis/train.py
```diff
@@ -122,9 +177,11 @@ def train_detector(model,
     else:
         model = MMDataParallel(model, device_ids=cfg.gpu_ids)
 
-    # build runner
+    # build optimizer
+    autoscale_lr(cfg, distributed, logger)
```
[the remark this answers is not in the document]
PeterH0323: Done, I use `auto_scale_lr` for unifying.

mmdet/apis/train.py
PeterH0323: Done

PeterH0323:
> This PR is almost done after resolving the remaining comments. Given its importance, we are inviting developers in other OpenMMLab repos to take a look and leave comments.

Okay. Looking forward to it.😄

ZwwWayne:
Invite @[MeowZheng](https://github.com/MeowZheng), @mzr1996, and @kennymckormick to take a look

mmdet/apis/train.py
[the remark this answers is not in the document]
hhaAndroid: I think it is needed. Because if the user is `pip install mmdet`, the configuration file will not be updated at the same time.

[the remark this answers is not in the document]
PeterH0323: Done.

```python
        logger.warning('Can not find "default_batch_size" in your '
                       'configuration file. Please update all the '
                       'configuration files to mmdet >= 2.23.0. '
                       'Disable automatic scaling of learning rate.')
```

docs/zh_cn/1_exist_data_model.md
```diff
@@ -487,17 +487,35 @@ pipelines = [
...
 目前，`configs/cityscapes` 文件夹下的配置文件都是使用 COCO 预训练权值进行初始化的。如果网络连接不可用或者速度很慢，你可以提前下载现存的模型。否则可能在训练的开始会有错误发生。
 
+### 学习率自动缩放
+
+**注意**：在配置文件中的学习率是在 8 块 GPU，每块 GPU 有 2 张图像（批大小为 8*2=16）的情况下设置的。其已经设置在`config/_base_/default_runtime.py` 中的 `default_batch_size`。当配置文件的批次大小为`16`时，学习率会基于该值进行自动缩放。同时，为了不影响其他基于 mmdet 的 codebase，启用自动缩放标志 `auto_scale_lr` 默认设置为 `False`。
+
+如果要启用此功能，需在命令添加参数 `--auto-scale-lr`。并且在启动命令之前，请检查下即将使用的配置文件的名称，因为配置名称指示默认的批处理大小。
+在默认情况下，批次大小是 `8 x 2 = 16`，例如：`faster_rcnn_r50_caffe_fpn_90k_coco.py` 或者 `pisa_faster_rcnn_x101_32x4d_fpn_1x_coco.py`；若不是默认批次，你可以在配置文件看到像 `_NxM_` 字样的，例如：`cornernet_hourglass104_mstest_32x3_210e_coco.py` 的批次大小是 `32 x 3 = 96`, 或者 `scnet_x101_64x4d_fpn_8x1_20e_coco.py` 的批次大小是 `8 x 1 = 8`。
+
+**请记住：如果使用不是默认批次大小的配置文件，请检查配置文件中的底部，会有 `default_batch_size`。另外，如果想使用自动缩放学习率的功能，请不要修改该值。**
```
MeowZheng, line 504:
```suggestion
**请记住：如果使用不是默认批次大小为`16`的配置文件，请检查配置文件中的底部，会有 `default_batch_size`。另外，如果想使用自动缩放学习率的功能，请不要修改该值。**
```

MeowZheng (approved):
LGTM

kennymckormick:
If we have a field named `default_batch_size`, then I think it would be better to have another field named `default_initial_lr`. 
This is to avoid users change the initial lr accidentally and being not aware that it will change the scaling behavior. 
Users modify `default_batch_size` and `default_initial_lr` if and only if they are intended to do so.

docs/zh_cn/1_exist_data_model.md
PeterH0323: Done

configs/_base_/default_runtime.py
```diff
@@ -19,3 +19,8 @@
 opencv_num_threads = 0
 # set multi-process start method as `fork` to speed up the training
 mp_start_method = 'fork'
+
+# The default batch size of configs in MMDet.
+default_batch_size = 16
```
mzr1996, line 24: The `default_batch_size` may be a little confusing. Although we have some comments to explain what it's for, users may considier it as the batch size setting but find the actual function is change the learning rate.
I think it's better to make it as a field of optimizer settings.

ZwwWayne: agree.
We can add comments above `optimizer.lr` to indicate users that if they want to modify optimizer.lr, they may also modify default_batch_size.

PeterH0323:
According above tow review opinion, 

from kennymckormick
> If we have a field named `default_batch_size`, then I think it would be better to have another field named `default_initial_lr`. This is to avoid users change the initial lr accidentally and being not aware that it will change the scaling behavior. Users modify `default_batch_size` and `default_initial_lr` if and only if they are intended to do so.

and 

from mzr1996
> The `default_batch_size` may be a little confusing. Although we have some comments to explain what it's for, users may considier it as the batch size setting but find the actual function is change the learning rate. I think it's better to make it as a field of optimizer settings.

I am consider that two ways to do: 

**New way**: Adding `default_batch_size` and `default_initial_lr` in `optimizer_config` in each config file.

image

**OR** using the **OLD way**, but add comments above optimizer.lr to indicate users that if they want to modify optimizer.lr, they may also modify default_batch_size. And at the  same time add `default_initial_lr` in each config file.

image

Please help me decide. I don't know which way is the best. @ZwwWayne

configs/_base_/default_runtime.py
wangruohui: If this option is only used for auto-scaling, I prefer setting something like `default_num_gpu=8` in config. So that user will know this config, by default, is trained on 8 GPU, and auto-scaling will be based on `world_size/ default_num_gpu`. 
Further, `default_num_gpu`, `samples_per_gpu`,  `world_size` are completely decoupled. They are respectively defined by mmdet's official configuration (a constant to user), user's training preference and user's distributed environment. In contrast, default_batch_size is `default_num_gpu * samples_per_gpu`, which is coupled.

PeterH0323:
According to above 2 review opinions,

from kennymckormick

If we have a field named default_batch_size, then I think it would be better to have another field named default_initial_lr. This is to avoid users change the initial lr accidentally and being not aware that it will change the scaling behavior. Users modify default_batch_size and default_initial_lr if and only if they are intended to do so.

and

from mzr1996

The default_batch_size may be a little confusing. Although we have some comments to explain what it's for, users may considier it as the batch size setting but find the actual function is change the learning rate. I think it's better to make it as a field of optimizer settings.

I using a cofig dict named `auto_scale_lr_config` to contain all the setting of auto scale lr:

image

And some config file which not using default setting will be add something at the bottom of the config file like this:

image

Thx for all the reviews, All make this PR more stronger 😄

configs/_base_/default_runtime.py
PeterH0323: > The `default_batch_size` may be a little confusing. Although we have some comments to explain what it's for, users may considier it as the batch size setting but find the actual function is change the learning rate. I think it's better to make it as a field of optimizer settings.

Done

PeterH0323: In my lastest commit, I made `default_num_gpu` as the specific setting of auto-scaling in `auto_scale_lr_config`. But reviews from both of you are really helpful to me that I decided to make all the config in one dict.

kennymckormick:
> According to above 2 review opinions,
> 
> from `kennymckormick`
> 
> > If we have a field named default_batch_size, then I think it would be better to have another field named default_initial_lr. This is to avoid users change the initial lr accidentally and being not aware that it will change the scaling behavior. Users modify default_batch_size and default_initial_lr if and only if they are intended to do so.
> 
> and
> 
> from `mzr1996`
> 
> > The default_batch_size may be a little confusing. Although we have some comments to explain what it's for, users may considier it as the batch size setting but find the actual function is change the learning rate. I think it's better to make it as a field of optimizer settings.
> 
> I using a config dict named `auto_scale_lr_config` to contain all the setting of auto scale lr in `default_runtime.py`:
> 
> image
> 
> And some config file which not using default setting will be add something at the bottom of the config file like this:
> 
> image
> 
> Thx for all the reviews, All make this PR more stronger 😄

Great, I think that's a good solution.

kennymckormick (approved):

gaotongxiao (commented):
It seems that if a user has set a new learning rate in the optimizer's config and turned auto-rescaling on, the rescaled default learning rate will override the customized one. Shall we always use the customized learning rate and warn users not to do both things together?

configs/yolox/yolox_tiny_8x8_300e_coco.py
```diff
@@ -51,3 +51,7 @@
     train=train_dataset,
     val=dict(pipeline=test_pipeline),
     test=dict(pipeline=test_pipeline))
+
+# NOTE: This variable is for automatically scaling LR,
+# USER SHOULD NOT CHANGE THIS VALUE.
+default_batch_size = 64  # (8 GPUs) x (8 samples per GPU)
```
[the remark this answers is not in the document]
PeterH0323: Done

docs/en/1_exist_data_model.md
```diff
@@ -510,6 +507,28 @@ Training requires preparing datasets too. See section [Prepare datasets](#prepar
...
+
+If you want to enable this feature, you need to add argument `--auto-scale-lr`. And you need to check the config name which you want to use before you process the command, because the config name indicates the default batch size.
+By default, it is `8 x 2 = 16 batch size`, like `faster_rcnn_r50_caffe_fpn_90k_coco.py` or `pisa_faster_rcnn_x101_32x4d_fpn_1x_coco.py`. In other cases, you will see the config file name have `_NxM_` in dictating, like `cornernet_hourglass104_mstest_32x3_210e_coco.py` which batch size is `32 x 3 = 96`, or `scnet_x101_64x4d_fpn_8x1_20e_coco.py` which batch size is `8 x 1 = 8`.
+
+**Please remember to check the bottom of the specific config file you want to use, it will have `auto_scale_lr_config.default_batch_size` if the batch size is not `16` and `auto_scale_lr_config.default_initial_lr` if the batch size is not `0.01`. If you can't find those values, check the config file which in `_base_=[xxx]` and you will find it. What's more please do not modify its values if you want to automatically scale the LR.**
```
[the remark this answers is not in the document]
PeterH0323: Done

PeterH0323:
> It seems that if a user has set a new learning rate in the optimizer's config and turned auto-rescaling on, the rescaled default learning rate will override the customized one. Shall we always use the customized learning rate and warn users not to do both things together?

Done. I add a warning message when the learning rate in the optimizer's config not equal to the `default_initial_lr`:

```python
        if cfg.optimizer.lr != default_initial_lr:
            logger.warning(
                'It seems that you changed '
                f'"cfg.optimizer.lr" to {cfg.optimizer.lr}. The '
                'automatically scaling learning rate base on '
                '"cfg.auto_scale_lr_config.default_initial_lr" = '
                f'{cfg.auto_scale_lr_config.default_initial_lr}. '
                'If you want to use the "cfg.optimizer.lr" as '
                'your learning rate, please disable the '
                'automatically scaling learning rate.')
```

But I still use the `default_initial_lr` to calculate the new learning rate. I think if the user wants to compare different LR for their experiment, they will change the LR, and they don't need to change it to the original when the want to use the auto scale LR.

gaotongxiao:
> > It seems that if a user has set a new learning rate in the optimizer's config and turned auto-rescaling on, the rescaled default learning rate will override the customized one. Shall we always use the customized learning rate and warn users not to do both things together?
> 
> Done. I add a warning message when the learning rate in the optimizer's config not equal to the `default_initial_lr`:
> 
> ```python
>         if cfg.optimizer.lr != default_initial_lr:
>             logger.warning(
>                 'It seems that you changed '
>                 f'"cfg.optimizer.lr" to {cfg.optimizer.lr}. The '
>                 'automatically scaling learning rate base on '
>                 '"cfg.auto_scale_lr_config.default_initial_lr" = '
>                 f'{cfg.auto_scale_lr_config.default_initial_lr}. '
>                 'If you want to use the "cfg.optimizer.lr" as '
>                 'your learning rate, please disable the '
>                 'automatically scaling learning rate.')
> ```
> 
> But I still use the `default_initial_lr` to calculate the new learning rate. I think if the user wants to compare different LR for their experiment, they will change the LR, and they don't need to change it to the original when they want to use the auto scale LR.

I think the behavior here is worth more discussion. From my understanding, when users start to tweak the learning rate in the optimizer's config, they are expecting some actual changes in the learning rate, and we should not use a default learning rate to override their choice, even though it's warned.

I also reviewed this PR thoroughly and found the current implementation only rescales the default learning rate which is not encouraged to be modified. Yet it never works for the learning rate that users are allowed to customize (in the optimizer's config). I guess the following behavior looks more consistent to me:

1. Throw away the `default_learning_rate` key, that is, we still load the learning rate from the optimizer's config.
2. Rescale the learning rate when `--auto-scale-lr` is on.

It might be a bit against what @kennymckormick has said, but I think if the user has specified `--auto-scale-lr` in their commands, then they know their learning rate will be changed according to the actual GPU number. We can also clearly document this behavior and always show users the actual learning rate to avoid potential confusion.

PeterH0323:
> I think the behavior here is worth more discussion. From my understanding, when users start to tweak the learning rate in the optimizer's config, they are expecting some actual changes in the learning rate, and we should not use a default learning rate to override their choice, even though we have warned them.
> 
> I also reviewed this PR thoroughly and found the current implementation only rescales the default learning rate which is not encouraged to be modified. Yet it never works for the learning rate that users are allowed to customize (in the optimizer's config). The logic is weird and I guess the following behavior looks more consistent to me:
> 
> 1. Throw away the `default_learning_rate` key, that is, we still load the learning rate from the optimizer's config.
> 2. Rescale the learning rate when `--auto-scale-lr` is on.
> 
> It might be a bit against what @kennymckormick has said, but I think if the user has specified `--auto-scale-lr` in their commands, then they know their learning rate will be changed according to the actual GPU number. We can also clearly document this behavior and always show users the actual learning rate to avoid potential confusion.

Your advice is valuable. I agree with it. In my latest commit, I use the `cfg.optimizer.lr` to calculate the new LR.

And I do some change of the warning message:

```python
            logger.warning(
                'It seems that you changed "cfg.optimizer.lr" to '
                f'{cfg.optimizer.lr} which is not the default initial lr '
                f'({default_initial_lr}) from the config file. The '
                'automatically scaling LR will use the "cfg.optimizer.lr" to'
                ' calculate the new LR. This may not lead to a best result of'
                ' the training. If you know what are you doing, ignore this '
                'warning message.')
```

PeterH0323:
> `default_initial_lr` hasn't been removed from the configs now. But you can hold and see if MMDet's maintainers also agree to my comments before making any more changes.

Yes, The `default_initial_lr` is not been removed, it used for trigering the warning message when the user change the optimizer.lr value.

mmdet/apis/train.py
```diff
@@ -68,6 +68,84 @@ def set_random_seed(seed, deterministic=False):
...
+                f'({default_initial_lr}) from the config file. The '
+                'automatically scaling LR will use the "cfg.optimizer.lr" to'
+                ' calculate the new LR. This may not lead to a best result of'
+                ' the training. If you know what are you doing, ignore this '
+                'warning message.')
+
+        # scale LR with
+        # [linear scaling rule](https://arxiv.org/abs/1706.02677)
+        scaled_lr = (batch_size / default_batch_size) * cfg.optimizer.lr
+        cfg.optimizer.lr = scaled_lr
```
ly015, line 138: Should this line be placed after L140?

PeterH0323: Done. Thx

LeoXing1996 (approved):

mmdet/apis/train.py
```diff
@@ -68,6 +68,84 @@ def set_random_seed(seed, deterministic=False):
...
+    # Get flag from config
+    auto_scale_lr_flag = cfg.auto_scale_lr_config.get('auto_scale_lr', False)
+    if auto_scale_lr_flag is False:
+        logger.info('Automatic scaling of learning rate (LR)'
+                    ' has been disabled.')
+        return
+
+    # Get default batch size from config
+    default_batch_size = cfg.auto_scale_lr_config.get('default_batch_size', 0)
+    if default_batch_size == 0:
```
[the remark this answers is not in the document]
jshilong: We usually use None to represent not found or not exist

jshilong (commented):
Just a concern: Would users be confused about the `lr` in the optimizer and the `default_initial_lr`?

mmdet/apis/train.py
```diff
@@ -68,6 +68,84 @@ def set_random_seed(seed, deterministic=False):
...
+
+    # Get flag from config
+    auto_scale_lr_flag = cfg.auto_scale_lr_config.get('auto_scale_lr', False)
+    if auto_scale_lr_flag is False:
+        logger.info('Automatic scaling of learning rate (LR)'
+                    ' has been disabled.')
+        return
+
+    # Get default batch size from config
+    default_batch_size = cfg.auto_scale_lr_config.get('default_batch_size', 0)
```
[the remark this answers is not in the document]
PeterH0323: Done

mmdet/apis/train.py
```diff
@@ -68,6 +68,84 @@ def set_random_seed(seed, deterministic=False):
...
+    default_initial_lr = cfg.auto_scale_lr_config.get('default_initial_lr', 0)
+    if default_initial_lr == 0:
+        logger.warning('Can not find "default_initial_lr" '
+                       f'{warning_msg}')
+        return
+
+    # Get gpu number
+    if distributed:
+        _, world_size = get_dist_info()
+        num_gpus = range(world_size)
```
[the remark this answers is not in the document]
PeterH0323: Done

mzr1996:
> Just a concern: Would users be confused about the `lr` in the optimizer and the `default_initial_lr`?

The same concern

PeterH0323:
> Just a concern: Would users be confused about the `lr` in the optimizer and the `default_initial_lr`?

Hi @jshilong and @mzr1996  , I consider this as well. So I set the `default_initial_lr` under a dict named `auto_scale_lr_config` to show it is only for the automatically scaling LR.

And according your review, I add some description in the doc:

> Something about the `optimizer.lr` and `auto_scale_lr_config.auto_scale_lr`:
> - `optimizer.lr`: It will be the base learning rate to calculate the new LR if you enable the `--auto-scale-lr`. On the contrary, it will be the LR directly. And you can change it if you know what you are doing.
> - `auto_scale_lr_config.auto_scale_lr`: It is the LR when mmdet release the config file, it will give a warning message when your `optimizer.lr` is not as same as `auto_scale_lr_config.auto_scale_lr`. **The most thing is that it CAN NOT be changed**.

Pls check it if my lastest commit can solve the concern of user. 😄

ZwwWayne:
After discussion, the reviewers reached an agreement about the modification. We suggest the usages as below:
```
auto_scale_lr = dict(enable=False, base_batch_size=16)
```
base_batch_size indicates the batch size that the LR will be scaled based on, when `enable=True`, auto_scale lr will be enabled. The whole config of auto scale lr is wrapped by the config `auto_scale_lr`.

PeterH0323:
> After discussion, the reviewers reached an agreement about the modification. We suggest the usages as below:
> 
> ```
> auto_scale_lr = dict(enable=False, base_batch_size=16)
> ```
> 
> base_batch_size indicates the batch size that the LR will be scaled based on, when `enable=True`, auto_scale lr will be enabled. The whole config of auto scale lr is wrapped by the config `auto_scale_lr`.

All done 😄 🎉

mmdet/apis/train.py
```diff
@@ -68,6 +68,67 @@ def set_random_seed(seed, deterministic=False):
...
+    if 'auto_scale_lr' not in cfg:
+        logger.warning(f'Can not find "auto_scale_lr" {warning_msg}')
+        return
+
+    # Get flag from config
+    auto_scale_lr_flag = cfg.auto_scale_lr.get('enable', False)
+    if auto_scale_lr_flag is False:
+        logger.info('Automatic scaling of learning rate (LR)'
+                    ' has been disabled.')
+        return
```
jbwang1997, line 93: Can merge those two branches togather.
```
if ('auto_scale_lr' not in cfg) or (not cfg.auto_scale_lr.get('enable', False)):
    ....
```
This will make the logic simpler.

mmdet/apis/train.py
```diff
@@ -68,6 +68,67 @@ def set_random_seed(seed, deterministic=False):
         torch.backends.cudnn.benchmark = False
 
 
+def auto_scale_lr(cfg, distributed, logger):
```
jbwang1997, line 71: We can delete some blank lines in this function to make this function compact.

PeterH0323:
> The config in `nas_fpn` also need to override default_batch_size.

Hi @jbwang1997 , It is need to overwrite to `8 x 8 = 64` ?

image

jbwang1997:
> > The config in `nas_fpn` also need to override default_batch_size.
> 
> Hi @jbwang1997 , It is need to overwrite to `8 x 8 = 64` ?
> 
> image

Yes

mmdet/apis/train.py
PeterH0323: Using `or` may cause an error when cfg has not `auto_scale_lr`.

PeterH0323: Using or may cause an error when cfg has not auto_scale_lr.

jbwang1997: If the first `('auto_scale_lr' not in cfg)` is True, the second part of `(not cfg.auto_scale_lr.get('enable', False))` will not call.

mmdet/apis/train.py
PeterH0323: Done

mmdet/apis/train.py
PeterH0323: Done

PeterH0323:
> The config in `nas_fpn` also need to override default_batch_size.

Done. In my lastest commit, I also added the `auto_scale_lr` to those config files which `sample_per_gpu` is not `2` and number of GPU is not specifying in README at the same time.

ZwwWayne (approved):

hubutui:
Great feature! Is there any plan to add this feature to other openmmlab projects?

YiqunChen1999:
Hi @PeterH0323, it seems that you don't consider the `GradientCumulativeOptimizerHook`. I think the equivalent batch size should be `num_gpus x samples_per_gpu x cumulative_iters`. Maybe you should take this into account.

PR #7502, opened 2022-03-22
[Fix] end_level is the index of the end input backbone level

Czm369:

## Motivation

end_level is the index of the end input backbone level, so `end_level=-1` is equal to `end_level=self.num_ins  - 1`

## Modification

To fix the bug, just change the if condition

To fix the bug, just change the if condition

To fix the bug,  we need to change the if condition
`if end_level == -1 or end_level == self.num_ins - 1`
and add 1 for some variables.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdet/models/necks/fpg.py
```diff
@@ -212,14 +212,14 @@ def __init__(self,
...
+        if end_level == -1 or end_level == self.num_ins - 1:
             self.backbone_end_level = self.num_ins
             assert num_outs >= self.num_ins - start_level
         else:
-            # if end_level < inputs, no extra level is allowed
-            self.backbone_end_level = end_level
-            assert end_level <= len(in_channels)
-            assert num_outs == end_level - start_level
+            # if end_level is not the last level, no extra level is allowed
+            self.backbone_end_level = end_level + 1
```
ZwwWayne, line 220: explain why +1 is necessary here.

tests/test_models/test_necks.py
```diff
@@ -13,6 +13,18 @@ def test_fpn():
...
+    FPN(in_channels=in_channels,
+        out_channels=out_channels,
+        start_level=0,
+        end_level=-1,
+        num_outs=5)
+    FPN(in_channels=in_channels,
+        out_channels=out_channels,
+        start_level=0,
+        end_level=3,
+        num_outs=5)
```
ZwwWayne, line 27: It seems that the updates are not fully covered by unit tests. Please add more unit tests to fix the issues.

zytx121 (approved):
LGTM

hhaAndroid:
@Czm369 ci failed

hhaAndroid (commented):
LGTM

ZwwWayne:
We should not only test assertion errors, but also test the cases when they can run normally.

PR #6464, opened 2021-11-09
Efficientnet

jiangyitong:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Add EfficientNet-b0 + FPN and EfficientNet-b3 + FPN
## Modification

add EfficientNet model in the mmdet/models/backbone/efficientnet.py
add two configs for efficientNet-b0 and efficientNet-b3 training
### acknowledgement

Thanks for David de la Iglesia Castro's contribution. We are grateful that he helps to implement the base efficientNet model and the corresponding tools:  https://github.com/daavoo/mmdetection

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #7514, opened 2022-03-23
[Feature] Support efficientnet in mmdetection.

jbwang1997:
## Motivation

Continue on PR #6464
Support EfficientNet-b3 backbone in mmdetection.
Add EfficientNet-b3 + FPN model.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


configs/efficientnet/retinanet_effb3_fpn_crop896_8x4_1x_coco.py
```diff
@@ -0,0 +1,87 @@
+_base_ = [
+    '../_base_/models/retinanet_r50_fpn.py',
+    '../_base_/datasets/coco_detection.py', '../_base_/default_runtime.py'
+]
+
+cudnn_benchmark = True
```
ZwwWayne, line 6: is this setting necessary?

configs/efficientnet/retinanet_effb3_fpn_crop896_8x4_1x_coco.py
```diff
@@ -0,0 +1,87 @@
...
+    train_cfg=dict(assigner=dict(neg_iou_thr=0.5)))
+# dataset settings
+img_norm_cfg = dict(
+    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
+train_pipeline = [
+    dict(type='LoadImageFromFile'),
+    dict(type='LoadAnnotations', with_bbox=True),
+    dict(
+        type='Resize',
+        img_scale=(896, 896),
```
ZwwWayne, line 40: we may write a img_size above line 35, then assign img_size in lines 40/43/54

mmdet/models/backbones/efficientnet.py
```diff
@@ -0,0 +1,409 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import copy
```
ZwwWayne, line 2: Is this file still necessary? Should we use the efficientnet backbone in mmcls?

configs/efficientnet/retinanet_effb3_fpn_crop896_8x4_1x_coco.py
jbwang1997: I directly copy the setting of retinanet_r50_fpn_crop640_50e_coco.py in nas_fpn.

mmdet/models/backbones/efficientnet.py
jbwang1997: It needs to change several places to use mmcls efficientnet backbone in mmdet.

ZwwWayne: If it is necessary, we should add comments to indicates the difference and explain the necessity.

configs/efficientnet/retinanet_effb3_fpn_crop896_8x4_1x_coco.py
ZwwWayne: OK, we can keep that for now. In the next step we should verify the effectiveness of this option and leave a comment to explain if this key is necessary, otherwise we can delete them.

mmdet/models/backbones/efficientnet.py
jbwang1997: Done.

configs/efficientnet/retinanet_effb3_fpn_crop896_8x4_1x_coco.py
jbwang1997: Done.

mmdet/models/backbones/efficientnet.py
jbwang1997: Done.

ZwwWayne (approved):

PR #7530, opened 2022-03-25
[Fix]fix the bug that mix_results may be None

Czm369:

## Motivation

Fix the bug that mix_results may be None.

## Modification

Some transform may return None, so we need to make sure that mix_results is not None.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,13 +401,16 @@ def __getitem__(self, idx):
...
+                    if not isinstance(indexes, collections.abc.Sequence):
+                        indexes = [indexes]
+                    mix_results = [
+                        copy.deepcopy(self.dataset[index]) for index in indexes
+                    ]
+                    if None not in mix_results:
+                        results['mix_results'] = mix_results
+                        break
 
             results = transform(results)
```
RangiLyu, line 415: some transform will return None here

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,13 +401,16 @@ def __getitem__(self, idx):
...
-                ]
-                results['mix_results'] = mix_results
+                while True:
+                    indexes = transform.get_indexes(self.dataset)
+                    if not isinstance(indexes, collections.abc.Sequence):
+                        indexes = [indexes]
+                    mix_results = [
+                        copy.deepcopy(self.dataset[index]) for index in indexes
+                    ]
+                    if None not in mix_results:
```
ZwwWayne, line 411: It seems that the transform will re-generate all samples even if there is only one None in mix_results, we may consider make it more efficient.

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,15 +401,26 @@ def __getitem__(self, idx):
...
-                if not isinstance(indexes, collections.abc.Sequence):
-                    indexes = [indexes]
-                mix_results = [
-                    copy.deepcopy(self.dataset[index]) for index in indexes
-                ]
-                results['mix_results'] = mix_results
-
-            results = transform(results)
+                while True:
+                    # RandomCrop and Albu may return None.
```
RangiLyu, line 405: This comment seems incorrect. RandomCrop and Albu do not belong to the load pipeline
'' Make sure the results passed the pipeline of the original dataset is not None.” would be more accurate.

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,15 +401,26 @@ def __getitem__(self, idx):
...
+                    mix_results = [
+                        copy.deepcopy(self.dataset[index]) for index in indexes
+                    ]
+                    if None not in mix_results:
+                        results['mix_results'] = mix_results
+                        break
+
+            while True:
+                # To confirm the results passed Mosaic, MixUp
+                # or CopyPaste is not None.
```
RangiLyu, line 419: "To confirm the results passed the pipeline of the wrapper is not None."

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,15 +401,26 @@ def __getitem__(self, idx):
...
-                indexes = transform.get_indexes(self.dataset)
-                if not isinstance(indexes, collections.abc.Sequence):
-                    indexes = [indexes]
-                mix_results = [
-                    copy.deepcopy(self.dataset[index]) for index in indexes
-                ]
-                results['mix_results'] = mix_results
-
-            results = transform(results)
+                while True:
```
RangiLyu, line 404: Need to add a limitation to these two loops to avoid infinite loop.

RangiLyu: And raise some detailed error messages to tell users their data processing pipeline may be incorrect.

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,15 +407,34 @@ def __getitem__(self, idx):
...
+                    mix_results = [
+                        copy.deepcopy(self.dataset[index]) for index in indexes
+                    ]
+                    if None not in mix_results:
+                        results['mix_results'] = mix_results
+                        break
+                else:
+                    raise RuntimeError(
+                        'the loading pipeline of the original dataset'
+                        ' always return None.')
```
RangiLyu, line 425:
```suggestion
                        'The loading pipeline of the original dataset'
                        ' always return None. Please check the correctness of the dataset and its pipeline.')
```

mmdet/datasets/dataset_wrappers.py
```diff
@@ -401,15 +407,34 @@ def __getitem__(self, idx):
...
+            for i in range(self.max_iters):
+                # To confirm the results passed the training pipeline
+                # of the wrapper is not None.
+                updated_results = transform(copy.deepcopy(results))
+                if updated_results is not None:
+                    results = updated_results
+                    break
+            else:
+                raise RuntimeError(
+                    'the training pipeline of the dataset wrapper'
```
RangiLyu, line 436: same as above

mmdet/datasets/dataset_wrappers.py
```diff
@@ -353,13 +353,18 @@ class MultiImageMixDataset:
             dynamically. Default to None. It is deprecated.
         skip_type_keys (list[str], optional): Sequence of type string to
             be skip pipeline. Default to None.
+        max_iters (int): The maximum number of iterations for calling a
+            transform. If the number of iterations is greater than
+            `max_iters`, but results is stillNone, then the iteration is
```
RangiLyu, line 358:
```suggestion
            `max_iters`, but results is still None, then the iteration is
```

mmdet/datasets/dataset_wrappers.py
```diff
@@ -353,13 +353,18 @@ class MultiImageMixDataset:
             dynamically. Default to None. It is deprecated.
         skip_type_keys (list[str], optional): Sequence of type string to
             be skip pipeline. Default to None.
+        max_iters (int): The maximum number of iterations for calling a
```
RangiLyu, line 356:
```suggestion
        max_refetch (int): The maximum number of iterations for calling a
```

RangiLyu (approved):

PR #7547, opened 2022-03-27
fix docstring in bbox_nms.py

yarkable:


mmdet/core/post_processing/bbox_nms.py
```diff
@@ -20,7 +20,7 @@ def multiclass_nms(multi_bboxes,
             contains scores of the background class, but this will be ignored.
         score_thr (float): bbox threshold, bboxes with scores lower than it
             will not be considered.
-        nms_thr (float): NMS IoU threshold
+        nms_cfg (float): NMS IoU threshold
```
jbwang1997, line 23: Thank you for your contribution. The `nms_cfg` now is a dict that contains the arguments of nms operations. Could you also update the description of `nms_cfg`?

yarkable: OK.

PR #7560, opened 2022-03-28
[Feature] Support memory profile hook

chhluo:
## Modification

Add memory profiler hook.

## Use cases (Optional)

image

### DIFF
--- a/mmdet/core/hook/__init__.py
+++ b/mmdet/core/hook/__init__.py
@@ -1,6 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from .checkloss_hook import CheckInvalidLossHook
 from .ema import ExpMomentumEMAHook, LinearMomentumEMAHook
+from .memory_profiler_hook import MemoryProfilerHook
 from .set_epoch_info_hook import SetEpochInfoHook
 from .sync_norm_hook import SyncNormHook
 from .sync_random_size_hook import SyncRandomSizeHook
@@ -10,5 +11,5 @@ from .yolox_mode_switch_hook import YOLOXModeSwitchHook
 __all__ = [
     'SyncRandomSizeHook', 'YOLOXModeSwitchHook', 'SyncNormHook',
     'ExpMomentumEMAHook', 'LinearMomentumEMAHook', 'YOLOXLrUpdaterHook',
-    'CheckInvalidLossHook', 'SetEpochInfoHook'
+    'CheckInvalidLossHook', 'SetEpochInfoHook', 'MemoryProfilerHook'
 ]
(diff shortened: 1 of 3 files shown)


mmdet/core/hook/memory_profiler_hook.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmcv.runner.hooks import HOOKS, Hook
+
+try:
+    import psutil
```
hhaAndroid, line 5: This way of writing is a bit problematic. Once the first library is not installed and the second library is installed, the user will be very confused.

mmdet/core/hook/memory_profiler_hook.py
```diff
@@ -0,0 +1,50 @@
...
+    Args:
+        interval (int): Checking interval (every k iterations).
+            Default: 50.
+    """
+
+    def __init__(self, interval=50):
+        if psutil is None and memory_usage is None:
+            raise RuntimeError('Please install psutil and memory_profiler.')
+        elif psutil is None:
+            raise RuntimeError('Please install psutil.')
```
hhaAndroid, line 26: We can directly specify the installation command.

mmdet/core/hook/memory_profiler_hook.py
```diff
@@ -0,0 +1,50 @@
...
+        elif memory_usage is None:
+            raise RuntimeError('Please install memory_profiler.')
+
+        self.interval = interval
+
+    def after_iter(self, runner):
```
hhaAndroid, line 32: Is the validation phase also required?

mmdet/core/hook/__init__.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from .checkloss_hook import CheckInvalidLossHook
 from .ema import ExpMomentumEMAHook, LinearMomentumEMAHook
+from .memory_profiler_hook import MemoryProfilerHook
```
hhaAndroid, line 4: Would it be more appropriate to put `mmcv` directly in this code? Because subsequent other repo also need

mmdet/core/hook/memory_profiler_hook.py
chhluo: Yes, there are differences between train phase and val phase, such as hooks called in difference phases.

mmdet/core/hook/__init__.py
chhluo: That's true. We should consider to put it in mmcv.

RangiLyu (approved):
LGTM. May add some documents to tell users about this tool in this PR or later.

ZwwWayne:
> LGTM. May add some documents to tell users about this tool in this PR or later.

We can add new documentation to mainly discuss useful hooks in mmdet. This can be done in future PR.

PR #7567, opened 2022-03-29
[Fix] Avoid invalid bbox after deform_sampling

jbwang1997:

## Motivation

After deform_sampling, some boxes will become illegal (the left-top point is at the right or bottom of the right-bottom point), which will make the GIoULoss negative.
Have no effect on performance (origin: 42.3, after: 42.2 & 42.4).

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/dense_heads/tood_head.py
+++ b/mmdet/models/dense_heads/tood_head.py
@@ -273,7 +273,7 @@ class TOODHead(ATSSHead):
             invalid_bbox_idx = (bbox_pred[:, [0]] > bbox_pred[:, [2]]) | \
                                (bbox_pred[:, [1]] > bbox_pred[:, [3]])
             invalid_bbox_idx = invalid_bbox_idx.expand_as(bbox_pred)
-            bbox_pred[invalid_bbox_idx] = reg_bbox[invalid_bbox_idx]
+            bbox_pred = torch.where(invalid_bbox_idx, reg_bbox, bbox_pred)
 
             cls_scores.append(cls_score)
             bbox_preds.append(bbox_pred)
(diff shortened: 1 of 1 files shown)


mmdet/models/dense_heads/tood_head.py
```diff
@@ -268,6 +268,13 @@ def forward(self, feats):
             reg_offset = self.reg_offset_module(feat)
             bbox_pred = self.deform_sampling(reg_bbox.contiguous(),
                                              reg_offset.contiguous())
+
+            # Avoid invalid bboxes after deform_sampling
+            invalid_bbox_idx = (bbox_pred[:, [0]] > bbox_pred[:, [2]]) | \
+                               (bbox_pred[:, [1]] > bbox_pred[:, [3]])
+            invalid_bbox_idx = invalid_bbox_idx.expand_as(bbox_pred)
+            bbox_pred[invalid_bbox_idx] = reg_bbox[invalid_bbox_idx]
+
```
ZwwWayne, line 277: 1. can also fix the docstring in line 287
2.  will this affect the performance?

jbwang1997: I have tested the [tood_r50_fpn_1x_coco.py](https://github.com/open-mmlab/mmdetection/blob/master/configs/tood/tood_r50_fpn_1x_coco.py), which shows some fluctuation. I am now trying to fix the random seed and test the results.

Dwrety: Good to see someone starts working on this issue. Is the current solution just zero-masking the `invalid_bbox`? I think this issue is because the initial prediction of `reg_offset` become so large that the sampling point actually goes outside the GT box. Has anyone tried to cap the `reg_offset` predictions for ltrb inside the GT box (only for anchor-free style bbox coder)? I don't think this violates the paper's concept.

jbwang1997: Good suggestion @Dwrety, but I think capping `reg_offset` in GT is too strict. Moreover, I think invalid boxes will still occur when adjacent predictions have a large difference. Currently, I replace invalid boxes with boxes before deform_sampling, which is certainly valid.

ZwwWayne (approved):

Issue #6926, opened 2021-12-29
FilterAnnotations used in large sclae jittoring


chhluo:
When train a panoptic segmentation model using large scale jittoring(I am not sure whether we should use it in pratice), the filtering method applied on gt_semantic_seg will cause error. gt_bboxes with shape (N, 4), gt_labels with shape (N, ), gt_masks.masks with shape (N, h, w), gt_semantic_seg with shape (h,w), the shape of gt_semantic_seg is different from others.

**Reproduction**
load coco panoptic dataset using following train pipeline will reproduce it.
```python
train_pipeline = [
    dict(type='LoadImageFromFile', to_float32=True),
    dict(
        type='LoadPanopticAnnotations',
        with_bbox=True,
        with_mask=True,
        with_seg=True),
    dict(type='RandomFlip', flip_ratio=0.5),
    # large sclae jittering
    dict(type='Resize', img_scale=image_size,
         ratio_range=(0.1, 2.0),
         multiscale_mode='range',
         keep_ratio=True),
    dict(type='RandomCrop',
         crop_size=image_size,
         crop_type='absolute',
         recompute_bbox=True,
         allow_negative_crop=True),
    dict(type='FilterAnnotations', min_gt_bbox_wh=(1e-2, 1e-2)),
    dict(type='Normalize', **img_norm_cfg),
    dict(type='Pad', size=image_size),
    dict(type='DefaultFormatBundle', img_to_float=True),
    dict(
        type='Collect',
        keys=['img', 'gt_bboxes', 'gt_labels', 'gt_masks', 'gt_semantic_seg']),
]
```
**Error traceback**
Traceback (most recent call last):
File "tools/train.py", line 185, in <module>
main()
File "tools/train.py", line 174, in main                                                                                                                                                                      
train_detector(
File "/data2/xxx/open-mmlab/mmdetection/mmdet/apis/train.py", line 203, in train_detector
runner.run(data_loaders, cfg.workflow)      
File "/home/xxx/.conda/envs/mmdet/lib/python3.8/site-packages/mmcv/runner/iter_based_runner.py", line 134, in run  
iter_runner(iter_loaders[i], **kwargs)                                            
File "/home/xxx/.conda/envs/mmdet/lib/python3.8/site-packages/mmcv/runner/iter_based_runner.py", line 59, in train                                                                                     
data_batch = next(data_loader)                                                
[...]

PR #7571, opened 2022-03-29
[Enhance] Mask2Former Instance Segm Only

PeterVennerstrom:
## Motivation

MaskFormer and Mask2Former currently support panoptic segmentation, but not instance segmentation only.  Minor changes enable training and evaluation using _base_/datasets/coco_instance.py.

## Modification

detectors/maskformer.py:
+ returns list of tuples typical of instance segm only methods when num_stuff_classes == 0
+ renames pan specific show_result method to _show_pan_result and overrides base class show_result() only when num_stuff_classes > 0
+ gt_semantic_seg argument in forward_train() is made optional and defaults to None

dense_heads/maskformer_head.py:
+ img_metas added to preprocess_gt() arguments
+ preprocess_gt() adjusted to expand gt_semantic_seg to list of None(s) when None to work with multi_apply()

models/utils/panoptic_gt_processing.py:
+ accepts additional img_metas argument
+ pads gt_masks with pad shape taken from img_metas rather than gt_semantic_seg
+ returns early when gt_semantic_seg == None

configs/mask2former/:
+ two coco_instance.py based configs included for mask2former


mmdet/models/utils/panoptic_gt_processing.py
```diff
@@ -3,21 +3,20 @@
...
+                           num_stuff, img_metas):
     """Preprocess the ground truth for a image.
 
     Args:
         gt_labels (Tensor): Ground truth labels of each bbox,
             with shape (num_gts, ).
         gt_masks (BitmapMasks): Ground truth masks of each instances
             of a image, shape (num_gts, h, w).
-        gt_semantic_seg (Tensor): Ground truth of semantic
+        gt_semantic_seg (Tensor | None): Ground truth of semantic
```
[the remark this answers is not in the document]
chhluo: should update docstring of returns

configs/mask2former/mask2former_swin-t-p4-w7-224_lsj_8x2_50e_coco_ins.py
```diff
@@ -1,4 +1,4 @@
-_base_ = ['./mask2former_r50_lsj_8x2_50e_coco.py']
+_base_ = ['./mask2former_r50_lsj_8x2_50e_coco_ins.py']
```
chhluo, line 1: It is better to keep the original configurations of the panoptic segmentation while making new instance segmentation configurations based on them. Keep the original configuration file name unchanged, and named the new instance segmentation configurations as 'mask2former_r50_lsj_8x2_50e_coco_ins'

chhluo: Is it appropriate to use the configs of panoptic segmengtation for instance segmentation training? We hope we can release competitive models.

mmdet/models/dense_heads/maskformer_head.py
```diff
@@ -143,13 +144,12 @@ def preprocess_gt(self, gt_labels_list, gt_masks_list, gt_semantic_segs):
             gt_masks_list (list[BitmapMasks]): Each is ground truth
                 masks of each instances of a image, shape
                 (num_gts, h, w).
-            gt_semantic_seg (Tensor): Ground truth of semantic
+            gt_semantic_seg (Tensor | None): Ground truth of semantic
```
chhluo, line 147: Need to specify when gt_semantic_seg is None.

tests/test_models/test_forward.py
```diff
@@ -815,7 +815,7 @@ def test_maskformer_forward():
 
 def test_mask2former_forward():
     model_cfg = _get_detector_cfg(
-        'mask2former/mask2former_r50_lsj_8x2_50e_coco.py')
+        'mask2former/mask2former_r50_lsj_8x2_50e_coco_pan.py')
```
chhluo, line 818: Need add unit test for instance segmentation

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
```diff
@@ -0,0 +1,234 @@
+_base_ = [
```
chhluo, line 1: Is it same as the config of instance segmentation on https://github.com/facebookresearch/Mask2Former

chhluo: There is no need to write so much code to configure, you can inherit the config file of panoptic segmentation and modify some options

mmdet/models/detectors/maskformer.py
```diff
@@ -43,6 +43,9 @@ def __init__(self,
         self.train_cfg = train_cfg
         self.test_cfg = test_cfg
 
+        if self.num_stuff_classes > 0:
+            self.show_result = self._show_pan_result
```
chhluo, line 47: Need to  add comments to explain that we use `SingleStageDetector.show_result` for instance segmentation.

tests/test_models/test_dense_heads/test_mask2former_head.py
```diff
@@ -11,10 +11,12 @@ def test_mask2former_head_loss():
     base_channels = 64
     img_metas = [{
         'batch_input_shape': (128, 160),
+        'pad_shape': (128, 160, 3),
```
chhluo, line 14: Need to design unit test for instance segmentation

mmdet/models/utils/panoptic_gt_processing.py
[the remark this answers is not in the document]
PeterVennerstrom: Updated the maskformer.py and maskformer_head.py as well.

mmdet/models/dense_heads/maskformer_head.py
PeterVennerstrom: Done

tests/test_models/test_forward.py
PeterVennerstrom: Reworked the unit test to test both configs.  Added tests when gt_semantic_seg=None and a test to check the output type (pan is dict, ins is tuple).

configs/mask2former/mask2former_swin-t-p4-w7-224_lsj_8x2_50e_coco_ins.py
PeterVennerstrom: Changed the names as described.

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
PeterVennerstrom: Thanks.  I will check the original implementation and make sure they have identical settings then inherit to reduce the amount of code.

mmdet/models/detectors/maskformer.py
PeterVennerstrom: Comment added.

tests/test_models/test_dense_heads/test_mask2former_head.py
PeterVennerstrom: Reworked the unit test to test panoptic and instance segmentation.  Please take a look at [L62](https://github.com/PeterVennerstrom/mmdetection/blob/bf6bc241e9cae60f05557ab14f72ec1c0e00ce61/tests/test_models/test_dense_heads/test_mask2former_head.py#L62).  Added a variable which is 100 when testing panoptic and 80 when testing instance seg.  Want to make sure this is still a valid test for instance seg.

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
PeterVennerstrom: Checked the original implementation and they use the same settings for instance segmentation and panoptic segmentation.  Compared their settings to ours and did not see any differences.  

Reworked the configs so the instance segmentation configs inherit from the primitive panoptic segmentation config.  The data pipeline and data definitions were fully redefined because inheriting from configs/_base_/datasets/coco_instance.py caused the builder to find duplicate keys.

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
```diff
@@ -0,0 +1,77 @@
...
+        num_things_classes=num_things_classes,
+        num_stuff_classes=num_stuff_classes),
+    test_cfg=dict(panoptic_on=False))
+
+# dataset settings
+image_size = (1024, 1024)
+img_norm_cfg = dict(
+    mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
+train_pipeline = [
+    dict(type='LoadImageFromFile', to_float32=True),
```
chhluo, line 20: There is an additional preprocessing in LSJ for instance segmentation compared to panoptic segmentation, see https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L179. And we also provided it in mmdet, see https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/configs/common/lsj_100e_coco_instance.py#L33

And make sure the data pipeline is same with the original repo.

mmdet/models/detectors/maskformer.py
chhluo: Sorry, I misremembered, it's `BaseDetector.show_result` instead of `SingleStageDetector.show_result`.

PeterVennerstrom: Good catch, updated.

.pre-commit-config.yaml
```diff
@@ -1,18 +1,18 @@
 repos:
   - repo: https://gitlab.com/pycqa/flake8.git
-    rev: 3.8.3
+    rev: 3.9.2
```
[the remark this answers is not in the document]
PeterVennerstrom: Thanks, didn't mean to.

mmdet/models/dense_heads/maskformer_head.py
```diff
@@ -143,13 +144,12 @@ def preprocess_gt(self, gt_labels_list, gt_masks_list, gt_semantic_segs):
...
                 segmentation with the shape (batch_size, n, h, w).
                 [0, num_thing_class - 1] means things,
                 [num_thing_class, num_class-1] means stuff,
-                255 means VOID.
-            target_shape (tuple[int]): Shape of output mask_preds.
-                Resize the masks to shape of mask_preds.
+                255 means VOID. Is None when training instance segmentation.
```
chhluo, line 151: It's better to write a complete sentence. `It's None when training instance segmentation` -> `Is None when training instance segmentation.`

.pre-commit-config.yaml
```diff
@@ -35,12 +35,12 @@ repos:
     hooks:
       - id: codespell
   - repo: https://github.com/myint/docformatter
-    rev: v1.3.1
+    rev: v1.4
```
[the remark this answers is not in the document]
PeterVennerstrom: Thanks, fixed.

mmdet/models/dense_heads/maskformer_head.py
PeterVennerstrom: Thanks, made these adjustments.

mmdet/models/detectors/maskformer.py
```diff
@@ -126,6 +131,9 @@ def simple_test(self, imgs, img_metas, **kwargs):
                     },
                     ...
                 ]
+
+            | list(tuple): Formatted bbox and mask results of multiple \
+                images when no panoptic segmentation classes exist.
```
chhluo, line 136:
```text
        Returns:
            list[dict[str, np.array | tuple]]: Semantic segmentation \
                results and panoptic segmentation results for each \
                image.
            .. code-block:: none
                [
                    {
                        'pan_results': np.array, # shape = [h, w]
                        'ins_results': tuple[list],
                        # semantic segmentation results are not supported yet
                        'sem_results': np.array
                    },
                    ...
                ]
            | list(tuple): Formatted bbox and mask results of multiple \
                images when no panoptic segmentation classes exist.
```
->

```text
        Returns:
            list[dict[str, np.array | tuple[list]] | tuple[list]]: Semantic segmentation \
                results and panoptic segmentation results of each \
                image for panoptic segmentation, or formatted bbox and \
                mask results of multiple images for instance segmentation.
            .. code-block:: none
                [# panoptic segmentation
                    {
                        'pan_results': np.array, # shape = [h, w]
                        'ins_results': tuple[list],
                        # semantic segmentation results are not supported yet
                        'sem_results': np.array
                    },                    
                    ...
                ]
                # or 
                [# instance segmentation
                   (
                    bboxes, # list[np.array]
                    masks # list[[np.array]]
                    ),
                   ... 
                ]
```

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
PeterVennerstrom: Added to the pipeline.

mmdet/models/detectors/maskformer.py
PeterVennerstrom: Added.

tests/test_models/test_forward.py
```diff
@@ -814,8 +814,13 @@ def test_maskformer_forward():
 
 
 def test_mask2former_forward():
-    model_cfg = _get_detector_cfg(
-        'mask2former/mask2former_r50_lsj_8x2_50e_coco.py')
+    # Test Panoptic Segmentation and Instance Segmentation
+    _mask2former_forward('mask2former/mask2former_r50_lsj_8x2_50e_coco.py')
```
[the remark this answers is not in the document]
PeterVennerstrom: Nice! Made the adjustment.

tests/test_models/test_dense_heads/test_mask2former_head.py
```diff
@@ -7,23 +7,111 @@
...
 
 def test_mask2former_head_loss():
-    """Tests head loss when truth is empty and non-empty."""
-    base_channels = 64
+    """Tests head loss when truth is empty and non-empty.
+
+    Tests head loss as Panoptic Segmentation and Instance Segmentation. Tests
+    forward_train and simple_test with masks and None as gt_semantic_seg
+    """
+    self = _init_model(num_stuff_classes=53)
```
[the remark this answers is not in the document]
PeterVennerstrom: Made the adjustment.

chhluo:
Do you have the computing resource to train mask2former for instance segmentation?

mmdet/models/detectors/maskformer.py
```diff
@@ -111,13 +116,12 @@ def simple_test(self, imgs, img_metas, **kwargs):
...
-                image.
-
+            list[dict[str, np.array | tuple[list]] | tuple[list]]:
+                Semantic segmentation results and panoptic segmentation \
+                results of each image for panoptic segmentation, or formatted \
+                bbox and mask results of each image for instance segmentation.
             .. code-block:: none
-
-                [
+                [# panoptic segmentation
```
chhluo, line 124: There are problems on rendering docs of this part, and need to fix it. Please change to 

```text
        Returns:
            list[dict[str, np.array | tuple[list]] | tuple[list]]:
                Semantic segmentation results and panoptic segmentation \
                results of each image for panoptic segmentation, or formatted \
                bbox and mask results of each image for instance segmentation.
            
            .. code-block:: none
                
                [
                    # panoptic segmentation
                    {
                        'pan_results': np.array, # shape = [h, w]
                        'ins_results': tuple[list],
                        # semantic segmentation results are not supported yet
                        'sem_results': np.array
                    },
                    ...
                ]

            or

            .. code-block:: none
            
                [
                    # instance segmentation
                    (
                        bboxes, # list[np.array]
                        masks # list[list[np.array]]
                    ),
                    ...
                ]

```

PeterVennerstrom:
> Do you have the computing resource to train mask2former for instance segmentation?

Only on a 4x GPU machine with limited memory (batch size 1).  A Mask2Former Swin-T model was trained for 50e which achieved:

 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.432
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=1000 ] = 0.660
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=1000 ] = 0.460
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.222
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.472
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.660
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.580
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=300 ] = 0.580
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=1000 ] = 0.580
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=1000 ] = 0.374
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=1000 ] = 0.617
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=1000 ] = 0.784

The data pipeline included extra data augmentation and was missing:
`dict(type='FilterAnnotations', min_gt_bbox_wh=(1e-2, 1e-2))`.

The Original Facebook implementation achieved AP 45.0.

chhluo:
> 
okay, I will help to train these two models in the next one or two weeks.

mmdet/models/detectors/maskformer.py
PeterVennerstrom: Adjusted.

chhluo:
please fix lint problems.

chhluo:
I ran mask2former_r50 for two times, resutls are 43.1 and 43.2 (target 43.7), and ran mask2former_swin_tiny for once, result is 44.7 (target 45.0). I doubt that there may be difference between our code and the original code, such as data loading and processing for training.

PeterVennerstrom:
Found a small difference in our filter empty annotations code:

`dict(type='FilterAnnotations', min_gt_bbox_wh=(1e-2, 1e-2))`

The original implementation uses 1e-5 for min height and width.  [1e-5 is the default argument in the function](https://github.com/facebookresearch/detectron2/blob/221448e4dcfbebf215b8d21ae7e4b1dfbf422d29/detectron2/data/detection_utils.py#L473) from Detection2.  

Filter_empty_instances is called by the Mask2Former code [here](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L179) without changing the default arguments.

Sorry I missed that.  I'll build the configs with the training commands for both Facebook Mask2Former and this implementation to look for other differences.

PeterVennerstrom:
They left a [comment ](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L178)suggesting the filter_empty_instances call happen after augmentation.  I think the augmentations can create empty instances, especially considering how extreme the ratio_range=(0.1, 2.0) resize is.  

Detectron2's [filter_empty_instances](https://github.com/facebookresearch/detectron2/blob/221448e4dcfbebf215b8d21ae7e4b1dfbf422d29/detectron2/data/detection_utils.py#L473) function also filters empty masks.  MMDet's [FilterAnnotations](https://github.com/open-mmlab/mmdetection/blob/d81990802cbd80f6d49fc35cdd255ebc353e4541/mmdet/datasets/pipelines/loading.py#L572) does not consider empty masks when filtering.

chhluo:
> They left a [comment ](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L178)suggesting the filter_empty_instances call happen after augmentation. I think the augmentations can create empty instances, especially considering how extreme the ratio_range=(0.1, 2.0) resize is. Our FilterAnnotations is placed after the resize/random crop like theirs.
> 
> Detectron2's [filter_empty_instances](https://github.com/facebookresearch/detectron2/blob/221448e4dcfbebf215b8d21ae7e4b1dfbf422d29/detectron2/data/detection_utils.py#L473) function also filters empty masks. MMDet's [FilterAnnotations](https://github.com/open-mmlab/mmdetection/blob/d81990802cbd80f6d49fc35cdd255ebc353e4541/mmdet/datasets/pipelines/loading.py#L572) does not consider empty masks when filtering.

In mmdet, mask is also filtered out, see https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/pipelines/loading.py#L601. Mask will be filtered out if its bbox with width or height less than 1e-2.  

In det2, they filter empty mask which has no true values. A mask has no true value if its bbox with width or height less than 1e-2.

So, I think these two filter ops are almost same.

mmdet/datasets/pipelines/loading.py
```diff
@@ -575,23 +575,45 @@ class FilterAnnotations:
...
-        h = gt_bboxes[:, 3] - gt_bboxes[:, 1]
-        keep = (w > self.min_gt_bbox_wh[0]) & (h > self.min_gt_bbox_wh[1])
+        keep = np.zeros(instance_num, dtype=bool)
+        if self.by_box:
+            w = gt_bboxes[:, 2] - gt_bboxes[:, 0]
+            h = gt_bboxes[:, 3] - gt_bboxes[:, 1]
+            keep += (w > self.min_gt_bbox_wh[0]) & (h > self.min_gt_bbox_wh[1])
+        if self.by_mask:
+            gt_masks = results['gt_masks']
+            keep += gt_masks.areas >= self.min_gt_mask_area
```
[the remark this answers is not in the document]
PeterVennerstrom: That makes sense.  No possibility of one true value in the mask the box is below the threshold since both implementations recompute boxes using the instance masks.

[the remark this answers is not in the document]
chhluo: I ran the lastest version, 43.2 for mask2former_r50.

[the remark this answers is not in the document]
chhluo: add unit test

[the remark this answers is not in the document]
chhluo: Is it better to use `keep =  keep & (gt_masks.areas >= self.min_gt_mask_area)` to align with det2?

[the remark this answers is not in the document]
chhluo: I compared the inference result and loss generated by our implementation with the original Mask2Former's, they are same. I ran mask2former_swin_t again to see if the result will reach 45.0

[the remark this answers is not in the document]
PeterVennerstrom: Interesting the loss and inference result are the same as the target model.  

Looking at our [default rescale setting](https://github.com/open-mmlab/mmdetection/blob/d81990802cbd80f6d49fc35cdd255ebc353e4541/mmdet/models/seg_heads/panoptic_fusion_heads/maskformer_fusion_head.py#L168) in maskformer_fusion_head.py simple_test().  It is set to False by default.  I don't see a config setting over-riding the default setting.  I may have missed rescale being set to True elsewhere in the code though.    

The original implementation always [rescales](https://github.com/facebookresearch/Mask2Former/blob/161be514814ea63440fae520d2ec72707b62ba54/mask2former/maskformer_model.py#L222-L227) the prediction to input image scale before post processing.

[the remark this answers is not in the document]
PeterVennerstrom: Noticed the [original FilterAnnotations returns results unfiltered](https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/pipelines/loading.py#L595-L599) if there are no results passing the threshold test and keep_empty=False.

It isn't clear to me why the FilterAnnotations would be applied if the results should be returned unfiltered.  Looks like those lines should occur [after results are filtered](https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/pipelines/loading.py#L600-L604).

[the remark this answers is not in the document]
PeterVennerstrom: Aligned FilterAnnotations with Det2.  It should work with boxes or masks only or both by &.  Added a test and changed the order of the keep_empty related logic as described above.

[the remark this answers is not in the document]
chhluo: > Noticed the [original FilterAnnotations returns results unfiltered](https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/pipelines/loading.py#L595-L599) if there are no results passing the threshold test and keep_empty=False.
> 
> It isn't clear to me why the FilterAnnotations would be applied if the results should be returned unfiltered. Looks like those lines should occur [after results are filtered](https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/pipelines/loading.py#L600-L604).

It will return util it get a non-empty result, see https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/datasets/custom.py#L218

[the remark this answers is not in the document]
chhluo: > I compared the inference result and loss generated by our implementation with the original Mask2Former's, they are same. I ran mask2former_swin_t again to see if the result will reach 45.0

two times for mask2former_swin_t: mask AP 44.6 and 44.7, three times for mask2former_r50: mask AP 43.2, 43.1, abd 43.2. The results are quite stabilizing. I ran inference by converting and loading the official weight of mask2former_r50,  got the almost same mask AP (43.7). And I also checked loss of two versions of the model with the same data sample and weights (I removed the random transforms: Resize, RandomFlip, RandomCrop and FilterAnnotations, which means I did no transforms to image and annoatations except normalization and padding), got the same loss. For now, we can completely ensure that our model structure is same with the original, and our test_pipeline setting is ok. The places I am not completely ensure are training settings (warmup, transforms in train_pipleine) and module initialization. (Maybe I missed something when I implement mask2former for panoptic segmentation).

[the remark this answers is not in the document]
PeterVennerstrom: [mask_former_instance_dataset_mapper.py](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/mask_former_instance_dataset_mapper.py#L156-L157) uses 128 as the image pad value.  

The Coco instance seg configs point at coco_instance_new_baseline_dataset_mapper.py instead.  [They create a padding mask](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L133-L148), but I need to verify the actual value used to pad the image.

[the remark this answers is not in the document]
PeterVennerstrom: Checked the padding values in the images at the resnet backbone's forward method for each implementation.  Just looked for repeated values.  

`print(img[0, :, 1023, 1023])`

Facebook: [0.0741, 0.2052, 0.4265]
Mmdet: [0., 0., 0.]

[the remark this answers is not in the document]
PeterVennerstrom: Added a config value to set the pad values directly.  I'm not sure how Det2 creates the pad values in the first place, but confirmed our config setting replicates them.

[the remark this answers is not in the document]
chhluo: > [mask_former_instance_dataset_mapper.py](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/mask_former_instance_dataset_mapper.py#L156-L157) uses 128 as the image pad value.
> 
> The Coco instance seg configs point at coco_instance_new_baseline_dataset_mapper.py instead. [They create a padding mask](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/coco_instance_new_baseline_dataset_mapper.py#L133-L148), but I need to verify the actual value used to pad the image.

Mask2Former actually used [COCOInstanceNewBaselineDatasetMapper](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/configs/coco/instance-segmentation/Base-COCO-InstanceSegmentation.yaml#L41), not [MaskFormerInstanceDatasetMapper](https://github.com/facebookresearch/Mask2Former/blob/c233619f7ea011cd565174a1b211bde1f43e38db/mask2former/data/dataset_mappers/mask_former_instance_dataset_mapper.py#L16)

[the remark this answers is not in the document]
PeterVennerstrom: Yes, I first noticed the difference looking at MaskFormerInstanceDatasetMapper, but compared the implementation to Mmdet using COCOInstanceNewBaselineDatesetMapper.

There was a difference in padding between our implementation and theirs.

mmdet/datasets/pipelines/loading.py
```diff
@@ -573,38 +573,71 @@ class FilterAnnotations:
     """Filter invalid annotations.
 
     Args:
-        min_gt_bbox_wh (tuple[int]): Minimum width and height of ground truth
+        min_gt_bbox_wh (tuple[float]): Minimum width and height of ground truth
             boxes.
+        min_gt_mask_area (int): Minimum foreground area of ground truth masks.
```
chhluo, line 578: Add description for defualt values of args.

mmdet/datasets/pipelines/loading.py
```diff
@@ -573,38 +573,71 @@ class FilterAnnotations:
...
+
+        keep = tests[0]
+        for t in tests[1:]:
+            keep = keep & t
+
+        keys = ('gt_bboxes', 'gt_labels', 'gt_masks', 'gt_semantic_seg')
+        for key in keys:
+            if key in results:
+                results[key] = results[key][keep]
+        if not tests[0].any():
```
chhluo, line 632: `tests[0].any()` -> `keep.any()`

tests/test_data/test_pipelines/test_loading.py
```diff
@@ -89,3 +92,25 @@ def test_load_webcam_img(self):
         assert results['img'].dtype == np.uint8
         assert results['img_shape'] == (288, 512, 3)
         assert results['ori_shape'] == (288, 512, 3)
+
+
+kwargs = (dict(min_gt_bbox_wh=(100, 100)),
+          dict(min_gt_bbox_wh=(100, 100), keep_empty=False),
```
chhluo, line 98: I think the following way is better 
```
def _build_filter_annotations_args():
    kwargs = (dict(min_gt_bbox_wh=(100, 100)),
              dict(min_gt_bbox_wh=(100, 100), keep_empty=False),
              dict(min_gt_bbox_wh=(1, 1)), dict(min_gt_bbox_wh=(.01, .01)),
              dict(min_gt_bbox_wh=(.01, .01),
                   by_mask=True), dict(by_mask=True),
              dict(by_box=False, by_mask=True))
    targets = (None, 0, 1, 2, 1, 1, 1)

    return list(zip(targets, kwargs))


@pytest.mark.parametrize('target, kwargs', _build_filter_annotations_args())
def test_filter_annotations(target, kwargs):
    filter_ann = FilterAnnotations(**kwargs)
    bboxes = np.array([[2., 10., 4., 14.], [2., 10., 2.1, 10.1]])
    raw_masks = np.zeros((2, 24, 24))
    raw_masks[0, 10:14, 2:4] = 1
    bitmap_masks = BitmapMasks(raw_masks, 24, 24)
    results = dict(gt_bboxes=bboxes, gt_masks=bitmap_masks)
    results = filter_ann(results)
    if results is not None:
        results = results['gt_bboxes'].shape[0]
    assert results == target
```

chhluo:
Please resolve the conflict. And I will run the lastest version in the next one or two days.

tests/test_data/test_pipelines/test_loading.py
PeterVennerstrom: Much better, thanks.

chhluo:
one good news and one bad news. mask2former_swin_tiny reach the target (45.0 mask AP), mask2former_r50 got 43.0 mask AP, filaed to reach the target (43.7 mask AP)

PeterVennerstrom:
Interesting swin-t achieved the target, but resnet50 did a bit worse than previous tests.  It appears possible the different scores are due to noise or due to the padding value alignment.  

Since resnet50 is not hitting the target I looked at the weights used, architecture and frozen batch normalization settings.  

They use Torchvision resnet50 weights, converted to Det2 naming scheme.  They link to resnet50-19c8e357.pth and the current Torchvision version installed with Torch 1.10 gives a file called resnet50-0676ba61.pth.  Confirmed the keys and weights are identical in both versions.  I couldn't find any architecture differences.  I'm still looking at how they freeze batch norm, but it appears identical to Mmdet.  

I see additional configs for panoptic in the latest dev branch.  I'll replicate those for instance seg tomorrow.  Training another swin and a resnet might isolate the issue to something resnet related if that is the case.

chhluo:
> Interesting swin-t achieved the target, but resnet50 did a bit worse than previous tests. It appears possible the different scores are due to noise or due to the padding value alignment.
> 
> Since resnet50 is not hitting the target I looked at the weights used, architecture and frozen batch normalization settings.
> 
> They use Torchvision resnet50 weights, converted to Det2 naming scheme. They link to resnet50-19c8e357.pth and the current Torchvision version installed with Torch 1.10 gives a file called resnet50-0676ba61.pth. Confirmed the keys and weights are identical in both versions. I couldn't find any architecture differences. I'm still looking at how they freeze batch norm, but it appears identical to Mmdet.
> 
> I see additional configs for panoptic in the latest dev branch. I'll replicate those for instance seg tomorrow. Training another swin and a resnet might isolate the issue to something resnet related if that is the case.

Thanks.

chhluo:
mask2former_swin_s 45.9 (target 46.3), mask2former_r101 43.9 and 44.0 (target 44.2).
Emmmmm, it's strange. There still exist some differences.

PeterVennerstrom:
The random nature of the num_points sampling might introduce noise to the training outcome.  Looking at the panoptic results there is a similar level of noise, but in some cases those results exceed the original implementation which gives more confidence there are no differences.

chhluo:
> The random nature of the num_points sampling might introduce noise to the training outcome. Looking at the panoptic results there is a similar level of noise, but in some cases those results exceed the original implementation which gives more confidence there are no differences.

Althougth random nature of the num_points sampling might introduce noise, the results of mask2former_r50 are 0.5 or 0.6 lower than the target value which is a large margin in instance segmentation. And the results of other models except mask2former_swin_tiny are all lower than the target values, this is strange.

PeterVennerstrom:
There is a [bit of discussion on similar issues](https://github.com/facebookresearch/Mask2Former/issues/46) with the replication using the original implementation.  

Unfortunately they still do a bit better than our runs so not conclusive.  

Reproduction using their implementation: 43.5, 43.2, 43.3 
Our runs: 43.1, 43.2, 43.2, 43.0

chhluo:
> There is a [bit of discussion on similar issues](https://github.com/facebookresearch/Mask2Former/issues/46) with the replication using the original implementation.
> 
> Unfortunately they still do a bit better than our runs so not conclusive.
> 
> Reproduction using their implementation: 43.5, 43.2, 43.3 Our runs: 43.1, 43.2, 43.2, 43.0


Before release panoptic segmentation models, I ran mask2former_r50 many times, 51.58, 51.9, 51.44, 51.53, 51.81, 51.96, 51.74, 51.78 (target 51.9 PQ). These values have a wide range, cover the target values. But for mask2forer_r50_ins, results are quite stable.  


Current results
|           | (Normalize, Pad)| (Pad, Normalize)| target |
| ------ | ----------------  | --------------- | ------ |
| r50     | 43.1, 43.2, 43.2 |     43.0            | 43.7   |
| r101   |                  | 43.9, **44.0** | 44.2   |
| swin-t | 45.0             | 44.7, 44.6      | 45.0   |
| swin-s |                  | 45.9            | 46.3   |

configs/mask2former/mask2former_r50_lsj_8x2_50e_coco_ins.py
```diff
@@ -0,0 +1,79 @@
...
+        multiscale_mode='range',
+        keep_ratio=True),
+    dict(
+        type='RandomCrop',
+        crop_size=image_size,
+        crop_type='absolute',
+        recompute_bbox=True,
+        allow_negative_crop=True),
+    dict(type='FilterAnnotations', min_gt_bbox_wh=(1e-5, 1e-5), by_mask=True),
+    dict(type='Pad', size=image_size, pad_val=pad_cfg),
```
ZwwWayne, line 38: Seems that pad image first than normalize does not follow the official code base https://github.com/facebookresearch/Mask2Former/blob/main/mask2former/maskformer_model.py#L194

PeterVennerstrom: [FixedSizeCrop()](https://github.com/facebookresearch/detectron2/blob/f62b8a42e4d995966d0e7d7eb4049f023860b5ff/detectron2/data/transforms/augmentation_impl.py#L302-L351) is used to pad to 1024 x 1024 with pad value 128 before the images enter the forward method of maskformer_model.

While ImageList.from_tensors() is called after normalization and can pad based on size divisibility it doesn't in this case because the padded image size is always divisible when set to 1024 x 1024.

PeterVennerstrom: They use pad 128 --> norm when training and norm --> pad 0 when testing.  Confirmed both cases in MaskFormer.forward() by printing the bottom right corner.  

`print(images.tensor[0, :, -1, -1])`

When training the padding is added by FixedSizeCrop() value 128 and when testing the padding is added by ImageList.from_tensors value 0 to comply with the size divisibility setting.  

Nice catch @ZwwWayne

chhluo: So in training, Pad 128 first and norm, while in testing, norm first and pad 0. I ran the test and found that the performance of [pad 128, norm] is the same as [norm, pad 0].

mitming:
Hi, @PeterVennerstrom @chhluo 

How is the progress of this work? How long will it take to solve this problem?

chhluo:
update

|  | (Normalize, Pad) | (Pad, Normalize) | target|
|-- | -- | -- | --|
| r50 | 43.1, 43.2, 43.2 | 43.0, 42.9 | 43.7|
| r101 |   | 43.9, **44.0**, 43.8 | 44.2|
| swin-t | 45.0 | 44.7, 44.6, 44.5, 44.7 | 45.0|
| swin-s |   | 45.9, **46.1** | 46.3|

liuyunfeng2016:
Hello, I'd like to ask how long it will take to merge this into the master branch

mitming:
> 

Hi, @chhluo

Have you updated the results in the past two days?

chhluo:
> > 
> 
> Hi, @chhluo
> 
> Have you updated the results in the past two days?

@mitming yes. We currently don't find any important difference between our implementation and the original code of Mask2Former.

chhluo:
@PeterVennerstrom , I am preparing to release Mask2Former for instance segmentaion now. There is a certain gap between the performance of the models I trained and the results mentioned in the paper. However the performance of the model trained with the official code is unstable, and the performance may be lower than the target results.

And I renamed config file for panoptic segmentation: `mask2former_xxx_coco.py` -> `mask2former_xxx_coco-panoptic.py`. `mask2former_xxx_coco.py` is for instance segmentation, while `mask2former_xxx_coco-panoptic.py` for panoptic segmentation.

I added config files for instance segmentation in PR #8032 , so please delete the configs in this PR.

chhluo:
@PeterVennerstrom , sorry, I made a mistake. Deleting config files is inappropriate. Please keep just the config files panoptic segmentation (you can copy them from dev branch).

PeterVennerstrom:
@chhluo Exciting news!  Thanks for all your work on this.  

No problem, I'll put the dev configs in.

mitming:
@PeterVennerstrom @chhluo 

Thanks for your contribution, could we use the [original pretrained weights](https://github.com/facebookresearch/Mask2Former/blob/main/MODEL_ZOO.md) on mmdet?

chhluo:
> @PeterVennerstrom @chhluo
> 
> Thanks for your contribution, could we use the [original pretrained weights](https://github.com/facebookresearch/Mask2Former/blob/main/MODEL_ZOO.md) on mmdet?

@mitming yes, but you should convert the keys in original pretrained weights then load them.

mitming:
> > @PeterVennerstrom @chhluo
> > Thanks for your contribution, could we use the [original pretrained weights](https://github.com/facebookresearch/Mask2Former/blob/main/MODEL_ZOO.md) on mmdet?
> 
> @mitming yes, but you should convert the keys in original pretrained weights then load them.

Got it, I will try it~ Thanks a lot

mitming:
Hi, @chhluo 

I cannot find Mask2Former for instance segmentation only in the dev branch. How could I use it?

chhluo:
> only in the dev branch. How coul

Not merged yet. https://github.com/PeterVennerstrom/mmdetection/tree/instance_only

mitming:
> > only in the dev branch. How coul
> 
> Not merged yet. https://github.com/PeterVennerstrom/mmdetection/tree/instance_only

Thanks a lot!

chhluo:
> > > only in the dev branch. How coul
> > 
> > 
> > Not merged yet. https://github.com/PeterVennerstrom/mmdetection/tree/instance_only
> 
> Thanks a lot!

There are two scripts for converting the original model weights.
See 

https://github.com/chhluo/mmdetection/blob/scripts/tools/model_converters/convert_mask2former_resnet_det2_2_mmdet.py

https://github.com/chhluo/mmdetection/blob/scripts/tools/model_converters/convert_mask2former_swin_det2_2_mmdet.py

mmdet/datasets/pipelines/loading.py
```diff
@@ -573,38 +573,72 @@ class FilterAnnotations:
...
-        min_gt_bbox_wh (tuple[int]): Minimum width and height of ground truth
-            boxes.
+        min_gt_bbox_wh (tuple[float]): Minimum width and height of ground truth
+            boxes. Default: (1., 1.)
+        min_gt_mask_area (int): Minimum foreground area of ground truth masks.
+            Default: 1
+        by_box (bool): Filter instances with bounding boxes not meeting the
+            min_gt_bbox_wh threshold. Default: True
+        by_mask (bool): Filter instances with masks not meeting
+        min_gt_mask_area threshold. Default: False
```
ZwwWayne, line 583: the indentation is wrong.

mmdet/datasets/pipelines/loading.py
```diff
@@ -573,38 +573,72 @@ class FilterAnnotations:
...
+                         & (h > self.min_gt_bbox_wh[1]))
+        if self.by_mask:
+            gt_masks = results['gt_masks']
+            tests.append(gt_masks.areas >= self.min_gt_mask_area)
+
+        keep = tests[0]
+        for t in tests[1:]:
+            keep = keep & t
+
+        keys = ('gt_bboxes', 'gt_labels', 'gt_masks', 'gt_semantic_seg')
```
ZwwWayne, line 629: why should gt_semantic_seg also be selected?

chhluo: `gt_semantic_seg` is in the original design of FilterAnnotations. Actually, `gt_semantic_seg` is unneccessary. There is a bug https://github.com/open-mmlab/mmdetection/issues/6926 when use FilterAnnotations in panoptic segmentation.

chhluo:
> > > only in the dev branch. How coul
> > 
> > 
> > Not merged yet. https://github.com/PeterVennerstrom/mmdetection/tree/instance_only
> 
> Hi @chhluo , I noticed this branch has deleted all the configs for instance only, what should I do?

@mitming see https://github.com/open-mmlab/mmdetection/pull/7571#issuecomment-1134752883

mitming:
Hi, @chhluo 

I tried to use [this script](https://github.com/chhluo/mmdetection/blob/scripts/tools/model_converters/convert_mask2former_resnet_det2_2_mmdet.py) to convert original weights into `maskformer2_R50_bs16_50ep.pth`

Then I tried to test the model with [this config](https://github.com/chhluo/mmdetection/blob/rename_mask2former_configs/configs/mask2former/mask2former_r50_lsj_8x2_50e_coco.py):
```bash
bash tools/dist_test.sh configs/mask2former/mask2former_r50_lsj_8x2_50e_coco.py maskformer2_R50_bs16_50ep.pth 8 --eval bbox segm
```

However, an error happened:
```
  File "tools/test.py", line 275, in <module>
    main()
  File "tools/test.py", line 267, in main
    metric = dataset.evaluate(outputs, **eval_kwargs)
  File "/home/tiger/mmdetection/mmdet/datasets/coco.py", line 641, in evaluate
    result_files, tmp_dir = self.format_results(results, jsonfile_prefix)
  File "/home/tiger/mmdetection/mmdet/datasets/coco.py", line 383, in format_results
    result_files = self.results2json(results, jsonfile_prefix)
  File "/home/tiger/mmdetection/mmdet/datasets/coco.py", line 331, in results2json
    raise TypeError('invalid type of results')
TypeError: invalid type of results
```

mmdet/datasets/pipelines/loading.py
PeterVennerstrom: Fixed, thanks

PeterVennerstrom:
@mitming the maskformer head has been adjusted with similar code to yours [here](https://github.com/PeterVennerstrom/mmdetection/blob/a947f61bab1ef2e41fe7ce108cdaf62489638b91/mmdet/models/detectors/maskformer.py#L174) to parse the results when operating in instance segmentation mode.  Please check if you are on the instance_only branch.

chhluo:
Please rebase on the dev branch.

chhluo:
@PeterVennerstrom unit test failed for mask2former instance segmentstion, because we removed the confile file for instance segmentation, please updated it. And sorry about that.

ZwwWayne (approved):

Issue #6788, opened 2021-12-14
Mobilenet v3


WassimBouatay:
I think it is a good idea to add mobilenet v3
It is better than mobilenet v2 and more efficient than most backbones present here.
Are you planning on adding it soon ?

RangiLyu:
Thanks for your suggestion!

WassimBouatay:
My suggestion has been moved to Done. What does this mean ? because I don't see any changes

PR #7586, opened 2022-03-30
[WIP] Support Mobilenetv3 (Draft PR)

HJoonKwon:
## Motivation

There was a [suggestion](https://github.com/open-mmlab/mmdetection/issues/6788) about supporting MobileNetV3 as a backbone in mmdet. And [the card for new feature](https://github.com/open-mmlab/mmdetection/projects/7#card-74874022) was created. 
I made MobileNetV3 backbone by referring to mmclassification and [the original paper](https://arxiv.org/abs/1905.02244) by personal interest. This is just a draft PR, so any feedback would be appreciated. If you are interested, then I will make other parts such as documentation, unit tests, and so on. 

## Modification

Being different from the classification backbone, the number of channels for all features between C4 and C5 was reduced by 2 for the object detection task as mentioned in the paper. Furthermore, the features were extracted from the intermediate expansion layers, not from the end of the InverseResidual blocks.(C4 and C5 are the terms for the feature extracting layers defined by the author) [Paddle Detection](https://github.com/PaddlePaddle/PaddleDetection/blob/release/2.4/ppdet/modeling/backbones/mobilenet_v3.py) already implemented this part so I took out some clues from this source. Except for this, most part is very similar to mmclassification model. The training configuration file is just an example that mimics the data pipeline from [ssdlite-mobilenetv2 config](https://github.com/open-mmlab/mmdetection/blob/master/configs/ssd/ssdlite_mobilenetv2_scratch_600e_coco.py). I haven't finished the training yet because I've been training the model with the pretrained backbone taken from mmclassification first for the comparison. Any training tricks or advice would be very appreciated! :) 

## BC-breaking (Optional)

No

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/utils/inverted_residual.py
+++ b/mmdet/models/utils/inverted_residual.py
@@ -20,6 +20,8 @@ class InvertedResidual(BaseModule):
         stride (int): The stride of the depthwise convolution. Default: 1.
         se_cfg (dict): Config dict for se layer. Default: None, which means no
             se layer.
+        return_list(bool): Return forward output as a list including feature
+            from the expanded layer. Default: False.
         with_expand_conv (bool): Use expand conv or not. If set False,
             mid_channels must be the same with in_channels.
             Default: True.
@@ -47,6 +49,7 @@ class InvertedResidual(BaseModule):
                  stride=1,
                  se_cfg=None,
                  with_expand_conv=True,
+                 return_list=False,
                  conv_cfg=None,
                  norm_cfg=dict(type='BN'),
                  act_cfg=dict(type='ReLU'),
@@ -62,6 +65,7 @@ class InvertedResidual(BaseModule):
             drop_path_rate) if drop_path_rate > 0 else nn.Identity()
         self.with_se = se_cfg is not None
         self.with_expand_conv = with_expand_conv
+        self.return_list = return_list
 
         if self.with_se:
             assert isinstance(se_cfg, dict)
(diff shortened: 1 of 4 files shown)


hhaAndroid:
@HJoonKwon Thank you very much for your contribution. Do we need to train MobileNetV3 from scratch due to modification of the backbone network?

HJoonKwon:
@hhaAndroid Yes, I think so!

mmdet/models/utils/inverted_residual.py
```diff
@@ -119,6 +124,9 @@ def _inner_forward(x):
 
             if self.with_res_shortcut:
                 return x + self.drop_path(out)
+
+            if self.return_list:
+                return [expanded_out, out]
```
jbwang1997, line 129: self.with_res_shortcut may shortcut self.return_list branch.

HJoonKwon: Yes.. I think I should fix it for generality. For now, self.with_res_shortcut is always False when self.return_list is True because the stride value is 2.

jbwang1997:
@HJoonKwon.
Hello. I see this pr still has a label of WIP. Do you have progress?

Regards,

HJoonKwon:
@jbwang1997 Sorry for the late update. I have trained the small model from scratch, but it only reached ~14 mAP on COCO dataset or so. I think I should go more above, so it should be pretrained with ImageNet and whatnot.. It's been delayed since I lack GPU resource. But I will definitely try (if it's not urgent) it.

jbwang1997:
It's not urgent. you can manage progress based on your schedule. by the way, mmclassification also has pretrained moblienet-v3 on ImageNet, you can directly load it in mmdetection.

Thanks again for your valuable contributions.

PR #7633, opened 2022-04-04
add solo style dicecost and modify MaskHungarianAssigner

FANG-MING:

## Motivation

add solo style `DiceCost`, as used in `K-net`, and fixed an error in `MaskHungarianAssigner` when `cls_pred` is `None`.

## Modification

add a `solo_style` option for diceloss init.

## BC-breaking (Optional)

never

## Use cases (Optional)

we can add `solo_style_dice_loss option  ` which is used in `SOLO` and `K-Net` when we init `DiceLoss`

### DIFF
--- a/mmdet/core/bbox/match_costs/match_cost.py
+++ b/mmdet/core/bbox/match_costs/match_cost.py
@@ -246,11 +246,11 @@ class DiceCost:
         eps (float, optional): default 1e-12.
     """
 
-    def __init__(self, weight=1., pred_act=False, eps=1e-3, solo_style_dice_loss=False):
+    def __init__(self, weight=1., pred_act=False, eps=1e-3, solo_style=False):
         self.weight = weight
         self.pred_act = pred_act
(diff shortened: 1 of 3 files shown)


mmdet/core/bbox/match_costs/match_cost.py
```diff
@@ -246,10 +246,11 @@ class DiceCost:
         eps (float, optional): default 1e-12.
     """
 
-    def __init__(self, weight=1., pred_act=False, eps=1e-3):
+    def __init__(self, weight=1., pred_act=False, eps=1e-3, solo_style=False):
         self.weight = weight
         self.pred_act = pred_act
         self.eps = eps
+        self.solo_style = solo_style
```
ZwwWayne, line 253: solo_style is not an appropriate name. You may check the name of option in dice loss. https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/dice_loss.py

PR #7668, opened 2022-04-08
[Feature] Support set dataloader args in config and and add function to handle config compatibility

jshilong:
## Describe the feature
- Add train_dataloader, val_dataloader, and test_dataloader in the data field of config so that users can specify different arguments for them.
- Add `cfg_compatibility` to handle the compatibility of config

## Motivation

Originally, the data field of config only allows the following usage:
```python
data = dict(
    samples_per_gpu=64, workers_per_gpu=4,
    train=dict(type='xxx', ...),
    val=dict(type='xxx', samples_per_gpu=4, ...),
    test=dict(type='xxx', ...),
)
```
Such a design only allows setting the different batch sizes in Val and test data loader.
Adding train_dataloader, val_dataloader, and test_dataloader in the data field of config allows more flexible usages.
A new example looks like below:
```python
data = dict(
    train=dict(type='xxx', ...),
    val=dict(type='xxx', ...),
    test=dict(type='xxx', ...),
    # Use different batch size during inference.
    train_dataloader=dict(samples_per_gpu=64, workers_per_gpu=4),
    val_dataloader=dict(samples_per_gpu=8, workers_per_gpu=2),
    test_dataloader=dict(samples_per_gpu=8, workers_per_gpu=2),
)
```
To achieve that, we should:

allow train_dataloader/val_dataloader/test_dataloader to be set in the data filed by modifying train_detector.
deprecate samples_per_gpu in data.val field by raising warnings if it is set
Related resources
See PR https://github.com/open-mmlab/mmclassification/pull/752.

## Midification
1. modifying train_detector to allow passing args in the loader field to the `build_dataloader`
2.  Add a `cfg_compatibility` to handle the compatibility

## BC
None


configs/_base_/datasets/coco_detection.py
```diff
@@ -29,6 +29,7 @@
         ])
 ]
 data = dict(
+    # default value for all dataloader
     samples_per_gpu=2,
```
ZwwWayne, line 33: We should deprecate this usage, so we can clean lines 32&33 in our own configs. Users will meet warning if they do not change their configs.

mmdet/apis/train.py
```diff
@@ -95,19 +95,30 @@ def train_detector(model,
...
+        runner_type=runner_type,
+        persistent_workers=False)
+    # update overall dataloader(for train, val and test) setting
+    train_dataloader_default_args.update({
+        k: v
+        for k, v in cfg.data.items() if k not in [
+            'train', 'val', 'test', 'train_dataloader', 'val_dataloader',
+            'test_dataloader'
+        ]
+    })
```
ZwwWayne, line 115: should raise warning if samples_per_gpu is in val

mmdet/apis/train.py
```diff
@@ -95,19 +95,30 @@ def train_detector(model,
...
+    train_loader_cfg = {
+        **train_dataloader_default_args,
+        **cfg.data.get('train_dataloader', {})
```
ZwwWayne, line 118: Will it cause error if train_data_loader has the same keys in train_data_loader_default_args?

tools/test.py
```diff
@@ -188,6 +170,49 @@ def main():
         distributed = True
         init_dist(args.launcher, **cfg.dist_params)
 
+    test_dataloader_default_args = dict(
```
ZwwWayne, line 173: can we extract a function so that the function can be reused in train and test?

tools/test.py
```diff
@@ -188,6 +170,49 @@ def main():
...
+        samples_per_gpu=1, workers_per_gpu=2, dist=distributed, shuffle=False)
+    # update overall dataloader(for train, val and test) setting
+    test_dataloader_default_args.update({
+        k: v
+        for k, v in cfg.data.items() if k not in [
+            'train', 'val', 'test', 'train_dataloader', 'val_dataloader',
+            'test_dataloader', 'samples_per_gpu'
```
VVsssssk, line 180: Do we need to add 'samples_per_gpu'?

jshilong: keep the same behavior as before

mmdet/apis/train.py
jshilong: The latter overwrites the former

configs/_base_/datasets/coco_detection.py
jshilong: In our and mmcls‘s design， the value in the data is the default value of all data_loader

ZwwWayne: leaving them here will cause ambiguity. Users will be confused about where should they set samples_per_gpu

ZwwWayne (commented):
Invite @mzr1996 to take a look.

tools/test.py
jshilong: The default keys for the different loader is different

configs/_base_/datasets/coco_detection.py
jshilong: I will add comments for the behavior

tools/test.py
```diff
@@ -188,6 +170,49 @@ def main():
...
+            [ds_cfg.pop('samples_per_gpu', 1) for ds_cfg in cfg.data.test])
+        test_dataloader_default_args['samples_per_gpu'] = samples_per_gpu
+        if samples_per_gpu > 1:
+            for ds_cfg in cfg.data.test:
+                ds_cfg.pipeline = replace_ImageToTensor(ds_cfg.pipeline)
+
+    train_loader_cfg = {
+        **test_dataloader_default_args,
+        **cfg.data.get('train_dataloader', {})
+    }
```
jbwang1997, line 214: Why here use `train_dataloader` but not `test_dataloader`?

jbwang1997, line 214: Why here use `train_dataloader` but not `test_dataloader`?

VVsssssk:
#https://github.com/open-mmlab/mmdetection3d/pull/1388

VVsssssk:
#[Feature]Add loader args to config https://github.com/open-mmlab/mmdetection3d/pull/1388

mmdet/apis/train.py
```diff
@@ -79,35 +78,26 @@ def train_detector(model,
...
+        seed=cfg.seed,
+        runner_type=runner_type,
+        persistent_workers=False)
+
+    train_loader_cfg = {
+        **train_dataloader_default_args,
+        **cfg.data.get('train_dataloader')
+    }
+
+    data_loaders = [build_dataloader(ds, **train_loader_cfg) for ds in dataset]
```
mzr1996, line 100: So this PR doens't support format as below?
```python
data=dict(
    samples_per_gpu=2,
    workers_per_gpu=2,
    train=...
)
```

jshilong: Yes, we decided to deprecate it in the data filed and I keep the compatibility in the function `cfg_compatibility`

mmdet/utils/cfg_compatibility.py
```diff
@@ -0,0 +1,125 @@
...
+                f'"samples_per_gpu"={cfg.data.samples_per_gpu}, "imgs_per_gpu"'
+                f'={cfg.data.imgs_per_gpu} is used in this experiments')
+        else:
+            warnings.warn('Automatically set "samples_per_gpu"="imgs_per_gpu"='
+                          f'{cfg.data.imgs_per_gpu} in this experiments')
+        cfg.data.samples_per_gpu = cfg.data.imgs_per_gpu
+    return cfg
+
+
+def loader_args_compatibility(cfg):
```
RangiLyu, line 52: Need to handle `persistent_workers`, which is used in YOLOX's configs.

mmdet/utils/cfg_compatibility.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import copy
+import warnings
+
+
+def cfg_compatibility(cfg):
```
RangiLyu, line 6: unit tests are required

tests/test_utils/test_compat_conig.py
```diff
@@ -0,0 +1,120 @@
+import pytest
```
VVsssssk, line 1: file name error

tests/test_utils/test_compat_conig.py
```diff
@@ -0,0 +1,120 @@
...
+        dict(
+            data=dict(
+                samples_per_gpu=1,
+                persistent_workers=True,
+                workers_per_gpu=1,
+                val=dict(samples_per_gpu=3),
+                test=dict(samples_per_gpu=2),
+                train=dict(),
+                train_dataloader=dict(samples_per_gpu=2))))
+    with pytest.raises(AssertionError):
```
ZwwWayne, line 83: Add comments to indicate what error is testing

tests/test_utils/test_compat_conig.py
```diff
@@ -0,0 +1,120 @@
...
+        dict(
+            data=dict(
+                samples_per_gpu=1,
+                persistent_workers=True,
+                workers_per_gpu=1,
+                val=dict(samples_per_gpu=3),
+                test=dict(samples_per_gpu=2),
+                train=dict(),
+                val_dataloader=dict(samples_per_gpu=2))))
+    with pytest.raises(AssertionError):
```
ZwwWayne, line 95: Add comments to indicate what error is testing, you can use 
`with pytest.raises(AssertionError, match='error message'):`

mmdet/utils/compat_config.py
```diff
@@ -0,0 +1,175 @@
...
+                      'in `train_dataloader`.')
+        samples_per_gpu = cfg.data.pop('samples_per_gpu')
+        assert 'samples_per_gpu' not in \
+               cfg.data.train_dataloader, ('`samples_per_gpu` are set '
+                                           'in `data` field and ` '
+                                           'data.train_dataloader` '
+                                           'at the same time. '
+                                           'Please only set it in '
+                                           '`data.train_dataloader`. ')
+        cfg.data.train_dataloader['samples_per_gpu'] = samples_per_gpu
```
mzr1996, line 81: I think just overriding the setting in the `data` field is enough, we don't need to raise a warning here.

jshilong: I believe we should follow such a principle that there is only one way when we set a  parameter.

ZwwWayne (approved):

mmdet/utils/compat_config.py
mzr1996: > I believe we should follow such a principle that there is only one way when we set a parameter.

Right, but these parameters are not with the same meaning. The `samples_per_gpu` in the `data` field is the global setting of all of the train/val/test dataloader. I know that in mmdet, `samples_per_gpu` only works for train dataloader, but it works as a global parameter in other codebases.
Another reason is that I think we don't need to deprecate these behaviors since the `data` field may be removed in the recent future.

wlf-darkmatter:
> ## Describe the feature
> * Add train_dataloader, val_dataloader, and test_dataloader in the data field of config so that users can specify different arguments for them.
> * Add `cfg_compatibility` to handle the compatibility of config
> 
> ## Motivation
> Originally, the data field of config only allows the following usage:
> 
> ```python
> data = dict(
>     samples_per_gpu=64, workers_per_gpu=4,
>     train=dict(type='xxx', ...),
>     val=dict(type='xxx', samples_per_gpu=4, ...),
>     test=dict(type='xxx', ...),
> )
> ```
> 
> Such a design only allows setting the different batch sizes in Val and test data loader. Adding train_dataloader, val_dataloader, and test_dataloader in the data field of config allows more flexible usages. A new example looks like below:
> 
> ```python
> data = dict(
>     train=dict(type='xxx', ...),
>     val=dict(type='xxx', ...),
>     test=dict(type='xxx', ...),
>     # Use different batch size during inference.
>     train_dataloader=dict(samples_per_gpu=64, workers_per_gpu=4),
>     val_dataloader=dict(samples_per_gpu=8, workers_per_gpu=2),
>     test_dataloader=dict(samples_per_gpu=8, workers_per_gpu=2),
> )
> ```
> 
> To achieve that, we should:
> 
> allow train_dataloader/val_dataloader/test_dataloader to be set in the data filed by modifying train_detector. deprecate samples_per_gpu in data.val field by raising warnings if it is set Related resources See PR [open-mmlab/mmclassification#752](https://github.com/open-mmlab/mmclassification/pull/752).
> 
> ## Midification
> 1. modifying train_detector to allow passing args in the loader field to the `build_dataloader`
> 2. Add a `cfg_compatibility` to handle the compatibility
> 
> ## BC
> None

It make an error raised when `test_dataloader ` not inside `cfg.data` that 
```
AttributeError: 'ConfigDict' object has no attribute 'test_dataloader'
```
it wou better to add an **if**  to tell if the `cfg.data` get the attribute `test_dataloader` to avoid error.

Issue #7495, opened 2022-03-22
Reading from the cfg.evaluation.gpu_collect field during testing


LutingWang:
**Describe the feature**

During testing, use `gpu_collect = args.gpu_collect or cfg.evaluation.gpu_collect`.

**Motivation**
During testing, the `cfg.evaluation.gpu_collect` does not take effect. Rather, users need to specify `gpu_collect` from the command line.

This behavior is inconsistent with the evaluation during training, where `cfg.evaluation.gpu_collect` decides whether to use `gpu_collect`.

**Additional context**
In `tools/test.py`, line 71-74

```python
    parser.add_argument(
        '--gpu-collect',
        action='store_true',
        help='whether to use gpu to collect results.')
```

In `mmdet/apis/train.py`, line 190-196

```python
        eval_cfg = cfg.get('evaluation', {})
        eval_cfg['by_epoch'] = cfg.runner['type'] != 'IterBasedRunner'
        eval_hook = DistEvalHook if distributed else EvalHook
        # In this PR (https://github.com/open-mmlab/mmcv/pull/1193), the
        # priority of IterTimerHook has been modified from 'NORMAL' to 'LOW'.
        runner.register_hook(
            eval_hook(val_dataloader, **eval_cfg), priority='LOW')
```

For backward compatibility, I think it may be reasonable to use `gpu_collect = args.gpu_collect or cfg.evaluation.gpu_collect` in `tools/test.py` line 238-239

```python
        outputs = multi_gpu_test(model, data_loader, args.tmpdir,
                                 args.gpu_collect)
```

If this seems OK, I can create a PR.

hhaAndroid:
@LutingWang Thanks for your feedback, we'll discuss if changes are needed.

hhaAndroid:
@LutingWang Yes. Your idea is right, Can you create a PR?

PR #7672, opened 2022-04-08
[Enhance] Support reading `gpu_collect` from cfg.evaluation.gpu_collect

LutingWang:
## Motivation

During testing, the `cfg.evaluation.gpu_collect` does not take effect. Rather, users need to specify `gpu_collect` from the command line.

This behavior is inconsistent with the evaluation during training, where `cfg.evaluation.gpu_collect` decides whether to use `gpu_collect`.

## Modification

`gpu_collect = args.gpu_collect or cfg.evaluation.gpu_collect` in `tools/test.py` line 238-239

```python
        outputs = multi_gpu_test(model, data_loader, args.tmpdir,
                                 args.gpu_collect)
```

## Checklist

## Reference

Issue #7495

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -238,7 +238,7 @@ def main():
             device_ids=[torch.cuda.current_device()],
             broadcast_buffers=False)
         outputs = multi_gpu_test(model, data_loader, args.tmpdir,
-                                 args.gpu_collect)
+                                 args.gpu_collect or cfg.evaluation.gpu_collect)
 
     rank, _ = get_dist_info()
     if rank == 0:
--- a/tools/test.py
+++ b/tools/test.py
@@ -237,8 +237,9 @@ def main():
             model.cuda(),
             device_ids=[torch.cuda.current_device()],
             broadcast_buffers=False)
-        outputs = multi_gpu_test(model, data_loader, args.tmpdir,
-                                 args.gpu_collect or cfg.evaluation.gpu_collect)
+        outputs = multi_gpu_test(
+            model, data_loader, args.tmpdir, args.gpu_collect
+            or cfg.evaluation.gpu_collect)
 
     rank, _ = get_dist_info()
     if rank == 0:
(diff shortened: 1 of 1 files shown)


tools/test.py
```diff
@@ -237,8 +237,9 @@ def main():
             model.cuda(),
             device_ids=[torch.cuda.current_device()],
             broadcast_buffers=False)
-        outputs = multi_gpu_test(model, data_loader, args.tmpdir,
-                                 args.gpu_collect)
+        outputs = multi_gpu_test(
+            model, data_loader, args.tmpdir, args.gpu_collect
+            or cfg.evaluation.gpu_collect)
```
ZwwWayne, line 242: what if the cfg.evaluation.gpu_collect is not set? Will it cause error？

ZwwWayne, line 242: We should handle the case when both args.gpu_collect and cfg.evaluation.gpu_collect is set. It might be better that args.gpu_collect has higher priority than cfg.evaluation.gpu_collect

LutingWang: Yes, it will. I'll fix it soon

LutingWang: args.gpu_collect presently cannot be set to false. So if args.gpu_collect is set, it does have a higher priority. 

The case in question is when cfg.evaluation.gpu_collect is true. Users cannot disable gpu_collect through args.gpu_collect. Is this behaviour acceptable?

LutingWang: args.gpu_collect presently cannot be set to false. So if args.gpu_collect is set, it does have a higher priority. 

The case in question is when cfg.evaluation.gpu_collect is true. Users cannot disable gpu_collect through args.gpu_collect. Is this behavior acceptable?

jbwang1997: Yes. I think this behavior is acceptable.

jbwang1997: We can use `args.gpu_collect or cfg.evaluation.get('gpu_collect', False)` here.

jbwang1997 (approved):
LGTM

Issue #7611, opened 2022-04-02
show_result function parameter out_file


wang-kangkang:
https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/inference.py#L245
model.show_result() has a parameter: "out_file", but in inference.py, it's ignored

chhluo:
Yes, it's better to add the param `out_file` in inference.py and `demo/image_demo.py`

PR #7676, opened 2022-04-08
[Enhance] Add param `out_file` in image_demo.py

chhluo:
## Motivation

solve #7611

## Modification

Add param `out_file`

### DIFF
--- a/demo/image_demo.py
+++ b/demo/image_demo.py
@@ -11,6 +11,7 @@ def parse_args():
     parser.add_argument('img', help='Image file')
     parser.add_argument('config', help='Config file')
     parser.add_argument('checkpoint', help='Checkpoint file')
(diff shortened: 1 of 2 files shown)


demo/image_demo.py
```diff
@@ -11,6 +11,7 @@ def parse_args():
     parser.add_argument('img', help='Image file')
     parser.add_argument('config', help='Config file')
     parser.add_argument('checkpoint', help='Checkpoint file')
+    parser.add_argument('--out-file', default='', help='Path to output file')
```
jbwang1997, line 14: Could we directly set the default as `None`?

chhluo: yes.

jbwang1997 (approved):
LGTM

PR #7690, opened 2022-04-11
[Enhance] Add comments for proposals in deformable detr

### LINKED ISSUE #7682: Why dose the Deformable DETR only use the score of the first class to choose the topk proposals?
In the two-stage mode of Deformable DETR, I notice that the code choose the topk foreground proposals only by the score of the first class (In cocodataset wo have 80 calss in total)? Although I notice that Deformable DETR uses sigmoid + BCE loss, I doubt if it is reasonable to choose topk foreground proposals only by the first class score?

               **topk_proposals = torch.topk(enc_outputs_class[..., 0], topk, dim=1)[1]**

chhluo:
## Motivation

fix #7682 

## Use cases (Optional)

### DIFF
--- a/mmdet/models/utils/transformer.py
+++ b/mmdet/models/utils/transformer.py
@@ -1007,6 +1007,10 @@ class DeformableDetrTransformer(Transformer):
                     self.decoder.num_layers](output_memory) + output_proposals
 
             topk = self.two_stage_num_proposals
+            # We only use the first class in enc_outputs_class as foreground,
+            # the other (num_classes - 1) classes are not used.
+            # See https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/dense_heads/deformable_detr_head.py#L241 # noqa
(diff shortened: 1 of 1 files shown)


Czm369:
LGTM

mmdet/models/utils/transformer.py
```diff
@@ -1007,6 +1007,10 @@ def forward(self,
                     self.decoder.num_layers](output_memory) + output_proposals
 
             topk = self.two_stage_num_proposals
+            # We only use the first class in enc_outputs_class as foreground,
```
ZwwWayne, line 1010: We only use the first channel in enc_outputs_class as foreground,
 the other (num_classes - 1) channels are actually not used.
Its targets are set to be 0s, which indicates the first class (foreground) because we use [0, num_classes - 1] to indicate class labels, background class is indicated by num_classes (similar convention in RPN).
This follows the official implementation of Deformable DETR.

PR #7832, opened 2022-04-26
Speedup the Video Inference by Accelerating data-loading Stage

chenxinfeng4:
## Motivation

The video inference was not efficient, because there are many "resize", "padding", "to rgb" and "normalize" operations in CPU workload. Besides, the 'img_metas' is calculated for every frame, which is nonefficient. The CPU workload is really high, but the speed is slow.

## Modification

I transformed the video "crop", "resize", "padding", "to rgb" in **ffmpeg-based video reader,** which is lite and CPU friendly. The video reader can support NVIDIA-VIDEO-DECODING if possible, to save more CPU workload. And the  "normalize" is done in GPU. In all, I reduced the data-loading time and CPU workload.

## Result
After this modification, the CPU workload can reduced from >1000% to 180%, and slightly improve the inference frame rate. It's significant in the case that dataloading is the heaviest cost than the network evalutation, such as the yolact.


chenxinfeng4:
You can compare the original `video_demo.py` with my `video_gpuaccel_demo.py`.  They have some argment inputs.
To see how much improved, please delete the post-operations in both codes. Because the post-operations are also CPU heavy.

````python
# post-operations in `video_demo.py` and my `video_gpuaccel_demo.py`.
model.show_result(xxx)

```

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,101 @@
+import argparse
```
hhaAndroid, line 1: add `# Copyright (c) OpenMMLab. All rights reserved.`

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,101 @@
...
+    parser.add_argument(
+        '--wait-time',
+        type=float,
+        default=1,
+        help='The interval of show (s), 0 is block')
+    args = parser.parse_args()
+    return args
+
+
+def prepare_cfg(model, ori_wh):
```
hhaAndroid, line 35:
```python
def prefetch_img_metas():
    ...
    return img_metas```

File: demo/video_gpuaccel_demo.py [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+    test_pipeline = Compose(cfg.data.test.pipeline)
+    data = {'img':np.zeros((h, w, 3), dtype=np.uint8)}
+    data = test_pipeline(data)
+    data_template = data['img_metas'][0].data
+    mean = torch.tensor(data_template['img_norm_cfg']['mean'])
+    std = torch.tensor(data_template['img_norm_cfg']['std'])
+    return data_template, mean, std
+
+
+def prepare_img(data_template, mean, std, frame):
```
hhaAndroid, line 48:
```python
def process_img(frame, img_metas):
     ...
     return data
```

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,101 @@
...
+    assert frame.shape==data_template['pad_shape']
+    frame_cuda = torch.from_numpy(frame.astype(np.float32)).cuda()
```
hhaAndroid, line 50: Why no need to resize. doesn't seem to run `test_pipeline`?

hhaAndroid:
@chenxinfeng4 Thank you very much for your contribution.

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,101 @@
+import argparse
+
+import cv2
+import mmcv
+import ffmpegcv
```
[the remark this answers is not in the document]
chenxinfeng4: I will fix it

demo/video_gpuaccel_demo.py
chenxinfeng4: I've done the resize  and padding in ffmpegcv

chenxinfeng4:
I tested the speed of data loading. It's about 230FPS , 800%CPU.
```python
    with torch.cuda.device(args.device), torch.no_grad():
        for frame_resize, frame_origin in zip(tqdm.tqdm(video_resize), video_origin):
            data = process_img(frame_resize, img_metas)
            continue
```
image
image

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,105 @@
...
+    parser.add_argument(
+        '--wait-time',
+        type=float,
+        default=1,
+        help='The interval of show (s), 0 is block')
+    args = parser.parse_args()
+    return args
+
+
+def prefetch_img_metas(model, ori_wh):
```
hhaAndroid, line 40: def prefetch_img_metas(cfg, ori_wh):

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,105 @@
...
+    data = test_pipeline(data)
+    img_metas = data['img_metas'][0].data
+    return img_metas
+
+
+def process_img(frame_resize, img_metas):
+    assert frame_resize.shape==img_metas['pad_shape']
+    frame_cuda = torch.from_numpy(frame_resize.astype(np.float32)).cuda()
+    frame_cuda = frame_cuda.permute(2, 0, 1) # HWC to CHW
+    mean = torch.from_numpy(img_metas['img_norm_cfg']['mean']).cuda()
```
hhaAndroid, line 55: it is my mistake. This value cannot be placed inside.

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,105 @@
...
+    w, h = ori_wh
+    cfg.data.test.pipeline[0].type = 'LoadImageFromWebcam'
+    test_pipeline = Compose(cfg.data.test.pipeline)
+    data = {'img':np.zeros((h, w, 3), dtype=np.uint8)}
+    data = test_pipeline(data)
+    img_metas = data['img_metas'][0].data
+    return img_metas
+
+
+def process_img(frame_resize, img_metas):
```
hhaAndroid, line 51: def process_img(frame_resize, mean, std, img_metas):

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,105 @@
...
+    assert frame_resize.shape==img_metas['pad_shape']
+    frame_cuda = torch.from_numpy(frame_resize.astype(np.float32)).cuda()
```
hhaAndroid, line 53: No need to convert to float, it must be float when norm is calculated.

`frame_cuda = torch.from_numpy(frame_resize).cuda()`

demo/video_gpuaccel_demo.py
chenxinfeng4: Inside place of `mean` and `std` won't affect the speed. I would like to keep that change.

demo/video_gpuaccel_demo.py
chenxinfeng4: The F.normalize need the float data type. 
```python
frame_cuda = torch.from_numpy(frame_resize).cuda()
frame_cuda = F.normalize(frame_cuda, mean=mean, std=std)
```
>Traceback (most recent call last):
  File "demo/video_gpuaccel_demo.py", line 103, in <module>
    main()
  File "demo/video_gpuaccel_demo.py", line 85, in main
    data = process_img(frame_resize, img_metas)
  File "demo/video_gpuaccel_demo.py", line 56, in process_img
    frame_cuda = F.normalize(frame_cuda, mean=mean, std=std)
  File "/home/liying_lab/chenxinfeng/.conda/envs/cbnet/lib/python3.7/site-packages/torchvision/transforms/functional.py", line 298, in normalize
    tensor.sub_(mean).div_(std)
RuntimeError: result type Float can't be cast to the desired output type Byte

I would keep the `frame_cuda = torch.from_numpy(frame_resize.astype(np.float32)).cuda()`

hhaAndroid: Then you can replace it with the following
`frame_cuda = torch.from_numpy(frame_resize).float().cuda()`

demo/video_gpuaccel_demo.py
hhaAndroid: Maybe because the amount of data is too small.  If other people's computers have poor performance, then it should have an impact. It is recommended to remove.

demo/video_gpuaccel_demo.py
hhaAndroid: And. Is it possible for this program to run on CPU？

chenxinfeng4: OK, I will swich to ` torch.from_numpy(a).cuda().float()`
```
    b = torch.from_numpy(a).float().cuda()     # slow; CPU 2,000%
    d = torch.from_numpy(a.astype(numpy.float32)).cuda() # fast; CPU 100%
    c = torch.from_numpy(a).cuda().float()    # fastest; CPU 100%
```

chenxinfeng4: This program can run in CPU by replacing the '.cuda() ' to '.to(device)' .

hhaAndroid: en. So it's better to change it to be more general.

ZwwWayne:
lint failed.

demo/video_gpuaccel_demo.py
RangiLyu: > en. So it's better to change it to be more general.

This script is called "gpuaccel_demo", so I think cuda only is acceptable.

RangiLyu (commented):
You can add some documents about this script in https://github.com/open-mmlab/mmdetection/blob/master/docs/en/1_exist_data_model.md

RangiLyu:
Please use pre-commit hook to fix the lint.

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,104 @@
...
+                                resize_keepratio = True,
+                                resize_keepratioalign = 'topleft',
+                                pix_fmt='rgb24')
+    video_writer = None
+    if args.out:
+        video_writer = ffmpegcv.VideoWriter(args.out, fps=video_origin.fps)
+
+    with torch.cuda.device(args.device), torch.no_grad():
+        for frame_resize, frame_origin in zip(tqdm.tqdm(video_resize), video_origin):
+            data = process_img(frame_resize, img_metas)
```
jbwang1997, line 86: `tqdm` is not a required three-part package in mmdet. We can use `mmcv.track_iter_progress` instead.

demo/video_gpuaccel_demo.py
```diff
@@ -0,0 +1,104 @@
...
+
+try:
+    import ffmpegcv
+except ImportError:
+    raise ImportError('Please install ffmpegcv with:\n\n    pip install ffmpegcv')
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='MMDetection video demo')
+    parser.add_argument('video', help='Video file')
```
jbwang1997, line 21: MMDetection video demo with GPU acceleration.

jbwang1997, line 21: MMDetection video demo with GPU acceleration.

hhaAndroid (approved):

hhaAndroid:
@chenxinfeng4 Thanks for your very fast response.

jbwang1997 (approved):
LGTM

ZwwWayne (approved):

chenxinfeng4:
I'm not an expert of git. Why the PR is block? What should I do next?

jbwang1997:
> I'm not an expert of git. Why the PR is block? What should I do next?

Hello @chenxinfeng4. 

Merging is controlled by maintainers. We will merge this pr as soon as possible. Seems the lint fails again. Please remember to fix it. 

Thanks for your quick response.

RangiLyu (approved):

ZwwWayne (approved):

PR #7891, opened 2022-05-02
[Enhance] Accelerate analyze result

### LINKED ISSUE #7631: run the tools/analysis_tools/analyze_results.py is so slow ,Is that normal？i feel it caculates with cpu.
```
(swin-t) E:\Object-Detection\Github\radar-detection>python tools/analysis_tools/analyze_results.py checkpoints/tood/devkit3_1calss_e500_af_r50_dcn_db/radar_tood.py checkpoints/tood/devkit3_1calss_e500_af_r50_dcn_db/pkl_and_cm/tood_95.pkl checkpoints/tood/devkit3_1calss_e500_af_r50_dcn_db/pkl_and_cm --show-score-thr 0.7
Done (t=0.00s)
creating index...
index created!
[>>>>>>>                                           ] 4/28, 0.0 task/s, elapsed: 209s, ETA:  1256s
```

chhluo:

## Motivation

Fix #7631

The time spent on [`eval_fn`](https://github.com/open-mmlab/mmdetection/blob/73b4e65a6a30435ef6a35f405e3474a4d9cfb234/tools/analysis_tools/analyze_results.py#L130) costs 1.5s, and on ['dataset.prepare_train_img(i)'](https://github.com/open-mmlab/mmdetection/blob/73b4e65a6a30435ef6a35f405e3474a4d9cfb234/tools/analysis_tools/analyze_results.py#L129) costs 0.01~0.05s. It is clear that we should accelerate the `eval_fn`. There are three nested loops in eval_fn: a loop for iou_thresh(0.5 ~ 0.95), a nested loop for classes, a nested loop for images. And the multi-processes design for fptp computing of list of images does not have expired effect when only one image passed in one time. The multi-proccesses design is unnecessary in this condition (one image a time).  But multi-proccesses design to compute mAP with different iou_thresh is necessary and effective.

image

After optimization, the time spend on `eval_fn` is reduced to less then one fifteenth of the original when `nproc = 4` (which is set by defualt). The evaluation time on results generated by retinanet_r50_fpn_1x_coco takes 10~15 mins.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/core/evaluation/mean_ap.py
+++ b/mmdet/core/evaluation/mean_ap.py
@@ -588,6 +588,11 @@ def eval_map(det_results,
     area_ranges = ([(rg[0]**2, rg[1]**2) for rg in scale_ranges]
                    if scale_ranges is not None else None)
 
+    if num_imgs > 1:
+        assert nproc > 0, 'nproc must be at least one.'
+        nproc = min(nproc, num_imgs)
+        pool = Pool(nproc)
+
     eval_results = []
     for i in range(num_classes):
         # get gt and det bboxes of this class
@@ -617,16 +622,12 @@ def eval_map(det_results,
             if ioa_thr is not None:
                 args.append([ioa_thr for _ in range(num_imgs)])
 
-            assert nproc > 0, 'nproc must be at least one.'
-            nproc = min(nproc, num_imgs)
-            pool = Pool(nproc)
             tpfp = pool.starmap(
                 tpfp_fn,
                 zip(cls_dets, cls_gts, cls_gts_ignore,
                     [iou_thr for _ in range(num_imgs)],
                     [area_ranges for _ in range(num_imgs)],
                     [use_legacy_coordinate for _ in range(num_imgs)], *args))
-            pool.close()
         else:
             # There is no need to use multi processes to process
             # when num_imgs = 1 .
@@ -685,6 +686,9 @@ def eval_map(det_results,
             'precision': precisions,
             'ap': ap
         })
+
+    if num_imgs > 1:
+        pool.close()
 
     if scale_ranges is not None:
         # shape (num_classes, num_scales)
--- a/mmdet/core/evaluation/mean_ap.py
+++ b/mmdet/core/evaluation/mean_ap.py
@@ -62,7 +62,8 @@ def tpfp_imagenet(det_bboxes,
                   gt_bboxes_ignore=None,
                   default_iou_thr=0.5,
                   area_ranges=None,
-                  use_legacy_coordinate=False):
+                  use_legacy_coordinate=False,
+                  **kwargs):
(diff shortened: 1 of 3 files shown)


mmdet/core/evaluation/mean_ap.py
```diff
@@ -611,13 +617,22 @@ def eval_map(det_results,
...
+                    [iou_thr for _ in range(num_imgs)],
+                    [area_ranges for _ in range(num_imgs)],
+                    [use_legacy_coordinate for _ in range(num_imgs)], *args))
+        else:
+            tpfp = tpfp_fn(cls_dets[0], cls_gts[0], cls_gts_ignore[0], iou_thr,
+                           area_ranges, use_legacy_coordinate)
+            tpfp = [
+                tpfp,
+            ]
+
```
jbwang1997, line 635: Can merge all num_imgs > 1 branches together as below.

```
if num_imgs > 1:
    # compute tp and fp for each image with multiple processes
    assert nproc > 0, 'nproc must be at least one.'
    pool = Pool(min(nproc, num_imgs))
    tpfp = pool.starmap(
        tpfp_fn,
        zip(cls_dets, cls_gts, cls_gts_ignore,
             [iou_thr for _ in range(num_imgs)],
             [area_ranges for _ in range(num_imgs)],
             [use_legacy_coordinate for _ in range(num_imgs)], *args))
    pool.close()
else:
    tpfp = tpfp_fn(cls_dets[0], cls_gts[0], cls_gts_ignore[0], iou_thr,
                   area_ranges, use_legacy_coordinate)
    tpfp = [
        tpfp,
    ]
```

tools/analysis_tools/analyze_results.py
```diff
@@ -39,11 +43,24 @@ def bbox_map_eval(det_result, annotation):
...
+        })
+        processes.append(p)
+
+    workers.close()
+    workers.join()
+
+    mean_aps = []
+    for p in processes:
+        mean_aps.append(p.get()[0])
+
```
jbwang1997, line 63: Can use pool.starmap instead.

chhluo: Passing parameters is cumbersome for pool.starmap here.
The result of 'apply_async' is out of order but more faster than `apply`.

jbwang1997 (approved):
LGTM

mmdet/core/evaluation/mean_ap.py
chhluo: The code style looks good, but creating process pool in the loop is of low efficiency and unnecessary. I test this.

image

So, I think the original way is better.

ZwwWayne:
Should try to add more unit tests to make sure that the AP calculation is not changed after modification.

Issue #7728, opened 2022-04-15
convert List[NDArray] to NDArray before converting to a tensor


hubutui:
**Describe the feature**
https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/models/dense_heads/yolox_head.py#L252-L286
https://github.com/open-mmlab/mmdetection/blob/3e2693151add9b5d6db99b944da020cba837266b/mmdet/models/dense_heads/yolo_head.py#L238-L270

`scale_factors` is a list of `np.ndarray`, it would be better to convert it a `nd.array` before converted to tensor as the warning described.

**Additional context**
```text
/usr/lib/python3.10/site-packages/mmdet/models/dense_heads/yolo_head.py:269: UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at  /build/python-pytorch/src/pytorch-1.11.0-cuda/torch/csrc/utils/tensor_new.cpp:207.)
```

BIGWangYuDong:
yes, you are right. PR is welcomed.

DLRook1e:
I also encountered the same problem. How can I solve it？

austinmw:
@BIGWangYuDong #7921

PR #7921, opened 2022-05-06
[Enhance] convert scale_factors from list to np.ndarray to prevent UserWarning

austinmw:
## Motivation

#7728

## Modification

Wrap lists in `np.array`

### DIFF
--- a/mmdet/models/dense_heads/yolo_head.py
+++ b/mmdet/models/dense_heads/yolo_head.py
@@ -3,6 +3,7 @@
 
 import warnings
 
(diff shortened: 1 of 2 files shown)


jbwang1997:
Thanks for your contribution. Seems the lint failed. You can refer to [this document](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to install pre-commit, which can automatically fix the lint error before commitment.

austinmw:
@jbwang1997 Thanks just created #7937

PR #8068, opened 2022-05-27
Allow category/single instance options in video demo

pallgeuer:
## Motivation

I want with the standard video demo to be able to set that only detections of a particular category are visualised (removing clutter of other visualised detections for example so the result can be better evaluated). I also want to be able to show only the single highest scoring detection of a particular class (useful if you for example know that you only care about the most prominent person in the video or whatever).

## Modification

I modified `demo/video_demo.py` to include the said features.

## Checklist

I used pre-commit etc. I tested using a command like:
```
python demo/video_demo.py VIDEO.mp4 configs/yolox/yolox_tiny_8x8_300e_coco.py https://download.openmmlab.com/mmdetection/v2.0/yolox/yolox_tiny_8x8_300e_coco/yolox_tiny_8x8_300e_coco_20211124_171234-b4047906.pth --score-thr 0.1 --show --wait-time 0 [--cat-id 57] [--single-inst]
```

### DIFF
--- a/demo/video_demo.py
+++ b/demo/video_demo.py
@@ -23,6 +23,16 @@ def parse_args():
         type=float,
         default=1,
         help='The interval of show (s), 0 is block')
+    parser.add_argument(
+        '--cat-id',
+        type=int,
+        default=0,
+        help='Category ID to visualise (0 => all)')
+    parser.add_argument(
+        '--single-inst',
+        action='store_true',
+        help=('For each category ID, only visualise the single maximum '
+              'scoring instance'))
     args = parser.parse_args()
     return args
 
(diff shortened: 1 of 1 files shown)


pallgeuer:
I get an empty box when I try to open the CLA in order to sign it...

pallgeuer:
I have signed the CLA successfully

demo/video_demo.py
```diff
@@ -23,6 +23,16 @@ def parse_args():
         type=float,
         default=1,
         help='The interval of show (s), 0 is block')
+    parser.add_argument(
+        '--cat-id',
+        type=int,
```
ZwwWayne, line 28: would it be better if it can be a list?

demo/video_demo.py
```diff
@@ -23,6 +23,16 @@ def parse_args():
...
+        default=0,
+        help='Category ID to visualise (0 => all)')
+    parser.add_argument(
+        '--single-inst',
```
ZwwWayne, line 32: --max-score-only would be more precise.

demo/video_demo.py
pallgeuer: I was being consistent with the way mmpose had an argument like that, but I agree that for mmdetection it would be better to be more flexible with the classes. My new commit addresses this (and I tested it).

pallgeuer:
I have updated the PR based on the comments. Is it good to merge now?

pallgeuer:
Any news on this?

PR #8090, opened 2022-05-30
Fix warning with torch.meshgrid

pallgeuer:
## Motivation

`torch.meshgrid()` has started raising a warning when not called with an explicit indexing parameter:
https://pytorch.org/docs/stable/generated/torch.meshgrid.html

```
python demo/image_demo.py demo/demo.jpg configs/yolox/yolox_s_8x8_300e_coco.py https://download.openmmlab.com/mmdetection/v2.0/yolox/yolox_s_8x8_300e_coco/yolox_s_8x8_300e_coco_20211121_095711-4592a793.pth
```
gives
```
/home/allgeuer/anaconda3/envs/mmdev/lib/python3.9/site-packages/torch/functional.py:445: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2157.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
```

## Modification

Provide `indexing='ij'` to all calls to `torch.meshgrid` that don't already specify `indexing`.

### DIFF
--- a/mmdet/utils/misc.py
+++ b/mmdet/utils/misc.py
@@ -6,8 +6,7 @@ import warnings
 
 import mmcv
 import torch
-from mmcv.utils import print_log
-from packaging import version
+from mmcv.utils import TORCH_VERSION, digit_version, print_log
 
 
 def find_latest_checkpoint(path, suffix='pth'):
@@ -78,8 +77,9 @@ def update_data_root(cfg, logger=None):
     cfg.data_root = dst_root
 
 
-_torch_version_meshgrid_indexing = version.parse(
-    torch.__version__) >= version.parse('1.10.0a0')
+_torch_version_meshgrid_indexing = (
+    'parrots' not in TORCH_VERSION
+    and digit_version(TORCH_VERSION) >= digit_version('1.10.0a0'))
 
 
 def torch_meshgrid_ij(*tensors):
--- a/mmdet/utils/misc.py
+++ b/mmdet/utils/misc.py
@@ -5,7 +5,9 @@ import os.path as osp
 import warnings
 
 import mmcv
+import torch
 from mmcv.utils import print_log
+from packaging import version
 
 
 def find_latest_checkpoint(path, suffix='pth'):
@@ -74,3 +76,14 @@ def update_data_root(cfg, logger=None):
 
     update(cfg.data, cfg.data_root, dst_root)
     cfg.data_root = dst_root
+
+
+_torch_version_meshgrid_indexing = version.parse(
+    torch.__version__) >= version.parse('1.10.0a0')
+
+
+def torch_meshgrid_ij(*tensors):
+    if _torch_version_meshgrid_indexing:
+        return torch.meshgrid(*tensors, indexing='ij')
+    else:
+        return torch.meshgrid(*tensors)  # Uses indexing='ij' by default
--- a/mmdet/core/anchor/point_generator.py
+++ b/mmdet/core/anchor/point_generator.py
@@ -3,6 +3,7 @@ import numpy as np
 import torch
 from torch.nn.modules.utils import _pair
 
+from mmdet.utils.misc import torch_meshgrid_ij
 from .builder import PRIOR_GENERATORS
 
 
@@ -68,7 +69,7 @@ class MlvlPointGenerator:
         return [1 for _ in range(len(self.strides))]
 
     def _meshgrid(self, x, y, row_major=True):
-        yy, xx = torch.meshgrid(y, x, indexing='ij')
+        yy, xx = torch_meshgrid_ij(y, x)
         if row_major:
             # warning .flatten() would cause error in ONNX exporting
             # have to use reshape here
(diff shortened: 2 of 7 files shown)


ZwwWayne:
Hi @pallgeuer ,
Thanks for your kind contribution. Since PyTorch only support `indexing` argument of meshgrid 1.10. Simply update the interface will cause mmdet cannot run on PyTorch<1.10. However, we need to support PyTorch >=1.5. Could you implement a meshgrid function to wrap torch.meshgrid? In this function, you can handle both cases so that  mmdet can still run on PyTorch < 1.10.
For example, implement and use the function below:
```
def meshgrid(x, y, indexing=None):
    # if pytorch version < 1.10:
    # use torch.meshgrid
    # else: use torch.meshgrid(x, y, indexing=xx) to avoid warning.
```

pallgeuer:
Good point. This affects many packages (mmdetection, mmpose, mmclassification, etc). How to best deal with that?
Create a wrapper for torch.meshgrid in mmcv?

ZwwWayne:
> Good point. This affects many packages (mmdetection, mmpose, mmclassification, etc). How to best deal with that?
> 
> Create a wrapper for torch.meshgrid in mmcv?

Since we do not want to increase the dependency on mmcv so fast, we can add a wrapper in mmcv first, then we can also copy the wrapper in mmdet and use that in mmcv if the mmcv version meets our demand, otherwise we can use that in mmdet.

pallgeuer:
I have updated the PR with the same solution as was merged for mmpose:
https://github.com/open-mmlab/mmpose/pull/1402

mmdet/utils/misc.py
```diff
@@ -74,3 +76,14 @@ def update(cfg, src_str, dst_str):
 
     update(cfg.data, cfg.data_root, dst_root)
     cfg.data_root = dst_root
+
+
+_torch_version_meshgrid_indexing = version.parse(
```
ZwwWayne, line 81: Following the version comparison style in MMCV https://github.com/open-mmlab/mmcv/blob/235c0253ab8806a2a2ee6954b4258a95358497ac/mmcv/parallel/distributed.py#L40, it might be better to write the code as below:
```python
from mmcv.utils import TORCH_VERSION, digit_version
if ('parrots' not in TORCH_VERSION
      and digit_version(TORCH_VERSION) >= digit_version('1.10')

```

ZwwWayne: > I can't seem to find much information on the parrots version of PyTorch - what should it do in that case? Does the `torch.meshgrid` method for parrots PyTorch support the indexing parameter or not and/or does it raise a warning?
> 
> I had a look and couldn't immediately identify what exactly `digit_version` does differently. In what case would the result of a version comparison like I coded it and your proposal differ?
> 
> `digit_version('1.10')` is not quite sufficient by the way, it does really need to be `digit_version('1.10.0a0')`. My own PyTorch installation is exactly inbetween those two and it makes a difference.

It is an internal version that we support. You are right that digit_version does not do anything different, it is just a convention. Then how about using `digit_version('1.10.0a0')`?

pallgeuer: I have updated the PR

ZwwWayne:
Hi. @pallgeuer ,
We have discussed the naming issue and wish the code to be changed as below:

```python

_torch_version_meshgrid_indexing = (
    'parrots' not in TORCH_VERSION
    and digit_version(TORCH_VERSION) >= digit_version('1.10.0a0'))


def torch_meshgrid(*tensors):
    """A wrapper of torch.meshgrid to compat different PyTorch versions.
    
    Since PyTorch 1.10.0a0, torch.meshgrid supports the arguments ``indexing``.
    So we implement a wrapper here to avoid warning when using high-version PyTorch and avoid compatibility issues when using previous versions of PyTorch.

    Args:
        xxx: (add description here for each argument)

    Return:
        
    """
    if _torch_version_meshgrid_indexing:
        return torch.meshgrid(*tensors, indexing='ij')
    else:
        return torch.meshgrid(*tensors)  # Uses indexing='ij' by default
```

To reduce the repeat work on your side, we suggest you directly update the PR in MMCV and we will review and merge it quickly.
Then we can help to update those interfaces in the remaining repositories like mmdet/mmpose/mmocr.

Thank you!

pallgeuer:
I have updated the mmcv PR.
I guess this means that the mmpose/etc packages will all have to depend on the latest version of mmcv soon then.

ZwwWayne:
Hi @pallgeuer ,
Thanks for your kind and patient contribution. I can imagine that it is not easy to process multiple similar PR across repos.
We will take over the following progress in MMCV and downstream repos. Furthermore, we will update our contribution documentation to make it more friendly to the community.

ZwwWayne:
MMCV already supports torch_meshgrid wrapper since 1.6.0 https://github.com/open-mmlab/mmcv/releases/tag/v1.6.0. We can upgrade the minimum requirement of MMCV since this PR.

Issue #3737, opened 2020-09-11
GCNet pre-trained model download failed


jianjiandandande:
I encountered this problem ' You have no right to access this object because of bucket acl.' when I downloaded the following pre-trained model.
![image](https://user-images.githubusercontent.com/15951818/92844751-c7ff2100-f418-11ea-9d93-65c3d9dbf454.png)

xvjiarui:
Hi @jianjiandandande 
I just uploaded them. You may check it.

PR #8103, opened 2022-06-01
Fix ResourceWarning in Wandb and Update README.md

hhaAndroid:

## Motivation

If the wandb version is lower than 0.12.10, the `wandb.log` will always print a [ResourceWarning](https://github.com/wandb/client/issues/2837). 

## Modification

Add wandb version judgment logic

## BC-breaking (Optional)

None

### DIFF
--- a/mmdet/core/hook/wandblogger_hook.py
+++ b/mmdet/core/hook/wandblogger_hook.py
@@ -10,6 +10,7 @@ from mmcv.runner import HOOKS
 from mmcv.runner.dist_utils import master_only
 from mmcv.runner.hooks.checkpoint import CheckpointHook
 from mmcv.runner.hooks.logger.wandb import WandbLoggerHook
+from mmcv.utils import digit_version
 
 from mmdet.core import DistEvalHook, EvalHook
 from mmdet.core.mask.structures import polygon_to_bitmap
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,22 +73,42 @@ The master branch works with **PyTorch 1.5+**.
...
+**2.25.0** was released in 1/6/2022:
 
-## Changelog
-
-**2.24.1** was released in 30/4/2022:
-
-- Support [Simple Copy Paste](configs/simple_copy_paste)
-- Support automatically scaling LR according to GPU number and samples per GPU
-- Support Class Aware Sampler that improves performance on OpenImages Dataset
+- Support dedicated `WandbLogger` hook
```
jbwang1997, line 80: `MMDetWandbHook`

mmdet/core/hook/wandblogger_hook.py
```diff
@@ -109,6 +110,23 @@ def __init__(self,
...
 
+    def import_wandb(self):
+        try:
+            import wandb
+            from wandb import init  # noqa
+
+            # Fix ResourceWarning when calling wandb.log in wandb v0.12.10.
+            # https://github.com/wandb/client/issues/2837
+            if digit_version(wandb.__version__) < digit_version('0.12.10'):
+                raise RuntimeError('The wandb version is required to be '
```
ZwwWayne, line 121: If we simply want to avoid warning, we should not raise runtimeerror but just raise a warning here.

hhaAndroid: OK

vansin:
Could wandb give some discount for mmdetection user to Subscriptions, and support the Alipay to pay? #3737

PR #8120, opened 2022-06-02
Add anchor pruning to projects documentation + Discussion about supporting the method in open-mmlab toolbox.

Mxbonn:
Dear MMDetection maintainers,

I have recently published a journal article about "Anchor Pruning for Object Detection." (https://doi.org/10.1016/j.cviu.2022.103445). As it was built using MMDetection I would like to contribute back to the main repository.

This pull requests consists of 2 parts:
Part 1 is the request to be added to the project list as can be seen in the changes of this pull request.

Part 2 is a discussion about whether you're open to supporting anchor pruning in the open-mmlab toolbox.
This would require 2 changes: 1) my adaption makes the AnchorGenerator more general to freely define each anchor in anchor-based object detectors. This would be a change that is fully compatible with all current models released by MMDetection and it would also open up the opportunity to match the SSD-Lite anchors precisely to the ones in the official TF repository.
2) The actual anchor pruning algorithm does seem to be a good candidate to be merged into the MMRazor toolbox. This would require a rewrite of my algorithm to match the templates of this repository but if you're open to supporting my pruning method than I'm happy to adapt my code.

Kind regards,
Maxim Bonnaerens


ZwwWayne:
Hi @Mxbonn ,
Thanks for your kind PR. It would be great if you would like to contribute.
I think the idea is great, we can update the AnchorGenerator to make it more general but compatible with the original functionality first. For the pruning part, it might need more consideration and I tend to migrating it to MMRazor for now.

ZwwWayne:
I will merge this PR first. We can create a new issue in MMRazor to discuss about that.

ZwwWayne (approved):

Issue #8110, opened 2022-06-01
Potential bugs when mmdetection runs on PyTorch < 1.8


serend1p1ty:
Thanks for your error report and we appreciate it a lot.

**Describe the bug**

There is a small [bug](https://github.com/pytorch/pytorch/issues/5059) of  PyTorch (version < 1.8).

**In short, when we use `DataLoadr` with `num_workers != 0`, these forked children process share the same `numpy` random seeds.**

A toy example, run with PyTorch 1.6.

```
import numpy as np
from torch.utils.data import Dataset, DataLoader

class RandomDataset(Dataset):
    def __getitem__(self, index):
        return np.random.randint(0, 1000, 3)

    def __len__(self):
        return 8

dataset = RandomDataset()
dataloader = DataLoader(dataset, batch_size=2, num_workers=2)
for batch in dataloader:
    print(batch)
```

Output:

```
tensor([[437, 650, 998],  # process 0
        [108,  34, 197]])
tensor([[437, 650, 998],  # process 1
        [108,  34, 197]])
tensor([[153, 629, 103],  # process 0
        [695, 102, 728]])
tensor([[153, 629, 103],  # process 1
        [695, 102, 728]])
```

Currently, mmdetection sets different `numpy` seeds for different workers according to the `worker_id`. However, the worker processes are killed at the end of each epoch, and all worker resources are lost. At the next epoch, mmdetection will set the same seeds as the previous epoch.

Example:

```
import numpy as np
from torch.utils.data import Dataset, DataLoader

class RandomDataset(Dataset):
    def __getitem__(self, index):
        return np.random.randint(0, 1000, 3)

    def __len__(self):
        return 8

def worker_init_fn(worker_id):
    np.random.seed(worker_id)

dataset = RandomDataset()
dataloader = DataLoader(dataset, batch_size=2, num_workers=2, worker_init_fn=worker_init_fn)

print("the first epoch is ok")
for batch in dataloader:
    print(batch)

print("the second epoch is the same as the first epoch")
for batch in dataloader:
    print(batch)
```

Output:

```
the first epoch is ok

tensor([[684, 559, 629],
        [192, 835, 763]])
tensor([[ 37, 235, 908],
        [ 72, 767, 905]])
tensor([[707, 359,   9],
        [723, 277, 754]])
tensor([[715, 645, 847],
        [960, 144, 129]])

the second epoch is the same as the first epoch

tensor([[684, 559, 629],
        [192, 835, 763]])
[...]

PR #8142, opened 2022-06-06
[Fix] Use torch.initial_seed to guarantee the randomness of numpy

jbwang1997:

## Motivation

solve https://github.com/open-mmlab/mmdetection/issues/8110
numpy will be set the same seed in each epoch when pytorch < 1.8

I reproduce the same bug.
test code:
image
the results:
image

after modification:
image

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/datasets/builder.py
+++ b/mmdet/datasets/builder.py
@@ -181,9 +181,7 @@ def build_dataloader(dataset,
                                    samples_per_gpu) if shuffle else None
         batch_sampler = None
 
-    init_fn = partial(
-        worker_init_fn, num_workers=num_workers, rank=rank,
-        seed=seed) if seed is not None else None
+    init_fn = partial(worker_init_fn, num_workers=num_workers, rank=rank)
 
     if (TORCH_VERSION != 'parrots'
             and digit_version(TORCH_VERSION) >= digit_version('1.7.0')):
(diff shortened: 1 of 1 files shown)


mmdet/datasets/builder.py
```diff
@@ -206,10 +204,15 @@ def build_dataloader(dataset,
     return data_loader
 
 
-def worker_init_fn(worker_id, num_workers, rank, seed):
+def worker_init_fn(worker_id, num_workers, rank):
+    # according to https://github.com/open-mmlab/mmdetection/issues/8110
+    # numpy may be set the same seed at each epoch.
+    # Thus, # we use the return torch.initail_seed here to make sure numpy
+    # is set a new seed in each epoch.
```
hhaAndroid, line 211: add https://pytorch.org/docs/stable/data.html#randomness-in-multi-process-data-loading link

mmdet/datasets/builder.py
```diff
@@ -206,10 +204,15 @@ def build_dataloader(dataset,
     return data_loader
 
 
-def worker_init_fn(worker_id, num_workers, rank, seed):
+def worker_init_fn(worker_id, num_workers, rank):
```
ZwwWayne, line 207: seems that the user seed is not used and the interface is changed. Try to avoid BC-breaking of the interface.

ZwwWayne:
Also invite @HAOCHENYE  to see if this need to be migrated into MMEngine.

HAOCHENYE:
> Also invite @HAOCHENYE to see if this need to be migrated into MMEngine.

Yes, `worker_init_fn` in `MMEngine` should also be updated.

PR #8144, opened 2022-06-06
[fix] fix error of 'unrecognized arguments: --load-from'

JoeyforJoy:
## Motivation

In the [document](https://mmdetection.readthedocs.io/en/latest/1_exist_data_model.html#train-predefined-models-on-standard-datasets) of mmdetection, users can use `--load-from` options to load the certain checkpoints when training. 

However, when I run the script of training as follow
```bash
python tools/train.py ${CONFIG_FILE} \
    --work-dir ${WORK_DIR} \
    --load-from ${CHECKPOINT_FILE} \
    --deterministic
```

The error happens like
```bash
train.py: error: unrecognized arguments: --load-from *****.pth
```
So, I fixed it by simply modify the `tools/train.py` script.

## Modification
- Add `--load-from` option in `parse_args()`
- Assign the `load_from` value for cfg

NOTE: If resume-from and load-from options are **both** setted, the training process will behave as the resume-from mode.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -29,6 +29,8 @@ def parse_args():
     parser.add_argument(
         '--resume-from', help='the checkpoint file to resume from')
     parser.add_argument(
+        '--load-from', help='the checkpoint file to load from')
+    parser.add_argument(
         '--auto-resume',
         action='store_true',
         help='resume from the latest checkpoint automatically')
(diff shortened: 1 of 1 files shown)


ZwwWayne:
For load_from, we recommend users to use `--cfg-options load_from=xxxx`. We will update the doc.

PR #8319, opened 2022-07-07
[Fix]: Fix Wandb hang error during training

zimoqingfeng:
## Motivation

I found the training procedure hanging when using the **configs/mask_rcnn/mask_rcnn_r50_fpn_1x_wandb_coco.py**

## Modification

Moving the **MMDetWandbHook** to custom_hooks could fix the problem.

## Env Info for reproducing

sys.platform: linux
Python: 3.7.3 (default, xx xx xx, xx) [GCC 8.3.0]
CUDA available: True
GPU 0,1,2,3,4,5,6,7: 
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.3, V11.3.109
GCC: x86_64-linux-gnu-gcc (Debian 8.3.0-6) 8.3.0
PyTorch: 1.10.0
TorchVision: 0.11.1+cu113
OpenCV: 4.5.1
MMCV: 1.5.3
MMCV Compiler: GCC 7.3
MMCV CUDA Compiler: 11.3
MMDetection: 2.25.0+56e42e7


zimoqingfeng:
Maybe @ayulockin can provide some advice? 

Thanks for your support in adding such a brilliant feature for mmdet, again!

ayulockin:
> Moving the MMDetWandbHook to custom_hooks could fix the problem.

Would love to understand the difference between `log_config` and `custom_hooks`.

zimoqingfeng:
> > Moving the MMDetWandbHook to custom_hooks could fix the problem.
> 
> Would love to understand the difference between `log_config` and `custom_hooks`.

Maybe @ZwwWayne @hhaAndroid could explain the difference? It seems like more details in MMCV packages.

PR #8323, opened 2022-07-07
[Fix]: Fix WandbLogger Hang Error for distributed training.

zimoqingfeng:
## Motivation

Fix WandbLogger Hang Error for **distributed** training.

### DIFF
--- a/mmdet/core/hook/wandblogger_hook.py
+++ b/mmdet/core/hook/wandblogger_hook.py
@@ -235,7 +235,6 @@ class MMDetWandbHook(WandbLoggerHook):
             # Log the table
             self._log_eval_table(runner.epoch + 1)
 
-    @master_only
     def after_train_iter(self, runner):
         if self.get_mode(runner) == 'train':
             # An ugly patch. The iter-based eval hook will call the
--- a/configs/mask_rcnn/mask_rcnn_r50_fpn_1x_wandb_coco.py
+++ b/configs/mask_rcnn/mask_rcnn_r50_fpn_1x_wandb_coco.py
@@ -4,11 +4,6 @@ _base_ = [
     '../_base_/schedules/schedule_1x.py', '../_base_/default_runtime.py'
 ]
 
-# Set evaluation interval
-evaluation = dict(interval=2)
-# Set checkpoint interval
-checkpoint_config = dict(interval=4)
-
 # yapf:disable
 log_config = dict(
     interval=50,


zimoqingfeng:
@ayulockin @hhaAndroid 
I found some problems in my previous PR and cancel it.
Just check this one, please.

mmdet/core/hook/wandblogger_hook.py
```diff
@@ -235,7 +235,6 @@ def after_train_epoch(self, runner):
             # Log the table
             self._log_eval_table(runner.epoch + 1)
 
-    @master_only
```
ayulockin, line 238: By removing this decorator, the method will be called in every process however the wandb run is initialized only in the master process ([source](https://github.com/open-mmlab/mmcv/blob/6a03918f55c1cc860bcb4d64a0e868ca13cf08ba/mmcv/runner/hooks/logger/wandb.py#L78)). 

I think it should throw error. 

Plus to handle distributed training properly we would be using `wandb_service` ([link](https://docs.wandb.ai/guides/track/advanced/distributed-training#wandb-service)) once it's available as default with the [client](https://github.com/wandb/client/search?q=service).

zimoqingfeng:
Anyone could solve it? @ayulockin @hhaAndroid

ayulockin:
Hey I can take a look @zimoqingfeng. Can you share your code that's hanging as a colab notebook? That will help reproduce the issue and find a solution.

zimoqingfeng:
> 

Just run "./tools/dist_train.sh configs/mask_rcnn/mask_rcnn_r50_fpn_1x_wandb_coco.py 8" can reproduce it. Can you reproduce? @ayulockin

mitming:
I have the same issue. Thank you very much; it helps a lot!

zimoqingfeng:
@ayulockin could you merge this PR?

ayulockin:
Hey, @zimoqingfeng, I am not a maintainer, so I can't merge your PR. I had reviewed the PR and left a comment long back.

zimoqingfeng:
@hhaAndroid @RangiLyu Could anyone check this PR? That seems like a long time.

hhaAndroid:
@ayulockin Is this PR ready to be merged?

ayulockin:
I don't think so. The PR is removing the `@master_only` decorator which I don't think is the right thing to do here.

I guess a hacky solution is proposed in this [issue](https://github.com/open-mmlab/mmsegmentation/issues/2137) where @haruishi43 has pointed this out - "I'm not really sure how but using only MMSegWandbHook does not timeout and trains. It times out only when MMSegWandbHook is used in conjunction with other loggers like TextLoggerHook."

Maybe something similar might be the issue here.

zimoqingfeng:
> I don't think so. The PR is removing the `@master_only` decorator which I don't think is the right thing to do here.
> 
> I guess a hacky solution is proposed in this [issue](https://github.com/open-mmlab/mmsegmentation/issues/2137) where @haruishi43 has pointed this out - "I'm not really sure how but using only MMSegWandbHook does not timeout and trains. It times out only when MMSegWandbHook is used in conjunction with other loggers like TextLoggerHook."
> 
> Maybe something similar might be the issue here.

I cant follow that, so how to fix it?

PR #8383, opened 2022-07-19
Upgrade onnxsim to at least 0.4.0

daquexian:
## Motivation

I'm the author of onnx-simplifier. Recently I published onnxsim v0.4.0. In the latest version, dynamic input shapes are natively supported with no need for extra flags like `dynamic_input_shapes`. `custom_lib` and `input_data` is also not needed.

### DIFF
--- a/tools/deployment/pytorch2onnx.py
+++ b/tools/deployment/pytorch2onnx.py
@@ -101,31 +101,18 @@ def pytorch2onnx(model,
 
     model.forward = origin_forward
 
-    # get the custom op path
-    ort_custom_op_path = ''
-    try:
-        from mmcv.ops import get_onnxruntime_op_path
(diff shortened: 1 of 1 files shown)


tools/deployment/pytorch2onnx.py
```diff
@@ -101,31 +101,17 @@ def pytorch2onnx(model,
 
     model.forward = origin_forward
 
-    # get the custom op path
-    ort_custom_op_path = ''
-    try:
-        from mmcv.ops import get_onnxruntime_op_path
-        ort_custom_op_path = get_onnxruntime_op_path()
```
daquexian, line 108: ort_custom_op_path is not used after upgrade.

RangiLyu (approved):
LGTM.
dalaoshi yyds

ZwwWayne (approved):

PR #8499, opened 2022-08-04
[Feature] Support ShuffledSequence

Czm369:
image

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/datasets/transforms/wrappers.py
+++ b/mmdet/datasets/transforms/wrappers.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Optional
 
 import numpy as np
 from mmcv.transforms import BaseTransform, Compose
+from mmcv.transforms.utils import cache_randomness
 
 from mmdet.registry import TRANSFORMS
 
@@ -55,6 +56,10 @@ class MultiBranch(BaseTransform):
 class ShuffledSequence(Compose):
     """Shuffle the transform Sequence."""
 
+    @cache_randomness
+    def _random_permutation(self):
+        return np.random.permutation(len(self.transforms))
+
     def transform(self, results: Dict) -> Optional[Dict]:
         """Transform function to apply transforms shuffled sequentially.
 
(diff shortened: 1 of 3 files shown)


mmdet/datasets/transforms/wrappers.py
```diff
@@ -48,3 +49,33 @@ def __repr__(self) -> str:
...
+    def transform(self, results: Dict) -> Optional[Dict]:
+        """Transform function to apply transforms shuffled sequentially.
+
+        Args:
+            results (dict): A result dict contains the results to transform.
+
+        Returns:
+            dict or None: Transformed results.
+        """
+        order = np.random.permutation(len(self.transforms))
```
RangiLyu, line 67: cache_randomness

mmdet/datasets/transforms/wrappers.py
```diff
@@ -48,3 +50,37 @@ def __repr__(self) -> str:
         repr_str = self.__class__.__name__
         repr_str += f'(branch_pipelines={list(self.branch_pipelines.keys())})'
         return repr_str
+
+
+@TRANSFORMS.register_module()
+class ShuffledSequence(Compose):
```
hhaAndroid, line 56: Can support passing in the `num_selected` parameter so that the user can randomly select part of the pipeline and then shuffled

Czm369: In fact, `RandAugment` has already supported randomly selecting part of the pipeline, so `ShuffledSequence` only simplifies the logic.

hhaAndroid (approved):

mmdet/datasets/transforms/wrappers.py
ZwwWayne: RandomOrder might be a more appropriate name

mmdet/datasets/transforms/wrappers.py
```diff
@@ -48,3 +50,37 @@ def __repr__(self) -> str:
...
+    def transform(self, results: Dict) -> Optional[Dict]:
+        """Transform function to apply transforms shuffled sequentially.
+
+        Args:
+            results (dict): A result dict contains the results to transform.
+
+        Returns:
+            dict or None: Transformed results.
+        """
+        order = self._random_permutation()
```
ZwwWayne, line 72: order -> inds

PR #8592, opened 2022-08-19
Rename config

wanghonglie:

## Motivation
rename config

## Modification

rename config


configs/yolox/yolox_l_8x8-300e_coco.py
```diff
@@ -0,0 +1,8 @@
+_base_ = './yolox_s_8x8-300e_coco.py'
```
ZwwWayne, line 1: yolox_s_8xb8-300e_coco.py

configs/albu_example/mask-rcnn_r50-fpn_albu-1x_coco.py
```diff
@@ -0,0 +1,68 @@
+_base_ = '../mask_rcnn/mask-rcnn_r50-fpn_1x_coco.py'
```
BIGWangYuDong, line 1: does mask_rcnn also need to rename to `mask-rcnn`

BIGWangYuDong:
- README.md in each config folder also need to rename.
- docs also need to check whether should change the config name
- Fix Unit Test

configs/albu_example/mask-rcnn_r50-fpn_albu-1x_coco.py
wanghonglie: Folders do not need to be renamed.

ZwwWayne:
continued in https://github.com/open-mmlab/mmdetection/pull/8603

PR #8596, opened 2022-08-19
[Feature] Add BaseDetDataset

BIGWangYuDong:
- Add BaseDetDataset to support load proposals in dataset.
- Support using FileClient to load annotations
- Support using FileClient in Metrics
- `mmcv.FileClient` -> `mmengine.fileio.FileClient`


mmdet/datasets/base_det_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+        The `proposals_list` should be a dict[img_path: proposals]
+        with the same length as `data_list`. And the `proposals` should be
+        a `dict` or :obj:`InstanceData` usually contains following keys.
+
+            - bboxes (np.ndarry): Has a shape (num_instances, 4),
+              the last dimension 4 arrange as (x1, y1, x2, y2).
+            - scores (np.ndarry): Classification scores, has a shape
+              (num_instance, ).
+        """
+        if not is_abs(self.proposal_file) and self.proposal_file:
```
hhaAndroid, line 85:
```suggestion
        if not is_abs(self.proposal_file):
```

mmdet/datasets/base_det_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+        if self._indices is not None:
+            self.data_list = self._get_unserialized_subset(self._indices)
+
+        # serialize data_list
+        if self.serialize_data:
+            self.data_bytes, self.data_address = self._serialize_data()
+
+        self._fully_initialized = True
+
+    def load_proposals(self) -> None:
```
hhaAndroid, line 73: Are unit tests added?

mmdet/datasets/base_det_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+              the last dimension 4 arrange as (x1, y1, x2, y2).
+            - scores (np.ndarry): Classification scores, has a shape
+              (num_instance, ).
+        """
+        if not is_abs(self.proposal_file) and self.proposal_file:
+            self.proposal_file = osp.join(self.data_root, self.proposal_file)
+        proposals_list = load(
+            self.proposal_file, file_client_args=self.file_client_args)
+        assert len(self.data_list) == len(proposals_list)
+        for data_info in self.data_list:
```
hhaAndroid, line 90: How to ensure that `data_list `and `proposals_list` are consistent? Whether to specify in the document?

BIGWangYuDong: proposal list is a dict( `img_path`: `proposals`(InstanceData)]

mmdet/datasets/base_det_dataset.py
BIGWangYuDong: This UT should add after Fast R-CNN is supported. We should first make sure the dumped proposal type, them we can add this UT

mmdet/datasets/base_det_dataset.py
```diff
@@ -0,0 +1,107 @@
...
+              (num_instance, ).
+        """
+        # TODO: Add Unit Test after fully support Dump-Proposal Metric
+        if not is_abs(self.proposal_file):
+            self.proposal_file = osp.join(self.data_root, self.proposal_file)
+        proposals_list = load(
+            self.proposal_file, file_client_args=self.file_client_args)
+        assert len(self.data_list) == len(proposals_list)
+        for data_info in self.data_list:
+            # TODO: add `file_name` into metafile during PackDetInputs
```
RangiLyu, line 92: which metafile?

hhaAndroid (approved):

mmdet/datasets/base_det_dataset.py
```diff
@@ -0,0 +1,107 @@
...
+        self._fully_initialized = True
+
+    def load_proposals(self) -> None:
+        """Load proposals from proposals file.
+
+        The `proposals_list` should be a dict[img_path: proposals]
+        with the same length as `data_list`. And the `proposals` should be
+        a `dict` or :obj:`InstanceData` usually contains following keys.
+
+            - bboxes (np.ndarry): Has a shape (num_instances, 4),
```
RangiLyu, line 80: Why the bbox in the instance is a list of float while the proposal is a ndarray? Should they be consistent?

mmdet/datasets/base_det_dataset.py
BIGWangYuDong: If we want to set `file_name` as a default key in DetDataSample, we should change all test pipelines:
```
test_pipeline = [
    ......
    dict(
        type='PackDetInputs',
        meta_keys=('img_id', 'img_path', 'ori_shape', 'img_shape',
                   'scale_factor', 'file_name'))
]
```
However, in my opinion, we can only change the RPN test pipeline.

mmdet/datasets/base_det_dataset.py
BIGWangYuDong: The boxes will be processed in LoadAnnotation and returned as np.ndarry. I think this is mainly because each dataset have different annotation type, it is hard to uniform. 
But proposals can be processed durning DumpProposalMetric, which we do not need to convert to list and re-convert back to np.ndarry.

BIGWangYuDong: We can discuss this in the PR of Fast R-CNN

BIGWangYuDong: the instance is a list but proposals is an `InstanceData`, they are not same. Also, the proposal will not be processed in dataset, maybe we can uniform the proposal type.

mmdet/datasets/openimages.py
```diff
@@ -379,6 +366,7 @@ def load_data_list(self) -> List[dict]:
             data_list.append(
                 dict(
                     img_path=osp.join(self.data_prefix['img'], filename),
+                    file_name=filename,
```
ZwwWayne, line 369: is this line necessary?

mmdet/datasets/coco.py
```diff
@@ -119,6 +119,7 @@ def parse_data_info(self, raw_data_info: dict) -> Union[dict, List[dict]]:
         else:
             seg_map_path = None
         data_info['img_path'] = img_path
+        data_info['file_name'] = img_info['file_name']
```
ZwwWayne, line 122: file_name should be unnecessary since img_path already indicated that.

BIGWangYuDong: The `image path` can be an absolute path. We should find another way to fix this potential error

BIGWangYuDong:
dataset and metric UT  all passed

ZwwWayne (approved):

PR #8617, opened 2022-08-23
[Refactor] Fix mdformat version to support python3.6

BIGWangYuDong:
- update the requirements of mmtrack, to install mmtrack-v1.x.  **NOTE** right now, cannot install mmtrack-v1.x by setting `-e` in the requirement file. It should find another way to fix this error
image

- update the mdformat version


requirements/tests.txt
```diff
@@ -9,7 +9,8 @@ isort==4.3.21
 # Note: used for kwarray.group_items, this may be ported to mmcv in the future.
 kwarray
 memory_profiler
--e git+https://github.com/open-mmlab/mmtracking#egg=mmtrack
+# -e git+https://github.com/open-mmlab/mmtracking@dev-1.x#egg=mmtrack
```
BIGWangYuDong, line 12: should we delete this line?

ZwwWayne: delete

PR #8648, opened 2022-08-27
Fix floordiv warning.

JiayuXu0:
## Motivation

## Modification

### DIFF
--- a/mmdet/utils/misc.py
+++ b/mmdet/utils/misc.py
@@ -5,7 +5,8 @@ import os.path as osp
 import warnings
 
 import mmcv
-from mmcv.utils import print_log
+import torch
+from mmcv.utils import TORCH_VERSION, digit_version, print_log
 
 
 def find_latest_checkpoint(path, suffix='pth'):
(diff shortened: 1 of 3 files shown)


RangiLyu:
The unit tests of pytorch<=1.7 failed. MMDetection needs to support all versions of pytorch >= 1.6.

JiayuXu0:
> 

If we add a conditional selection (torch version>=1.8 or  torch version <1.8) before calculation, is it ok？

ZwwWayne:
We can implement a floordiv wrapper that handles different pytorch versions, and use that wrapper uniformly in the head.

JiayuXu0:
> 

Thanks for your suggestion. I have modify the code and wait for your review.

mmdet/utils/misc.py
```diff
@@ -74,3 +75,15 @@ def update(cfg, src_str, dst_str):
...
     cfg.data_root = dst_root
+
+
+_torch_version_div_indexing = (
+    'parrots' not in TORCH_VERSION
+    and digit_version(TORCH_VERSION) >= digit_version('1.8'))
+
+
+def floordiv(dividend, divisor, rounding_mode='trunc'):
+    if _torch_version_div_indexing:
```
hhaAndroid, line 86: Actually, you can use `torch.floor_divide` directly

JiayuXu0: [torch.floor_divide()](https://pytorch.org/docs/stable/generated/torch.floor_divide.html#torch.floor_divide) is deprecated and will be removed in a future PyTorch release.
https://pytorch.org/docs/stable/generated/torch.floor_divide.html

RangiLyu (approved):

BIGWangYuDong:
@JiayuXu0 Hi, you can checkout from dev-3.x and create a new PR to dev-3.x. Thanks for your PR!

PR #8670, opened 2022-08-30
[Fix] Enable to set float mlp_ratio in SwinTransformer

i-aki-y:
## Motivation

In the original implementation, the `mlp_ratio` parameter in the `SwinTransformer`  can be float,  but the current mmdet's implementation does not support non-integer `mlp_ratio` (an Exception will be raised if a non-integer value is given).

I have found the same fix in mmsegmentation:
see, https://github.com/open-mmlab/mmsegmentation/pull/1274

## Modification

When a float `mlp_ratio` is given, dims of mlp's out channels are truncated to an integer.

## BC-breaking (Optional)

I think this does not break BC.

## Checklist


ZwwWayne:
We need to migrate this PR in MMDet 3.x

Issue #8214, opened 2022-06-17
FilterAnnotations __repr__ function throws an AttributeError


jalexhurt:
**Describe the bug**
The `__repr__` function of the `FilterAnnotations` class attempts to access an attribute `always_keep` of the class, which does not exist, and causing an `AttributeError`, as seen here: https://github.com/open-mmlab/mmdetection/blob/ca11860f4f3c3ca2ce8340e2686eeaec05b29111/mmdet/datasets/pipelines/loading.py#L643

**Reproduction**

1. What command or script did you run?

```none
from mmdet.datasets import PIPELINES

print(PIPELINES.get("FilterAnnotations")())
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?

No modifications required to reproduce bug

4. What dataset did you use?

No dataset was required to reproduce the bug

**Environment**
```

{'CUDA available': False,
 'GCC': 'gcc (Debian 10.2.1-6) 10.2.1 20210110',
 'MMCV': '1.5.2',
 'MMCV CUDA Compiler': 'not available',
 'MMCV Compiler': 'GCC 10.2',
 'OpenCV': '4.6.0',
 'PyTorch': '1.9.0+cpu',
 'PyTorch compiling details': 'PyTorch built with:\n'
                              '  - GCC 7.3\n'
                              '  - C++ Version: 201402\n'
                              '  - Intel(R) Math Kernel Library Version '
                              '2020.0.0 Product Build 20191122 for Intel(R) 64 '
                              'architecture applications\n'
                              '  - Intel(R) MKL-DNN v2.1.2 (Git Hash '
                              '98be7e8afa711dc9b66c8ff3504129cb82013cdb)\n'
                              '  - OpenMP 201511 (a.k.a. OpenMP 4.5)\n'
                              '  - NNPACK is enabled\n'
                              '  - CPU capability usage: AVX2\n'
                              '  - Build settings: BLAS_INFO=mkl, '
                              'BUILD_TYPE=Release, '
                              'CXX_COMPILER=/opt/rh/devtoolset-7/root/usr/bin/c++, '
                              'CXX_FLAGS= -Wno-deprecated '
                              '-fvisibility-inlines-hidden -DUSE_PTHREADPOOL '
                              '-fopenmp -DNDEBUG -DUSE_KINETO '
                              '-DLIBKINETO_NOCUPTI -DUSE_FBGEMM -DUSE_QNNPACK '
[...]

PR #8136, opened 2022-06-05
FilterAnnotations  bug fix

MiXaiLL76:
Fix #8131 
FilterAnnotation bug fix for https://github.com/open-mmlab/mmdetection/issues/8131

PR #8673, opened 2022-08-30
Refactor filtering mask, fix attribute error

christian-hiebl:
## Motivation

Filtering annotations for a dataset is great (besides the gt_bboxes_ignore option) I changed the implementation of the keep mask and added the attribute error fix in repr as also discussed here [#8214](https://github.com/open-mmlab/mmdetection/issues/8214) but not included in [#8136](https://github.com/open-mmlab/mmdetection/pull/8136)

## Modification

Change of filter annotation masking from appending boolean mask to a list and iterating, to now a bumpy array. Additionally small fix discussed but not included from #8214

## BC-breaking (Optional)

I don't think it will break BC.

## Checklist

### DIFF
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -603,30 +603,26 @@ class FilterAnnotations:
             assert 'gt_bboxes' in results
             gt_bboxes = results['gt_bboxes']
             instance_num = gt_bboxes.shape[0]
+            keep = np.ones(gt_bboxes.shape[0], dtype=bool)
         if self.by_mask:
             assert 'gt_masks' in results
             gt_masks = results['gt_masks']
             instance_num = len(gt_masks)
+            keep = np.ones(len(gt_masks), dtype=bool)
 
         if instance_num == 0:
             return results
 
-        tests = []
         if self.by_box:
             w = gt_bboxes[:, 2] - gt_bboxes[:, 0]
             h = gt_bboxes[:, 3] - gt_bboxes[:, 1]
-            tests.append((w > self.min_gt_bbox_wh[0])
-                         & (h > self.min_gt_bbox_wh[1]))
+            keep = keep & ((w > self.min_gt_bbox_wh[0])
+                           & (h > self.min_gt_bbox_wh[1]))
         if self.by_mask:
             gt_masks = results['gt_masks']
-            tests.append(gt_masks.areas >= self.min_gt_mask_area)
-
-        keep = tests[0]
-        for t in tests[1:]:
-            keep = keep & t
+            keep = keep & (gt_masks.areas >= self.min_gt_mask_area)
 
         keep = keep.nonzero()[0]
-
         keys = ('gt_bboxes', 'gt_labels', 'gt_masks')
         for key in keys:
             if key in results:
@@ -642,4 +638,4 @@ class FilterAnnotations:
             f'min_gt_mask_area={self.min_gt_mask_area},' \
             f'by_box={self.by_box},' \
             f'by_mask={self.by_mask},' \
-            f'always_keep={self.always_keep})'
+            f'keep_empty={self.keep_empty})'
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -603,13 +603,11 @@ class FilterAnnotations:
             assert 'gt_bboxes' in results
             gt_bboxes = results['gt_bboxes']
             instance_num = gt_bboxes.shape[0]
-            keep = np.ones(gt_bboxes.shape[0], dtype=bool)
         if self.by_mask:
             assert 'gt_masks' in results
             gt_masks = results['gt_masks']
             instance_num = len(gt_masks)
-            keep = np.ones(len(gt_masks), dtype=bool)
-
+        keep = np.ones(instance_num, dtype=bool)
         if instance_num == 0:
             return results
 


christian-hiebl:
@jbwang1997 Do you know why the CUDA build could be failing? Locally I made the changes for this commit with a non CUDA torch and mmcv version. But when the tests are run here, this shouldn't be affected correct?

jbwang1997:
> @jbwang1997 Do you know why the CUDA build could be failing? Locally I made the changes for this commit with a non CUDA torch and mmcv version. But when the tests are run here, this shouldn't be affected correct?

Seem the CI failed when installing the dependence. I have restarted the CI. Let's wait for the results.

christian-hiebl:
> > @jbwang1997 Do you know why the CUDA build could be failing? Locally I made the changes for this commit with a non CUDA torch and mmcv version. But when the tests are run here, this shouldn't be affected correct?
> 
> Seem the CI failed when installing the dependence. I have restarted the CI. Let's wait for the results.

Ok great, thank you for the restart!

jbwang1997 (approved):
LGTM

mmdet/datasets/pipelines/loading.py
```diff
@@ -603,30 +603,26 @@ def __call__(self, results):
             assert 'gt_bboxes' in results
             gt_bboxes = results['gt_bboxes']
             instance_num = gt_bboxes.shape[0]
+            keep = np.ones(gt_bboxes.shape[0], dtype=bool)
         if self.by_mask:
             assert 'gt_masks' in results
             gt_masks = results['gt_masks']
             instance_num = len(gt_masks)
+            keep = np.ones(len(gt_masks), dtype=bool)
```
BIGWangYuDong, line 611: Seems that `instance_num` and `keep` will over-write if self.by_box and self.by_mask is True.

BIGWangYuDong: len(gt_masks) -> instance_num

BIGWangYuDong:
@christian-hiebl Thanks for your contribution. I think ` gt_bboxes.shape[0] -> instance_num` and `len(gt_masks) -> instance_num` is necessary to fix. Not sure if there is a more elegant way to fix the over-write problem. 
I think maybe we can set `keep` out of the conditions. Such as:
```python
if self.by_box:
    assert 'gt_bboxes' in results
    gt_bboxes = results['gt_bboxes']
    instance_num = gt_bboxes.shape[0]
if self.by_mask:
    assert 'gt_masks' in results
    gt_masks = results['gt_masks']
    instance_num = len(gt_masks)
keep = np.ones(instance_num, dtype=bool)
```
This PR looks good to me.

mmdet/datasets/pipelines/loading.py
```diff
@@ -603,30 +603,26 @@ def __call__(self, results):
             assert 'gt_bboxes' in results
             gt_bboxes = results['gt_bboxes']
             instance_num = gt_bboxes.shape[0]
+            keep = np.ones(gt_bboxes.shape[0], dtype=bool)
         if self.by_mask:
             assert 'gt_masks' in results
             gt_masks = results['gt_masks']
             instance_num = len(gt_masks)
```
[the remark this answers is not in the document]
christian-hiebl: Yes, I think it should be. And the test cases with both mask instances and gt_bbox instances passes.

christian-hiebl:
> @christian-hiebl Thanks for your contribution. I think maybe we can set `keep` out of the conditions. Such as:
> 
> ```python
> if self.by_box:
>     assert 'gt_bboxes' in results
>     gt_bboxes = results['gt_bboxes']
>     instance_num = gt_bboxes.shape[0]
> if self.by_mask:
>     assert 'gt_masks' in results
>     gt_masks = results['gt_masks']
>     instance_num = len(gt_masks)
> keep = np.ones(instance_num, dtype=bool)
> ```
> 
> Not sure if there is a more elegant way to fix the over-write problem.

Thank you for the suggestion! In the __init__ of the FilterAnnotations class there is an assertion, that the filtering is done by either mask or box, so the instance_num won't be over-written.

BIGWangYuDong:
> > @christian-hiebl Thanks for your contribution. I think maybe we can set `keep` out of the conditions. Such as:
> > ```python
> > if self.by_box:
> >     assert 'gt_bboxes' in results
> >     gt_bboxes = results['gt_bboxes']
> >     instance_num = gt_bboxes.shape[0]
> > if self.by_mask:
> >     assert 'gt_masks' in results
> >     gt_masks = results['gt_masks']
> >     instance_num = len(gt_masks)
> > keep = np.ones(instance_num, dtype=bool)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure if there is a more elegant way to fix the over-write problem.
> 
> Thank you for the suggestion! In the **init** of the FilterAnnotations class there is an assertion, that the filtering is done by either mask or box, so the instance_num won't be over-written.

@christian-hiebl 
```
assert by_box or by_mask
```
if by_box and by_mask all set True, it also works.

christian-hiebl:
@BIGWangYuDong You are right, 'or' is inclusive so it should also work when both if statements are true. Currently it passes in the test case that a mask and box is given. Also, the instance_num for both is the same, so it will not be a problem.

mmdet/datasets/pipelines/loading.py
```diff
@@ -607,26 +607,20 @@ def __call__(self, results):
...
-
+        keep = np.ones(instance_num, dtype=bool)
```
BIGWangYuDong, line 610: How about add a assert before to avoid some potential error?
such as:
```
instance_num = None
if self.by_box:
    if instance_num is not None:
        assert instance_num == box_num
if self.by_mask:
        if instance_num is not None:
            assert instance_num == mask_num

```

PR #8717, opened 2022-09-03
Fix floordiv warning.

JiayuXu0:
## Motivation

## Modification


JiayuXu0:
@BIGWangYuDong Some tests fail on circleci, but  it doesn't seem to have anything to do with my changes .Do I need to make further modifications?

ZwwWayne:
Hi @JiayuXu0 ,
Thanks for your kind PR. We have fixed the issue of CI in #8707 . Please rebase your code to the newest dev-3.x.

PR #8722, opened 2022-09-03
Add mmengine installation to the get started docs

argenisleon:
Needed to use the download command in mim

## Motivation
Fix the code samples download

## Modification

Added `pip installation mmengine`


ZwwWayne:
Hi @argenisleon ,
Thanks for your kind PR. However, the master branch does not need to install mmengine, and we have add it in the docs of 3.x https://github.com/open-mmlab/mmdetection/blob/dev-3.x/docs/en/get_started.md.
Therefore, this PR does not need to be merged and is closed.

PR #8738, opened 2022-09-06
Fix floordiv warning.

JiayuXu0:
## Motivation

## Modification


JiayuXu0:
@ZwwWayne The code is re-summited here.

ZwwWayne:
Hi @JiayuXu0 ,
Thanks for your kind PR. Would you like to fix the conflict so that we can merge it?

RangiLyu (approved):

ZwwWayne (dismissed):

ZwwWayne (approved):

PR #8743, opened 2022-09-06
update SSDAnchorGenerator

MambaWong:
update SSDAnchorGenerator to support: [Caffe PriorBoxLayer](https://github.com/weiliu89/caffe/blob/289b10ee5cb844278ad235c9fe4280d29f9af311/src/caffe/layers/prior_box_layer.cpp)
1. only min_sizes (list[list[float]]) or (list[float]) as input for base sizes on each level.
2. min_sizes (list[list[float]]) and max_sizes (list[list[float]]) as input for (minimum, maximum) sizes on each level.


ZwwWayne:
Hi @MambaWong ,
Thanks for your kind PR. Could you also reveal why scale_major should be removed in the PR messages?

MambaWong:
Hi @ZwwWayne 
In the description of 'class SSDAnchorGenerator':
> scale_major (bool): Whether to multiply scales first when generating
            base anchors. If true, the anchors in the same row will have the
            same scales. **It is always set to be False in SSD.**

'scale_major' is always set to be False in SSD.

PR #8769, opened 2022-09-11
add missing install command

franchesoni:


BIGWangYuDong:
Thanks for your contribution. 

Right now, mmdet v2.x do not need to install mmengine. The installation of mmdet 3.0 has already update to using mim to install mmengine and mmcv.

PR #8770, opened 2022-09-11
remove documentation for load-from option

franchesoni:


ZwwWayne:
Hi @franchesoni ,

image

docs/en/1_exist_data_model.md
```diff
@@ -582,13 +582,6 @@ This tool accepts several optional arguments, including:
 - `--resume-from ${CHECKPOINT_FILE}`: Resume from a previous checkpoint file.
 - `--options 'Key=value'`: Overrides other settings in the used config.
 
-**Note**:
```
BIGWangYuDong, line 585: The difference between load-from and resume-from should add somewhere, 
right now, maybe we can tell users that if they want to use load-from, they should use `--cfg-options load_from=${CHECKPOINT_FILE}` (for example)

PR #8777, opened 2022-09-13
[Feature] K-Net mask cost

chhluo:

## Motivation

## Modification

1. Add `MaskCost` for K-Net.
2. Update `panoptic_gt_processing`
3. Update docstring in mask_hungarian_assigner
4. Update `MaskFormerHead.preprocess_gt`.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

5. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
6. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
7. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmdet/core/bbox/match_costs/match_cost.py
+++ b/mmdet/core/bbox/match_costs/match_cost.py
@@ -357,3 +357,42 @@ class CrossEntropyLossCost:
             raise NotImplementedError
 
         return cls_cost * self.weight
+
+
+@MATCH_COST.register_module()
+class MaskCost:
+    """MaskCost.
+
+    Args:
+        weight (int | float, optional): loss_weight
+        use_sigmoid (bool, optional): Whether the prediction uses sigmoid
+                of softmax. Defaults to True.
+    """
+
+    def __init__(self, weight=1., use_sigmoid=True):
+        self.weight = weight
+        self.use_sigmoid = use_sigmoid
+
+    def __call__(self, mask_preds, gt_masks):
+        """
+        Args:
+            mask_preds (Tensor): Mask prediction logits in shape
+                (num_query, h, w).
+            gt_masks (Tensor): Ground truth in shape (num_gt, h, w).
+        Returns:
+            Tensor: Ccost matrix with weight in shape (num_query, num_gt).
+        """
+
+        if self.use_sigmoid:
+            mask_preds = mask_preds.sigmoid()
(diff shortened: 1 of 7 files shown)


mmdet/core/bbox/match_costs/match_cost.py
```diff
@@ -357,3 +357,42 @@ def __call__(self, cls_pred, gt_labels):
...
 
         return cls_cost * self.weight
+
+
+@MATCH_COST.register_module()
+class MaskCost:
+    """MaskCost.
+
+    Args:
+        weight (int | float, optional): loss_weight
```
BIGWangYuDong, line 367: optional here is strange, `weight` need be float or int. Add defaluts value. same as `use_sigmoid`

mmdet/models/utils/panoptic_gt_processing.py
```diff
@@ -2,10 +2,14 @@
...
 
-def preprocess_panoptic_gt(gt_labels, gt_masks, gt_semantic_seg, num_things,
-                           num_stuff, img_metas):
+def preprocess_panoptic_gt(gt_labels,
+                           gt_masks,
+                           gt_semantic_seg,
+                           num_things,
+                           num_stuff,
+                           img_metas,
+                           return_seperate=False):
```
BIGWangYuDong, line 11: doc strings need be updated

mmdet/models/utils/panoptic_gt_processing.py
```diff
@@ -51,18 +57,28 @@ def preprocess_panoptic_gt(gt_labels, gt_masks, gt_semantic_seg, num_things,
...
+    if return_seperate:
+        if len(stuff_labels_list) > 0:
+            stuff_masks = torch.stack(stuff_masks_list, dim=0)
+            stuff_labels = torch.stack(stuff_labels_list, dim=0)
+        else:
+            stuff_masks = gt_semantic_seg.new_zeros(
+                size=(0, ) + gt_semantic_seg.shape[-2:])
+            stuff_labels = gt_semantic_seg.new_zeros(size=(0, ))
+
+        return things_labels, things_masks, stuff_labels, stuff_masks
```
BIGWangYuDong, line 73: the return is different, may occur some potential error. Maybe we can add `None` to force set return value that not needed when return_seperate is False

PR #8784, opened 2022-09-13
Update loading.py

shuxp:
fix bug in test_dataloader. 
because test_dataloader has no 'instance' key.
check 'instance' whether in results.keys()

### DIFF
--- a/mmdet/datasets/transforms/loading.py
+++ b/mmdet/datasets/transforms/loading.py
@@ -260,10 +260,9 @@ class LoadAnnotations(MMCV_LoadAnnotations):
         """
         gt_bboxes = []
         gt_ignore_flags = []
-        if 'instances' in results.keys():
-            for instance in results['instances']:
-                gt_bboxes.append(instance['bbox'])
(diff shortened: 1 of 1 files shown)


mmdet/datasets/transforms/loading.py
```diff
@@ -260,9 +260,10 @@ def _load_bboxes(self, results: dict) -> None:
         """
         gt_bboxes = []
         gt_ignore_flags = []
-        for instance in results['instances']:
-            gt_bboxes.append(instance['bbox'])
-            gt_ignore_flags.append(instance['ignore_flag'])
+        if 'instances' in results.keys():
+            for instance in results['instances']:
+                gt_bboxes.append(instance['bbox'])
+                gt_ignore_flags.append(instance['ignore_flag'])
```
RangiLyu, line 266:
```suggestion
        for instance in results.get('instances', []):
            gt_bboxes.append(instance['bbox'])
            gt_ignore_flags.append(instance['ignore_flag'])
```

RangiLyu (commented):
should also update `_process_masks`

PR #8792, opened 2022-09-14
[Document] update location of `test_cfg` and `train_cfg`

DrRyanHuang:
Add `test_cfg` and `train_cfg` to `model`.


ZwwWayne:
Hi @xxx,
Thanks for your kind PR. It seems that there are conflicts between your PR and the targeted branch. Could you solve the conflicts so that we could eventually merge this PR?

You can find more details in the place shown as below:
image

PR #8814, opened 2022-09-16
[Feature]Implement dump results

wanghonglie:
Implement dump results

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -108,8 +108,17 @@ def main():
         dump_metric = dict(type='DumpResults', out_file_path=args.dump)
         if isinstance(cfg.test_evaluator, (list, tuple)):
             cfg.test_evaluator = list(cfg.test_evaluator).append(dump_metric)
-        else:
-            cfg.test_evaluator = [cfg.test_evaluator, dump_metric]
+        elif isinstance(cfg.test_evaluator, dict):
+            if 'metric' in cfg.test_evaluator:
+                if isinstance(cfg.test_evaluator.metric, (list, tuple)):
+                    cfg.test_evaluator.metric = list(
+                        cfg.test_evaluator.metric).append(dump_metric)
+                else:
+                    cfg.test_evaluator.metric = [
+                        cfg.test_evaluator.metric, dump_metric
+                    ]
+            else:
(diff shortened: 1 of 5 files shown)


chhluo (dismissed):

tools/test.py
```diff
@@ -97,6 +101,16 @@ def main():
...
         cfg = trigger_visualization_hook(cfg, args)
 
+    # Dump predictions
+    if args.dump is not None:
+        assert args.dump.endswith(('.pkl', '.pickle')), \
+            'The dump file must be a pkl file.'
+        dump_metric = dict(type='DumpResults', out_file_path=args.dump)
+        if isinstance(cfg.test_evaluator, (list, tuple)):
+            cfg.test_evaluator = list(cfg.test_evaluator).append(dump_metric)
+        else:
```
ZwwWayne, line 111: What if the evaulator has a type and the metric is wrapped inside like
```python
test_evaluator = dict(
    type='xxxevaluator',
    metric=[dict(type='CocoMetric'), ])
```

RangiLyu (dismissed):

tools/test.py
```diff
@@ -19,6 +19,10 @@ def parse_args():
     parser.add_argument(
         '--work-dir',
         help='the directory to save the file containing evaluation metrics')
+    parser.add_argument(
+        '--dump',
```
ZwwWayne, line 23: Consider to use `--format-only` or some keys that exists in MMDet 2.x for compatibility.

tools/test.py
```diff
@@ -97,6 +101,25 @@ def main():
     if args.show or args.show_dir:
         cfg = trigger_visualization_hook(cfg, args)
 
+    # Dump predictions
+    if args.dump is not None:
+        assert args.dump.endswith(('.pkl', '.pickle')), \
```
ZwwWayne, line 106: Try to implement a function to encapsulate lines 106-121

tools/test.py
RangiLyu: We follow the MMCls's args. Do we have any conventions?

ZwwWayne (approved):

ZwwWayne (approved):

PR #8897, opened 2022-09-28
Use MMEval VOCMeanAP and OIDMeanAP

ice-tong:
TBD


mmdet/evaluation/metrics/openimages_metric.py
```diff
@@ -1,237 +1,131 @@
...
+                ['bboxes'].cpu().numpy(),  # noqa: E501
+                'scores': data_sample['pred_instances']
+                ['scores'].cpu().numpy(),  # noqa: E501
+                'labels':
+                data_sample['pred_instances']['labels'].cpu().numpy()
+            }
+            predictions.append(pred)
+
+            gt = {
+                'instances': data_sample['instances'],
```
RangiLyu, line 75: Can we get gt from the data_sample['gt_instances'] by adding 'LoadAnnotation' pipeline?
So that the gt and pred can be the same format.

image

mmdet/evaluation/metrics/voc_metric.py
```diff
@@ -1,86 +1,67 @@
...
-        scale_ranges (List[tuple], optional): Scale ranges for evaluating
-            mAP. If not specified, all bounding boxes would be included in
-            evaluation. Defaults to None.
-        metric (str | list[str]): Metrics to be evaluated. Options are
-            'mAP', 'recall'. If is list, the first setting in the list will
-             be used to evaluate metric.
-        proposal_nums (Sequence[int]): Proposal number used for evaluating
-            recalls, such as recall@100, recall@1000.
-            Default: (100, 300, 1000).
-        eval_mode (str): 'area' or '11points', 'area' means calculating the
```
BIGWangYuDong, line 30: It is suggested to keep these parameters because it is easy for users to check parameters. And do not need to check in mmeval.VOCMeanAP. 
maybe we can use 
```
super().__init__(
            classwise_result=True,
            dist_backend=dist_backend,
            use_legacy_coordinate=True,
            xxx=xxx,
            xxx=xxx,...
)
```

mmdet/evaluation/metrics/voc_metric.py
```diff
@@ -91,85 +72,60 @@ def process(self, data_batch: dict, data_samples: Sequence[dict]) -> None:
...
-            and the values are corresponding results.
+        This method would be invoked by ``mmengine.Evaluator``.
         """
-        logger: MMLogger = MMLogger.get_current_instance()
-        gts, preds = zip(*results)
-        eval_results = OrderedDict()
-        if self.metric == 'mAP':
-            assert isinstance(self.iou_thrs, list)
-            dataset_type = self.dataset_meta.get('DATASET_TYPE')
-            if dataset_type in ['VOC2007', 'VOC2012']:
```
BIGWangYuDong, line 125: Does this logic still keep in mmeval?

ice-tong: `dataset_type` is used to choose [which `tpfp_fn` to use](https://github.com/open-mmlab/mmdetection/blob/e63459234d27b90a770e0a4bcc126b4371e5d2ad/mmdet/evaluation/functional/mean_ap.py#L614) and [get the class names for print map summary](https://github.com/open-mmlab/mmdetection/blob/e63459234d27b90a770e0a4bcc126b4371e5d2ad/mmdet/evaluation/functional/mean_ap.py#L771).

MMEval is no need to choose `tpfp_fn` from inputs now and get the class names from `dataset_meta['CLASSES']` directly. So I remove this stuff.

mmdet/evaluation/metrics/voc_metric.py
ice-tong: Updated

mmdet/evaluation/metrics/voc_metric.py
ice-tong: Update a simple check of dataset type and eval mode

BIGWangYuDong:
LGTM

mmdet/evaluation/metrics/voc_metric.py
```diff
@@ -1,86 +1,104 @@
...
+                 scale_ranges: Optional[List[Tuple]] = None,
+                 num_classes: Optional[int] = None,
+                 eval_mode: str = 'area',
+                 use_legacy_coordinate: bool = True,
+                 nproc: int = 4,
+                 drop_class_ap: bool = True,
+                 dist_backend: str = 'torch_cuda',
+                 **kwargs) -> None:
+
+        metric = kwargs.pop('metric', None)
```
BIGWangYuDong, line 65: We may delete these warning

PR #8942, opened 2022-10-05
make compatible with `ConfigDict` and `dict`

MambaWong:
If `AnchorHead` is used as the net head and `dict` is used as the `train_cfg` parameter, an error will be reported. 
That is because `dict` does not support python object attributes interface like `train_cfg.assigner`, but `ConfigDict` does.

```python
cfg = dict(
    type='mmdet.AnchorHead',
    num_classes=80,
    in_channels=256,
    train_cfg=dict(
        assigner=dict(type='MaxIoUAssigner', pos_iou_thr=0.5,
                      neg_iou_thr=0.3)))

anchor_head = MODELS.build(cfg)
```
> AttributeError: class `AnchorHead` in mmdet/models/dense_heads/anchor_head.py: 'dict' object has no attribute 'assigner'

### DIFF
--- a/mmdet/models/dense_heads/anchor_head.py
+++ b/mmdet/models/dense_heads/anchor_head.py
@@ -86,10 +86,10 @@ class AnchorHead(BaseDenseHead):
         self.train_cfg = train_cfg
         self.test_cfg = test_cfg
         if self.train_cfg:
-            self.assigner = TASK_UTILS.build(self.train_cfg.assigner)
+            self.assigner = TASK_UTILS.build(self.train_cfg['assigner'])
             if train_cfg.get('sampler', None) is not None:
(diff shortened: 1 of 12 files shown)


ZwwWayne:
Hi @MambaWong ,
Thanks for your kind PR. Could elaborate more in the PR message to indicate its necessity? For example, in which case the dict will be passed to the head

BIGWangYuDong:
Can we covert dict to ConfigDict in front of these code, so that we can avoid fixing these codes? Maybe the code will more elegant?

BIGWangYuDong:
LGTM, we can check all codes to avoid this potential error

RangiLyu (approved):

ZwwWayne (approved):

PR #8944, opened 2022-10-05
Fix bugs when ploting loss curve by analyze_logs.py

sltlls:

## Motivation

When ploting loss_cls of some run using analyze_logs.py, the numbers of xs and ys in function plot_curve(line 83,84) are not the same! Therefore causing error when running plt.plot().
After debug, I found that this problem was caused by the quantity mismatch of values of key 'step' and key 'loss_cls' in log_dicts returned by function load_json_logs. And this problem occured because validation lines in json log file producing after training also contains key 'step', causing the number of key 'step' in log_dicts 12 more than that in key 'loss_cls'.

## Modification

Simply add val_flag variable to judge whether current line is validation line, if it is, do not add the step info into the log_dicts.

## BC-breaking (Optional)

no

### DIFF
--- a/tools/analysis_tools/analyze_logs.py
+++ b/tools/analysis_tools/analyze_logs.py
@@ -170,6 +170,7 @@ def load_json_logs(json_logs):
             epoch = 1
             for i, line in enumerate(log_file):
                 log = json.loads(line.strip())
+                val_flag = False
                 # skip lines only contains one key
                 if not len(log) > 1:
                     continue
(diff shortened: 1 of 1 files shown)


jbwang1997:
Hello, thanks for your contribution. Seems the lint has failed. Please follows this [tutorial](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md) to install pre-commit, which can automatically fix the lint error.

sltlls:
do I need to recommit the PR after installing pre-commit in my local repository?

sltlls:
Ok, I have already pushed it after updating my local repository

wanghonglie (approved):

RangiLyu (approved):

PR #8947, opened 2022-10-05
[Fix] Fix the bug that MMDetWandbHook cannot log evaluation results

RockeyCoss:

## Motivation

Fix the bug that MMDetWandbHook cannot log evaluation results

## Modification

Fix the bug that MMDetWandbHook cannot log evaluation results

## Description

EvalHook calls all LoggerHook instances before evaluation. Since MMDetWandbHook class is a subclass of LoggerHook, `after_train_iter` hook is called twice in the iteration that performs evaluation. MMDetWandbHook logs the loss information in the first call and the evaluation results in the second call. If `self.commit` is True in the first call, then the second call will not log the evaluation results successfully because the metrics of this step are already persisted. See https://docs.wandb.ai/guides/track/log#stepwise-and-incremental-logging for more details.

I encountered this problem while training a panoramic segmentation model using MMDetWandbHook. I apologize for not verifying if this bug occurs when training instance segmentation models, as my GPU resources are pretty tight. 

This bug can be demonstrated using the following simple example:

script_a.py:
```python
import wandb

wandb.init(project='project_name',
                 entity='entity_name',
                 name='use commit == True')

for i in range(100):
    wandb.log({'a': i}, step=i, commit=True)
    wandb.log({'b': i**2}, step=i, commit=True)
```

script_b.py:
```python
import wandb

wandb.init(project='project_name',
                 entity='entity_name',
                 name='use commit == False')

for i in range(100):
    wandb.log({'a': i}, step=i, commit=False)
    wandb.log({'b': i**2}, step=i, commit=True)
```

Running these two scripts and you will find that wandb only logs the points of `a` in script_a.py while wandb can log points of `a` and `b` in script_b.py

### DIFF
--- a/mmdet/core/hook/wandblogger_hook.py
+++ b/mmdet/core/hook/wandblogger_hook.py
@@ -110,6 +110,9 @@ class MMDetWandbHook(WandbLoggerHook):
         self.log_evaluation = (num_eval_images > 0)
         self.ckpt_hook: CheckpointHook = None
         self.eval_hook: EvalHook = None
+        self.skip_flag = False
+        self.last_eval_iter = -100
+        self.temp_commit = None
 
     def import_wandb(self):
         try:
@@ -237,6 +240,27 @@ class MMDetWandbHook(WandbLoggerHook):
 
     @master_only
     def after_train_iter(self, runner):
+        # EvalHook calls all LoggerHook instances before evaluation. Since
+        # MMDetWandbHook class is a subclass of LoggerHook, `after_train_iter`
+        # hook is called twice in the iteration that performs evaluation.
+        # MMDetWandbHook logs the loss information in the first call and the
+        # evaluation results in the second call. If `self.commit` is True in
+        # the first call, then the second call will not log the evaluation
+        # results successfully because the metrics of this step are already
+        # persisted. See
+        # https://docs.wandb.ai/guides/track/log#stepwise-and-incremental-logging
+        # for more details.
+        if self.log_evaluation and self.eval_hook._should_evaluate(runner) \
+                and self.last_eval_iter != runner.iter:
+            self.skip_flag = True
+            self.temp_commit = self.commit
+            self.commit = False
+            self.last_eval_iter = runner.iter
+        else:
+            if self.skip_flag:
+                self.commit = self.temp_commit
+            self.skip_flag = False
+
         if self.get_mode(runner) == 'train':
             # An ugly patch. The iter-based eval hook will call the
             # `after_train_iter` method of all logger hooks before evaluation.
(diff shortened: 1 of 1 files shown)


hhaAndroid:
@RockeyCoss Thank you very much for your contribution. can you fix the ci？ Invite @ayulockin to take a look.

RockeyCoss:
> @RockeyCoss Thank you very much for your contribution. can you fix the ci？ Invite @ayulockin to take a look.

OK

ayulockin:
Hey @RockeyCoss, thank you for using it for panoptic segmentation. Since wandb doesn't support panoptic segmentatio yet (as you rightly pointed out) I didn't test it.

By evaluation result you mean the result controlled by this config param:
```
# We can set the evaluation interval to reduce the evaluation times
cfg.evaluation.interval = 2
```

If yes I tested the official colab [notebook](https://github.com/open-mmlab/mmdetection/blob/master/demo/MMDet_InstanceSeg_Tutorial.ipynb) with `MMDetWandbHook` with and without your fix. In the case of segmentation/detection the evaluation results are the same with ([wandb run page](https://wandb.ai/ayut/mmdetection-test/runs/3sykjrlc/overview?workspace=user-ayut)) and without ([wandb run page](https://wandb.ai/ayut/mmdetection-test/runs/1ibmonim/overview?workspace=user-ayut)) your fix.

If I am not correct can you help me understand what evaluation is in the context of MM ecosystem.

Also the changes made in this PR looks good to me. cc: @hhaAndroid

RockeyCoss:
Hey @ayulockin, thanks for testing my pr. It seems that the example from [notebook](https://github.com/open-mmlab/mmdetection/blob/master/demo/MMDet_InstanceSeg_Tutorial.ipynb) uses EpochBasedRunner instead of IterBasedRunner. My pr only affects the training process that uses IterBasedRunner. Only when using IterBasedRunner for training is MMDetWandbHook called twice in the iteration that calls EvalHook. You could check this: [EvalHook](https://github.com/open-mmlab/mmcv/blob/e417035f5d473b9f85d15ba01267d48d7f30e71e/mmcv/runner/hooks/evaluation.py#L261). Since MMDetWandbHook is a subclass of LoggerHook, MMDetWandbHook is called by the EvalHook's `after_train_iter` function. Then, since the priority of EvalHook is higher than MMDetWandbHook, MMDetWandbHook is called after EvalHook finishes its job. Thus, MMDetWandbHook is called a total of two times in one iteration, which caused the problem mentioned in the `Description` part of my pr information.

ayulockin:
Ah, thanks for correcting me; my bad. I will test it with `IterBasedRunner` and let you know. :)

PR #8948, opened 2022-10-06
Fixed inconsistency in 'nn.Conv2d' weight channels

MambaWong:
When `stacked_convs` is 0, the value of `feat_channels` and `in_channels` may be different.


wanghonglie:
fix the lint failed on CircleCI

PR #8980, opened 2022-10-09
Fix CircleCI and readthedoc build failed

hhaAndroid:

## Motivation

Fix CI

## Modification

Note: 

The following writing will cause `readthedoc` to fail to build and the code can't use relative paths

`## [MemoryProfilerHook](../../../mmdet/engine/hooks/memory_profiler_hook.py)`

The correct writing is as follows

```shell
## MemoryProfilerHook

[Memory profiler hook](https://github.com/open-mmlab/mmdetection/blob/3.x/mmdet/engine/hooks/memory_profiler_hook.py) 
```

## BC-breaking (Optional)

None

### DIFF
--- a/tests/test_models/test_seg_heads/test_panoptic_fpn_head.py
+++ b/tests/test_models/test_seg_heads/test_panoptic_fpn_head.py
@@ -15,7 +15,7 @@ class TestPanopticFPNHead(unittest.TestCase):
             num_things_classes=2,
             num_stuff_classes=2,
             in_channels=32,
-            inner_channels=1)
+            inner_channels=32)
         head.init_weights()
         assert_allclose(head.conv_logits.bias.data,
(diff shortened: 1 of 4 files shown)


RangiLyu (approved):

tests/test_models/test_seg_heads/test_panoptic_fpn_head.py
```diff
@@ -14,8 +14,8 @@ def test_init_weights(self):
         head = PanopticFPNHead(
             num_things_classes=2,
             num_stuff_classes=2,
-            in_channels=1,
```
ZwwWayne, line 17: consider use a smaller number to reduce GPU memory cost

hhaAndroid: Just initialized, not really running.

PR #9000, opened 2022-10-11
Update loading.py

Daa98:
RLE 'counts' is in str type.


BIGWangYuDong:
image
gt_mask['counts'] is a list in coco dataset

PR #9004, opened 2022-10-12
feat(device): Add automatic device select for samplers

Qiza-lyhm:
## Motivation

Add device select for samplers, thus they can use MLU or other devices.

## Modification

Add ```get_device``` and ```sync_random_seed(seed, device)``` to infinite_sampler and class_aware_sampler.

## BC-breaking (Optional)

No BC-breaking.

## Use cases (Optional)

No cases.

## Checklist


Qiza-lyhm:
This PR has tested by ```python -m torch.distributed.launch --nproc_per_node=8 tools/train.py configs/retinanet/retinanet_r50_fpn_90k_coco.py  --launcher pytorch``` with MLU backend.

mmdet/datasets/samplers/class_aware_sampler.py
```diff
@@ -58,7 +59,8 @@ def __init__(self,
         # Must be the same across all workers. If None, will use a
         # random seed shared among workers
         # (require synchronization among all workers)
-        self.seed = sync_random_seed(seed)
+        device = get_device()
+        self.seed = sync_random_seed(seed, device)
```
ZwwWayne, line 63: can we simply call get_device inside the function of sync_random_seed? Thus, we do not need to modify the code of samplers and only need to update the function sync_random_seed

ZwwWayne: We can use the proposed logic.

PR #9015, opened 2022-10-13
[Fix] Fix bugs of `empty_instances` when predicting without nms in roi_head.

jbwang1997:

## Motivation

When `rcnn_test_cfg is None` the score of a empty instance need to have shape of (0, num_classes + 1). But now, it has shape (0, ). This will cause soft teacher crashes sometimes.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/utils/misc.py
+++ b/mmdet/models/utils/misc.py
@@ -149,7 +149,8 @@ def empty_instances(batch_img_metas: List[dict],
             Defaults to False.
         num_classes (int): num_classes of bbox_head. Defaults to 80.
         score_per_cls (bool):  Whether to generate class-aware score for
-            the empty instance. Defaults to False.
+            the empty instance. ``score_per_cls`` will be True when the model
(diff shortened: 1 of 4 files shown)


mmdet/models/utils/misc.py
```diff
@@ -145,6 +147,9 @@ def empty_instances(batch_img_metas: List[dict],
         box_type (str or type): The empty box type. Defaults to `hbox`.
         use_box_type (bool): Whether to warp boxes with the box type.
             Defaults to False.
+        num_classes (int): num_classes of bbox_head. Defaults to 80.
+        score_per_cls (bool):  Whether to generate class-aware score for
```
hhaAndroid, line 151: Need to explain in detail when is true？

hhaAndroid (approved):

ZwwWayne (approved):

PR #8847, opened 2022-09-20
fix voc07+12 training

Nioolek:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Fixed an issue where VOC07 +12 dataset could not be trained.
`ConcatDataset` in `mmengine`  judge whether datasets metainfo is equal.
`VOC07 DATASET_TYPE` is `VOC2007`, but `VOC12 DATASET_TYPE` is `VOC2012`.
580562bf8e6964e6e9435aedd1bf4f7
This resulted in an inability to train when using the VOC07+12 dataset:
image

## Modification

Methods:
- add `def metainfo(self)` in `VOCDataset` to delete `DATASET_TYPE`.
- Dont judge and warn in `voc_metric.py`

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #9058, opened 2022-10-18
[Fix] Add `ignore_key` in voc dataset to fix error

BIGWangYuDong:
## Motivation

VOCDataset has `DATASET_TYPE` key in metainfo which is different in VOC2007 and VOC2012. This will get an error when using ConcatDataset. Hence, we add an `ignore_key` in ConcatDataset to fix this error.

## Modification

add an `ignore_key` in ConcatDataset to fix this error.

This PR should be merged after https://github.com/open-mmlab/mmengine/pull/556  has been merged in MMEngine.

Before merging this PR, https://github.com/open-mmlab/mmdetection/pull/8847 can be temporarily fixed VOC ConcatDataset error.

### NOTE
This PR change the minimum version of mmengine to 0.3.0.

### DIFF
--- a/configs/_base_/datasets/voc0712.py
+++ b/configs/_base_/datasets/voc0712.py
@@ -38,6 +38,9 @@ train_dataloader = dict(
         times=3,
         dataset=dict(
             type='ConcatDataset',
+            # VOCDataset will add `DATASET_TYPE` in dataset.metainfo,
+            # which will get error if using Concatdataset. Adding
+            # `ignore_keys` can avoid this error.
             ignore_keys=['DATASET_TYPE'],
             datasets=[
                 dict(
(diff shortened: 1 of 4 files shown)


configs/_base_/datasets/voc0712.py
```diff
@@ -38,6 +38,7 @@
         times=3,
         dataset=dict(
             type='ConcatDataset',
+            ignore_keys=['DATASET_TYPE'],
```
RangiLyu, line 41: add comments to tell users why ignore_keys are needed here

RangiLyu (dismissed):

BIGWangYuDong:
The mmengine_minimum_version should be updated to 0.3.0 to avoid bc breaking

PR #9084, opened 2022-10-20
Set default values for meta keys in LoadMultiChannelImageFromFiles

wanghonglie:

## Motivation
resolve of https://github.com/open-mmlab/mmdetection/pull/2330

## Modification
Fix LoadMultiChannelImageFromFiles to set default values for meta keys similar to LoadImageFromFile.
Refactor Image loading classes to reduce code duplication

### DIFF
--- a/mmdet/datasets/pipelines/loading.py
+++ b/mmdet/datasets/pipelines/loading.py
@@ -64,7 +64,6 @@ class LoadImageFromFile:
         Returns:
             dict: The dict contains loaded image and meta information.
         """
-
         if self.file_client is None:
             self.file_client = mmcv.FileClient(**self.file_client_args)
         img, filename = self._load_image(results)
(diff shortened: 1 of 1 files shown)


mmdet/datasets/pipelines/loading.py
```diff
@@ -45,36 +44,44 @@ def __init__(self,
...
+        img, filename = self._load_image(results)
 
         results['filename'] = filename
         results['ori_filename'] = results['img_info']['filename']
         results['img'] = img
         results['img_shape'] = img.shape
         results['ori_shape'] = img.shape
+        # Set initial values for default meta_keys
+        num_channels = 1 if len(img.shape) < 3 else img.shape[2]
+        if self.__class__.__name__ == 'LoadMultiChannelImageFromFiles':
```
ZwwWayne, line 78: why cannot we write the logic in the module of LoadMultiChannelImageFromFiles?

PR #9078, opened 2022-10-19
fix jpg->png

wanghonglie:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

use seg_suffix to resolve the https://github.com/open-mmlab/mmdetection/pull/6490

## Modification

Please briefly describe what modification is made in this PR.

PR #9088, opened 2022-10-20
fix jpg->png in dev-3.x

wanghonglie:

## Motivation

migrate https://github.com/open-mmlab/mmdetection/pull/9078 to dev-3.x

### DIFF
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -115,7 +115,7 @@ class CocoDataset(BaseDetDataset):
         if self.data_prefix.get('seg', None):
             seg_map_path = osp.join(
                 self.data_prefix['seg'],
-                img_info['filename'].rsplit('.', 1)[0] + self.seg_suffix)
+                img_info['filename'].rsplit('.', 1)[0] + self.seg_map_suffix)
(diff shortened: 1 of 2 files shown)


mmdet/datasets/coco.py
```diff
@@ -115,7 +115,7 @@ def parse_data_info(self, raw_data_info: dict) -> Union[dict, List[dict]]:
         if self.data_prefix.get('seg', None):
             seg_map_path = osp.join(
                 self.data_prefix['seg'],
-                img_info['file_name'].replace('jpg', 'png'))
```
ZwwWayne, line 118: we can use seg_map_suffix now to be consistent with mmseg.

mmdet/datasets/coco.py
```diff
@@ -115,7 +115,7 @@ def parse_data_info(self, raw_data_info: dict) -> Union[dict, List[dict]]:
...
+                img_info['filename'].rsplit('.', 1)[0] + self.seg_map_suffix)
```
ZwwWayne, line 118: is the change from file_name to filename correct? Did you verify the modification?

Issue #6884, opened 2021-12-24
Is eval_recalls(proposal_nums = 1) same as COCO ARmax=1?? 


bhyun-kim:
Would eval_recalls (in mmdetection/mmdet/core/evaluation/recall.py) return same results as COCO ARmax = 1 if I use it with proposal_nums = 1 ??

PR #3497, opened 2020-08-05
Add metric items for COCO

xvjiarui:
This PR added `metric_items` option for COCO evaluation. 
It also fixed #3468

PR #9105, opened 2022-10-21
[bugfix] set default max_dets to same values as pycocotools

ritosonn:
## Motivation

The COCO dataset recall evaluation in mmdetection differs from the [original pycocotools](https://cocodataset.org/#detection-eval). This is because the mmdetection's default `maxDets` (`[100, 500, 1000]` ) are not equal to the others (`[1, 10, 100]`).

This change fixes issue https://github.com/open-mmlab/mmdetection/issues/6884.

## Modification

- Rename `proposal_nums` to `max_dets`
  - `proposal_nums` seems to be used in the first stage of two-stage object detection (many proposals will appear).
  - The name `max_dets` is compatible to coco evaluation.
- Change the default values to `cocoEval.params.maxDets` from `[100, 500, 1000]` to `[1, 10, 100]`
  - [This change](https://github.com/open-mmlab/mmdetection/pull/3497/files#diff-136b3f31689b4e74b64162ee5eb155777202e00231010a752df486a097a8633dR452) incorrectly sets all case to `[100, 500, 1000]` 
  - We can fix it to set `cocoEval.params.maxDets` only if `metric == "proposal"`. I didn't because you'll change this variable in other situations (such as detection of a large number of objects).

## BC-breaking (Optional)

- Default recall evaluation of COCO dataset (measured after https://github.com/open-mmlab/mmdetection/pull/3497/commits/e20c5c2c83b9deac1ea24915955cea6b56632458 )
  - If you want to get the previous value, you can set `maxDets=[1, 10, 100]` or simply see recall@100 (because recall@[100, 500, 1000] will get the same value in most experiments).

## Use cases (Optional)

COCO evaluation

## Checklist

### DIFF
--- a/mmdet/datasets/coco.py
+++ b/mmdet/datasets/coco.py
@@ -390,7 +390,7 @@ class CocoDataset(CustomDataset):
                           metrics,
                           logger=None,
                           classwise=False,
-                          proposal_nums=(100, 300, 1000),
+                          max_dets=(1, 10, 100),
                           iou_thrs=None,
                           metric_items=None):
         """Instance segmentation and object detection evaluation in COCO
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Hi @ritosonn ,
Thanks for your kind PR. The current change will introduce BC-breaking because the proposal_nums is no longer used. To avoid that and provide a more smooth experience to the users, could you also keep the argument of proposal_nums? When evaluating AP, we can use max_det to be align with pycocotools. When evaluating proposal AR, we can still use proposal_nums.

DiYaGo-zzh:
Hi @ritosonn ,
I have a question, I wonder if this maxDets is just in a different format than the print of pycocotools? That is, will the change of maxDets affect the calculation of its specific indicator values?And when i try to align with pycocotools and change coco.py file from your work, but when i use tools/test.py to calculate the indicator, it still prints in the previous format，as shown in the image. If i want use tools/test.py to align with pycocotools format,whether you need to modify other files?
图片

PR #9176, opened 2022-10-28
[Fix] Fix `pixel_decoder_type` discrimination in MaskFormer Head.

MengzhangLI:
## Motivation

Currently, MMDetection does not treat inter-codebases using modules very well. For example, in upstream codebase which `pip install mmdet` and uses type `mmdet.PixelDecoder`, the error would happen in this line because this `if else` would discriminate `mmdet.PixelDecoder` and `PixelDecoder` although they are the same.

image

### DIFF
--- a/mmdet/models/dense_heads/maskformer_head.py
+++ b/mmdet/models/dense_heads/maskformer_head.py
@@ -103,7 +103,7 @@ class MaskFormerHead(AnchorFreeHead):
         self.transformer_decoder = MODELS.build(transformer_decoder)
         self.decoder_embed_dims = self.transformer_decoder.embed_dims
         pixel_decoder_type = pixel_decoder.get('type')
-        if pixel_decoder_type == 'PixelDecoder' and (
+        if pixel_decoder_type.endswith('PixelDecoder') and (
                 self.decoder_embed_dims != in_channels[-1]
                 or enforce_decoder_input_project):
             self.decoder_input_proj = Conv2d(


ZwwWayne:
Hi @MengzhangLI 

image

mmdet/models/dense_heads/maskformer_head.py
```diff
@@ -103,7 +103,7 @@ def __init__(self,
         self.transformer_decoder = MODELS.build(transformer_decoder)
         self.decoder_embed_dims = self.transformer_decoder.embed_dims
         pixel_decoder_type = pixel_decoder.get('type')
-        if pixel_decoder_type == 'PixelDecoder' and (
+        if pixel_decoder_type.endswith('PixelDecoder') and (
```
RangiLyu, line 106: better to check the class type of `self.pixel_decoder` instead of the string

mmdet/models/dense_heads/maskformer_head.py
```diff
@@ -102,10 +103,10 @@ def __init__(self,
...
         self.transformer_decoder = MODELS.build(transformer_decoder)
         self.decoder_embed_dims = self.transformer_decoder.embed_dims
-        pixel_decoder_type = pixel_decoder.get('type')
-        if pixel_decoder_type == 'PixelDecoder' and (
-                self.decoder_embed_dims != in_channels[-1]
-                or enforce_decoder_input_project):
+        if isinstance(
+                self.pixel_decoder,
+                PixelDecoder) and (self.decoder_embed_dims != in_channels[-1]
+                                   or enforce_decoder_input_project):
```
Li-Qingyun, line 109: It brings unit test error: 
图片

I set track here, and found the conflict:

图片

When the self.pixel_decoder is the `TransformerEncoderPixelDecoder` which is inherited from the `PixelDecoder`. 

The `isinstance(self.pixel_decoder)` returns `True`, while the original `pixel_decoder_type == 'PixelDecoder'` returns `False`.

I tried replace it with `self.pixel_decoder._get_name() == 'PixelDecoder'`.
i.e.
```Python
pixel_decoder_type = self.pixel_decoder._get_name()
```
The ut pass.

Li-Qingyun (approved):

mmdet/models/dense_heads/maskformer_head.py
RangiLyu, line 109:
```suggestion
        if type(self.pixel_decoder) == PixelDecoder and (
                self.decoder_embed_dims != in_channels[-1]
                or enforce_decoder_input_project):
```

RangiLyu (commented):
LGTM

PR #9179, opened 2022-10-29
[Enhance] Add box_type support for `DynamicSoftLabelAssigner`

liuyanyi:

## Motivation

Add box_type support for Dynamic Soft Label Assigner.

Allow mmrotate to use `DynamicSoftLabelAssigner` directly with out any modify.

Also i found that the `find_inside_points` method results are slightly different than the original version, `find_inside_points` allow the points on the edge.
https://github.com/open-mmlab/mmdetection/blob/816637e1677146d790b4b4a1fc05c400d9260dfb/mmdet/structures/bbox/horizontal_boxes.py#L333-L334

I think we could add a `keep_edge` or `eps` arg in `find_inside_points` to control that behavior, also looking forward to your suggestions.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/task_modules/assigners/dynamic_soft_label_assigner.py
+++ b/mmdet/models/task_modules/assigners/dynamic_soft_label_assigner.py
@@ -7,6 +7,7 @@ from mmengine.structures import InstanceData
 from torch import Tensor
 
 from mmdet.registry import TASK_UTILS
+from mmdet.structures.bbox import BaseBoxes
 from mmdet.utils import ConfigType
 from .assign_result import AssignResult
 from .base_assigner import BaseAssigner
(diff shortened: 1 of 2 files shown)


mmdet/models/task_modules/assigners/dynamic_soft_label_assigner.py
```diff
@@ -82,11 +83,16 @@ def assign(self,
                                                    dtype=torch.long)
 
         prior_center = priors[:, :2]
-        lt_ = prior_center[:, None] - gt_bboxes[:, :2]
-        rb_ = gt_bboxes[:, 2:] - prior_center[:, None]
+        if isinstance(gt_bboxes, BaseBoxes):
```
RangiLyu, line 86: add unit tests to cover this case

ZwwWayne (approved):

Issue #9159, opened 2022-10-27
SoloV2 dealing with empty images at Train Time


Ahmed14974:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmdetection/issues) and [Discussions](https://github.com/open-mmlab/mmdetection/discussions) but cannot get the expected help.
- [X] I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
- [X] The bug has not been fixed in the [latest version (master)](https://github.com/open-mmlab/mmdetection) or [latest version (3.x)](https://github.com/open-mmlab/mmdetection/tree/dev-3.x).

### 💬 Describe the reimplementation questions

Our use case has a custom dataset, where we want to use SoloV2 for instance segmentation. 
For some of the images in the training set, there are no objects to be detected, and so the COCO dataset json file we create has no annotations to correspond to that image. If we run any SoloV2 model with such a training set, it gives an error. 

Does current SoloV2 implementation have support for empty training images? If not then how should we implement it?

### Environment

sys.platform: linux
Python: 3.9.11 (main, Mar 29 2022, 19:08:29) [GCC 7.5.0]
CUDA available: True
GPU 0,1,2,3: NVIDIA RTX A4000
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.1, V11.1.105
GCC: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
PyTorch: 1.12.0
PyTorch compiling details: PyTorch built with:
  - GCC 9.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.6.0 (Git Hash 52b5f107dd9cf10910aaa19cb47f3abf9b349815)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.3
[...]

PR #9185, opened 2022-10-31
[Fix] Fix solov2 cannot dealing with empty gt image

BIGWangYuDong:
Solve PR: https://github.com/open-mmlab/mmdetection/issues/9159

`loss_mask` may empty when `num_pos = 0`. Hence, fix the num_pos=0  case to support dealing with empty gt image

### DIFF
--- a/mmdet/models/dense_heads/solov2_head.py
+++ b/mmdet/models/dense_heads/solov2_head.py
@@ -559,7 +559,7 @@ class SOLOV2Head(SOLOHead):
         if num_pos > 0:
             loss_mask = torch.cat(loss_mask).sum() / num_pos
         else:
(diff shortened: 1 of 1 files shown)


Czm369 (approved):
LGTM

mmdet/models/dense_heads/solov2_head.py
```diff
@@ -559,7 +559,7 @@ def loss(self,
         if num_pos > 0:
             loss_mask = torch.cat(loss_mask).sum() / num_pos
         else:
-            loss_mask = torch.cat(loss_mask).mean()
+            loss_mask = mask_feats.new_zeros(1).mean()
```
ZwwWayne, line 562: usually we use the results.sum() * 0 to include all parameters in the graph

Issue #8962, opened 2022-10-08
Make scipy as a default dependency in runtime


ZwwWayne:
### What is the problem this feature will solve?

Since DETR series become very popular, bipart matching becomes a default.
Therefore, it would be better to make scipy a default runtime dependency rather than optional

### What is the feature you are proposing to solve the problem?

Move scipy from https://github.com/open-mmlab/mmdetection/blob/master/requirements/optional.txt#L3 to runtime.txt

BIGWangYuDong:
should update both dev-3.x and dev branch

PR #9187, opened 2022-10-31
[Enhance] Make scipy as a default dependency in runtime

BIGWangYuDong:
Since DETR series become very popular, bipart matching becomes a default.
Therefore, it would be better to make scipy a default runtime dependency rather than optional

Resolve PR: https://github.com/open-mmlab/mmdetection/issues/8962


Czm369 (dismissed):
LGTM

ZwwWayne:
should also clean these try excee in the code

BIGWangYuDong:
Done

RangiLyu (approved):
LGTM

ZwwWayne (approved):

PR #9259, opened 2022-11-07
[Fix]: fix (Cached) MixUp padding

plutoyuxie:

### DIFF
--- a/mmdet/datasets/transforms/transforms.py
+++ b/mmdet/datasets/transforms/transforms.py
@@ -2464,9 +2464,9 @@ class MixUp(BaseTransform):
         ori_img = results['img']
         origin_h, origin_w = out_img.shape[:2]
         target_h, target_w = ori_img.shape[:2]
-        padded_img = np.zeros(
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Hi @plutoyuxie ,
Thanks for your PR. Could you explain more.about the motivation of this PR, e.g., why do you need to change the padded part should be ones instead of zeros？

RangiLyu (approved):

plutoyuxie:
> Hi @plutoyuxie , Thanks for your PR. Could you explain more.about the motivation of this PR, e.g., why do you need to change the padded part should be ones instead of zeros？

Dear @ZwwWayne，below is my motivation of this PR:

When I use `browse_dataset.py` to visual check `CachedMixUp` augmentation in my own project, some kind of artificial grey strips appeared. 
The padded area is not filled with the given `pad_val` properly. 
It is not clear in coco dataset, but very obvious when a x-ray dataset was used in my project, while the image background is white and `pad_val=255.0` is needed in my case.

PR #9265, opened 2022-11-08
[Update] Rename `utils/typing.py` to `utils/typing_utils.py`

BIGWangYuDong:
Rename `utils/typing.py` to `utils/typing_utils.py`. 

After renaming, `collect_env` can run normally.

## Note
This change may introduce BC breaking, so downstream repo such as MMYOLO needs to pay attention to synchronous changes.

It can simply use `from mmdet.utils import xxx(in typing_utils)`, which will not occur BC-breaking


hhaAndroid (approved):

ZwwWayne:
suggest to also ad UT to test collect_env to avoid such an issue in the future.

PR #9284, opened 2022-11-09
[Enhance]: Add resume from specific checkpoint in CLI

JosonChan1998:

## Motivation

Add resume from a specific checkpoint in CLI.
It would be better to add an option for resuming from a specific checkpoint.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -26,6 +26,8 @@ def parse_args():
         action='store_true',
         help='enable automatically scaling LR.')
     parser.add_argument(
+        '--resume-from', help='the checkpoint file to resume from')
+    parser.add_argument(
         '--resume',
         action='store_true',
         help='resume from the latest checkpoint in the work_dir automatically')
@@ -101,7 +103,11 @@ def main():
                                '"auto_scale_lr.base_batch_size" in your'
                                ' configuration file.')
 
+    # resume is determined in this priority: resume from > auto_resume
     cfg.resume = args.resume
+    if args.resume_from is not None:
+        cfg.resume = True
+        cfg.load_from = args.resume_from
 
     # build the runner from config
     if 'runner_type' not in cfg:


tools/train.py
```diff
@@ -101,7 +103,11 @@ def main():
                                '"auto_scale_lr.base_batch_size" in your'
                                ' configuration file.')
 
+    # resume is determined in this priority: resume from > auto_resume
```
BIGWangYuDong, line 106: Is that necessary to add `load-from` too?

JosonChan1998: I don't think it's necessary. Since `load-from` may generally use to load pretrain weights from configs, right?

JosonChan1998: @BIGWangYuDong Hi! Still need the `load-from` option?

RangiLyu: No need to add load-from in this PR.

README.md
```diff
@@ -77,9 +77,9 @@ Apart from MMDetection, we also released [MMEngine](https://github.com/open-mmla
 
 **v3.0.0rc3** was released in 4/11/2022:
 
-- Support training detection models in Detectron2
```
BIGWangYuDong, line 80: please remember to merge the latest `dev-3.x` branch

JosonChan1998: Done

tools/train.py
```diff
@@ -103,6 +105,12 @@ def main():
     if args.resume:
         cfg.resume = args.resume
 
+    # resume is determined in this priority: resume from > auto_resume
+    cfg.resume = args.resume
```
RangiLyu, line 109: This line will cause resume conflict. Please refer to the logic in https://github.com/open-mmlab/mmdetection/pull/9287

JosonChan1998: Done. I have removed this line.

RangiLyu (approved):
LGTM

ZwwWayne:
Hi @JosonChan1998 ,
Thanks for your kind PR. It seems that there are conflicts between your PR and the targeted branch. Could you solve the conflicts so that we could eventually merge this PR?

You can find more details in the place shown as below:
image

JosonChan1998:
> Hi @JosonChan1998 , Thanks for your kind PR. It seems that there are conflicts between your PR and the targeted branch. Could you solve the conflicts so that we could eventually merge this PR?
> 
> You can find more details in the place shown as below: image

Done.

RangiLyu:
After our discussion, we suggest following MMClassification's argparse like this:
https://github.com/open-mmlab/mmclassification/blob/75e502ed755bacdbe72aacef797bfe5438c5dc09/tools/train.py#L19-L26

This makes the resume arg to support both "auto-resume" and "resume from a specific location".
```python
    # resume training
    if args.resume == 'auto':
        cfg.resume = True
        cfg.load_from = None
    elif args.resume is not None:
        cfg.resume = True
        cfg.load_from = args.resume
```
You can refer to these codes to change your PR.

JosonChan1998:
> After our discussion, we suggest following MMClassification's argparse like this: https://github.com/open-mmlab/mmclassification/blob/75e502ed755bacdbe72aacef797bfe5438c5dc09/tools/train.py#L19-L26
> 
> This makes the resume arg to support both "auto-resume" and "resume from a specific location".
> 
> ```python
>     # resume training
>     if args.resume == 'auto':
>         cfg.resume = True
>         cfg.load_from = None
>     elif args.resume is not None:
>         cfg.resume = True
>         cfg.load_from = args.resume
> ```
> 
> You can refer to these codes to change your PR.

That looks good. I have changed it following your suggestion.

RangiLyu (approved):

PR #9341, opened 2022-11-17
[Community] Add 'Projects/' folder, and the first example project.

RangiLyu:
# Motivation
Introducing new models and features into OpenMMLab's algorithm libraries has long been complained to be troublesome due to the rigorous requirements on code quality, which could hinder the fast iteration of SOTA models and might discourage potential contributors from sharing their latest outcome here.

# Modification
This PR adds a new projects/ folder, which will be a place for some experimental models/features. Implementations inside might be not quite perfect but already fine to produce some exciting results. We hope that this PR can help us better embrace the contribution from our community. We also add the first example project to illustrate what we expect a good project to have.

### DIFF
--- a/projects/example_project/README.md
+++ b/projects/example_project/README.md
@@ -1,0 +1,110 @@
+# Dummy ResNet Wrapper
+
+This is an example README for community `projects/`. We have provided detailed explanations for each field in the form of html comments, which are visible when you read the source of this README file. If you wish to submit your project to our main repository, then all the fields in this README are mandatory for others to understand what you have achieved in this implementation. For more details, read our [contribution guide](https://mmdetection.readthedocs.io/en/3.x/notes/contribution_guide.html) or approach us in [Discussions](https://github.com/open-mmlab/mmdetection/discussions).
+
+## Description
+
+<!-- Share any information you would like others to know. For example:
+Author: @xxx.
+This is an implementation of \[XXX\]. -->
+
+This project implements a dummy ResNet wrapper, which literally does nothing new but prints "hello world" during initialization.
+
+## Usage
+
+<!-- For a typical model, this section should contain the commands for training and testing. You are also suggested to dump your environment specification to env.yml by `conda env export > env.yml`. -->
+
+### Training commands
+
+In MMDetection's root directory, run the following command to train the model:
+
+```bash
+python tools/train.py projects/example_project/configs/faster-rcnn_dummy-resnet_fpn_1x_coco.py
+```
+
+### Testing commands
+
+In MMDetection's root directory, run the following command to test the model:
+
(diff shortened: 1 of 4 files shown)


projects/example_project/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+  - [ ] Training-time correctness
+
+    <!-- If you are reproducing the result from a paper, checking this item means that you should have trained your model from scratch based on the original paper's specification and verified that the final result matches the report within a minor error range. -->
+
+- [ ] Milestone 3: Good to be a part of our core package!
+
+  - [ ] Type hints and docstrings
+
+    <!-- Ideally *all* the methods should have [type hints](https://www.pythontutorial.net/python-basics/python-type-hints/) and [docstrings](https://google.github.io/styleguide/pyguide.html#381-docstrings). [Example](https://github.com/open-mmlab/mmdetection/blob/5b0d5b40d5c6cfda906db7464ca22cbd4396728a/mmdet/datasets/transforms/transforms.py#L41-L169) -->
```
hhaAndroid, line 92: https://github.com/open-mmlab/mmdetection/blob/5b0d5b40d5c6cfda906db7464ca22cbd4396728a/mmdet/datasets/transforms/transforms.py   

This link can be replaced

projects/example_project/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+  - [ ] Unit tests
+
+    <!-- Unit tests for each module are required. [Example](https://github.com/open-mmlab/mmdetection/blob/5b0d5b40d5c6cfda906db7464ca22cbd4396728a/tests/test_datasets/test_transforms/test_transforms.py#L35-L88) -->
```
hhaAndroid, line 96: same

projects/example_project/README.md
RangiLyu: Replace with what?

hhaAndroid: https://github.com/open-mmlab/mmdetection/blob/3.x/mmdet/datasets/transforms/transforms.py

RangiLyu: > https://github.com/open-mmlab/mmdetection/blob/3.x/mmdet/datasets/transforms/transforms.py

If the code is updated, the line permalink will mismatch. It must link to a specific commit hash.

projects/example_project/README.md
ZwwWayne: I think we could use permlink

projects/example_project/README.md
ZwwWayne: I think we can keep the current link for now.

hhaAndroid (approved):

PR #9391, opened 2022-11-27
clip saturation value to valid range [0, 1]

MambaWong:
For image(type=float32), after convert bgr to hsv by opencv, valid saturation value range is [0, 1].

### DIFF
--- a/mmdet/datasets/transforms/transforms.py
+++ b/mmdet/datasets/transforms/transforms.py
@@ -981,6 +981,10 @@ class PhotoMetricDistortion(BaseTransform):
         # random saturation
         if saturation_flag:
             img[..., 1] *= saturation_value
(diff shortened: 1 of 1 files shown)


mmdet/datasets/transforms/transforms.py
```diff
@@ -981,6 +981,10 @@ def transform(self, results: dict) -> dict:
         # random saturation
         if saturation_flag:
             img[..., 1] *= saturation_value
+            # For image(type=float32), after convert bgr to hsv by opencv,
+            # valid saturation value range is [0, 1]
+            if saturation_value > 1:
+                img[..., 1][img[..., 1] > 1] = 1
```
RangiLyu, line 987:
```suggestion
                img[..., 1] = img[..., 1].clip(0, 1)
```

RangiLyu (approved):
LGTM

Issue #8145, opened 2022-06-07
deadlock using Wandb


linnanwang:
Thanks for your error report and we appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**
Hello mmdet developers,

We found the training loop can be dead lock in some places if we use multiGPU training and enable wandb tracking. Single GPU works perfectly fine. I only tested with YOLOX. Please see the command below.

**Reproduction**

1. What command or script did you run?

```
./tools/dist_train.sh ./configs/yolox/yolox_s_8x8_300e_coco.py 2
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?
No
3. What dataset did you use?
MSCOCO

**Environment**

1. Please run `python mmdet/utils/collect_env.py` to collect necessary environment information and paste it here.
sys.platform: linux
Python: 3.8.13 (default, Mar 28 2022, 11:38:47) [GCC 7.5.0]
CUDA available: True
GPU 0,1: Quadro GV100
CUDA_HOME: /usr/local/cuda
NVCC: Build cuda_11.3.r11.3/compiler.29745058_0
GCC: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
PyTorch: 1.10.0
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.2.3 (Git Hash 7336ca9f055cf1bfa13efb658fe15dc9b41f0740)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.3
[...]

PR #9476, opened 2022-12-10
Fixed deadlock issue related with MMDetWandbHook

MilkClouds:
Co-authored-by: WangYudong <yudong.wang@akane.waseda.jp>

## Motivation

https://github.com/open-mmlab/mmdetection/issues/8145
To summarize problem, when we use `MMDetWandbHook` after `TextLoggerHook`, problem such as deadlock or `AssertionError: loss log variables are different across GPUs!` occurs. These are because of same reason and I summarized the reason in https://github.com/open-mmlab/mmdetection/issues/8145#issuecomment-1345343076 here.

## Modification

Made MMDetWandbHook to clear `runner.log_buffer` on every process, including process for GPU 1,2,3...

I'm wondering if I should consider `after_val_epoch` of `MMDetWandbHook`, because since there is not `after_val_epoch` on `MMDetWandbHook`, I didn't deal with this method. If someone implement `after_val_epoch` in `MMDetWandbHook` with `@master_only`, same error can occurs.

I guess it is sufficient to mention above lines in `MMDetWandbHook` as I did, but it may be a potential threat.
```
    # for the reason of this double-layered structure, refer to
    # https://github.com/open-mmlab/mmdetection/issues/8145#issuecomment-1345343076
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/core/hook/wandblogger_hook.py
+++ b/mmdet/core/hook/wandblogger_hook.py
@@ -202,10 +202,14 @@ class MMDetWandbHook(WandbLoggerHook):
             # Log ground truth data
             self._log_data_table()
 
-    @master_only
+    # for the reason of this double-layered structure, refer to
+    # https://github.com/open-mmlab/mmdetection/issues/8145#issuecomment-1345343076
     def after_train_epoch(self, runner):
         super(MMDetWandbHook, self).after_train_epoch(runner)
+        self._after_train_epoch(runner)
 
+    @master_only
+    def _after_train_epoch(self, runner):
         if not self.by_epoch:
             return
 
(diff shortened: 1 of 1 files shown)


MilkClouds:
There was already a similar commit on mmcls https://github.com/open-mmlab/mmclassification/pull/1242, which I didn't know! (this commit was only 5 days ago)
I spent quite a long time trying to figure out the cause of the bug, but there was someone who already figured out the causes :( 
Anyway, please merge this PR. Even though detailed implementation is different with https://github.com/open-mmlab/mmclassification/pull/1242, purpose is the same.


Also I guess we don't need separating `after_train_epoch`, since runner's output buffer is cleared before epoch. I'll test once and update PR.

hhaAndroid (approved):

MilkClouds:
I've verified my commit.

PR #9481, opened 2022-12-13
Adjust FocalLoss and QualityFocalLoss to fit MMYOLO

kitecats:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/losses/gfocal_loss.py
+++ b/mmdet/models/losses/gfocal_loss.py
@@ -1,4 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from functools import partial
+
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -52,7 +54,7 @@ def quality_focal_loss(pred, target, beta=2.0):
 
 
 @weighted_loss
-def quality_focal_loss_tensor_target(pred, target, beta=2.0):
+def quality_focal_loss_tensor_target(pred, target, beta=2.0, activated=False):
     """`QualityFocal Loss <https://arxiv.org/abs/2008.13367>`_
     Args:
         pred (torch.Tensor): The prediction with shape (N, C), C is the
@@ -61,11 +63,17 @@ def quality_focal_loss_tensor_target(pred, target, beta=2.0):
             classification score with shape (N, C), C is the number of classes.
         beta (float): The beta parameter for calculating the modulating factor.
             Defaults to 2.0.
+        activated (bool): Whether the input is activated.
+            If True, it means the input has been activated and can be
+            treated as probabilities. Else, it should be treated as logits.
+            Defaults to False.
     """
     # pred and target should be of the same size
     assert pred.size() == target.size()
-
-    pred_sigmoid = pred.sigmoid()
+    if activated:
+        pred_sigmoid = pred
+    else:
+        pred_sigmoid = pred.sigmoid()
 
     scale_factor = pred_sigmoid
     target = target.type_as(pred)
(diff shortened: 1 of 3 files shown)


mmdet/models/losses/gfocal_loss.py
```diff
@@ -166,8 +167,10 @@ def forward(self,
             pred (torch.Tensor): Predicted joint representation of
                 classification and quality (IoU) estimation with shape (N, C),
                 C is the number of classes.
-            target (tuple([torch.Tensor])): Target category label with shape
-                (N,) and target quality label with shape (N,).
+            target (Union(tuple([torch.Tensor]),Torch.Tensor)): The type is
+                tuple ,it should be includedTarget category label with
```
hhaAndroid, line 171:
```suggestion
                tuple, it should be included target category label with
```

mmdet/models/losses/gfocal_loss.py
```diff
@@ -166,8 +167,10 @@ def forward(self,
...
+                shape (N,) and target quality label with shape (N,).The type
+                is torch.Tensor, target should be One-Hot form.
```
hhaAndroid, line 173:
```suggestion
                is torch.Tensor, the target should be one-hot form with soft weights.
```

mmdet/models/losses/gfocal_loss.py
```diff
@@ -184,6 +188,24 @@ def forward(self,
                 calculate_loss_func = quality_focal_loss_with_prob
             else:
                 calculate_loss_func = quality_focal_loss
+
+            if isinstance(target, torch.Tensor):
+                # this means that target is already in One-Hot form.
```
RangiLyu, line 193: QFL uses soft targets instead of binary One-Hot labels. Please fix the comments.

mmdet/models/losses/gfocal_loss.py
```diff
@@ -184,6 +188,24 @@ def forward(self,
...
+                assert pred.dim() == target.dim()
+                if target.dim() == 4:
```
RangiLyu, line 195: Should this method accept so many different types & shapes of input? Why not flatten the targets outside the loss function?

mmdet/models/losses/gfocal_loss.py
```diff
@@ -184,6 +188,24 @@ def forward(self,
...
+                pred = pred.reshape(-1, target.shape[-1])
+                target = target.reshape(-1, target.shape[-1])
+
+                # get category label and quality label info
+                pos_ind, pos_value = torch.max(target, dim=-1)
+                # set bg label
+                pos_ind[~(pos_value > 0)] = target.shape[-1]
+                target = (pos_ind.long(), pos_value.type(pred.dtype))
+
             loss_cls = self.loss_weight * calculate_loss_func(
```
hhaAndroid, line 209: Instead of modifying this class, add a new `quality_focal_loss_tensor_target` function to support any dim tensor target

mmdet/models/losses/gfocal_loss.py
hhaAndroid: This is because bce loss supports arbitrary dimensions. We hope to implement loss direct replacement

mmdet/models/losses/gfocal_loss.py
hhaAndroid: Yes

hhaAndroid (approved):

ZwwWayne (approved):

PR #9495, opened 2022-12-15
[Fix] Fix yolof training stability

twmht:

## Motivation

the training of yolof is unstable, when `cls_scores` is large, `cls_scores.exp()` may be overflowed, thus we can't clamp by INF.

Without this fix it's very hard to converage when training widerface. 

## Modification

clamp INF before exp

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdet/models/dense_heads/yolof_head.py
```diff
@@ -121,8 +121,11 @@ def forward_single(self, x: Tensor) -> Tuple[Tensor, Tensor]:
         # implicit objectness
         objectness = objectness.view(N, -1, 1, H, W)
         normalized_cls_score = cls_score + objectness - torch.log(
-            1. + torch.clamp(cls_score.exp(), max=INF) +
-            torch.clamp(objectness.exp(), max=INF))
+            1. + torch.clamp(
```
hhaAndroid, line 124: Thank you very much for your feedback. The formatting of this is not very good, can you simply write a function to beautify it?

twmht: sounds good.

ZwwWayne:
Hi @twmht ,
It would be great if you could also provide benchmark results.

PR #9498, opened 2022-12-15
Deprecating `np.bool` Type Alias

zachcoleman:

## Motivation

Gets rid of instances of `np.bool` to increase compatibility with newer versions of `numpy`. For more details see: https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations. 

_Note: `np.bool` has just been an alias for `bool`_ 

## Modification

Replaces all `np.bool` instances w/ `bool`.

## BC-breaking (Optional)

Not applicable

## Use cases (Optional)

Not applicable

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/PyCQA/flake8
-    rev: 3.8.3
+    rev: 5.0.4
     hooks:
       - id: flake8
   - repo: https://github.com/PyCQA/isort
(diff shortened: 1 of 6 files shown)


hhaAndroid:
@zachcoleman Hi, I would like to ask if there is a problem if the user version is lower than 1.20?

zachcoleman:
> Details

@hhaAndroid There should be no problem. `np.bool` is just an alias and another variable pointing to `bool`. As an example here's numpy v1.19.5:


and another version v1.16.6 (this version is ~3 years old):


I believe using `bool` is the most compatible way to define these arrays across versions. There is work for `1.25.0` (scheduled for June 2023) to add `np.bool` back into numpy. 

I will follow up when I get an answer [here](https://github.com/numpy/numpy/issues/22021) to try and determine the best course forward. My instinct is that `bool` is probably the best way to do this for <1.20, 1.20-1.24, and 1.25+ based on the logic to map `bool` accordingly (see [here](https://numpy.org/doc/stable/user/basics.types.html#:~:text=bool%20means%20np.bool_%2C)).

RangiLyu (approved):

PR #9503, opened 2022-12-16
delete parrots-awared code of roi extractor

chenyuwang814:
## Motivation

SenseParrots has already aligned some apis of autograd with pytorch so that requiring grad of "feats" is no longer needed by parrots, which will also cause an error.

## Modification

We deleted a bunch of parrots-awared code within roi extractor so that the parameter "feats" no longer requires grad.


ZwwWayne:
Hi @chenyuwang814 ,

image

RangiLyu (approved):

zytx121 (approved):
LGTM

ZwwWayne:
We need to migrate this PR into MMDet 3.x

PR #9518, opened 2022-12-20
[WIP] [Feature] Refactor assigner and sampler

BIGWangYuDong:
Refactor AssignerResult and SamplingResult. 

- keep AssignerResult and SamplingResult have the same logic, which means can deprecate `pseudo sampler`
- the input of SamplingResult changed to InstanceData, which can have better scalability, (can deprecate MaskSamplingResult)
- Use AvoidCUDAOOM instead of `assign_on_cpu` in assigner

### DIFF
--- a/mmdet/models/task_modules/samplers/sampling_result.py
+++ b/mmdet/models/task_modules/samplers/sampling_result.py
@@ -1,14 +1,12 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import warnings
-
 import numpy as np
 import torch
+from mmengine.structures import InstanceData
 from torch import Tensor
 
 from mmdet.structures.bbox import BaseBoxes, cat_boxes
 from mmdet.utils import util_mixins
 from mmdet.utils.util_random import ensure_rng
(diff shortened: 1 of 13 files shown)


mmdet/models/task_modules/samplers/sampling_result.py
```diff
@@ -84,60 +82,118 @@ class SamplingResult(util_mixins.NiceRepr):
...
+        self.neg_inds = neg_inds
+        self.num_pos = max(pos_inds.numel(), 1)
+        self.num_neg = max(neg_inds.numel(), 1)
+        self.avg_factor_with_neg = avg_factor_with_neg
+        self.avg_factor = self.num_pos + self.num_neg \
+            if avg_factor_with_neg else self.num_pos
+
+        self.pos_assigned_gt_inds = pred_instances.gt_inds[pos_inds] - 1
+
+    def update_sampling_result(self,
```
RangiLyu, line 102: This logic should be placed in the sampler, instead of in a method of the AssignResult.

PR #9537, opened 2022-12-28
[Fix] Deprecating old type alias due to new version of numpy

RangeKing:
## Motivation

Deprecating old type alias due to new version of numpy.

## Modification

1. Replace all `np.int` with `np.int64`.
2. Replace all `np.bool` with `bool`.


RangiLyu (approved):

zytx121 (approved):
LGTM

PR #9670, opened 2023-01-29
fix amp pooling overflow

zylo117:

## Motivation

Pooling, mostly avg pooling and max pooling, in amp(float16) mode may overflow, because when every element in one channel sums up, it might be greater than the upper limit of float16, which is only 65536, then it overflows and starts from the lower limit. This is very common when the weights are not converged.

I trained coco with rtmdet from scratch in amp mode, I encounter nan loss issue. I thought this may be overflowing. And when I turn off amp mode, no more nan issue. So I confirmed this is pooling overflow issue, I know it because I have fixed it in another repo before.

## Modification

Detect whether input tensor dtype in pooling is float16 or float32, if float16, force casting it to float32 and do pooling, then cast the result back to float16.

I tested by training rtmdet on coco from scratch, no more nan. But in other network, this issue may still exist.

### DIFF
--- a/mmdet/models/backbones/csp_darknet.py
+++ b/mmdet/models/backbones/csp_darknet.py
@@ -115,7 +115,11 @@ class SPPBottleneck(BaseModule):
 
     def forward(self, x):
         x = self.conv1(x)
-        x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1)
+        if x.dtype == torch.float16:
+            x = x.float()
+            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1).half()
+        else:
+            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1)
         x = self.conv2(x)
         return x
 
--- a/mmdet/models/backbones/csp_darknet.py
+++ b/mmdet/models/backbones/csp_darknet.py
@@ -116,7 +116,7 @@ class SPPBottleneck(BaseModule):
     def forward(self, x):
         x = self.conv1(x)
         with torch.autocast(enabled=False):
-            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], 
+            x = torch.cat([x] + [pooling(x) for pooling in self.poolings],
                           dim=1)
         x = self.conv2(x)
         return x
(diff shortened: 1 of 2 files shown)


mmdet/models/backbones/csp_darknet.py
```diff
@@ -115,7 +115,11 @@ def __init__(self,
 
     def forward(self, x):
         x = self.conv1(x)
-        x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1)
+        if x.dtype == torch.float16:
+            x = x.float()
+            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1).half()
+        else:
+            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1)
```
ZwwWayne, line 122:
```suggestion
        with torch.autocast(enabled=False):
            x = torch.cat([x] + [pooling(x) for pooling in self.poolings], dim=1)
```

ZwwWayne: can we simplified the modification as the suggestion?

zylo117: looks good to me

ZwwWayne:
Hi @zylo117 ,
The overall PR looks good to me now and can be merged If the modification has been verified to resolve the issue in your case. Furthermore, please rebase your code to the most recent dev-3.x, which resolves the lint issue.

zylo117:
> Hi @zylo117 , The overall PR looks good to me now and can be merged If the modification has been verified to resolve the issue in your case. Furthermore, please rebase your code to the most recent dev-3.x, which resolves the lint issue.

What is 
"""
yapf.....................................................................Failed
- hook id: yapf
- files were modified by this hook

"""
I don't know what kind of formatter you are using. And when I format the code with yapf with either pep8 or google style, many original lines were changed. I think the original file was using custom formatter. Can't the ci perform auto formatting?

RangiLyu:
Thanks for your contribution! Please run `pip install pre-commit` and then run `pre-commit run --all-files` to fix the lint.

zylo117:
now that the unit test failed. I think it has nothing to do with my commits?

ZwwWayne:
> now that the unit test failed. I think it has nothing to do with my commits?

Yes, you are right. We will fix it in other PR.

hhaAndroid:
@zylo117 Both rtmdet and yolov5 are trained with amp. I have never encountered this problem. Is there any reference for this? Thank you!

PR #9693, opened 2023-01-30
make sure conv2 of CSPNeXtBlock controlled by use_depthwise

zylo117:
## Motivation

Whether use_depthwise=False/True, self.conv2 of CSPNeXtBlock is depthwise conv, it's not controlled by this argument.

## Modification
```        
conv = DepthwiseSeparableConvModule if use_depthwise else ConvModule

self.conv2 = DepthwiseSeparableConvModule(
=>
self.conv2 = conv(
```

### DIFF
--- a/mmdet/models/layers/csp_layer.py
+++ b/mmdet/models/layers/csp_layer.py
@@ -126,7 +126,7 @@ class CSPNeXtBlock(BaseModule):
             padding=1,
             norm_cfg=norm_cfg,
             act_cfg=act_cfg)
-        self.conv2 = DepthwiseSeparableConvModule(
+        self.conv2 = conv(
             hidden_channels,
             out_channels,
             kernel_size,


RangiLyu:
The second conv module of CSPNeXtBlock must be DepthwiseSeparableConvModule.

zylo117:
> The second conv module of CSPNeXtBlock must be DepthwiseSeparableConvModule.

I see.
So can we consider adding a argument that forces all conv modules being vanilla conv?

I have trained rtmdet-tiny with full vanilla conv, mAP 41.5 looks good to me, even a little better than original rtmdet-tiny.
Here is the log and the weights.

https://drive.google.com/file/d/1KFbvVh0BMwmMVTwRnaqvC55gvsod9GRv/view?usp=sharing

Some edge devices, like amlogic npu, doesn't support per_channel quantization, when using depthwise conv, the result will be completely wrong. So it's very important to eliminate depthwise conv before depolyment. 

If real-time model like rtmdet can't deploy on edge devices, then rtmdet is not very practical.

RangiLyu:
Vanilla conv with 5x5 kernel size will increase the number of parameters by a large amount. This is why the full vanilla conv version achieves better mAP.

RangiLyu:
If you want to use an argument to control the type of conv2 layer, you need to use another argument name rather than `use_depthwise`.

PR #9143, opened 2022-10-26
[Feature] Support auto import modules from registry.

RangiLyu:
## Motivation

The registry now supports auto-import modules from the given location.

register_all_modules before running is no longer needed. The modules will be lazy-imported during building.

- [x] This PR can be merged after https://github.com/open-mmlab/mmengine/pull/643. The MMEngine version should be updated.

resolve https://github.com/open-mmlab/mmyolo/issues/312

PR #9751, opened 2023-02-10
[DOCS] import  `register_all_modules` and call it 

chg0901:
## Motivation

We should register all modules in mmdet into the registries before runing the rest codes

I have seen similar pull in another notebook, but this problem doesn't exist in [3.x](https://github.com/open-mmlab/mmdetection/edit/3.x/docs/en/user_guides/inference.md)

## Modification
import  `register_all_modules` and call `register_all_modules()` before runing the rest codes

```
from mmdet.utils import register_all_modules

# Register all modules in mmdet into the registries
register_all_modules()
```

## Checklist

### DIFF
--- a/docs/en/user_guides/inference.md
+++ b/docs/en/user_guides/inference.md
@@ -17,8 +17,11 @@ import mmcv
 from mmcv.transforms import Compose
 from mmengine.utils import track_iter_progress
 from mmdet.registry import VISUALIZERS
+from mmdet.utils import register_all_modules
 from mmdet.apis import init_detector, inference_detector
 
+# Register all modules in mmdet into the registries
+register_all_modules()
 
 # Specify the path to model config and checkpoint file
 config_file = 'configs/faster_rcnn/faster-rcnn_r50-fpn_1x_coco.py'


docs/en/user_guides/inference.md
```diff
@@ -17,8 +17,11 @@ import mmcv
 from mmcv.transforms import Compose
 from mmengine.utils import track_iter_progress
 from mmdet.registry import VISUALIZERS
+from mmdet.utils import register_all_modules
```
RangiLyu, line 20: from mmengine.registry import init_default_scope

docs/en/user_guides/inference.md
```diff
@@ -17,8 +17,11 @@ import mmcv
...
 from mmdet.apis import init_detector, inference_detector
 
+# Register all modules in mmdet into the registries
+register_all_modules()
```
RangiLyu, line 24: replace this with
```
init_default_scope(config.get('default_scope', 'mmdet'))
```
after load the config

RangiLyu: We no longer need to call register_all_modules after introducing the 'auto_import' in PR https://github.com/open-mmlab/mmdetection/pull/9143.

ZwwWayne: yes

Issue #8746, opened 2022-09-07
ConcatDataset didn't have attribute `update_skip_type_keys` like MultiImageMixDataset


ykk648:
When I trained yolox with 3 datasets, it fails when `num_last_epochs` starts, releated issue: #5997 #6055

`
No mosaic and mixup aug now!
train_loader.dataset.update_skip_type_keys(self.skip_type_keys)
raise AttributeError
`

jbwang1997:
Only `MultiImageMixDataset` has update_skip_type_keys, please warp your dataset with `MultiImageMixDataset`.

ykk648:
> 

that means only single dataset can be loaded during last aug-free epoches of yolox training, maybe `update_skip_type_keys` should added to `ConcatDataset`, I dont have time to do this so thank you for your reply.

PR #9752, opened 2023-02-10
fix: add update skip type keys interface for concat dataset

wep21:

## Motivation

resolve https://github.com/open-mmlab/mmdetection/issues/8746

## Modification

add update_skip_type_keys into concat dataset to support concat dataset training with yolox hooks.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/datasets/dataset_wrappers.py
+++ b/mmdet/datasets/dataset_wrappers.py
@@ -149,6 +149,19 @@ class ConcatDataset(_ConcatDataset):
             self.datasets[0].data_infos = original_data_infos
             return eval_results
 
+    def update_skip_type_keys(self, skip_type_keys):
+        """Update skip_type_keys. It is called by an external hook.
+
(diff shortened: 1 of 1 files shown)


mmdet/datasets/dataset_wrappers.py
```diff
@@ -149,6 +149,19 @@ def evaluate(self, results, logger=None, **kwargs):
             self.datasets[0].data_infos = original_data_infos
             return eval_results
 
+    def update_skip_type_keys(self, skip_type_keys):
+        """Update skip_type_keys. It is called by an external hook.
+
+        Args:
+            skip_type_keys (list[str], optional): Sequence of type
```
ZwwWayne, line 156:
```suggestion
            skip_type_keys (list[str]): Sequence of type string to be skipped in pipeline
```

PR #9784, opened 2023-02-16
[Fix] Fix VOC metrics

BIGWangYuDong:
1. Fixed VOC metrics, add `eval_mode` in `eval_map`, or it will used `area` when calculating VOC2007, which is wrong.

### DIFF
--- a/mmdet/evaluation/metrics/dump_det_results.py
+++ b/mmdet/evaluation/metrics/dump_det_results.py
@@ -17,11 +17,6 @@ class DumpDetResults(DumpResults):
     segmentation masks into RLE format.
 
     Args:
-        keep_gt (bool): Whether to dump `gt_instances` simultaneously. It
-            should be True if offline VOCMetric is used. Defaults to False.
-        keep_gt_ignore (bool): Whether to dump `ignored_instances`
-            simultaneously. It should be True if offline VOCMetric is used.
-            Defaults to False.
         out_file_path (str): Path of the dumped file. Must end with '.pkl'
             or '.pickle'.
         collect_device (str): Device name used for collecting results from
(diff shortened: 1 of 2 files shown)


mmdet/evaluation/metrics/dump_det_results.py
```diff
@@ -17,20 +17,35 @@ class DumpDetResults(DumpResults):
     segmentation masks into RLE format.
 
     Args:
+        keep_gt (bool): Whether dumped `gt_instances` simultaneously. It
```
RangiLyu, line 20:
```suggestion
        keep_gt (bool): Whether to dump `gt_instances` simultaneously. It
```

mmdet/evaluation/metrics/dump_det_results.py
```diff
@@ -17,20 +17,35 @@ class DumpDetResults(DumpResults):
...
+            should be True if offline VOCMetric is used. Defaults to False.
+        keep_gt_ignore (bool): Whether dumped `ignored_instances`
```
RangiLyu, line 22:
```suggestion
        keep_gt_ignore (bool): Whether to dump `ignored_instances`
```

mmdet/evaluation/metrics/dump_det_results.py
```diff
@@ -17,20 +17,35 @@ class DumpDetResults(DumpResults):
...
+        self.keep_gt_ignore = keep_gt_ignore
+
     def process(self, data_batch: dict, data_samples: Sequence[dict]) -> None:
         """transfer tensors in predictions to CPU."""
         data_samples = _to_cpu(data_samples)
         for data_sample in data_samples:
             # remove gt
-            data_sample.pop('gt_instances', None)
-            data_sample.pop('ignored_instances', None)
+            if not self.keep_gt:
```
RangiLyu, line 45: I'm wondering whether we should keep gt in the dumped results. Because it is duplicated with the dataset annotation and is not compressed.
We can modify the `eval_metric.py` to load the annotation during offline evaluation.

RangiLyu (approved):

PR #9850, opened 2023-02-28
[Refactor] Refactor CityScapesMetric by using MMEval

BIGWangYuDong:
Related PR: https://github.com/open-mmlab/mmeval/pull/98

Before:
Iaj9THTFd2

After:
lCygZ2KAN8

### DIFF
--- a/mmdet/evaluation/metrics/cityscapes_metric.py
+++ b/mmdet/evaluation/metrics/cityscapes_metric.py
@@ -1,37 +1,30 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import os
+import itertools
 import os.path as osp
-import shutil
-from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


mmdet/evaluation/metrics/cityscapes_metric.py
```diff
@@ -40,8 +30,8 @@ class CityScapesMetric(BaseMetric):
             evaluation. It is useful when you want to format the result
             to a specific format and submit it to the test server.
             Defaults to False.
-        keep_results (bool): Whether to keep the results. When ``format_only``
-            is True, ``keep_results`` must be True. Defaults to False.
+        classwise (bool): Whether to return the computed  results of each
```
RangiLyu, line 33:
```suggestion
        classwise (bool): Whether to return the computed results of each
```

Issue #9878, opened 2023-03-05
[Bug] VOCDataset print wrong class names in some cases


LoSealL:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmdetection/issues) and [Discussions](https://github.com/open-mmlab/mmdetection/discussions) but cannot get the expected help.
- [X] I have read the [FAQ documentation](https://mmdetection.readthedocs.io/en/latest/faq.html) but cannot get the expected help.
- [X] The bug has not been fixed in the [latest version (master)](https://github.com/open-mmlab/mmdetection) or [latest version (3.x)](https://github.com/open-mmlab/mmdetection/tree/dev-3.x).

### Task

I'm using the official example scripts/configs for the officially supported tasks/models/datasets.

### Branch

master branch https://github.com/open-mmlab/mmdetection

### Environment

sys.platform: win32
Python: 3.8.7 (tags/v3.8.7:6503f05, Dec 21 2020, 17:59:51) [MSC v.1928 64 bit (AMD64)]
CUDA available: True
GPU 0: NVIDIA GeForce GTX 1660 Ti
CUDA_HOME: C:\opt\CUDA\v11.6
NVCC: Cuda compilation tools, release 11.6, V11.6.124
MSVC: 用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.35.32215 版
GCC: n/a
PyTorch: 1.13.1+cu116
PyTorch compiling details: PyTorch built with:
  - C++ Version: 199711
  - MSVC 192829337
  - Intel(R) Math Kernel Library Version 2020.0.2 Product Build 20200624 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.6.0 (Git Hash 52b5f107dd9cf10910aaa19cb47f3abf9b349815)
  - OpenMP 2019
  - LAPACK is enabled (usually provided by MKL)
  - CPU capability usage: AVX2
  - CUDA Runtime 11.6
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_37,code=compute_37
[...]

PR #9879, opened 2023-03-05
[Fix] Fix VOCDataset print wrong class names (#9878)

LoSealL:
## Motivation

To fix a print error when evaluating VOCDataset (#9878).

## Modification

Assign `ds_name` by `self.CLASSES` directly without the condition on `year` attribute.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/core/evaluation/mean_ap.py
+++ b/mmdet/core/evaluation/mean_ap.py
@@ -532,7 +532,8 @@ def eval_map(det_results,
              tpfp_fn=None,
              nproc=4,
              use_legacy_coordinate=False,
-             use_group_of=False):
+             use_group_of=False,
+             mode='area'):
     """Evaluate mAP of a dataset.
 
     Args:
@@ -572,6 +573,9 @@ def eval_map(det_results,
             Default: False.
         use_group_of (bool): Whether to use group of when calculate TP and FP,
             which only used in OpenImages evaluation. Default: False.
+        mode (str): 'area' or '11points', 'area' means calculating the area
+            under precision-recall curve, '11points' means calculating the
+            average precision of recalls at [0, 0.1, ..., 1]. Default: 'area'.
 
     Returns:
         tuple: (mAP, [dict, dict, ...])
@@ -677,7 +681,6 @@ def eval_map(det_results,
             recalls = recalls[0, :]
             precisions = precisions[0, :]
             num_gts = num_gts.item()
-        mode = 'area' if dataset != 'voc07' else '11points'
         ap = average_precision(recalls, precisions, mode)
         eval_results.append({
             'num_gts': num_gts,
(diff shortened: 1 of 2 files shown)


BIGWangYuDong:
This is not a good way to fix this potential error. voc07 is still necessary when calculating AP:

image

Hence, seems it should keep voc07.

LoSealL:
> This is not a good way to fix this potential error. voc07 is still necessary when calculating AP:
> 
> image
> 
> Hence, seems it should keep voc07.

Move `mode` as an input argument to `eval_map` and make its default value to `'area'`.

BIGWangYuDong:
> > This is not a good way to fix this potential error. voc07 is still necessary when calculating AP:
> > image
> > Hence, seems it should keep voc07.
> 
> Move `mode` as an input argument to `eval_map` and make its default value to `'area'`.

Yep, that's a way to solve this problem, but you still need to check whether the dataset is VOC2007

LoSealL:
I don't think it's a good idea to check dataset name (or type) in evaluating function, otherwise there are massive information or even conflict information in that scope.

Besides, we've already told the function to use mode `11points` to calculate AUC, and we do use this mode in VOCDataset iff it's voc2007, so if we are using `VOCDataset` of year 2007, then in the evaluation it must be voc07 and use the `11points` mode.

mmdet/core/evaluation/mean_ap.py
```diff
@@ -572,6 +573,9 @@ def eval_map(det_results,
             Default: False.
         use_group_of (bool): Whether to use group of when calculate TP and FP,
             which only used in OpenImages evaluation. Default: False.
+        mode (str): 'area' or '11points', 'area' means calculating the area
+            under precision-recall curve, '11points' means calculating the
+            average precision of recalls at [0, 0.1, ..., 1]. Default: 'area'.
```
BIGWangYuDong, line 578:
```suggestion
            average precision of recalls at [0, 0.1, ..., 1]. Defaults to 'area'.
```

BIGWangYuDong:
I think the logic is OK now, a single comment need fix

mmdet/core/evaluation/mean_ap.py
LoSealL: Change to `Defaults to` style? But at line 575 and others they use `Default:` style

BIGWangYuDong: Yes, you can change all `Default:` to `Defaults to` in this file.

BIGWangYuDong: We are going to unify using `defaults to` in mmdet

LoSealL: I've changed all `"Default:"` to `"Defaults to"` in this file.

BIGWangYuDong:
@LoSealL LGTM now. Would you mind having a test on VOC2007, to check whether it is OK? 

Also, lint and ci are failed. 

image

please follow https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md to fix them

LoSealL:
> @LoSealL LGTM now. Would you mind having a test on VOC2007, to check whether it is OK?
> 
> Also, lint is failed.
> 
> image
> 
> please follow https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md#2-configure-pre-commit to fix them

Fixing

BIGWangYuDong (approved):
LGTM

ZwwWayne (approved):

PR #10018, opened 2023-03-24
[Feature] Support defrom detr fp16 training

RangiLyu:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/dense_heads/detr_head.py
+++ b/mmdet/models/dense_heads/detr_head.py
@@ -8,6 +8,7 @@ from mmcv.cnn import Linear
 from mmcv.cnn.bricks.transformer import FFN
 from mmengine.model import BaseModule
 from mmengine.structures import InstanceData
+from mmengine.runner import amp
 from torch import Tensor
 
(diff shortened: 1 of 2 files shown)


mmdet/models/dense_heads/detr_head.py
```diff
@@ -223,13 +224,13 @@ def loss_by_feat(
...
             'for batch_gt_instances_ignore setting to None.'
-
-        losses_cls, losses_bbox, losses_iou = multi_apply(
-            self.loss_by_feat_single,
-            all_layers_cls_scores,
-            all_layers_bbox_preds,
-            batch_gt_instances=batch_gt_instances,
-            batch_img_metas=batch_img_metas)
+        with amp.autocast(enabled=False):
+            losses_cls, losses_bbox, losses_iou = multi_apply(
```
hhaAndroid, line 228: Don't we need to find out exactly which operators are not suitable for fp16? Setting the entire loss to fp32 shouldn't be efficient enough？

PR #10070, opened 2023-04-03
[Fix] Fix inference_detector fail on non-cuda and non-cpu devices

Metal-joker:
## Motivation

`mmdet.apis.inference_detector` will fail on non-cuda and non-cpu devices because of this https://github.com/open-mmlab/mmdetection/blob/e9cae2d0787cd5c2fc6165a6061f92fa09e48fb1/mmdet/apis/inference.py#L146-L153

`torch.Tensor.is_cuda` will return `False` when tensor is not on cuda device, so when this tensor is on other non-cuda devices, this code will not move tensor to the same device with the model weight and this will cause `inference_detector` crash.

## Modification

use `mmcv.device.scatter` to move the input to device when model weight is not on cpu.

## BC-breaking (Optional)

This change is unlikely to cause any issues, but I have tested it on both MLU device and CPU to be sure.

### DIFF
--- a/mmdet/apis/inference.py
+++ b/mmdet/apis/inference.py
@@ -5,8 +5,9 @@ from pathlib import Path
 import mmcv
 import numpy as np
 import torch
+from mmcv.device import scatter
 from mmcv.ops import RoIPool
-from mmcv.parallel import collate, scatter
+from mmcv.parallel import collate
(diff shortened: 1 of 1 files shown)


Metal-joker:
ci seems on hold, did I miss something?

PR #10116, opened 2023-04-09
[Fix] YOLOXModeSwitchHook does not switch the mode when resumed from the checkpoint after switched

i-aki-y:
## Motivation

The current `YOLOXModeSwitchHook` does not switch the mode when it is resumed from the checkpoint that was saved after the mode switching. Because the switch is triggered only once at the condition `(epoch + 1) ==  max_epochs - num_last_epochs`. So if we resume the training from the checkpoint saved after the condition, the switch will not be triggered. Note that the mode switch is implemented by changing the dataset state that is not saved in the checkpoint.

I think it is a bug that the behavior of the training strategy unexpectedly changed after the resumption. 

## Modification

I tried to make the condition more robust by adding a condition that checks whether a mode switch has taken place.
1. whether a trigger epoch has arrived.
2. whether a mode switch has taken place.

## BC-breaking (Optional)

I think there is no BC-breaking. 

## Use cases (Optional)

This is a bug fix.

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
4. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls.
5. The documentation has been modified accordingly, like docstring or example tutorials.


mmdet/datasets/dataset_wrappers.py
```diff
@@ -167,3 +167,13 @@ def update_skip_type_keys(self, skip_type_keys):
             isinstance(skip_type_key, str) for skip_type_key in skip_type_keys
         ])
         self._skip_type_keys = skip_type_keys
+
+    def has_all_skip_type_keys(self, skip_type_keys):
```
hhaAndroid, line 171: This will result in backwards compatibility breaking.

hhaAndroid: " think there is a simpler solution, which is to introduce a 'has_switch' flag in this hook to handle it.

i-aki-y: @hhaAndroid  Thank you for your review. I fixed this.

i-aki-y:
@hhaAndroid  The `PipelineSwitchHook` looks to use the same logic.
Can I apply the same fix when this fix could be approved?

hhaAndroid:
> @hhaAndroid The `PipelineSwitchHook` looks to use the same logic. Can I apply the same fix when this fix could be approved?

Yes

hhaAndroid:
@i-aki-y  Please fix the CI

i-aki-y:
@hhaAndroid I added a fix of PipelineSwitchHook, and CI looked to be fixed by restart.

hhaAndroid (approved):

PR #10206, opened 2023-04-22
[Feature] Support Label-Studio SAM Semi-Label

vansin:
RTMDet predict the bbox as the SAM Prompt，and SAM out the Mask, the Label Studio refine the mask.

image


hhaAndroid:
@vansin mmdet just needs to add a reference link

Issue #10323, opened 2023-05-13
a bug in losses/utils.py


haocaia:
When I use ATSSHead and loss_bbox="SmoothL1Loss", I get the following error: 
  ```
File "mmdetection/mmdet/models/losses/utils.py", line 52, in weight_reduce_loss
RuntimeError: The size of tensor a (4) must match the size of tensor b (4299) at non-singleton dimension 1 
```
I found that this is due to the first dimension missing from the input weight's shape:
When calculating loss_cls, the shape of weight is [num_anchors, 1]
```
loss: torch.Size([389376, 63]) # 63 is the number of class
weight: torch.Size([389376, 1])                                                                                                                                                                                                      
```       
When calculating loss_bbox, the shape of weight is [num_anchor]
```
loss: torch.Size([4299, 4])                                                                                               
weight: torch.Size([4299])        
``` 

since the first dimension is missing, a RuntimeError is generated when calculating loss * weight.
```
# mmdetection/mmdet/models/losses/utils.py
# 48 lines
...
# if weight is specified, apply element-wise weight
    if weight is not None:
        loss = loss * weight
...
```

I don't know if this error is general (whether it is in setting other heads or losses).
But setting ATSSHead+SmoothL1Loss raises this error.

I rewrote SmoothL1Loss, adding `weight = weight.unsqueeze(-1)` in forward() solved the problem.

PR #10423, opened 2023-06-01
[fix] fix `pred` and `weight` dims unmatch in [Smooth]L1Loss

MambaWong:
#4982 #10323 All similar problems are caused by `weight`. Should we take them to `weighted_loss` for processing.

### DIFF
--- a/mmdet/models/losses/smooth_l1_loss.py
+++ b/mmdet/models/losses/smooth_l1_loss.py
@@ -96,6 +96,10 @@ class SmoothL1Loss(nn.Module):
         Returns:
             Tensor: Calculated loss
         """
+        if weight is not None and not torch.any(weight > 0):
+            if pred.dim() == weight.dim() + 1:
+                weight = weight.unsqueeze(1)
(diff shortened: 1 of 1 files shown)


mmdet/models/losses/smooth_l1_loss.py
```diff
@@ -96,6 +96,10 @@ def forward(self,
         Returns:
             Tensor: Calculated loss
         """
+        if weight is not None and not torch.any(weight > 0):
+            if pred.dim() == weight.dim() + 1:
+                weight = weight.unsqueeze(1)
+            return (pred * weight).sum()  # 0
```
hhaAndroid, line 102:
```suggestion
            return (pred * weight).sum()
```

mmdet/models/losses/smooth_l1_loss.py
```diff
@@ -149,6 +153,10 @@ def forward(self,
...
```
hhaAndroid, line 159:
```suggestion
            return (pred * weight).sum()
```

PR #10517, opened 2023-06-15
[Docs] Fix wrong doc in Mosaic and CachedMosaic

jamiechoi1995:
Image size after mosaic should be twice the input size

### DIFF
--- a/mmdet/datasets/transforms/transforms.py
+++ b/mmdet/datasets/transforms/transforms.py
@@ -2217,13 +2217,13 @@ class Mosaic(BaseTransform):
     Modified Keys:
 
     - img
-    - img_shape
+    - img_scale
     - gt_bboxes (optional)
     - gt_bboxes_labels (optional)
     - gt_ignore_flags (optional)
 
     Args:
-        img_scale (Sequence[int]): Image size after mosaic pipeline of single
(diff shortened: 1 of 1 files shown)


mmdet/datasets/transforms/transforms.py
```diff
@@ -2217,13 +2217,13 @@ class Mosaic(BaseTransform):
     Modified Keys:
 
     - img
-    - img_shape
+    - img_scale
```
hhaAndroid, line 2220:
```suggestion
    - img_shape
```

mmdet/datasets/transforms/transforms.py
```diff
@@ -3278,13 +3278,13 @@ class CachedMosaic(Mosaic):
...
```
hhaAndroid, line 3281:
```suggestion
    - img_shape
```

mmdet/datasets/transforms/transforms.py
hhaAndroid: You may have misunderstood, this field in docstring is to indicate which keys in result dictionary will be modified.

jamiechoi1995: Oh, I see, it is the result['img_shape']

hhaAndroid (approved):

PR #10648, opened 2023-07-15
Fix: remove some unicode chars from `en/` docs

evdcush:
## Motivation

I am trying to `make latexpdf` the `en/` docs.
Fullwidth chars cause issues for `latexpdf` build targets, so I replaced them with their normal (half-width) variants.

## Modification

- Added some missing `papertype` indicators in docs
- Replaced some unicode fullwidth chars with their halfwidth variants in docs and python code

## Use cases

You used to be able to download the docs as `pdf` until `v3.1.0`.
I don't know what broke this, but at least some of the work done in this branch should help narrow down probable causes of the pdf build breakage.

## Checklist

1. ✅ Pre-commit or other linting tools are used to fix the potential lint issues.
2. ✅ The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. ✅ If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMPreTrain.
4. ✅ The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmdet/models/detectors/glip.py
+++ b/mmdet/models/detectors/glip.py
@@ -58,8 +58,8 @@ def remove_punctuation(text: str) -> str:
         str: The text with punctuation removed.
     """
     punctuation = [
-        '|', ':', ';', '@', '(', ')', '[', ']', '{', '}', '^', '\'', '\"',
-        '\'', '`', '?', '$', '%', '#', '!', '&', '*', '+', ',', '.'
+        '|', ':', ';', '@', '(', ')', '[', ']', '{', '}', '^', '\'', '\"', '`',
+        '?', '$', '%', '#', '!', '&', '*', '+', ',', '.'
     ]
     for p in punctuation:
         text = text.replace(p, '')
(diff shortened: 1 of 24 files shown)


evdcush:
How long do PRs typically remain in the `pr_stage_test/hold` phase?

mmdet/models/detectors/glip.py
```diff
@@ -58,8 +58,8 @@ def remove_punctuation(text: str) -> str:
         str: The text with punctuation removed.
     """
     punctuation = [
-        '|', ':', ';', '@', '(', ')', '[', ']', '{', '}', '^', '\'', '\"', '’',
```
hhaAndroid, line 61: This part should not need to be modified, right?

evdcush: Reverted this diff in https://github.com/open-mmlab/mmdetection/pull/10648/commits/e95c7063bf1a7fce6f21ba748cd3415ece5167aa 🙆🏻‍♂️

hhaAndroid (approved):

PR #10733, opened 2023-08-02
Update documentation about the overlaps shape

ShaneXiangH:
## Motivation

Updated documentation about the overlaps shape. The previous documentation is outdated.


mmdet/models/task_modules/assigners/max_iou_assigner.py
```diff
@@ -165,7 +165,7 @@ def assign_wrt_overlaps(self, overlaps: Tensor,
 
         Args:
             overlaps (Tensor): Overlaps between k gt_bboxes and n bboxes,
-                shape(k, n).
+                shape(k, n, 1).
```
hhaAndroid, line 168: Please confirm once again if it is correct. Thank you.

PR #10868, opened 2023-08-31
Setting scope for image reader in video_demo

apatsekin:
Otherwise running it with mmyolo models triggers error, trying to find that module undery mmyolo scope


apatsekin:
@RangiLyu could you take a look

hhaAndroid (approved):

Issue #10722, opened 2023-07-29
The code may have some bugs


li624120638:
mmdet/models/data_preprocessors/data_preprocessor.py
DetDataPreprocessor._get_pad_shape():
`
        elif isinstance(_batch_inputs, torch.Tensor):
            ...
            pad_h = int(
                np.ceil(_batch_inputs.shape[1] / self.pad_size_divisor)) * self.pad_size_divisor
            pad_w = int( np.ceil(_batch_inputs.shape[2] / self.pad_size_divisor)) * self.pad_size_divisor
           ...
`
might be
`
elif isinstance(_batch_inputs, torch.Tensor):
            ...
            pad_h = int(
                np.ceil(_batch_inputs.shape[2] / self.pad_size_divisor)) * self.pad_size_divisor
            pad_w = int( np.ceil(_batch_inputs.shape[3] / self.pad_size_divisor)) * self.pad_size_divisor
           ...
`

hhaAndroid:
@li624120638  Yes, thank you very much for your feedback. May I ask if you can submit a pull request to fix it?

PR #10997, opened 2023-10-03
Fix Pad Shape

agpeshal:
## Motivation
#10722 

## Modification

Fix the bug and add test

## Checklist


mmdet/models/data_preprocessors/data_preprocessor.py
```diff
@@ -118,7 +118,7 @@ def forward(self, data: dict, training: bool = False) -> dict:
         Returns:
             dict: Data in the same format as the model input.
         """
-        batch_pad_shape = self._get_pad_shape(data)
+        batch_pad_shape = self.get_pad_shape(data)
```
hhaAndroid, line 121: Please do not modify the method name.

agpeshal: sure! I reverted the change the name.

hhaAndroid (dismissed):

PR #11104, opened 2023-10-29
Enable customization of kernel size in CSPLayer and DarknetBottleneck while preserving previously default kernel size

handleandwheel:

## Motivation

The default kernel size is 3 in DarknetBottleneck and in CSPLayer when using DarknetBottleneck. However, some network design requires larger kernel size (e.g. [Yolov6 Lite, where the kernel size is 5](https://github.com/meituan/YOLOv6/blob/main/configs/yolov6_lite/README.md)).

## Modification

I enable customization of kernel size in CSPLayer and DarknetBottleneck by adding input parameter `kernel_size` at the end of the formal parameter list, and set to 3 by default.

## BC-breaking (Optional)

This PR doesn't break the BC of the downstream repos.

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet/models/layers/csp_layer.py
+++ b/mmdet/models/layers/csp_layer.py
@@ -201,7 +201,8 @@ class CSPLayer(BaseModule):
                  init_cfg: OptMultiConfig = None) -> None:
         super().__init__(init_cfg=init_cfg)
         block = CSPNeXtBlock if use_cspnext_block else DarknetBottleneck
-        kernel_size = (5 if use_cspnext_block else 3) if kernel_size is None else kernel_size
+        kernel_size = (5 if use_cspnext_block else 3) \
+                        if kernel_size is None else kernel_size
         mid_channels = int(out_channels * expand_ratio)
         self.channel_attention = channel_attention
         self.main_conv = ConvModule(


mmdet/models/layers/csp_layer.py
```diff
@@ -44,7 +47,8 @@ def __init__(self,
                  norm_cfg: ConfigType = dict(
                      type='BN', momentum=0.03, eps=0.001),
                  act_cfg: ConfigType = dict(type='Swish'),
-                 init_cfg: OptMultiConfig = None) -> None:
+                 init_cfg: OptMultiConfig = None,
+                 kernel_size: int = 3,) -> None:
```
hhaAndroid, line 51:
```suggestion
                 kernel_size: int = 3) -> None:
```

hhaAndroid: This parameter should be placed before the `init_cfg`.

mmdet/models/layers/csp_layer.py
```diff
@@ -191,9 +197,11 @@ def __init__(self,
                  norm_cfg: ConfigType = dict(
                      type='BN', momentum=0.03, eps=0.001),
                  act_cfg: ConfigType = dict(type='Swish'),
-                 init_cfg: OptMultiConfig = None) -> None:
+                 init_cfg: OptMultiConfig = None,
+                 kernel_size: Union[int, None] = None) -> None:
```
hhaAndroid, line 201:
```suggestion
                 kernel_size: Option[int] = None) -> None:
```

PR #11108, opened 2023-10-31
fix incorrect kl_div_loss

ZaberKo:
The original `knowledge_distillation_kl_div_loss` in LD was mistakenly implementated as mentioned in https://github.com/HikariTJU/LD/issues/74.

## Motivation

Fix the above issue. Now the kl_div is equal to the definition of KL Div.

## Modification

Change the `knowledge_distillation_kl_div_loss`. And modify the corresponding `loss_weight`, so the final behavior should be identical to the original one.

## BC-breaking (Optional)

No.

### DIFF
--- a/configs/ld/ld_r18-gflv1-r101_fpn_1x_coco.py
+++ b/configs/ld/ld_r18-gflv1-r101_fpn_1x_coco.py
@@ -49,7 +49,7 @@ model = dict(
             loss_weight=1.0),
         loss_dfl=dict(type='DistributionFocalLoss', loss_weight=0.25),
         loss_ld=dict(
-            type='KnowledgeDistillationKLDivLoss', loss_weight=0.25, T=10),
+            type='KnowledgeDistillationKLDivLoss', loss_weight=0.25/(16+1), T=10),
         reg_max=16,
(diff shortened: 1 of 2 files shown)


ZaberKo:
Hi, it seems that the CI got stuck, any ideas? @hhaAndroid

hhaAndroid:
@ZaberKo Thank you very much for the feedback. I see that the author's comments don't necessarily indicate a need for modification. Do you think it is necessary to make the change? Perhaps training it this way will yield better results or there might be no difference at all?

hhaAndroid:
@ZaberKo The main concern is that making such a fix now would result in significant backward compatibility breaking, causing issues for existing users.

ZaberKo:
> @ZaberKo Thank you very much for the feedback. I see that the author's comments don't necessarily indicate a need for modification. Do you think it is necessary to make the change? Perhaps training it this way will yield better results or there might be no difference at all?

There is no difference in the final training mathematically.

However, the incorrect kl_div will cause a misusage for future developments. Currently, I didn't find any other models using this function except LD, so I don't think it breaks BC. I adivse we need fix it as soon as possible before any new algorithms using this function.

ZaberKo:
@hhaAndroid Hi, is there any new progress?

PR #11117, opened 2023-11-01
[Feature] Add optional score threshold option to coco_error_analysis.py

guyleaf:

## Motivation

In some models, e.g., DETR, it does not support filtering score after prediction.
By adding score threshold option, we could filter the detection bboxes in flexible way.

It is also a common case to visualize the actual performance in dataset.
For example, positive bbox with low confident score will affect the AP and AR going higher.

## Modification

* Add optional score threshold to coco_error_analysis.py
* Filter the detection bboxes with the score threshold when user set it.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


hhaAndroid:
@guyleaf Please fix the lint

guyleaf:
> @guyleaf Please fix the lint

Done.

hhaAndroid (approved):

PR #11198, opened 2023-11-20
Large Image utilities need a local implementation of shift_bbox

SajjadPSavoji:
## Motivation
MMdetection is adding support for large image inference including a demo.
As of today this implementation is heavily dependant on Sahi codebase and maintenance.
I observed an inconsistency between Sahi postprocesing implementation and MMDetection codebase. 
And I am proposing a adding a local function to extend the support alongside stability. 

## Bug
MMDetection models are able to output boxes with negative coordinate values. While this is the case Sahi Bbox class does not support negative values for coordinates.(That's a shame right :) )

Because of the aforementioned problem when using MMDet models(in my case YoloX) the occasional bbox with negative coordinates(this is negative coordinates in the patch not frame) will throw an exception.

## Modification

I am proposing an easy solution by implementing a local version of `shift_bboxes` function instead of importing it from Sahi.

The modifications include writing a new function for `shift_bboxes` and removing the import from Sahi part in `shift_predictions`.

NOTE: shifting masks are still done by Sahi functions

## BC-breaking (Optional)
Modification is seamless and will not break any downstream tasks due to python dynamic linking.

## Use cases (Optional)
This functionality is use in the large image demo.

## Checklist

### DIFF
--- a/mmdet/utils/large_image.py
+++ b/mmdet/utils/large_image.py
@@ -23,6 +23,22 @@ def shift_rbboxes(bboxes: torch.Tensor, offset: Sequence[int]):
     shifted_bboxes[:, 0:2] = shifted_bboxes[:, 0:2] + offset_tensor
     return shifted_bboxes
 
+def shift_bboxes(bboxes: torch.Tensor, offset: Sequence[int]):
+    """Shift bboxes with offset.
+
+    Args:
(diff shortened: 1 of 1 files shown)


sparshgarg23:
any updates on this,i know mmyolo has it but not sure if mmdetection ever got to implement this given that the libraey has become stale.

PR #11251, opened 2023-12-05
Fix bug in convert_coco_format

honeybadger78:

## Motivation

For a simple test, I ran the instance segmentation demo.ipynb, but encountered a problem in the convert_balloon_to_coco function. By fixing this error, i hope to avoid confusion for beginners using mmdetection.

## Modification

Previously, using the convert_balloon_to_coco function resulted in the following error.

TypeError: "tasks" must be a tuple object or a sequence object, but got <class 'dict_values'>

To fix this, I changed the type to list.

The changes are as follows:
data_infos.values() --> list(data_infos.values())

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


honeybadger78:
@hhaAndroid  could you check my PR?

hhaAndroid (approved):

PR #11280, opened 2023-12-13
Fixed mask shape after Albu postprocess

ilcopione:
## Motivation

If the augmented image (mask) is for example rotated by 90 or 270 deg. using "RandomRotate90" and the original image (mask) is rectangular, it could end up with height and width inverted, therefore we need to use dimensions of the augmented mask (image) to convert.

## Modification

Take the augmented mask shape instead of the original mask shape. 

## BC-breaking (Optional)

None.

## Use cases (Optional)

RandomRotate90

## Checklist


hhaAndroid:
@ilcopione Please fix the lint

ilcopione:
@hhaAndroid, thanks for your reply. I've fixed the lint as you requested!!

hhaAndroid (approved):

Issue #11422, opened 2024-01-24
Albumentations Postprocess Index Error after Release 3.3.0


AItairs:
Hi, 
I noticed this pull request https://github.com/open-mmlab/mmdetection/pull/11280 changes the Albu postprocessing. 
In my pipeline results['masks'] can be empty at that stage which results in an index error for results['masks'][0].shape[0].

hhaAndroid:
@AItairs Thank you. Can you help fix it?

PR #11432, opened 2024-01-25
fix albumentations mask postprocessing

AItairs:
## Motivation

Fix https://github.com/open-mmlab/mmdetection/issues/11422

## Modification

Use augmented image shape instead of first mask shape.

## Comment

Reopend pull request to fix commit author


GeorgePearse:
Why has this not been merged? Completely fixes my problem

PR #11487, opened 2024-02-21
Add missing entry of 'isthing' for  coco panoptic in customize dataset

FranklynJey:
## Motivation

I created a customized dataset for coco panoptic segmentation by following mmdetection/docs/en/advanced_guides/customize_dataset.md.

In the description, we can find

```
categories': [  # including both foreground categories and background categories
    {'id': 0, 'name': 'person'},
    ...
```
 
A dataset created like this will cause errors, as the attribute 'isthing' is required. I first realized it in _browse_dataset.py_ which breaks using the panoptic coco dataset format according to customized_dataset.md.

See 

Screenshot from 2024-02-21 12-05-00

This holds  even if _metainfo_ is defined as 

```
metainfo = dict(
        classes = ('class1','class2',... ),
        thing_classes = ('class1',... ),
        stuff_classes = ('class2',...)
        )

```
and is given to _train_dataloader_ etc.

## Modification

I would recommend changing the line 

```
categories': [  # including both foreground categories and background categories
    {'id': 0, 'name': 'person'},
```
to

```
categories': [  # including both foreground categories and background categories
    {'id': 0, 'name': 'person', 'isthing': 1},
```
so that it matches the original coco panoptic description and prevents errors in mmdetection. 

## PS
Thank you guys so much for your awesome work in mmdetection!

## Checklist

### DIFF
--- a/docs/en/advanced_guides/customize_dataset.md
+++ b/docs/en/advanced_guides/customize_dataset.md
@@ -395,7 +395,7 @@ The annotation json files in COCO Panoptic format has the following necessary ke
 ]
 
 'categories': [  # including both foreground categories and background categories
-    {'id': 0, 'name': 'person'},
+    {'id': 0, 'name': 'person', 'isthing': 1},
     ...
  ]
 ```


FranklynJey:
@jbwang1997  did you have time to look on it ? :)

PR #11789, opened 2024-06-12
Update __init__.py

JohnConnor123:
Fixed a bug due to which mmdet was not installed for some Windows 10 users.

## Motivation
I was one of the users who experienced this bug. All options did not work, and the only working solution to the problem in the project issue on github is to change <2.2.0 to <=2.2.0 in the assert function. For version 2.1.0 an error occurred in all installation methods.

## Modification
<2.2.0 to <=2.2.0 in the assert function

## BC-breaking (Optional)
No


perry-smartsizer:
Really looking forward to this fix!

Kulikovpavel:
Hello!
Also need this patch, is it possible to merge?

PR #11809, opened 2024-06-21
fix:Fix two of the CO-DETR config files

Zzzan:
## Motivation

The motivation of this PR is to fix one of the config files for CO-DETR project. Currently, running
```
python tools/train.py /home/baiye/mamba-test/mmdetection-main/projects/CO-DETR/configs/codino/co_dino_5scale_swin_l_16xb1_3x_coco.py
```
throws a following exception:
```
AssertionError: Each parameter scheduler should contain the key type, but got {'milestones': [30]}
```
I assume that when modifying a particular item within a list, users are required to add a type key to distinguish it from other possible items in that particular list.

## Modification

This PR resolves the error by adding the type key that is inherited from some of the base configs:
co_dino_5scale_swin_l_16xb1_3x_coco.py
co_dino_5scale_r50_lsj_8xb2_3x_coco.py

## BC-breaking (Optional)
N/A
## Use cases (Optional)
Anyone extending the modified config can now do so without running into this error.

## Checklist

### DIFF
--- a/projects/CO-DETR/configs/codino/co_dino_5scale_r50_lsj_8xb2_3x_coco.py
+++ b/projects/CO-DETR/configs/codino/co_dino_5scale_r50_lsj_8xb2_3x_coco.py
@@ -1,4 +1,4 @@
 _base_ = ['co_dino_5scale_r50_lsj_8xb2_1x_coco.py']
 
-param_scheduler = [dict(milestones=[30])]
+param_scheduler = [dict(type='MultiStepLR', milestones=[30])]
 train_cfg = dict(max_epochs=36)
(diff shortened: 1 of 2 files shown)


projects/CO-DETR/configs/codino/co_dino_5scale_r50_lsj_8xb2_3x_coco.py
```diff
@@ -1,4 +1,4 @@
 _base_ = ['co_dino_5scale_r50_lsj_8xb2_1x_coco.py']
 
-param_scheduler = [dict(milestones=[30])]
+param_scheduler = [dict(type='MultiStepLR', milestones=[30])]
```
MiXaiLL76, line 3: Maybe it would be better to do it like in base?

https://github.com/open-mmlab/mmdetection/blob/main/projects/CO-DETR/configs/codino/co_dino_5scale_r50_lsj_8xb2_1x_coco.py#L335
```py
max_epochs = 12
train_cfg = dict(max_epochs=max_epochs)

param_scheduler = [
    dict(
        type='MultiStepLR',
        begin=0,# Defaults to 0.
        end=max_epochs, # Defaults to INF.
        by_epoch=True, # Defaults to True
        milestones=[30],
        gamma=0.1) # gamma=0.1 - default
]
```

PR #12063, opened 2024-12-05
[Bug Fix] fix eval_recalls for compatibility with NumPy > 1.23

risingClouds:
## Motivation

In NumPy versions prior to 1.23, creating an array from nested lists of varying shapes would result in an object array by default. Starting from NumPy 1.23, such usage raises a ValueError instead. This PR addresses the issue in the eval_recalls function to ensure compatibility with NumPy > 1.23

## Modification

Replaced `all_ious = np.array(all_ious)` with `all_ious = np.array(all_ious, dtype=object)` in `mmdet/evaluation/functional/recall.py`, specifically in the `eval_recall` function (line 111).

## BC-breaking (Optional)

No, this modification does not break backward compatibility. The function is adjusted to maintain consistent behavior across NumPy versions.

## Checklist
1. ✅ Pre-commit linting tools have been used to ensure code quality.
2. ✅ Unit tests have been updated to verify the fix and ensure compatibility across different NumPy versions.
3. ✅ Tested with downstream projects to ensure no regressions (e.g., MMDet).
4. ✅ Updated the documentation and added comments in the code explaining the change.

### DIFF
--- a/mmdet/evaluation/functional/recall.py
+++ b/mmdet/evaluation/functional/recall.py
@@ -108,7 +108,7 @@ def eval_recalls(gts,
                 img_proposal[:prop_num, :4],
                 use_legacy_coordinate=use_legacy_coordinate)
         all_ious.append(ious)
-    all_ious = np.array(all_ious,dtype=object)
+    all_ious = np.array(all_ious, dtype=object)
     recalls = _recalls(all_ious, proposal_nums, iou_thrs)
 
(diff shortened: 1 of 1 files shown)


lauriebax:
If this MR doesn't get merged here, feel free to open it against our VBTI fork at https://github.com/VBTI-development/onedl-mmdetection - we're actively maintaining updated versions with latest PyTorch support and would welcome the contribution!

PR #12304, opened 2025-02-09
Make MMCV and MMEngine Upper Version Bound Inclusive

anglesideangle:
## Motivation

When trying to build mmdetection from source, the build was unsuccessful because "`AssertionError: MMCV==2.2.0 is used but incompatible. Please install mmcv>=2.0.0rc4, <2.2.0.`" This may be intended behavior, but I suspect maximum version is supposed to be inclusive.

## Modification

I replaced < with <= for maximum version checking of mmcv and mmengine.


anglesideangle:
Looks like this is a duplicate of #11789

